PR #800, opened 2015-01-20
JBTM-2333 Update to remove the jacorb dependency

zhfeng:
!MAIN !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB

NO_TEST


mmusgrov:
Hi Amos,

Did you get rid of the calls to initialize the orb in TransportFactory (OrbManagement.getInstance)

blacktie/jatmibroker-xatmi/src/main/java/org/jboss/narayana/blacktie/jatmibroker/core/transport/OrbManagement.java
```diff
@@ -26,9 +26,9 @@
 public class OrbManagement implements Runnable {
     private static final Logger log = LogManager.getLogger(OrbManagement.class);
     private static final String CorbaOrbClassProp = "org.omg.CORBA.ORBClass";
-    private static final String CorbaOrbClassValue = "org.jacorb.orb.ORB";
+    private static final String CorbaOrbClassValue = "com.sun.corba.se.impl.orb.ORBImpl";
```
zhfeng: Hi Mike,

I just changed to use the com.sun.corba.se.impl.orb.ORBImpl
@tomjenkinson does it make sense ?

Amos

tomjenkinson: Sure, but I think we should really try to consider whether we need OrbManagement at all as the code should not really be using ORB anymore as we use RTS to propagate the transaction. There is one use case within the app server were the transaction is propagated over EJB and if we need OrbManagement for that then sure, if the change you made works then happy to go with that. If we don't need the OrbManagement class for that I would just remove it but that is my personal opinion and I haven't reviewed the code itself.

mmusgrov: I agree that we no longer need OrbManagement. The test I referred to in yesterdays meeting uses JTS to propagate a transaction initiated by BlackTie between two servers but that is different from requiring BlackTie to support OTS - so my apologies for introducing the confusion.

So, although we initialize the OrbManagement function, we never actually use CORBA anywhere in the BlackTie code base. So it is up to you whether you want to make the change here or simply create a new JIRA for removing CORBA and fix it another time.

mmusgrov: But if you do decide to remove CORBA I think your commit which replaces JacORB with JDK orb should be kept just in case we ever do need to go back to using CORBA/OTS as the transport.

zhfeng: Thanks a lot for your input !

I'd reviewed the codebase and found that https://github.com/jbosstm/narayana/blob/master/blacktie/jatmibroker-xatmi/src/main/java/org/jboss/narayana/blacktie/jatmibroker/core/tx/TransactionImpl.java which also uses the OrbManagement. I'm not very sure that we should remove it ?

zhfeng: I prefer to NOT remove the OrbManagement and create another JIRA for removing CORBA as Mike's suggestion.

mmusgrov: So we still need to port the JATMI code - best to leave that for another day

zhfeng: Thanks Mike ! I get it

PR #803, opened 2015-01-28
JBTM-2336

tomjenkinson:
!BLACKTIE !XTS !QA_JTS_JACORB

### DIFF
--- a/ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/ConnectionImple.java
+++ b/ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/ConnectionImple.java
@@ -379,7 +379,8 @@ public class ConnectionImple implements Connection
 
 	        if (!delayClose)  // close now
 	        {
-	            _transactionalDriverXAConnectionConnection.closeCloseCurrentConnection();
+	            if (_theModifier != null)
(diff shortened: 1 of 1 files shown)


ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/ConnectionImple.java
```diff
@@ -379,7 +379,8 @@ public void close() throws SQLException
 
 	        if (!delayClose)  // close now
 	        {
-	            _transactionalDriverXAConnectionConnection.closeCloseCurrentConnection();
+	            if (_theModifier != null)
```
mmusgrov: So if the driver is not in the modifier list then you do not register an AC to close the connection. Similarly, at this line, if its not in the modifier list then you do not call close on the TransactionalDriverXAConnection connection. But, as I recall, the original bug was that TransactionalDriverXAConnection#close() was not getting called so I would say the fix is not sufficient.

tomjenkinson: Well, I have merged this now so it should close the regression I caused. What I propose is that we create a new bug to look into this case as it was not addressed before either. Would you like to raise the Jira? We should reference this Jira for hints if we don't tackle it straight away

PR #806, opened 2015-02-17
[JBTM-2341] Extra statistics

mmusgrov:
!PERF !BLACKTIE !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
@@ -1506,6 +1506,8 @@ public class BasicAction extends StateManager
             return Abort();
         }
 
+        Long startTime = TxStats.enabled() ? System.nanoTime() : null;
+
         if (pendingList != null)
         {
             /*
@@ -1558,7 +1560,10 @@ public class BasicAction extends StateManager
             {
                 if (heuristicDecision != TwoPhaseOutcome.HEURISTIC_ROLLBACK)
                 {
-                    TxStats.getInstance().incrementCommittedTransactions();
+                    if (startTime == null)
+                        TxStats.getInstance().incrementCommittedTransactions(1L);
+                    else
+                        TxStats.getInstance().incrementCommittedTransactions(System.nanoTime() - startTime);
                 }
             }
         }
@@ -1853,6 +1858,8 @@ public class BasicAction extends StateManager
         }
         else
         {
+            Long startTime = TxStats.enabled() ? System.nanoTime() : null;
+
             criticalStart();
 
             actionStatus = ActionStatus.COMMITTING;
@@ -1926,7 +1933,11 @@ public class BasicAction extends StateManager
 
             if (TxStats.enabled()) {
                 if (heuristicDecision != TwoPhaseOutcome.HEURISTIC_ROLLBACK) {
-                    TxStats.getInstance().incrementCommittedTransactions();
+                    // NB statistics monitoring could have been dynamically enabled after starting this transaction
+                    if (startTime == null)
+                        TxStats.getInstance().incrementCommittedTransactions(1L);
+                    else
+                        TxStats.getInstance().incrementCommittedTransactions(System.nanoTime() - startTime);
                 }
             }
 
@@ -1993,11 +2004,14 @@ public class BasicAction extends StateManager
 
         /*
            * To get to this stage we had to try to commit, which means that we're
-           * rolling back because of a resource problem.
+           * rolling back because of a resource problem or an internal error.
            */
 
         if (TxStats.enabled()) {
-            TxStats.getInstance().incrementResourceRollbacks();
+            if (internalError)
+                TxStats.getInstance().incrementSystemRollbacks();
+            else
+                TxStats.getInstance().incrementResourceRollbacks();
             TxStats.getInstance().incrementAbortedTransactions();
         }
     }
@@ -2103,7 +2117,7 @@ public class BasicAction extends StateManager
             if (getStore() == null)
             {
                 actionStatus = ActionStatus.ABORT_ONLY;
-
+                internalError = true;
                 return TwoPhaseOutcome.PREPARE_NOTOK;
             }
         }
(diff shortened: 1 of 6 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
```diff
@@ -1558,7 +1560,10 @@ protected synchronized int End (boolean reportHeuristics)
             {
                 if (heuristicDecision != TwoPhaseOutcome.HEURISTIC_ROLLBACK)
                 {
-                    TxStats.getInstance().incrementCommittedTransactions();
+                    if (startTime == null)
```
tomjenkinson: I don't think startTime can be null here as line 1509 sets this if TxStats.enabled we are in 1159 which also checks .enabled()

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
```diff
@@ -1926,7 +1933,11 @@ protected synchronized final void phase2Commit (boolean reportHeuristics) throws
 
             if (TxStats.enabled()) {
                 if (heuristicDecision != TwoPhaseOutcome.HEURISTIC_ROLLBACK) {
-                    TxStats.getInstance().incrementCommittedTransactions();
+                    // NB statistics monitoring could have been dynamically enabled after starting this transaction
+                    if (startTime == null)
```
tomjenkinson: same comment about startTime not being able to be null

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
mmusgrov: stats can be enabled dynamically so they they can be disabled when the txn is started and then enabled mid flight.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TxStats.java
```diff
@@ -184,9 +204,37 @@ void incrementAbortedTransactions()
...
+
+		if (duration == 0)
+			duration = 1; // a duration of zero is disallowed - round up to minimum unit
+
+		do {
+			prev = commitStatistic.get();
+
+			long prevCount = prev.getNumberOfCommittedTransactions();
+			long newCount = prevCount + 1;
+			// nextAvg = (duration  + prevCount * prevAvg) / (prevCount + 1)
```
tomjenkinson: Is this intended to be left in?

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TxStats.java
```diff
@@ -184,9 +204,37 @@ void incrementAbortedTransactions()
...
+	void incrementCommittedTransactions(long duration) {
+		TxCommitStatistic prev, next;
+
+		if (duration == 0)
+			duration = 1; // a duration of zero is disallowed - round up to minimum unit
+
+		do {
+			prev = commitStatistic.get();
+
+			long prevCount = prev.getNumberOfCommittedTransactions();
```
tomjenkinson: looks good in general - can't we just call getNumberOfCommittedTransactions instead of storing it in TxCommitStatistic? Then you could just an atomicreference to a long rather than the new TxCommitStatistic. Overall my question is to the need for TxCommitStatistic instead of just averageCommitTime stored as a long

mmusgrov: The average commit time depends upon how many previous commits there were. So when a new transaction is committed the new average will be:

newAverage = (duration  + prevCount * prevAvg) / (prevCount + 1)

where duration is how long it took to commit the new transaction.

Therefore I need to (atomically) get the current number of committed transactions (prevCount) and the current average (prevAvg) in order to calculate the newAverage.

To atomically read (or write) a pair of values we need an AtomicReference (which is quicker that using synchronized blocks).

tomjenkinson: Sure, ok. I was sacrificing accuracy for a possibly faster implementation but my approach would be more affected by skew with a low number of transactions with widely distributed values which might be what the user is interested in sampling

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TxStats.java
```diff
@@ -184,9 +204,37 @@ void incrementAbortedTransactions()
...
 	}
 
+	/**
+	 * @Deprecated as of 5.0.5.Final use {@link #incrementCommittedTransactions(long)}} instead
+	 */
+	@Deprecated
 	void incrementCommittedTransactions()
 	{
-        numberOfCommittedTransactions.incrementAndGet();
+        incrementCommittedTransactions(0L);
```
tomjenkinson: is it worth using 1L here to avoid the mutation on l224

mmusgrov: Nice tip

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TxStatsMBean.java
```diff
@@ -53,6 +53,13 @@
     long getNumberOfCommittedTransactions();
 
     /**
+     * Returns the average time it is taking to commit a transaction.
+     * Note that a small number of stuck transactions can skew the overall average.
+     * @return the average time, in nanoseconds, it has taken to commit a transaction.
+     */
+    long getAverageCommitTime();
```
tomjenkinson: Hi Mike, as per the comment on the forum I don't think this Javadoc definitively clarifies it is the timing of the commit() call not the time from begin() to commit() completing. I will reopen the issue for that aspect to be covered

PR #809, opened 2015-03-03
JBTM-2346 Remove the H2 datasource from the test war as it is provided b...

tomjenkinson:
!XTS !BLACKTIE !QA_JTA !QA_JTS_JACORB


mmusgrov:
The fix looks good, except, what happens if wildfly remove or replace the default H2 datasource?

PR #813, opened 2015-03-11
JBTM-2349. Close CompensationContext only after all participants are done

Gytis:
NO_TESTS

### DIFF
--- a/compensations/src/main/java/org/jboss/narayana/compensations/impl/ParticipantImpl.java
+++ b/compensations/src/main/java/org/jboss/narayana/compensations/impl/ParticipantImpl.java
@@ -32,11 +32,16 @@ import org.jboss.narayana.compensations.api.ConfirmationHandler;
 import org.jboss.narayana.compensations.api.TransactionLoggedHandler;
 
 import javax.enterprise.inject.spi.BeanManager;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * @author paul.robinson@redhat.com 22/03/2013
  */
 public class ParticipantImpl implements BusinessAgreementWithParticipantCompletionParticipant, ConfirmCompletedParticipant {
+
+    private static final Map<Object, AtomicInteger> PARTICIPANT_COUNTERS = new HashMap<>();
 
     private Class<? extends CompensationHandler> compensationHandler;
     private Class<? extends ConfirmationHandler> confirmationHandler;
@@ -55,6 +60,8 @@ public class ParticipantImpl implements BusinessAgreementWithParticipantCompleti
 
         beanManager = BeanManagerUtil.getBeanManager();
         applicationClassloader = Thread.currentThread().getContextClassLoader();
+
+        incrementParticipantsCounter();
     }
 
     private <T extends Object> T instantiate(Class<T> clazz) {
@@ -92,9 +99,10 @@ public class ParticipantImpl implements BusinessAgreementWithParticipantCompleti
             ConfirmationHandler handler = instantiate(confirmationHandler);
             handler.confirm();
 
-            CompensationContext.close(currentTX);
             Thread.currentThread().setContextClassLoader(origClassLoader);
         }
+
+        decrementParticipantsCounter();
     }
 
     @Override
@@ -115,12 +123,13 @@ public class ParticipantImpl implements BusinessAgreementWithParticipantCompleti
                 CompensationHandler handler = instantiate(compensationHandler);
                 handler.compensate();
 
-                CompensationContext.close(currentTX);
                 Thread.currentThread().setContextClassLoader(origClassLoader);
             }
         } catch (Exception e) {
             e.printStackTrace();
         }
+
+        decrementParticipantsCounter();
     }
 
     @Override
@@ -137,5 +146,40 @@ public class ParticipantImpl implements BusinessAgreementWithParticipantCompleti
     @Override
     public void error() throws SystemException {
 
+    }
+
+    /**
+     * Increments the counter of the Compensations participants in the transaction.
+     */
+    private void incrementParticipantsCounter() {
+
+        synchronized (PARTICIPANT_COUNTERS) {
+            final AtomicInteger counter = PARTICIPANT_COUNTERS.get(currentTX);
+
+            if (counter == null) {
+                PARTICIPANT_COUNTERS.put(currentTX, new AtomicInteger(1));
+            } else {
+                counter.incrementAndGet();
+            }
+        }
+    }
+
+    /**
+     * Decrements the counter of the Compensations participants in the transaction.
+     * CompensationContext of the current transaction is destroyed once the counter reaches 0.
+     */
+    private void decrementParticipantsCounter() {
+
+        synchronized (PARTICIPANT_COUNTERS) {
+            final AtomicInteger counter = PARTICIPANT_COUNTERS.get(currentTX);
+
+            if (counter == null || counter.decrementAndGet() > 0) {
+                return;
+            }
+
+            PARTICIPANT_COUNTERS.remove(currentTX);
+        }
+
+        CompensationContext.close(currentTX);
     }
 }
--- a/compensations/src/test/java/org/jboss/narayana/compensations/functional/compensationScoped/CompensationScopedTest.java
+++ b/compensations/src/test/java/org/jboss/narayana/compensations/functional/compensationScoped/CompensationScopedTest.java
@@ -147,6 +147,22 @@ public abstract class CompensationScopedTest {
         Assert.assertEquals(false, MyConfirmationHandler.dataAvailable);
     }
 
+    @Test
+    public void testTwoCompensationHandlers() throws Exception {
+
+        ParticipantCompletionCoordinatorRules.setParticipantCount(6);
+
+        MyCompensationHandler.expectedData = "blah";
+        MyConfirmationHandler.expectedData = "blah";
+
+        getBAControler().beginBusinessActivity();
+        service.doWork("blah");
+        service.doWork("blah");
+        getBAControler().cancelBusinessActivity();
+
+        Assert.assertTrue(MyCompensationHandler.dataAvailable);
+        Assert.assertFalse(MyConfirmationHandler.dataAvailable);
+    }
 
     @Test
     public void testConfirmationHandler() throws Exception {
@@ -162,6 +178,23 @@ public abstract class CompensationScopedTest {
 
         Assert.assertEquals(false, MyCompensationHandler.dataAvailable);
         Assert.assertEquals(true, MyConfirmationHandler.dataAvailable);
+    }
+
+    @Test
+    public void testTwoConfirmationHandlers() throws Exception {
+
+        ParticipantCompletionCoordinatorRules.setParticipantCount(6);
+
+        MyCompensationHandler.expectedData = "blah";
+        MyConfirmationHandler.expectedData = "blah";
+
+        getBAControler().beginBusinessActivity();
+        service.doWork("blah");
+        service.doWork("blah");
+        getBAControler().closeBusinessActivity();
+
+        Assert.assertFalse(MyCompensationHandler.dataAvailable);
+        Assert.assertTrue(MyConfirmationHandler.dataAvailable);
     }
 
 }
--- a/compensations/src/test/java/org/jboss/narayana/compensations/functional/compensationScoped/MyCompensationHandler.java
+++ b/compensations/src/test/java/org/jboss/narayana/compensations/functional/compensationScoped/MyCompensationHandler.java
@@ -43,6 +43,8 @@ public class MyCompensationHandler implements CompensationHandler {
 
         if (myData != null && myData.getData() != null && myData.getData().equals(expectedData)) {
             dataAvailable = true;
+        } else {
+            dataAvailable = false;
         }
     }
 }
--- a/compensations/src/test/java/org/jboss/narayana/compensations/functional/compensationScoped/MyConfirmationHandler.java
+++ b/compensations/src/test/java/org/jboss/narayana/compensations/functional/compensationScoped/MyConfirmationHandler.java
@@ -43,6 +43,8 @@ public class MyConfirmationHandler implements ConfirmationHandler {
 
         if (myData != null && myData.getData() != null && myData.getData().equals(expectedData)) {
             dataAvailable = true;
+        } else {
+            dataAvailable = false;
         }
     }
 }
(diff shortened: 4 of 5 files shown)


Gytis:
@mmusgrov, @paulrobinson could you take a look at this, please?

paulrobinson:
This looks right to me.

compensations/src/main/java/org/jboss/narayana/compensations/impl/CompensationContextLifetimeManager.java
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * Notify the manager that new participant was created.
+     *
+     * @param currentTX
+     */
+    public synchronized void participantCreated(final Object currentTX) {
+        AtomicInteger counter = counters.get(currentTX);
+
+        if (counter == null) {
+            counter = new AtomicInteger(1);
```
mmusgrov: Why not call counters.put(currentTX, counter); here

Gytis: Though will save a line of code :) do you think it would be cleaner to have two lines of "counters.put(currentTX, counter);"?

tomjenkinson: I expect Mike is suggesting you would remove the .put on line 61

compensations/src/test/java/org/jboss/narayana/compensations/functional/compensationScoped/MyCompensationHandler.java
```diff
@@ -43,6 +43,8 @@ public void compensate() {
 
         if (myData != null && myData.getData() != null && myData.getData().equals(expectedData)) {
             dataAvailable = true;
+        } else {
+            dataAvailable = false;
```
tomjenkinson: just checking, is this bit required due to the changes? I saw the test change so wondered about exisiting client compatibility. Same question for the other data|Available = false changes

compensations/src/main/java/org/jboss/narayana/compensations/impl/CompensationContextLifetimeManager.java
Gytis: There is an else clause below after which "put" is also needed

tomjenkinson: Ok, it must be something subtle because it seems the get on line 53 would retrieve the counter so it should already be in the counters map and hence shouldn't need the put on line 61?

compensations/src/main/java/org/jboss/narayana/compensations/impl/CompensationContextLifetimeManager.java
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Class used to make sure that CompensationContext is closed once all participants have completed.
+ *
+ * @author <a href="mailto:gytis@redhat.com">Gytis Trikleris</a>
+ */
+public final class CompensationContextLifetimeManager {
+
+    private static CompensationContextLifetimeManager INSTANCE = new CompensationContextLifetimeManager();
+
+    private final Map<Object, AtomicInteger> counters;
```
mmusgrov: This class is maintaining a counter on behalf of each instance of ParticipantImpl (ie there is one counter in the map for each instance of ParticipantImpl). Rather than have a global class with global synchronizations why not simply maintain the counter inside ParticipantImpl.java

compensations/src/test/java/org/jboss/narayana/compensations/functional/compensationScoped/MyCompensationHandler.java
Gytis: This I have added first thing, to prove that this fails.
The issue was, if there were two instances of this handler, second instance would have myData as null, because of the bug. And because of that it wouldn't change dataAvailability, which was set to true by the first handler instance. So the test would always pass even with the bug

compensations/src/main/java/org/jboss/narayana/compensations/impl/CompensationContextLifetimeManager.java
```diff
@@ -0,0 +1,86 @@
...
+     */
+    public synchronized void participantDestroyed(final Object currentTX) {
+        AtomicInteger counter = counters.get(currentTX);
+
+        if (counter == null) {
+            return;
+        }
+
+        if (counter.decrementAndGet() > 0) {
+            counters.put(currentTX, counter);
```
mmusgrov: The entry (currentTX, counter) is already in the map so you do not need to put it in again.

compensations/src/test/java/org/jboss/narayana/compensations/functional/compensationScoped/MyCompensationHandler.java
tomjenkinson: got you, thanks for confirming

Gytis: fyi same applies for all handlers

compensations/src/main/java/org/jboss/narayana/compensations/impl/CompensationContextLifetimeManager.java
Gytis: I do need to update it with decremented value "counter.decrementAndGet()"

compensations/src/main/java/org/jboss/narayana/compensations/impl/ParticipantImpl.java
```diff
@@ -92,9 +94,10 @@ public void close() throws WrongStateException, SystemException {
             ConfirmationHandler handler = instantiate(confirmationHandler);
             handler.confirm();
 
-            CompensationContext.close(currentTX);
             Thread.currentThread().setContextClassLoader(origClassLoader);
         }
+
+        CompensationContextLifetimeManager.getInstance().participantDestroyed(currentTX);
```
mmusgrov: These lifecycle methods can throw exceptions. I do not fully understand the semantics of such exceptions but you may need to put the  participantDestroyed calls inside a finally block.

compensations/src/main/java/org/jboss/narayana/compensations/impl/CompensationContextLifetimeManager.java
Gytis: Ah, I imagine it wouldn't change the instance in the map... In that case you both are right. Same applies for the Mikes comment bellow then

compensations/src/test/java/org/jboss/narayana/compensations/functional/compensationScoped/CompensationScopedTest.java
```diff
@@ -147,6 +147,22 @@ public void testCompensationHandler() throws Exception {
...
+
+        MyCompensationHandler.expectedData = "blah";
+        MyConfirmationHandler.expectedData = "blah";
+
+        getBAControler().beginBusinessActivity();
+        service.doWork("blah");
+        service.doWork("blah");
+        getBAControler().cancelBusinessActivity();
+
+        Assert.assertEquals(true, MyCompensationHandler.dataAvailable);
```
mmusgrov: Why not use AssertTrue here

compensations/src/main/java/org/jboss/narayana/compensations/impl/CompensationContextLifetimeManager.java
Gytis: Yeah, these were my two options. Just thought this might be leaner to not pollute ParticipantImple class too much... And that's the main thing I want the feedback on, because both implementations are pretty much the same

compensations/src/test/java/org/jboss/narayana/compensations/functional/compensationScoped/CompensationScopedTest.java
Gytis: good point

Gytis: I just blindly copied old tests, didn't even think about this

compensations/src/main/java/org/jboss/narayana/compensations/impl/CompensationContextLifetimeManager.java
mmusgrov: The counter is already in the map. You are just mutating its state - ie the instance itself has not changed

compensations/src/main/java/org/jboss/narayana/compensations/impl/ParticipantImpl.java
Gytis: What exception do you have in mind? CompensationContextLifetimeManager class comes with this PR

compensations/src/test/java/org/jboss/narayana/compensations/functional/compensationScoped/CompensationScopedTest.java
tomjenkinson: I was just about to mention that I can see where the pattern was copied from on line 147 ;)

compensations/src/main/java/org/jboss/narayana/compensations/impl/CompensationContextLifetimeManager.java
Gytis: Ok I'm updating all usages of the Map

compensations/src/main/java/org/jboss/narayana/compensations/impl/CompensationContextLifetimeManager.java
mmusgrov: But now all transaction are having to potentially block waiting for other threads inside synchronized void participantDestroyed to finish.

But you are polluting the object name space and the logic rightfully belongs inside ParticipantImple

Gytis: Agree on name space. I have that implementation as well.
However, synchronization would still be needed, because counters map would be a static singleton.

mmusgrov: My proposal was that you would no longer need a global map. Instead you have a local fied in ParticipantImple for the counter which is decremented/incremented in the various ParticipantImple methods. And since the counter is atomic you most likely do not need any synchronization whatsoever.

compensations/src/main/java/org/jboss/narayana/compensations/impl/ParticipantImpl.java
mmusgrov: I refer to the methods in ParticipantImpl. For example:
public void close() throws WrongStateException, SystemException

If the body of the method throws one of these exceptions then your participantDestroyed call is not invoked.

compensations/src/main/java/org/jboss/narayana/compensations/impl/CompensationContextLifetimeManager.java
Gytis: Counter needs to be incremented once participant is created i.e. inside the constructor. It counts how many compensations participants are enlisted to the particular transaction, so that the context would be destroyed only when all of them complete.

compensations/src/main/java/org/jboss/narayana/compensations/impl/ParticipantImpl.java
Gytis: Ah yes I know that. I wasn't sure about this one, because before context destruction was without try/catch as well. So if e.g. close operation fails context does not close.

compensations/src/main/java/org/jboss/narayana/compensations/impl/CompensationContextLifetimeManager.java
mmusgrov: Yes but my point is that you are maintaining one counter per instance of ParticipantImple. Therefore store the counter as a local field of the class - you not need a map for this purpose.

Gytis: No, I maintain one counter per currentTX

compensations/src/main/java/org/jboss/narayana/compensations/impl/ParticipantImpl.java
mmusgrov: OK, I did preface my comment with "I do not fully understand the semantics of such exceptions ..." 

Does the specification say that the calls will be reissued if any/all of these exception types are thrown? If not then you would probably require the finally block.

Gytis: Well it's WS-BA participant, so this might be recalled by recovery. However, I'm not sure about the life cycle of the context. To be honest, my first instinct was to add it to the final block, but I removed it later.

compensations/src/main/java/org/jboss/narayana/compensations/impl/CompensationContextLifetimeManager.java
mmusgrov: I see, so there can be many ParticipantImple's with the same currentTX. In that case I agree you need a global map. Unless there is some class somewhere whose instances are in 1-1 correspondence the currentTX. If there is I would store the counter there and avoid the global map. I would definitely strive to remove the need to introduce global variables.

compensations/src/main/java/org/jboss/narayana/compensations/impl/ParticipantImpl.java
tomjenkinson: Maybe @paulrobinson can help us out here?

compensations/src/main/java/org/jboss/narayana/compensations/impl/CompensationContextLifetimeManager.java
Gytis: The only option I could think of would be LocalBAControler and RemoteBAControler (they call participant constructor). However, those methods are stateless and BAControllerFactory creates new instance on every request...

mmusgrov: OK go with the global then.

The costliest bit with respect to locking the whole object is the call to:
CompensationContext.close(currentTX);
Could make this call without holding the global lock on CompensationContextLifetimeManager?

Gytis: Yeah CompensationContext.close(currentTX) is save to call without synchronization

Gytis: So now need to decide if to keep CompensationContextLifetimeManager or move the functionality to static methods of ParticipantImpl?

mmusgrov: Good, so you move the CompensationContext.close() call outside of the synchronized block.

I think that the location of the global map is now down to your own personal preference.

Gytis:
Crash recovery tests failed because of WFLY-4424. Compensations bit passed

Gytis:
@mmusgrov, is the PERF failure caused by yesterday's commit?

mmusgrov:
You can disable performance regression checks using !PERF

Gytis:
I know, just forgot about it. Anyway, it's good for review now.

PR #815, opened 2015-03-18
[JBTM-2347] propagate narayana.sh args to benchmark.sh

mmusgrov:
NO_TEST


mmusgrov:
@tomjenkinson I am merging the PR since it only affects PR performance testing and I want to check the behaviour on Jespers PR 807

We can continue the benchmarking discussion via email (or via JBTM-2347)

mmusgrov:
Awesome indeed. I think now we can expect to see the occasional spurious failure due to performance regressions.

PR #816, opened 2015-03-19
JBTM-2353 Compensation handler should not throw exception, if there is no transaction

Gytis:
!PERF !MAIN !BLACKTIE !QA_JTA !QA_JTS_JACORB


mmusgrov:
Looks good Gytis (sorry for the delay - I've been on PTO)

PR #817, opened 2015-03-20
JBTM-2347 typo in PR comment text

mmusgrov:
!XTS !BLACKTIE !QA_JTA !QA_JTS_JACORB !MAIN PERFORMANCE_IMPROVEMENT


mmusgrov:
PERF profile tests failed because I have made a change to the JMHARGS passed to the job but that config is on haverstraw which is not on the redhat network - I will update the config tomorrow. The change was to set

JMHARGS="-i 6 -wi 10 -f 10 -r 60" instead of JMHARGS="-i 6 -wi 10 -f 10 -r 60 -rf csv -rff "

Passing  -rf csv -rff via the job config was error prone so I moved it into benchmark.sh

PR #819, opened 2015-03-25
JBTM-2222 Do not move the GAV for txframework

tomjenkinson:
!BLACKTIE !QA_JTA !QA_JTS_JACORB !PERF !MAIN

Requires: https://github.com/jbosstm/jboss-as/pull/36


Gytis:
Hi Tom,

overall I think it looks correct.
Only not sure about *.sagas.* test. This doesn't test neither compensations nor txframework directly, but it's useful. And it will be gone once we remove txframework completely. So maybe it might be good to keep it in compensations, just in case?

Gytis:
sure, no problem.

PR #820, opened 2015-03-30
JBTM-2355 Upgrade Narayana to use Wildfly 9.0.0.CR1-SNAPSHOT

zhfeng:


mmusgrov:
Please ignore the PERF profile failure - I needed to update the perf job arguments on the local network which I have just done.

PR #821, opened 2015-04-07
JBTM-2356. REST-AT transaction should not be removed, if participant fails

Gytis:
!BLACKTIE !XTS !QA_JTA !QA_JTS_JACORB !PERF

### DIFF
--- a/rts/at/tx/src/main/java/org/jboss/jbossts/star/resource/Transaction.java
+++ b/rts/at/tx/src/main/java/org/jboss/jbossts/star/resource/Transaction.java
@@ -323,7 +323,9 @@ public class Transaction extends AtomicAction
             }
         }
 
-        coordinator.removeTxState(arjunaStatus, this, enlistmentIds);
+        if (failedList == null || failedList.size() == 0) {
+            coordinator.removeTxState(arjunaStatus, this, enlistmentIds);
+        }
 
         return super.afterCompletion(arjunaStatus, report_heuristics);
     }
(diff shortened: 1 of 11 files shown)


Gytis:
One is the fix and another was the reproducer. It was not necessary because the problem was in the different module. But I didn't want to throw it out, so I thought I'll keep it as a separate commit.

rts/at/tx/src/main/java/org/jboss/jbossts/star/resource/Transaction.java
```diff
@@ -323,7 +323,9 @@ protected boolean afterCompletion(int arjunaStatus, boolean report_heuristics) {
             }
         }
 
-        coordinator.removeTxState(arjunaStatus, this, enlistmentIds);
+        if (failedList == null || failedList.size() == 0) {
```
Gytis: @mmusgrov, do you think I should add heuristicList to this check as well?

mmusgrov: I would say no because such participants are ignored by the recovery system. 

However, the participant is not allowed to forget the heuristic choice it made until it is told to by the system. This should happen by the standard Narayana recovery mechanisms (ie the REST-AT coordinator does not do it explicitly). Maybe we should consider having a test for that (ie it would verify that the participant is eventually told to forget its decision).

mmusgrov: Wait, I already have a tests for that:
https://github.com/jbosstm/narayana/blob/master/rts/at/tx/src/test/java/org/jboss/jbossts/star/test/SpecTest.java#L426
and
https://github.com/jbosstm/narayana/blob/master/rts/at/tx/src/test/java/org/jboss/jbossts/star/test/SpecTest.java#L371

PR #822, opened 2015-04-09
JBTM-2101 Blacktie integration CSTest hang

zhfeng:
!XTS !QA_JTA !QA_JTS !QA_JTS_JACORB !PERF !MAIN

### DIFF
--- a/blacktie/integration-tests/src/test/java/org/jboss/narayana/blacktie/jatmibroker/xatmi/CSControl.java
+++ b/blacktie/integration-tests/src/test/java/org/jboss/narayana/blacktie/jatmibroker/xatmi/CSControl.java
@@ -27,6 +27,8 @@ import junit.framework.TestCase;
 
 import org.apache.log4j.LogManager;
 import org.apache.log4j.Logger;
+import org.jboss.narayana.blacktie.jatmibroker.xatmi.Connection;
+import org.jboss.narayana.blacktie.jatmibroker.xatmi.ConnectionFactory;
 
 public abstract class CSControl extends TestCase {
     static final Logger log = LogManager.getLogger(CSControl.class);
@@ -40,6 +42,7 @@ public abstract class CSControl extends TestCase {
     private String CS_EXE;
     private String REPORT_DIR;
     private static int sid = 1;
+    private Connection connection;
 
     protected boolean isSunOS = false;
     protected boolean isWinOS = false;
@@ -54,7 +57,11 @@ public abstract class CSControl extends TestCase {
             if (server != null) {
                 log.debug("destroying server process");
                 server.interrupt();
-                server.getProcess().destroy();
+
+                log.debug("send shutdown command to the BTDomainAdmin");
+                X_OCTET sendbuf = (X_OCTET) connection.tpalloc("X_OCTET", null);
+                sendbuf.setByteArray("shutdown,testsui,0,".getBytes());
+                connection.tpcall("BTDomainAdmin", sendbuf, 0);
                 log.debug("destroyed server process");
                 server.getProcess().waitFor();
             }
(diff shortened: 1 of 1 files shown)


mmusgrov:
It should be safe for yoyu to ignore the PERF profile failure which I think is a result of misconfiguration and I will investigate that.

zhfeng:
Thanks Mike, I added "!PERF"

zhfeng:
The CI builds have still hang when running the CSTests. I will keep investigating.

mmusgrov:
FYI the PERF hang is fixed by JBTM-2361

zhfeng:
I just disable the tests until I make it work on my local machine

zhfeng:
The failure is due to JBTM-2353 and it needs to wait https://github.com/jbosstm/narayana/pull/827 for merging and re-test again

zhfeng:
re-test again

zhfeng:
rebase and test it again

blacktie/integration-tests/src/test/java/org/jboss/narayana/blacktie/jatmibroker/xatmi/CSControl.java
```diff
@@ -53,10 +54,14 @@ public void tearDown() {
         try {
             if (server != null) {
                 log.debug("destroying server process");
-                server.interrupt();
-                server.getProcess().destroy();
-                log.debug("destroyed server process");
+                try {
+                    int pid = getUnixPID(server.getProcess());
+                    Runtime.getRuntime().exec("kill -INT " + pid).waitFor();
```
tomjenkinson: Can you add a note as to why this is required? I am suprised that server.interrupt() does not propagate to the process.

mmusgrov: The test launches a C program which in turn spawns a C process to run a blacktie server. Before the fix the kill signal kills the CSControl process but not the spawned Blacktie server.

tomjenkinson: Ah, can cs.exe propagate the interrupt then? Trap the interrupt and send it to the server.exe? 

The issue with the current fix is it won't work on Windows. I realise its a test issue but it could happen on windows too.

mmusgrov: Yes that is definitely the preferred solution, Amos?

zhfeng: cs.exe works both in server and client.
runServer() to start a server process with the command "cs.exe -s testsui -p 12342 -i sid" 
runTest(name) to start a client process with the command "cs.exe name"

It needs to shutdown the server by sending the SIGINT to the server process. In term of windows which has not the signal, it looks like need "Ctrl+Break" .

zhfeng: I'm searching for the windows solution but it looks like no easy way to do it.

tomjenkinson: Can't we send an admin command to the server?

zhfeng: Ah, yeah, it looks like a good idea ! we have the btadmin tool to start/stop the server.

PR #825, opened 2015-04-16
JBTM-2367 Moved ClassloadingUtility to common folder

tomjenkinson:


mmusgrov:
The PERF failure was caused by JBTM-2368 which is now fixed. It is in the perf repo so you do not need to rebase.

PR #826, opened 2015-04-17
Revert "JBTM-2356. RTS integration recovery test"

Gytis:
This reverts commit 5b278de35b9095a27e27514cca4eea51d8e231cf.


Gytis:
I don't remember what was the agreement on reviewing PRs with reverted commits. But just in case I'm assigning you, @mmusgrov

mmusgrov:
It looks like you just reverted the test but not the fix - was that intentional?

Gytis:
yes, there are two commits. One with the fix and unit test. And another with integration test.
Do you think that I should revert fix too? It's here fyi https://github.com/jbosstm/narayana/commit/530344c33714086f3fd4e3b0b3341e543f35f34b

Gytis:
I did it because, my current suspicion is that the problem is in the integration test. What do you think?

mmusgrov:
I thought you were committing a fix without a test but the other commit you refer to, namely 530344c, does contain a test for:
    // Transaction should not be removed because of the failure
in CoordinatorTest#testFailureInSecondParticipantDuringCommit. So that seems like a reasonable basic test for the fix. How about raising a new JIRA for the integration test and link it to the current JIRA.

Gytis:
But I am removing that integration test. Also Tom suggested last week to reuse JBTM-2356 for the integration test if I put it back in in the future, since it's only a test

mmusgrov:
@Gytis No that is not true you are not reverting it. The test I refer to ( CoordinatorTest#testFailureInSecondParticipantDuringCommit.) is in the same commit that has the fix (it is not in this PR that is doing the revert). But what I am saying is that this unit test will probably be good enough from a process point of view.

@tomjenkinson JBTM-2356 is the bug and Gytis is submitting a fix for that (commit 530344c) without a proper integration test. We need to be track his progress on that, hence either JBTM-2356 is left unresolved or Gytis raises a separate JIRA for the integration test.

Gytis:
ok I agree, I think new JIRA to add integration test would be good

PR #827, opened 2015-04-17
JBTM-2363 Blacktie c++ tx client build build failure

zhfeng:
NO_TEST

### DIFF
--- a/blacktie/tx/src/main/cpp/HttpTxManager.cxx
+++ b/blacktie/tx/src/main/cpp/HttpTxManager.cxx
@@ -207,7 +207,7 @@ bool HttpTxManager::recover(XAWrapper *resource)
 	if (ri.status_code == 200) {
 		LOG4CXX_DEBUG(httptxlogger, "recovery: told TM about participant "
 			<< resource->get_name() << " xid: " << xid);
-
+		_branches.erase(resource->get_name());
 		return true;
 	} else {
 		// TODO what about the other HTTP codes
@@ -379,6 +379,7 @@ bool HttpTxManager::handle_request(
 					LOG4CXX_DEBUG(httptxlogger, "Return 200 OK with body: " <<
 						status << " XA status: " << res);
 					code = 200;
+					_branches.erase(branch->get_name());
 					break;
 				case XAER_PROTO:
 					// TODO Do all PROTO errors mean rollback for example (at least with ORACLE)
--- a/blacktie/tx/src/test/cpp/TestTransactions.cxx
+++ b/blacktie/tx/src/test/cpp/TestTransactions.cxx
@@ -512,6 +512,7 @@ void TestTransactions::test_wait_for_recovery()
 {
 	int nsecs = 180;
 	int nrecs, nrecs1 = count_log_records();
+	const int SLEEP_INTERVAL = 10;
 	fault_t fault2 = {0, 102, O_XA_COMMIT, XA_OK, F_CB, (void *) recovery_cb2};
 	fault_t fault3 = {0, 102, O_XA_ROLLBACK, XA_OK, F_CB, (void *) recovery_cb2};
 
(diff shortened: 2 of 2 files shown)


blacktie/tx/src/test/cpp/TestTransactions.cxx
```diff
@@ -527,11 +527,14 @@ void TestTransactions::test_wait_for_recovery()
...
-	while ((rcCnt2 == 0 || (nrecs = count_log_records()) == nrecs1) && nsecs != 0) {
-		if (nsecs-- % 10 == 0) {
-			btlogger("TestTransactions::test_run_recovery sleeping for %d seconds", nsecs);
+	nrecs = count_log_records();
+	if(nrecs > 0 ) {
+		while ((rcCnt2 == 0 || (nrecs = count_log_records()) == nrecs1) && nsecs != 0) {
+			if (nsecs-- % 10 == 0) {
+				btlogger("TestTransactions::test_run_recovery sleeping for %d seconds", nsecs);
+			}
+			doSix(1);
```
mmusgrov: The previous test makes no sense unless you move the sleep (doSix) inside the if block

blacktie/tx/src/test/cpp/TestTransactions.cxx
```diff
@@ -527,11 +527,14 @@ void TestTransactions::test_wait_for_recovery()
 	(void) dummy_rm_add_fault(fault2);
 	(void) dummy_rm_add_fault(fault3);
 	BT_ASSERT_EQUAL(TX_OK, tx_open());
-	while ((rcCnt2 == 0 || (nrecs = count_log_records()) == nrecs1) && nsecs != 0) {
-		if (nsecs-- % 10 == 0) {
-			btlogger("TestTransactions::test_run_recovery sleeping for %d seconds", nsecs);
+	nrecs = count_log_records();
+	if(nrecs > 0 ) {
+		while ((rcCnt2 == 0 || (nrecs = count_log_records()) == nrecs1) && nsecs != 0) {
```
mmusgrov: This test is waiting for recovery to happen. The test for recovery is the negation of the first condition:
    (rcCnt2 == 0 || (nrecs = count_log_records()) == nrecs1)
I think the intent of the second condition (ie nsecs != 0) is to only wait for a maximum of nsecs seconds before giving up. But the way the if condition is written means that the loop will always wait for nsecs but the loop should terminate as soon as recovery is detected (ie when the first condition is false). Why not something similar to:

		const int SLEEP_INTERVAL = 10;
		
		for (int i = 0; i < nsecs; i += SLEEP_INTERVAL) {
			if (rcCnt2 == 0 || (nrecs = count_log_records()) == nrecs1)
			    break;
			btlogger("TestTransactions::test_run_recovery sleeping for another %d seconds", SLEEP_INTERVAL);
			doSix(SLEEP_INTERVAL)
		}

blacktie/tx/src/main/cpp/HttpTxManager.cxx
```diff
@@ -207,7 +207,7 @@ bool HttpTxManager::recover(XAWrapper *resource)
 	if (ri.status_code == 200) {
 		LOG4CXX_DEBUG(httptxlogger, "recovery: told TM about participant "
 			<< resource->get_name() << " xid: " << xid);
-
+		_branches.erase(resource->get_name());
```
mmusgrov: Well spotted.

I think we should also be removing the branch after a successful commit or abort. Do you agree?

blacktie/tx/src/test/cpp/TestTransactions.cxx
zhfeng: Thanks Mike,

the rcCnt2 > 0 means that the recovery_cb2 () has been called but it still needs to wait for removing the transaction logs. So I think the if condition should be
``` 
if ((nrecs = count_log_records()) < nrecs1)
    break;
```

blacktie/tx/src/main/cpp/HttpTxManager.cxx
zhfeng: I agree and will take a look at the codes

blacktie/tx/src/main/cpp/HttpTxManager.cxx
```diff
@@ -379,6 +379,7 @@ bool HttpTxManager::handle_request(
 					LOG4CXX_DEBUG(httptxlogger, "Return 200 OK with body: " <<
 						status << " XA status: " << res);
 					code = 200;
+					_branches.erase(branch->get_name());
```
zhfeng: I add  to remove the branch

mmusgrov: OK I think that should be enough.

blacktie/tx/src/main/cpp/HttpTxManager.cxx
mmusgrov: This change is wrong (see Amos' comment on the associated JIRA JBTM-2363)

blacktie/tx/src/main/cpp/HttpTxManager.cxx
mmusgrov: Similarly here too - this change is wrong (see Amos' comment on the associated JIRA JBTM-2363)

blacktie/tx/src/main/cpp/HttpTxManager.cxx
zhfeng: I think this change looks good

blacktie/tx/src/main/cpp/HttpTxManager.cxx
zhfeng: this change is wrong as the branch looks like it has been delete after calling the do_commit or do_rollback. I have a fix in https://github.com/jbosstm/narayana/pull/828

PR #828, opened 2015-04-22
JBTM-2363 Update to fix the bug of removing the branch

zhfeng:

### DIFF
--- a/blacktie/tx/src/main/cpp/HttpTxManager.cxx
+++ b/blacktie/tx/src/main/cpp/HttpTxManager.cxx
@@ -379,7 +379,6 @@ bool HttpTxManager::handle_request(
 					LOG4CXX_DEBUG(httptxlogger, "Return 200 OK with body: " <<
 						status << " XA status: " << res);
 					code = 200;
-					_branches.erase(branch->get_name());
 					break;
 				case XAER_PROTO:
 					// TODO Do all PROTO errors mean rollback for example (at least with ORACLE)


blacktie/tx/src/main/cpp/HttpTxManager.cxx
```diff
@@ -379,7 +376,7 @@ bool HttpTxManager::handle_request(
 					LOG4CXX_DEBUG(httptxlogger, "Return 200 OK with body: " <<
 						status << " XA status: " << res);
 					code = 200;
-					_branches.erase(branch->get_name());
+					_branches.erase(wid);
```
mmusgrov: The logic for cleaning up branches is driven from the resources (in XAWrapper using the method XAWrapper::set_complete). Please can you explain why you think that code is not working.

zhfeng: yeah, the XAResourceManager::set_complete(XAWrapper* resource)
https://github.com/jbosstm/narayana/blob/master/blacktie/tx/src/main/cpp/XAResourceManager.cxx#L328

it delete the ra and the branch->get_name() could get the wrong memory reference.

mmusgrov: So do you agree that the fix is wrong. I still think we need to find the root cause. The intent is to clean up branches from two places:

* during normal processing XAWrapper::set_complete does the clean up
* during recovery XAResourceManager::recover does the clean up

What do you mean by your second comment. Can you point me at the exact bit of code you are referring to.

XAResourceManager::recoverRTS

mmusgrov: You applied a fix to erase the branch here:
https://github.com/jbosstm/narayana/blob/master/blacktie/tx/src/main/cpp/HttpTxManager.cxx#L210
but that was not required since the branch was already cleaned here:
https://github.com/jbosstm/narayana/blob/master/blacktie/tx/src/main/cpp/XAResourceManager.cxx#L187

mmusgrov: When I wrote the recovery clean up code I see I added the following comment:
    
    // TODO have we caught all race conditions - recover can cause ra to call back
    // here via XAResourceManager::set_complete

I am just guessing but is it possible that we are seeing such a race condition?

zhfeng: no, I don't thinks so. I agree that the fix was wrong and the root cause might be the variable _branches in HttpTxManager and branches_ in XAResourceManager look like not sync. when XAResourceManager clean the branches and there were still the references in the HttpTxManager. I just wonder that why we need to have these two ?

zhfeng: https://github.com/jbosstm/narayana/blob/master/blacktie/tx/src/main/include/HttpTxManager.h#L56
```
 typedef std::map<std::string, XAWrapper*> XABranchMap;
 XABranchMap _branches;
```
https://github.com/jbosstm/narayana/blob/master/blacktie/tx/src/main/include/XAResourceManager.h#L61
```
typedef std::map<XID, XAWrapper *> XABranchMap;
XABranchMap branches_;
```
It looks like duplicated ?

mmusgrov: The one in XAResourceManager if for recovering branches and the one in HttpTxManager is for active branches. I think I separated them out because XAResourceManager is the class that drives recovery (and not HttpTxManager). If they need to be in synch then maybe we could revisit the design (but I am not sure they do need to be in synch).

You said TestTransactions::test_recovery() failed with SIGABORT - do you know which bit of code was responsible for that?

mmusgrov: Also the key types are different because HttpTxManager and XAResourceManager need to keep track of XA resources for different reasons.

zhfeng: * https://github.com/jbosstm/narayana/blob/master/blacktie/tx/src/main/cpp/XAResourceManagerFactory.cxx#L213
run_recovery after createRMs
* https://github.com/jbosstm/narayana/blob/master/blacktie/tx/src/main/cpp/XAResourceManagerFactory.cxx#L172
resource manager is running recovery
* https://github.com/jbosstm/narayana/blob/master/blacktie/tx/src/main/cpp/XAResourceManager.cxx#L183
 calling txMangaer to send the recovery to the TM
1. the TM return 404 before merging the fix of JBTM-2356 and return false
2. the TM return 200 when merging the fix of JBTM-2356 and return true. So the rm delete the branch as it has been recovered but the HttpTxManager has still contains this branch.
https://github.com/jbosstm/narayana/blob/master/blacktie/tx/src/main/cpp/XAResourceManager.cxx#L189
3. when HttpTxManager::request_handler() receive the COMMIT 
https://github.com/jbosstm/narayana/blob/master/blacktie/tx/src/main/cpp/HttpTxManager.cxx#L349
the branch pointer should have been deleted and the SIGABORT happens.

Does it clear ?

zhfeng: I propose to add the removing function in the HttpTxManager and when the XAResourceManager deletes the branch, it notifies the HttpTxManager to remove it too.
```
void HttpTxManager::remove_branch(const char * name) {
    _branches.erase(name); // does it need to add the lock ?
}
```
https://github.com/jbosstm/narayana/blob/master/blacktie/tx/src/main/cpp/XAResourceManager.cxx#L189
```
bool XAResourceManager::recoverRTS(XID& bid, const char* rc) {
...
    atmibroker::tx::TxManager::get_instance()->remove_branch(ra->get_name());
    delete ra;
}
```
https://github.com/jbosstm/narayana/blob/master/blacktie/tx/src/main/cpp/XAResourceManager.cxx#L328
```
void XAResourceManager::set_complete(XAWrapper* resource) {
...
    LOG4CXX_TRACE(xarmlogger, (char*) "removing branch");
    atmibroker::tx::TxManager::get_instance()->remove_branch(ra->get_name());
    delete ra;
}
```

zhfeng: Does it make sense ?

mmusgrov: The analysis you gave makes sense except at the third bullet. After that call your original fix will delete the branch at
https://github.com/jbosstm/narayana/blob/master/blacktie/tx/src/main/cpp/HttpTxManager.cxx#L210

mmusgrov: I think the fix at line:
https://github.com/jbosstm/narayana/blob/master/blacktie/tx/src/main/cpp/HttpTxManager.cxx#L382
is the culprit.

zhfeng: sure, I will try this fix.

mmusgrov:
Looks like the fix worked.

zhfeng:
OK, I will merge it.

PR #831, opened 2015-04-28
JBTM-2378 cleaned up some of the provided dependencies that are no longe...

tomjenkinson:
...r required
https://github.com/jbosstm/jboss-as/pull/37

### DIFF
--- a/ArjunaCore/arjuna/pom.xml
+++ b/ArjunaCore/arjuna/pom.xml
@@ -85,6 +85,7 @@
       <groupId>org.hornetq</groupId>
       <artifactId>hornetq-core</artifactId>
       <version>${version.org.hornetq}</version>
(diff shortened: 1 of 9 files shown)


Gytis:
If you don't mind I'll run tests locally to check if just merged TXFramework/Compensations fix didn't break. Will save some time from running all CI axis

Gytis:
Just one question. I've noticed that you have left Arjuna dependencies in JTA as compile scoped (same for Arjuna and JTA in JTS). Is it because they can be run standalone?

I'm asking, because I was thinking of marking them all as provided in XTS (already did it in compensations and txf).

Gytis:
Other than that looks good

Gytis:
So I guess it's ok to mark them as provided for XTS, since we do not use it standalone?

PR #834, opened 2015-04-30
JBTM-2385 Updated to treat TM::resume(null) as TM::suspend like AtomicAc...

tomjenkinson:
!PERF


ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/BaseTransactionManagerDelegate.java
```diff
@@ -116,8 +116,12 @@ public Transaction getTransaction()
     public void resume(final Transaction transaction)
         throws InvalidTransactionException, IllegalStateException, SystemException
     {
-        transactionManager.resume(transaction) ;
-        notifyAssociationListeners(transaction, EnumSet.of(EventType.ASSOCIATED));
+        if (transaction == null) {
+            suspend(); // This is what AtomicAction does
```
mmusgrov: Very odd behaviour but I agree we need to be consistent with AtomicAction

PR #835, opened 2015-04-30
[JBTM-2379] Mark TXBridge and XTS dependencies as provided where possible

Gytis:
!MAIN !BLACKTIE !QA_JTA !QA_JTS_JACORB !PERF
Requires WildFly PR: https://github.com/jbosstm/jboss-as/pull/39


Gytis:
Ok, I'll add servlet and jbossws for jbossxts instead of modifying all submodules.
As for txbridge I think jta, jbossxts, and jbossxts api should be marked as provided.

You're right about txf and compensations. They used to rely on transitive dependencies.

Gytis:
merged

PR #836, opened 2015-05-01
JBTM-2380 Updated RTS so that it does not require exclusions when embedd...

tomjenkinson:
https://github.com/jbosstm/jboss-as/pull/40

!PERF !QA_JTA !QA_JTS_JACORB !XTS !BLACKTIE

### DIFF
--- a/rts/at/tx/pom.xml
+++ b/rts/at/tx/pom.xml
@@ -62,6 +62,7 @@
       <groupId>log4j</groupId>
       <artifactId>log4j</artifactId>
       <version>${version.log4j}</version>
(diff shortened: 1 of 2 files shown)


Gytis:
I think resteasy-jaxrs and resteasy-client should be set as provided or test (I don't remember in which part of the code they are used).
Those dependencies are in both files. I checked other modules of RTS and they are ok.

Gytis:
resteasy-client should be ok as test dependency. While resteasy-jaxrs might have to be as provided

Gytis:
Looks good

PR #838, opened 2015-05-05
JBTM-2265 BlackTie Jatmibroker TestTPConversation hang

zhfeng:
!XTS !QA_JTA !QA_JTS !QA_JTS_JACORB !PERF

### DIFF
--- a/blacktie/jatmibroker-xatmi/src/main/java/org/jboss/narayana/blacktie/jatmibroker/core/transport/hybrid/stomp/StompManagement.java
+++ b/blacktie/jatmibroker-xatmi/src/main/java/org/jboss/narayana/blacktie/jatmibroker/core/transport/hybrid/stomp/StompManagement.java
@@ -7,7 +7,6 @@ import java.io.OutputStream;
 import java.net.Socket;
 import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
@@ -23,6 +22,10 @@ public class StompManagement {
     private static final Logger log = LogManager.getLogger(StompManagement.class);
 
     private static List<Socket> disconnectedConnections = new ArrayList<Socket>();
+
+    private static final byte[] COLON = {':'};
+    private static final byte[] EOL = {'\n'};
+    private static final byte[] EOM = {'\0', '\n', '\n'};
 
     public static void close(Socket socket, OutputStream outputStream, InputStream inputStream) throws IOException {
         log.debug("close");
@@ -68,26 +71,24 @@ public class StompManagement {
 
     public static void send(Message message, OutputStream outputStream) throws IOException {
         log.trace("Writing on: " + outputStream);
-        StringBuffer toSend = new StringBuffer(message.getCommand().toString());
-        toSend.append("\n");
+        synchronized (outputStream) {
+            outputStream.write(message.getCommand().getBytes());
+            outputStream.write(EOL);
 
-        Iterator<String> keys = message.getHeaders().keySet().iterator();
-        while (keys.hasNext()) {
-            String key = keys.next();
-            String value = message.getHeaders().get(key);
-            toSend.append(key);
-            toSend.append(":");
-            toSend.append(value);
-            toSend.append("\n");
+            for (Map.Entry<String, String> header : message.getHeaders().entrySet()) {
+                outputStream.write(header.getKey().getBytes());
+                outputStream.write(COLON);
+                outputStream.write(header.getValue().getBytes());
+                outputStream.write(EOL);
+            }
+            outputStream.write(EOL);
+            
+            if(message.getBody() != null) {
+                outputStream.write(message.getBody());
+            }
+            outputStream.write(EOM);
         }
-        toSend.append("\n");
-        outputStream.write(toSend.toString().getBytes());
 
-        if (message.getBody() != null) {
-            outputStream.write(message.getBody());
-        }
-
-        outputStream.write("\000\n\n".getBytes());
         log.trace("Wrote on: " + outputStream);
     }
 


blacktie/jatmibroker-xatmi/src/main/java/org/jboss/narayana/blacktie/jatmibroker/core/transport/hybrid/stomp/StompManagement.java
```diff
@@ -67,28 +67,30 @@ public static Socket connect(String host, int port, String username, String pass
...
-        toSend.append("\n");
-
-        Iterator<String> keys = message.getHeaders().keySet().iterator();
-        while (keys.hasNext()) {
-            String key = keys.next();
-            String value = message.getHeaders().get(key);
-            toSend.append(key);
-            toSend.append(":");
-            toSend.append(value);
+        synchronized (outputStream) {
```
mmusgrov: I have not studied the issue in any detail but just looking at the code changes I don't think you need to synchronize on the whole method, maybe just on the bit that writes to the socket.

In fact, if you assemble the data and do a single write I would expect OutputStream to write it out in a single go - ie I don't think you need any synchronization at all.

mmusgrov: For example the javadoc for java.nio.channels.SocketChannel says:

"Socket channels are safe for use by multiple concurrent threads. They support concurrent reading and writing, though at most one thread may be reading and at most one thread may be writing at any given time. The connect and finishConnect methods are mutually synchronized against each other, and an attempt to initiate a read or write operation while an invocation of one of these methods is in progress will block until that invocation is complete."

The javadoc for java.net.Socket isn't as definitive but I would expect/hope for something similar or you could use NIO instead.

zhfeng: Thanks to point it out and I do agree that it just needs to synchronize on the writes to the socket. I think the write on java.net.Socket is not thread-safe so even if we assemble the data and do a single write, it could not guarantee an atomic operation. The NIO SocketChannel is great and I think it needs to change more codes. So it would be better to do in the other JIRA.

mmusgrov: Why can't you use

    java.net.socket.getChannel().write()

which probably requires fewer changes than you are making in the current PR.

mmusgrov: I also think it is best to assemble the data into the buffer and then write it to the socket in one go. So using getChannel and assembling the data before writing we would have:

        ByteBuffer toSend = ByteBuffer.wrap(message.getCommand().getBytes());

        if (message.getBody() != null)
            toSend.put(message.getBody());
     
        toSend.put("\000\n\n".getBytes()); 
       
        socket.getChannel().write(toSend);

Implementing in this way avoids the need for synchronization.

mmusgrov:
See my previous comment

zhfeng:
Good idea ! I will take a look at it.

blacktie/jatmibroker-xatmi/src/main/java/org/jboss/narayana/blacktie/jatmibroker/core/transport/hybrid/stomp/StompManagement.java
```diff
@@ -81,14 +81,17 @@ public static void send(Message message, OutputStream outputStream) throws IOExc
             toSend.append("\n");
         }
         toSend.append("\n");
-        outputStream.write(toSend.toString().getBytes());
+        synchronized (outputStream) {
+            outputStream.write(toSend.toString().getBytes());
```
mmusgrov: This was the comment I made on the previous outdated diff:

I also think it is best to assemble the data into the buffer and then write it to the socket in one go. So using getChannel and assembling the data before writing we would have:

    ByteBuffer toSend = ByteBuffer.wrap(message.getCommand().getBytes());

    if (message.getBody() != null)
        toSend.put(message.getBody());

    toSend.put("\000\n\n".getBytes()); 

    socket.getChannel().write(toSend);
Implementing in this way avoids the need for synchronization.

blacktie/jatmibroker-xatmi/src/main/java/org/jboss/narayana/blacktie/jatmibroker/core/transport/hybrid/stomp/StompManagement.java
```diff
@@ -81,14 +81,17 @@ public static void send(Message message, OutputStream outputStream) throws IOExc
...
 
-        if (message.getBody() != null) {
-            outputStream.write(message.getBody());
-        }
+            if (message.getBody() != null) {
+                outputStream.write(message.getBody());
+            }
 
-        outputStream.write("\000\n\n".getBytes());
+            outputStream.write("\000\n\n".getBytes());
```
mmusgrov: If we are sending "\000\n\n" a lot of times it may be beneficial to pull it out as a constant:

    static final byte[] eom =  "\000\n\n".getBytes()

blacktie/jatmibroker-xatmi/src/main/java/org/jboss/narayana/blacktie/jatmibroker/core/transport/hybrid/stomp/StompManagement.java
zhfeng:
```
socket.getChannel().write(toSend); 
```
This throws the Null Exception and I think it's causes by we just create the socket in blocking mode. see StompMangement.connect()

```
Socket socket = new Socket(host, port);
InputStream inputStream = socket.getInputStream();
OutputStream outputStream = socket.getOutputStream();
```
I'm trying to do 

```
SocketChannel channel=SocketChannel.open();
channel.connect(new InetSocketAddress(host,port));
channel.finishConnect();

Socket socket = channel.socket();
```
It looks good but the SompManagement.receive() now is blocking on reading from the input stream. I think it could also need to change the receive method by using the nio channel and this would require more code changes. So I agree assemble the data and synchronize the write operation.

blacktie/jatmibroker-xatmi/src/main/java/org/jboss/narayana/blacktie/jatmibroker/core/transport/hybrid/stomp/StompManagement.java
zhfeng: It looks good to me.

zhfeng:
I'm sorry that it looks like a wrong in assemble the data and I will try to use the SocketChannel solution.

blacktie/jatmibroker-xatmi/src/main/java/org/jboss/narayana/blacktie/jatmibroker/core/transport/hybrid/stomp/StompManagement.java
zhfeng: There could be a wrong in assemble the data and I will try to fix it.

mmusgrov: OK understood, this approach requires more changes than I anticipated:

> A socket will have a channel if, and only if, the channel itself was created via the SocketChannel.open or 
> ServerSocketChannel.accept methods.

So, yes, I agree with your original approach of synchronizing on the stream.

blacktie/jatmibroker-xatmi/src/main/java/org/jboss/narayana/blacktie/jatmibroker/core/transport/hybrid/stomp/StompManagement.java
mmusgrov: So what do you think of assembling all the data into toSend and then synchronize on a single outputStream.write?

zhfeng: I agree and I'm working on this.

blacktie/jatmibroker-xatmi/src/main/java/org/jboss/narayana/blacktie/jatmibroker/core/transport/hybrid/stomp/StompManagement.java
```diff
@@ -5,6 +5,7 @@
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.Socket;
+import java.nio.ByteBuffer;
```
mmusgrov: These changes seem to be a hybrid of using NIO and your original approach. I was expecting something like:

https://github.com/mmusgrov/narayana/commit/d9f23e80986ffb0638d974c2a6d6f9b0a48eae6f

mmusgrov:
@tomjenkinson Nice, I am impressed with the speed with which you updated 109 files

blacktie/jatmibroker-xatmi/src/main/java/org/jboss/narayana/blacktie/jatmibroker/core/transport/hybrid/stomp/StompManagement.java
zhfeng: Thanks Mike, please see my comments on your commit.

mmusgrov: OK, so taking into account your comments on my commit, what do you think about my approach?

zhfeng: yeah, it looks good to me. I will update the PR with these changes.

zhfeng: There are two test failures with these changes.
TestTPCall.test_tpcall_x_c_type() and TestTPCall.test_tpcall_x_common().

I think it might be related to appending the message.body. The two tests allocate a struct buffer.
```
public void test_tpcall_x_common()  {
        X_COMMON dptr = (X_COMMON) connection.tpalloc("X_COMMON", "deposit");

        dptr.setLong("acct_no", 12345678);
        dptr.setShort("amount", (short) 50);
}
public void test_tpcall_x_c_type() {
        X_C_TYPE aptr = (X_C_TYPE) connection.tpalloc("X_C_TYPE", "acct_info");

        aptr.setLong("acct_no", 12345678);
        aptr.setByteArray("name", "TOM".getBytes());

        float[] foo = new float[2];
        foo[0] = 1.1F;
        foo[1] = 2.2F;
        aptr.setFloatArray("foo", foo);

        double[] balances = new double[2];
        balances[0] = 1.1;
        balances[1] = 2.2;
        aptr.setDoubleArray("balances", balances);
}
```

SO I think it makes sense to use the ByteBuffer rather than StringBuffer to assemble the data.
How do you think about this ?

mmusgrov: Maybe but it seems very wasteful (and inefficient) copying all the data into one array and then copying it into another before sending. So I would prefer one of the following three options:

1. Copy the data into one array and send that down the OutputStream in  a single synchronized write. I realise that two tests failed with this approach. I wonder if there is any befefit in debugging why).
2. Go with your original approach and perform the output in multiple writes and enclose them in a synchronized block.
3. Use a SocketChannel and do not use synchronization (I realise this is a larger change so probably isn't an option at the moment).

@tomjenkinson do you have an opinion on which option to implement - this is an often used routine so it needs to perform well.

tomjenkinson: I think I would take a look in the debugger at the content of the byte array you are sending before and after patch to see if you can work out why the byte[] is different.

If making the change just that syncing the outputStream would work I would consider that too initially. We haven't looked into.

Finally. does _not_ allocating a StringBuffer and just allocating a bytebuffer (which you need to do anyway) and copying in to that work?

Last observation is that my inspection of the patch also concurs it looks functionally equivalent. I wonder if there is a bug in the test that is exposed by closing this threading issue?

zhfeng: There are two threads intend to write into the socket when closing the service dispatcher.
http://albany.eng.hst.ams2.redhat.com/view/Narayana/job/narayana/814/PROFILE=BLACKTIE,jdk=jdk7.latest,label=linux64el6/artifact/blacktie/jatmibroker-xatmi/target/test-classes/blacktie.log
```
2015-04-29 04:39:39,986 [ConvService-Dispatcher-0]           TRACE (server.ServiceDispatcher :72  ) - Processed
2015-04-29 04:39:39,986 [ConvService-Dispatcher-0]           DEBUG (transport.Message        :48  ) - Acking message: ID:588db268-ee21-11e4-bf70-4fe43b4f0ed4
2015-04-29 04:39:39,986 [ConvService-Dispatcher-0]           TRACE (stomp.StompManagement    :70  ) - Writing on: java.net.SocketOutputStream@57c9af4d
...
2015-04-29 04:39:39,989 [main]           TRACE (server.ServiceDispatcher :116 ) - closing: ConvService-Dispatcher-0
2015-04-29 04:39:39,989 [main]           TRACE (server.ServiceDispatcher :118 ) - Closing receiver
2015-04-29 04:39:39,989 [main]           DEBUG (stomp.StompReceiverImpl  :131 ) - close
2015-04-29 04:39:39,989 [main]           DEBUG (stomp.StompReceiverImpl  :136 ) - closing socket: Socket[addr=localhost/127.0.0.1,port=61613,localport=60114]
2015-04-29 04:39:39,989 [main]           DEBUG (stomp.StompManagement    :28  ) - close
2015-04-29 04:39:39,989 [main]           TRACE (stomp.StompManagement    :70  ) - Writing on: java.net.SocketOutputStream@57c9af4d
2015-04-29 04:39:39,990 [main]           TRACE (stomp.StompManagement    :91  ) - Wrote on: java.net.SocketOutputStream@57c9af4d
2015-04-29 04:39:39,990 [main]           DEBUG (stomp.StompManagement    :35  ) - Sent disconnect
2015-04-29 04:39:39,990 [main]           TRACE (stomp.StompManagement    :96  ) - Reading from: java.net.SocketInputStream@be83fed
2015-04-29 04:39:39,990 [ConvService-Dispatcher-0]           TRACE (stomp.StompManagement    :91  ) - Wrote on: java.net.SocketOutputStream@57c9af4d
2015-04-29 04:39:39,990 [ConvService-Dispatcher-0]           DEBUG (transport.Message        :55  ) - Acked message: ID:588db268-ee21-11e4-bf70-4fe43b4f0ed4
2015-04-29 04:39:39,990 [ConvService-Dispatcher-0]           DEBUG (server.ServiceDispatcher :103 ) - Close the thread
```
It looks like the mess stomp messages sending and read from the socket is hanging as the server does not know these messages. So I agree to assemble all the data and synchronize in one write operation. Option 1 looks good and it needs to use ByteBuffer rather than StringBuffer to assemble the data. Option 2 is less code changes. Option 3 is a large change as Mike said.

mmusgrov: So is you plan to implement option 1 (without the need to move data between buffers)?

zhfeng: yeah, I plan to use the bytebuffer to copy all the data.

zhfeng:
The ByteBuffer only supports the fixed size by using
```
ByteBuffer toSend = ByteBuffer.allocate(capability);
```
but there is the dynamical headers in the message and it needs a extra loop to calculate the size of headers such like 
```
int capability = message.getCommand().length() + 1;
for (Map.Entry<String, String> entry : message.getHeaders().entrySet()) {
     capability += entry.getKey().length() + entry.getValue().length() + 2;
}
capability += 1;
capability += message.getBody() == null ? 0 : message.getBody().length;
capability += EOM.length;

ByteBuffer toSend = ByteBuffer.allocate(capability);

toSend.put(message.getCommand().toString().getBytes());
toSend.put(EOL);

for (Map.Entry<String, String> entry : message.getHeaders().entrySet()) {
    toSend.put(entry.getKey().getBytes());
    toSend.put(":".getBytes());
    toSend.put(entry.getValue().getBytes());
    toSend.put(EOL);
}
toSend.put(EOL);

if (message.getBody() != null) {
    toSend.put(message.getBody());
}

toSend.put(EOM);
synchronized (outputStream) {
    outputStream.write(toSend.array());
}
``` 

how do you think about this ?

zhfeng:
The two failures due to use the StringBuffer to append the message body. So it looks good to use ByteBuffer but it needs a extra loop to calculate the headers (key + value) size. Another way is to use ByteArrayOutputStream in https://github.com/zhfeng/narayana/commit/6af8915056d729a20f9ab0b2d6538ed1037d9a75 to assemble the data, Mike thinks it could be bad to copy a new array.
```
synchronized (outputStream) {
    //outputStream.write(toSend.toByteArray());  //it create a newly byte array
    toSend.writeTo(outputStream); //this could be better as it does not copy the array
}
```

does it make sense ?

mmusgrov:
Yes writeTo is perfect. Thanks

mmusgrov:
I lot of the discussion on this PR arose because of the need to take a synchronization lock around multiple writes to the OutputStream which I deemed inefficient. I know I said using writeTo looks good but I have just looked at the source for ByteArrayOutputStream and all of its write methods are synchronized so this is an even worse solution than the original proposal. 

I still think we can improve the implementation - all we need is an efficient way of filling a byte array.

mmusgrov:
ByteBuffers are designed with efficiency in mind. It seems that it does not directly allocate the memory at creation time. If this is how it works then one possibility is to set the capacity to a very large value and check that you are not exceeding the capacity as you write into it.

mmusgrov:
I took a more detailed look at ByteBuffer and it too ends up either doing an array copy or in the case of ByteBuffer.allocateDirect() it uses native memory and so does not expose the backing byte array.

So it looks like your original approach is best (augmented with some static final byte[] constants for newline, colon and end of message sequence).

mmusgrov:
To clarify when I say your original approach I mean the one where you bracketed all the output stream writes inside a synchronized block:

    // string characters are usually 2 bytes each - can we use single bytes here or is it defined by
    // the stomp protocol. If not then we should be explicitly setting the encoding to match the protocol
    static final byte[] NEWLN =  "\n".getBytes();
    static final byte[] COLON =  ":".getBytes();
    static final byte[] EOM = "\000\n\n".getBytes();

    public void send(Message message, OutputStream outputStream) throws IOException {
        synchronized (outputStream) {
            outputStream.write(message.getCommand().getBytes());
            outputStream.write(NEWLN);

            for (Map.Entry<String, String> header : message.getHeaders().entrySet()) {
                outputStream.write(header.getKey().getBytes());
                outputStream.write(COLON);
                outputStream.write(header.getValue().getBytes());
                outputStream.write(NEWLN);
            }

            outputStream.write(NEWLN);
            outputStream.write(message.getBody());
            outputStream.write(EOM);
        }
    }

mmusgrov:
Wait, I checked where we are using StompManagement.send() and each usage is from synchronized methods in StompSenderImpl:

  public synchronized void send(...)
  public synchronized void close(...)

so I don't see how synchronizing on outputStream will help unless I have missed some other paths into the StompManagement.send(). Are you aware of other ways of using the socket (other than from StompManagement) that could cause of the interleaving of messages.

mmusgrov:
StompReceiverImpl is the other caller of StompManagement.send but it uses a different OutputStream object from the one used in StompSenderImpl so synchronizing on it will not protect against interleaving writes to the socket. So (since you have discounted the use of socket channels) you will need to synchronize the whole method instead.

zhfeng:
Message.ack() is also calling StompManagement.send and it uses the outputStream setting by the StompReceiverImpl.receive(). This issue happens between Message.ack and StompReciverImpl.close which are calling StompManagement.send at the same time. So it looks like synchronize on the outputStream is enough.

zhfeng:
In term of the static string characters, see the stomp specification 
https://stomp.github.io/stomp-specification-1.2.html#Augmented_BNF
```
NULL                = <US-ASCII null (octet 0)>
LF                  = <US-ASCII line feed (aka newline) (octet 10)>
CR                  = <US-ASCII carriage return (octet 13)>
EOL                 = [CR] LF 
OCTET               = <any 8-bit sequence of data>
```

zhfeng:
So can we define these three bytes as
```
static final byte[] NEWLN = {'\n'};
static final byte[] COLON = {':'};
static final byte[] EOM= {'\0', '\n', '\n'};
```

mmusgrov:
Please ignore the failure in the JACOCO profile since it should not have been included in the btny-pulls-narayana job. I have now removed that axis.

mmusgrov:
And the failure on the wildfly RTS integration tests: 

  MAIN profile tests failed (http://albany.eng.hst.ams2.redhat.com/job/btny-pulls-  narayana/PROFILE=MAIN,jdk=jdk8.latest,label=linux/1076/): RTS AS Integration Test failed

is caused by a bug I introduced whilst fixing the codeCoverage job failure JBTM-2404

mmusgrov:
Thanks for the link to the spec which says the encoding is us-ascii which is a subset of UTF-8. So your suggestion of defining:

    static final byte[] NEWLN = {'\n'};
    static final byte[] COLON = {':'};
    static final byte[] EOM= {'\0', '\n', '\n'};

looks good.

I do have an observation: the spec says that the message body and headers are sequences of octets so the sender and receiver need to agree on the character encoding. The code in StompManagement.send uses the String.getBytes() method and the javadoc for that says it uses "the platform's default charset" ie sender and receiver need to agree on this character set. I wonder if things would work if, for example, a blacktie client running in an Arabic country talks to a blacktie service running in the US.

I think it is probably overkill for us to define the character set on these getBytes("char set") calls but it is something to be aware of if we ever do come across bugs which look like messages being corrupted.

mmusgrov:
The code changes look good but I think you need to rebase your branch to get the latest jboss-as.version change.

PR #840, opened 2015-05-06
[JBTM-2213] Minimize deprecation warnings in txframework

Gytis:
!PERF !BLACKTIE !MAIN !QA_JTA !QA_JTS_JACORB

### DIFF
--- a/txframework/pom.xml
+++ b/txframework/pom.xml
@@ -76,6 +76,13 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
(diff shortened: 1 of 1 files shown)


mmusgrov:
I like your fix but how will our users figure out what is and what isn't deprecated.

Gytis:
This should only minimize warnings when building Narayana. It shouldn't influence user software builds. Also, it doesn't completely remove warnings, but minimizes from ~50-100 lines to 3-5 lines.

mmusgrov:
Yes

Gytis:
merged

PR #843, opened 2015-05-12
Drop old ORB references during reload

tadamski:


mmusgrov:
The spurious performance regression is unrelated to your changes and can safely be ignored.

Thanks for the fix, when you are ready can you add the label "Ready For Review".

mmusgrov:
Tomasz, I will merge your fix when you are ready

tadamski:
I have added the Jiras to the commit name. Fix is ready to merge. When it will be released?

PR #845, opened 2015-05-13
JBTM-2404 Include a JACOCO profile

mmusgrov:
JACOCO

https://issues.jboss.org/browse/JBTM-2404


mmusgrov:
Unable to test via a PR since the update is to scripts/hudson/narayana.sh
I will raise another one to test that the new profile actually works.

PR #849, opened 2015-05-15
JBTM-2413 use vm_stat on OS X builds

mmusgrov:
NO_TEST


mmusgrov:
Closing because I fixed the same issue in JBTM-2369

PR #853, opened 2015-05-20
JBTM-2418 Check for null in BaseTransactionManagerDelegate

mmusgrov:
MAIN


mmusgrov:
The first run of the MAIN profile ran on the mac and failed for an unrelated reason (malloc error - pointer being freed was not allocated). I re-ran it manually on the same mac and it passed.
The second run of the MAIN profile ran on linux and passed.

I therefore regarding it as a spurious error unrelated to my commit.

mmusgrov:
The public interface to register listeners in is the spi. I do mention there that transaction cannot be null. The private method is implementation so only of interest to developers who should be aware of the requirements (especially now that I have written tests for it). 

public interface TransactionListenerRegistry {
    /**
     * @param transaction the transaction that the caller is interested in receiving events for (must not be null)
     * @param listener an object that will be invoked when events of type {@link EventType}
     *                 occur
     * @param types a collection of events that the listener is interested in
     * @throws TransactionTypeNotSupported if the passed in transaction type does not support the listeners feature
     */
    void addListener (Transaction transaction, TransactionListener listener, EnumSet<EventType> types) throws TransactionTypeNotSupported;
}

PR #855, opened 2015-05-28
JBTM-1339 JBTM-2423 New profile to use openjdk-orb

mmusgrov:
!BLACKTIE !PERF !XTS !QA_JTA !QA_JTS_JACORB

The changes are twofold:
- I add a jar to the bootclasspath for testing a particular profile - ie it is isolated and only kicks in during surefire tests.
- secondly, when we run an orb we do some clean up after the run loop completes and that is not (CORBA) spec compliant so I removed that code. This meant there was a risk that destroy on the root poa would not get called so I added that check to the shutdown method on our ORB portability layer wrapper.

### DIFF
--- a/ArjunaJTS/idl/pom.xml
+++ b/ArjunaJTS/idl/pom.xml
@@ -69,6 +69,17 @@
       </modules>
     </profile>
     <profile>
+      <id>jts-openjdk</id>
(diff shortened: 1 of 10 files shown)


nmcl:
Looks ok to me.

PR #870, opened 2015-06-15
JBTM-2398 New volatile stores for uid lookup via type

mmusgrov:
!XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !PERF !BLACKTIE

### DIFF
--- a/ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/ObjectStoreTest.java
+++ b/ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/ObjectStoreTest.java
@@ -31,10 +31,16 @@ package com.hp.mwtests.ts.arjuna.objectstore;
  * $Id: ObjectStoreTest.java 2342 2006-03-30 13:06:17Z  $
  */
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
 
+import com.arjuna.ats.internal.arjuna.common.UidHelper;
 import org.junit.Test;
 
 import com.arjuna.ats.arjuna.common.ObjectStoreEnvironmentBean;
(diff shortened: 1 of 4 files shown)


ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/ObjectStoreTest.java
```diff
@@ -49,15 +55,6 @@
 import com.arjuna.ats.arjuna.state.InputObjectState;
 import com.arjuna.ats.arjuna.state.OutputObjectState;
 import com.arjuna.ats.arjuna.utils.FileLock;
-import com.arjuna.ats.internal.arjuna.objectstore.ActionStore;
```
tomjenkinson: Any reason to remove the individual imports and replace with *?

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/TypedVolatileStore.java
```diff
@@ -0,0 +1,115 @@
...
+
+    @Override
+    public String getStoreName() {
+        return TypedVolatileStore.class.getName();
+    }
+
+    @Override
+    public boolean allTypes(InputObjectState buff) throws ObjectStoreException {
+        Set<String> types = new HashSet<>(stateTypes.values());
+        OutputObjectState store = new OutputObjectState();;
```
tomjenkinson: ";;"

mmusgrov:
I am going to benchmark this current implementation with one that adds allObjUids support directly to the existing volatile store implementations.

mmusgrov:
I ran the benchmarks and updated the associated JIRA with the results: the performance of the new stores are significantly different thus warranting the introduction of sub types.

mmusgrov:
OK a config switch is a good compromise.

nmcl:
Playing Devil's Advocate: what if some implementation wants to instantiate two instances, one which has the new allObjUids and one which doesn't? A config option would apply to the entire running JVM.

mmusgrov:
@nmcl The Volatile stores take a config bean in the constructor:

    public VolatileStore(ObjectStoreEnvironmentBean objectStoreEnvironmentBean) 

so provided different instances of ObjectStoreEnvironmentBean are passed into the constructors it is possible to have both stores configured in the same JVM. But I will add a junit test to verify that.

nmcl:
I'm ok.

PR #871, opened 2015-06-16
JBTM-2308 fix windows path issue

mmusgrov:
!XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !PERF !BLACKTIE

### DIFF
--- a/ArjunaJTS/jtax/tests/classes/com/hp/mwtests/ts/jta/jts/tools/JTSOSBTestBase.java
+++ b/ArjunaJTS/jtax/tests/classes/com/hp/mwtests/ts/jta/jts/tools/JTSOSBTestBase.java
@@ -43,6 +43,7 @@ import org.junit.BeforeClass;
 import org.junit.AfterClass;
 
 import javax.management.*;
+import java.io.File;
 import java.io.IOException;
 import java.util.*;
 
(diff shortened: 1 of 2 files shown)


ArjunaJTS/jtax/tests/classes/com/hp/mwtests/ts/jta/jts/tools/JTSOSBTestBase.java
```diff
@@ -132,20 +144,27 @@ protected void assertBeanWasCreated(ArjunaTransactionImple txn) {
 
         MBeanServer mbs = JMXServer.getAgent().getServer();
 
+        showAllMBeans(mbs);
+
 		try {
-			String type = txn.type();
+			String type = txn.type().replace(File.separator, "/");
+
 			if (type.charAt(0) == '/')
```
gytis: I think you should compare with File.separator instead of '/'

mmusgrov: No. On the previous line I have called replace(File.separator, "/") so type will only contain '/' characters.

gytis: Ah got you, was too lazy to check the syntax of replace... I thought you're changing "/" to File.separator.

PR #877, opened 2015-06-24
JBTM-2437 Update to replace 'activemq-ra' with 'hornetq-ra'

zhfeng:


mmusgrov:
hornetq has donated their code base to apache so you need to be using activemq-ra rather than hornetq-ra

mmusgrov:
@tomjenkinson OK

PR #883, opened 2015-07-07
JBTM-2440 profile to disable BMUnitRunner tests with ibmjdk

mmusgrov:
NO_TEST


mmusgrov:
Closing since the NO_TEST profile is ignored on windows and the relevant module has successfully build (and tested)

PR #889, opened 2015-07-22
JBTM-2464 replace hornetq-journal with artemis equivalent

mmusgrov:


mmusgrov:
Hi Gytis

I have assigned the reviewer even though the TXBRIDGE TESTS failed since that is unrelated to the name change of the hornetq-journal. 

The perf profile also failed for an unrelated reason (I raised  JBTM-2465 to cover that)

gytis:
@mmusgrov , was the txbridge test hanging?

gytis:
@mmusgrov , maybe we should look into renaming classes and packages since HornetqJournalStore class is not valid any more?

mmusgrov:
@gytis It was the txbridge hanging.

HornetqJournalStore is still valid. I am not going to rename class files since that will impact external users and as such we would need to depreciate it for at least 2 releases before removing it.

gytis:
@tomjenkinson , I agree. Name should be more generic in case the journal provider changes again in the future.

gytis:
Anyway, I think it is going to be confusing in the future, since HornetQ is not available any more but our store is still called that way.

mmusgrov:
I disagree - it's just a name for another store implementation. I see no benefit in deprecating and introducing another name.

mmusgrov:
You could apply the same argument to the package name arjuna and I do not sense any desire to change that.

gytis:
There is no benefit in functionality for sure. What I mean it might be clear now what it is, but for somebody in the future might not make any sense why is it called that way. At the end of the day, shouldn't the name be descriptive?

mmusgrov:
I never heard any complaints about the HornetqStore name before. The name has never said anything about what the user can expect from the store - as I keep saying is just a name.

Our users already know the HornetqStore is a fast store-  why would they suddenly become confused.

Now if we were in the market of branding our store names and wished it to be associated with the Artemis brand then you might have a point. But we aren't.

gytis:
I'm not suggesting to use Artemis name, but use more generic one like JournalStore as Tom suggested. Because in my opinion current name suggests that this store uses HornetQ which it doesn't any more.

Also, I'm not suggesting to rename everything now. I understand the complication it brings as well as the deprecation rules. Just trying to put it there that it would be good to rename it once it is possible.

mmusgrov:
But why the sudden concern now - HornetqStore never described the characteristics of the store so any arguments for a name change would have been valid from the outset. There are lots of examples where I think we could improve the names of classes and packages but it would be madness to go about renaming them all.

gytis:
Well, the store was implemented using Hornetq journal, thus the name made sense. Now the Hornetq journal isn't there any more so the name is misleading.
Also, it's more of the observation rather than concern.

mmusgrov:
OK, so my nth reasonable argument is that as a user of narayana it would definitely vex me to have to update code based on such a trivial justification.

gytis:
Fair enough and I under stand that the reason is trivial. But at the same time all hornetq store classes are in the internal package, which user shouldn't be using directly in the first place.
The only annoying bit for the user to change would be configuration.

PR #890, opened 2015-07-31
Start embedded RM if -recovery is passed to transaction sever

gytis:
https://issues.jboss.org/browse/JBTM-2470
NO_TEST

### DIFF
--- a/ArjunaJTS/jts/classes/com/arjuna/ats/jts/TransactionServer.java
+++ b/ArjunaJTS/jts/classes/com/arjuna/ats/jts/TransactionServer.java
@@ -31,6 +31,7 @@
 
 package com.arjuna.ats.jts;
 
+import com.arjuna.ats.arjuna.recovery.RecoveryManager;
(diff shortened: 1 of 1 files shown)


ArjunaJTS/jts/classes/com/arjuna/ats/jts/TransactionServer.java
```diff
@@ -109,6 +110,10 @@ static void doWork (String[] args, boolean exitOnComplete) throws Exception
 		else
 		    System.exit(0);
 	    }
+
+            if (args[i].compareTo("-recovery") == 0) {
+                RecoveryManager.manager().startRecoveryManagerThread();
```
mmusgrov: You should update the help command:
    System.out.println("Usage: [-otsname <name>] [-help] [-version]");

gytis: Good point, didn't think of that one.

PR #901, opened 2015-08-17
JBTM-2485 Updated to use the nexus plugin for deployment

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2485

!QA_JTA !QA_JTS_JACORB !PERF

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -22,11 +22,8 @@
     <plugins>
       <plugin>
         <inherited>false</inherited>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-deploy-plugin</artifactId>
-        <configuration>
-          <skip>false</skip>
-        </configuration>
+        <groupId>org.sonatype.plugins</groupId>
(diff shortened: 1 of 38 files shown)


common/pom.xml
```diff
@@ -8,7 +8,7 @@
   <parent>
     <groupId>org.jboss.narayana</groupId>
     <artifactId>narayana-all</artifactId>
-    <version>5.2.3.Final-SNAPSHOT</version>
+    <version>foo2</version>
```
gytis: I didn't know we've changed the versioning pattern :)

pom.xml
```diff
@@ -22,7 +22,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.jboss.narayana</groupId>
   <artifactId>narayana-all</artifactId>
-  <version>5.2.3.Final-SNAPSHOT</version>
+  <version>foo2</version>
```
gytis: Same as above

common/pom.xml
tomjenkinson: whoops! fixed (and below) and will reassign back

PR #903, opened 2015-08-24
Updated to delay lookups

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2499

!XTS !QA_JTA !QA_JTS_JACORB !MAIN !PERF


zhfeng:
The changes look good to me but I think the root cause should be the StompConnectService has to be started after the ActiveMQService. I'm taking a look at the codes to if it is possible to add the dependency of ActiveMQService when installing the StompConnectService.

zhfeng:
I update the https://github.com/jbosstm/narayana/pull/904 to add the dependency.

PR #910, opened 2015-09-14
JBTM-2507 Disable XTS CR test on IBM JDK

mmusgrov:
!BLACKTIE !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !MAIN

https://issues.jboss.org/browse/JBTM-2507

Skip XTS crash rec tests when running with the IBM java compiler


mmusgrov:
Gytis,

I think the PR test runs are superfluous since the fix only kicks in when running on the IBM compiler so I am going to kill the jobs but I would still appreciate a review of the changes

PR #911, opened 2015-09-16
JBTM-2514 Invoke orb shutdown when stoping the txn service

mmusgrov:
!XTS !BLACKTIE !QA_JTA  !QA_JTS_JACORB !PERF 

https://issues.jboss.org/browse/JBTM-2514

PR #913, opened 2015-09-17
JBTM-2514 Do not shutdown external orb

mmusgrov:
!XTS !BLACKTIE !QA_JTA !QA_JTS_JACORB !PERF

https://issues.jboss.org/browse/JBTM-2514

This PR follows on from https://github.com/jbosstm/narayana/pull/911 since we also need to avoid shutting down the external orb when shutting down our own

### DIFF
--- a/ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/jts/ExternalORBWrapper.java
+++ b/ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/jts/ExternalORBWrapper.java
@@ -1,0 +1,202 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2015, Red Hat Middleware LLC, and individual contributors
+ * as indicated by the @author tags.
+ * See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ * This copyrighted material is made available to anyone wishing to use,
+ * modify, copy, or redistribute it subject to the terms and conditions
+ * of the GNU Lesser General Public License, v. 2.1.
+ * This program is distributed in the hope that it will be useful, but WITHOUT A
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+ * PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more details.
+ * You should have received a copy of the GNU Lesser General Public License,
+ * v.2.1 along with this distribution; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA  02110-1301, USA.
+ *
+ * @author JBoss Inc.
+ */
+package com.arjuna.ats.jbossatx.jts;
+
+import org.omg.CORBA.NVList;
+import org.omg.CORBA.*;
+import org.omg.CORBA.ORBPackage.InvalidName;
+
+import java.util.Properties;
+import java.applet.Applet;
+
+/**
+ * Wrapper for external orbs to avoid shutting them down when narayana shuts down its own orbs.
+ * External orbs are passed into the following methods:
+ * {@link TransactionManagerService#start} and {@ink RecoveryManagerService#start}
+ */
+class ExternalORBWrapper extends org.omg.CORBA.ORB {
+    private final ORB orb;
+
+    public ExternalORBWrapper(ORB orb) {
+        this.orb = orb;
+    }
+    protected void set_parameters(String[] args, Properties props) {
+        ;
+    }
+    protected void set_parameters(Applet app, Properties props) {
+        ;
+    }
+    public void connect(org.omg.CORBA.Object obj) {
+        orb.connect(obj);
+    }
+    public void destroy( ) {
+        // do not delegate destroy
+    }
+    public void disconnect(org.omg.CORBA.Object obj) {
+        orb.disconnect(obj);
+    }
+    public String[] list_initial_services() {
+        return orb.list_initial_services();
+    }
+    public org.omg.CORBA.Object resolve_initial_references(String object_name) throws InvalidName {
+        return orb.resolve_initial_references(object_name);
+    }
+    public String object_to_string(org.omg.CORBA.Object obj) {
+        return orb.object_to_string(obj);
+    }
+    public org.omg.CORBA.Object string_to_object(String str) {
+        return orb.string_to_object(str);
+    }
+    public NVList create_list(int count) {
+        return orb.create_list(count);
+    }
+    public NVList create_operation_list(org.omg.CORBA.Object oper) {
+        return orb.create_operation_list(oper);
+    }
+    public NamedValue create_named_value(String s, Any any, int flags) {
+        return orb.create_named_value(s, any, flags);
+    }
+    public ExceptionList create_exception_list() {
+        return orb.create_exception_list();
+    }
+    public ContextList create_context_list() {
+        return orb.create_context_list();
+    }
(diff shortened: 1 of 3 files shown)


mmusgrov:
The unit tests passed. The failure is due to "Deployment Cannot deploy: test.war" which we know about so please can you review the second part of the fix.

ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/jts/ORBWrapper.java
```diff
@@ -0,0 +1,209 @@
...
+        ;
+    }
+    protected void set_parameters(Applet app, Properties props) {
+        ;
+    }
+    public void connect(org.omg.CORBA.Object obj) {
+        orb.connect(obj);
+    }
+    public void destroy( ) {
+//        orb.destroy();
```
tomjenkinson: how do you actually destroy the orb? same question for shutdown really.

ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/jts/ORBWrapper.java
```diff
@@ -0,0 +1,209 @@
...
+        return ORB.init();
+    }
+    public static ORB init(String[] args, Properties props) {
+        return ORB.init(args, props);
+    }
+    public static ORB init(Applet app, Properties props) {
+        return ORB.init(app, props);
+    }
+    protected void set_parameters(String[] args, Properties props) {
+        ;
```
tomjenkinson: why no passthrough? Might be worth adding a comment

ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/jts/RecoveryManagerService.java
```diff
@@ -54,11 +54,18 @@ public RecoveryManagerService(org.omg.CORBA.ORB theCorbaORB) throws Exception
...
 
         ORBManager.setORB(orb);
         ORBManager.setPOA(oa);
     }
+
+    @Override
+    public void stop() throws Exception {
+        super.stop();
+
+        ORB.getInstance("jboss-atx").shutdown();
```
tomjenkinson: is it always that ORB name?

ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/jts/ORBWrapper.java
mmusgrov: The point of the fix is that when we are supplied the orb by an external subsytem we do not destroy or shut it down - that is the job of whoever supplied us with the orb (these changes only apply to TransactionManagerService in the integration classes)

ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/jts/ORBWrapper.java
mmusgrov: These are static methods so no need to pass through

ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/jts/RecoveryManagerService.java
mmusgrov: It must be the same one as was constructed in the start method so I guess we could do with a final field for that.

ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/jts/ORBWrapper.java
tomjenkinson: Do they even need defining then? and they don't look static

ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/jts/RecoveryManagerService.java
tomjenkinson: I am with you thanks - I should have checked.

ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/jts/ORBWrapper.java
tomjenkinson: Who is calling destroy at the wrong point then, is it us? Why is destroy ever called at the wrong point?

tomjenkinson: Just trying to understand and make sure its not possible that an orb can be left running

ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/jts/ORBWrapper.java
mmusgrov: You are correct, I will remove them. Perhaps I should rename the class too: How about ORBWrapper -> ExternalORBWrapper

ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/jts/ExternalORBWrapper.java
```diff
@@ -0,0 +1,202 @@
...
+ * {@link TransactionManagerService#start} and {@ink RecoveryManagerService#start}
+ */
+class ExternalORBWrapper extends org.omg.CORBA.ORB {
+    private final ORB orb;
+
+    public ExternalORBWrapper(ORB orb) {
+        this.orb = orb;
+    }
+    protected void set_parameters(String[] args, Properties props) {
+        ;
```
tomjenkinson: sounds like a good idea to remove them

mmusgrov: Done - is it ready for merge yet?

ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/jts/TransactionManagerService.java
```diff
@@ -92,7 +100,7 @@ public void start(org.omg.CORBA.ORB theCorbaORB) throws Exception
     public void stop() {
         super.stop();
 
-        ORB.getInstance("jboss-atx").shutdown();
+        ORB.getInstance(ORB_NAME).shutdown();
```
tomjenkinson: It's funny, because this shutdown() is a no-op in ExternalORBWrapper

mmusgrov: ORB.getInstance(ORB_NAME) returns our orbportability wrapper which calls our shutdown hooks. The external orb is set in ORBBase and it is that one that we want to avoid calling close on (since it may still be in use by other systems)

mmusgrov: The goal of the fix is:

- ensure that our shutdown hooks are called when the TM service is stopped
- ensure that we do not shutdown an externally supplied orb

tomjenkinson: Thanks Mike, that makes perfect sense now. I will mark it for merge but acknowledge CI is broken

PR #919, opened 2015-09-29
JBTM-1854 Don't initialise two orbs

mmusgrov:
!JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF

(just test MAIN and QA_JTS_JDKORB)

https://issues.jboss.org/browse/JBTM-1854

The primary change is the new class ArjunaJTS/jts/tests/classes/com/hp/mwtests/ts/jts/utils/ServerORB.java together with updates to all the test server classes that have main methods to use it.

### DIFF
--- a/scripts/hudson/narayana.sh
+++ b/scripts/hudson/narayana.sh
@@ -590,13 +590,13 @@ function qa_tests_once {
         ok=0
         for i in `seq 1 $QA_STRESS`; do
           echo run $i;
-          ant -f run-tests.xml -Dtest.name=$QA_TESTGROUP -Dtest.methods="$QA_TESTMETHODS" onetest -Dcode.coverage=$codeCoverage;
+          ant -f run-tests.xml $QA_PROFILE -Dtest.name=$QA_TESTGROUP -Dtest.methods="$QA_TESTMETHODS" onetest -Dcode.coverage=$codeCoverage;
           if [ $? -ne 0 ]; then
             ok=1; break;
           fi
         done
       else
-        ant -f run-tests.xml -Dtest.name=$QA_TESTGROUP -Dtest.methods="$QA_TESTMETHODS" onetest -Dcode.coverage=$codeCoverage;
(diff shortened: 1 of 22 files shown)


scripts/hudson/narayana.sh
```diff
@@ -590,13 +590,13 @@ function qa_tests_once {
         ok=0
         for i in `seq 1 $QA_STRESS`; do
           echo run $i;
-          ant -f run-tests.xml -Dtest.name=$QA_TESTGROUP -Dtest.methods="$QA_TESTMETHODS" onetest -Dcode.coverage=$codeCoverage;
+          ant -f run-tests.xml $QA_PROFILE -Dtest.name=$QA_TESTGROUP -Dtest.methods="$QA_TESTMETHODS" onetest -Dcode.coverage=$codeCoverage;
```
tomjenkinson: Hi Mike, wondered what QA_PROFILE does?

gytis: Enables journal store if needed.

tomjenkinson: thanks - it wasn't clear from the diff but I guess this is CI integration aspect

mmusgrov:
Ah right, that should have been documented somewhere. We have been testing it with:

    QA_PROFILE="-Dprofile=hornetq"

so that the qa tests run with hornetq/artemis-journal

mmusgrov:
Although we normally consider narayana.sh as self documenting

mmusgrov:
I had to kill the mac job because Paul wanted to move the NCL jenkins cluster. But the pass on linux should suffice.

gytis:
Everything looks good. Just that we don't test QA_JTS_JDKORB.

mmusgrov:
Ah yes the pull job only does QA_JTS_JACORB that is an error - we should be using QA_JTS_JDKORB. I will update the job.

mmusgrov:
So can I change the label and assignee?

PR #920, opened 2015-09-30
[JBTM-2519] XTS privileged actions

gytis:
https://issues.jboss.org/browse/JBTM-2519

!MAIN !BLACKTIE !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTA !PERF

### DIFF
--- a/XTS/WS-C/dev/src/com/arjuna/webservices11/wscoor/client/WSCOORClient.java
+++ b/XTS/WS-C/dev/src/com/arjuna/webservices11/wscoor/client/WSCOORClient.java
@@ -1,5 +1,7 @@
 package com.arjuna.webservices11.wscoor.client;
 
+import com.arjuna.webservices11.util.PrivilegedServiceFactory;
+import com.arjuna.webservices11.util.PrivilegedServiceHelper;
 import com.arjuna.webservices11.wsaddr.AddressingHelper;
 import org.jboss.ws.api.addressing.MAP;
 import org.oasis_open.docs.ws_tx.wscoor._2006._06.ActivationPortType;
@@ -10,8 +12,6 @@ import org.oasis_open.docs.ws_tx.wscoor._2006._06.RegistrationService;
 import javax.xml.ws.BindingProvider;
 import javax.xml.ws.soap.AddressingFeature;
 import javax.xml.ws.wsaddressing.W3CEndpointReference;
-import java.security.AccessController;
-import java.security.PrivilegedAction;
 import java.util.Map;
 
 /**
@@ -41,8 +41,7 @@ public class WSCOORClient
     private static synchronized ActivationService getActivationService()
     {
         if (activationService.get() == null) {
-            // activationService.set(new ActivationService(null, new QName("http://docs.oasis-open.org/ws-tx/wscoor/2006/06", "ActivationService")));
-            activationService.set(new ActivationService());
+            activationService.set(PrivilegedServiceFactory.getInstance(ActivationService.class).getService());
         }
         return activationService.get();
     }
@@ -54,17 +53,16 @@ public class WSCOORClient
     private static synchronized RegistrationService getRegistrationService()
     {
         if (registrationService.get() == null) {
-            // registrationService.set(new RegistrationService(null, new QName("http://docs.oasis-open.org/ws-tx/wscoor/2006/06", "RegistrationService")));
-            registrationService.set(new RegistrationService());
+            registrationService.set(PrivilegedServiceFactory.getInstance(RegistrationService.class).getService());
         }
         return registrationService.get();
     }
 
-    public static ActivationPortType getActivationPort(MAP map,
-                                                       String action)
+    public static ActivationPortType getActivationPort(MAP map, String action)
     {
-        ActivationService service = getActivationService();
-        ActivationPortType port = service.getPort(ActivationPortType.class, new AddressingFeature(true, true));
+        final ActivationService service = getActivationService();
+        final ActivationPortType port = PrivilegedServiceHelper.getInstance().getPort(service, ActivationPortType.class,
+                new AddressingFeature(true, true));
         BindingProvider bindingProvider = (BindingProvider)port;
         String to = map.getTo();
         /*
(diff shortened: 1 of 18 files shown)


XTS/WS-C/dev/src/com/arjuna/webservices11/wscoor/client/WSCOORClient.java
```diff
@@ -41,12 +42,27 @@
     private static synchronized ActivationService getActivationService()
     {
         if (activationService.get() == null) {
-            // activationService.set(new ActivationService(null, new QName("http://docs.oasis-open.org/ws-tx/wscoor/2006/06", "ActivationService")));
-            activationService.set(new ActivationService());
+            setActivationService();
```
tomjenkinson: I have a query about the style you used throughout this patch. I see you added "private static synchronized" methods throughout and called them from places where you replaced the original setter. I wondered if it would read better if the calls were still inline. Is there a security style/precedent that you were following or was it personal preference?

XTS/WS-C/dev/src/com/arjuna/webservices11/wscoor/client/WSCOORClient.java
```diff
@@ -41,12 +42,27 @@
...
+    private static synchronized void setActivationService()
+    {
+        // activationService.set(new ActivationService(null, new QName("http://docs.oasis-open.org/ws-tx/wscoor/2006/06", "ActivationService")));
+        if (System.getSecurityManager() == null) {
+            activationService.set(new ActivationService());
+        } else {
+            AccessController.doPrivileged(new PrivilegedAction<Void>() {
+                @Override
+                public Void run() {
+                    activationService.set(new ActivationService());
```
tomjenkinson: Does this set really require security manager permissions? Is it not maybe something inside ActivationService ctor that needs the permissions instead?

XTS/WS-C/dev/src/com/arjuna/webservices11/wscoor/client/WSCOORClient.java
gytis: Well, I did it in order to not overcomplicate the method and make it do only single job.

tomjenkinson: The reason I asked is I just don't see setActivationService() as ever being reused - I wondered if you had seen David using this pattern or something

XTS/WS-C/dev/src/com/arjuna/webservices11/wscoor/client/WSCOORClient.java
tomjenkinson: I think this is the main consideration - does that setter really need security privs?

gytis: All these services are extensions of javax.xml.ws.Service class. They don't do much other than call super methods (this particular constructor also only calls super). But since they're always created from the client, I thought it'd be better to check permissions in the client rather than scatter them through the bunch of different classes.

XTS/WS-C/dev/src/com/arjuna/webservices11/wscoor/client/WSCOORClient.java
gytis: I saw it in WildFly as well as in other places. I think in general it is preferred to have small methods with single task.
Why do you think it's bad to split them up?

gytis: I think it also makes it more readable

XTS/WS-C/dev/src/com/arjuna/webservices11/wscoor/client/WSCOORClient.java
tomjenkinson: I think you should add a comment then as its not clear that it is the fact they extend javax.xml.ws.Service and need to load that class which causes the problem.

XTS/WS-C/dev/src/com/arjuna/webservices11/wscoor/client/WSCOORClient.java
tomjenkinson: I find it harder to read as you have to track around a class. I tend to just extract to methods I can see being reused. Anyway, you are the implementor - I was checking why you did it that way is all (considering the previous method performed the function inline in the method.

gytis: Yes, but then there was only one line while now there are eleven.

XTS/WS-C/dev/src/com/arjuna/webservices11/wscoor/client/WSCOORClient.java
```diff
@@ -41,12 +42,27 @@
...
-            activationService.set(new ActivationService());
+            setActivationService();
         }
         return activationService.get();
     }
 
+    private static synchronized void setActivationService()
+    {
+        // activationService.set(new ActivationService(null, new QName("http://docs.oasis-open.org/ws-tx/wscoor/2006/06", "ActivationService")));
+        if (System.getSecurityManager() == null) {
```
mmusgrov: How often is  setActivationService called - if lots of times then there might be some performance benefit from cacheing. If not then no point. This is a general point about sprinkling privileged actions throughout our code base and it might be beneficial to make sure we are doing things efficiently.

gytis: I'm not sure I understood your concern. Do you mean that this comparison with null is inefficient?

gytis: In this particular case, I don't think we can extract this privileged action, because it needs to access activationService.
However, I found a couple of places where I can extract privileged actions to their own classes and use singleton instances.

XTS/WS-C/dev/src/com/arjuna/webservices11/wscoor/client/WSCOORClient.java
gytis: How about instead of using "new ActivationService()" I'll move that to the separate factory class. I could handle the security privileges there.
As a bonus I wouldn't need these extracted set methods and this would address your concern from above too.

tomjenkinson: TBH I am OK with what you have (but happy to review other changes if you like). All I meant is it looks odd until I knew about the extension point of ws.Service causing a permissions issue so a comment to that effect would also be fine.

gytis: I've decided to only add the comment, because there are many different services and they are only set once. Thus creating separate classes for that would be a bit of an overkill.

mmusgrov:
The changes look OIK. I did have a general comment: I see lots of methods and classes here that are almost identical and I think something could be done to make the code more readable and modular but I guess that just my opinion.

gytis:
I agree, I don't like the design of most of them either. I think maybe generic factory class for those services which need privileges could work too. I'll look into that and maybe that will improve readability.

gytis:
As you said before, there were a lot of repetitive methods in there. So I tried to extract them where possible. e.g. PrivilegedServiceFactory

mmusgrov:
Nice implementation, it looks much more elegant.

PR #921, opened 2015-10-02
JBTM-2476 Updated to print out stack activity of multiple threads when cancelled (e.g. by reaper)

tomjenkinson:
!XTS !BLACKTIE !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB
https://issues.jboss.org/browse/JBTM-2476

Can be verified in SimpleTest in the JTA surefire logs that a stack trace was printed

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
@@ -3363,6 +3363,19 @@ public class BasicAction extends StateManager
                     tsLogger.i18NLogger.warn_coordinator_BasicAction_57(get_uid());
                 } else {
                     tsLogger.i18NLogger.warn_coordinator_BasicAction_58(get_uid());
+
+                    // This was added for JBTM-2476 - it could be used during commit
+                    // but as the commit path needs to be optimum and getStackTrace is expensive
+                    // I did not include it
+                    for (Thread entry : _childThreads.values()) {
+                        StackTraceElement[] stackTrace = entry.getStackTrace();
+                        StringBuilder sb = new StringBuilder();
+                        for (StackTraceElement element : stackTrace) {
+                            sb.append(element.toString());
+                            sb.append("\n");
+                        }
+                        tsLogger.i18NLogger.warn_multiple_threads(objectUid, entry.getName(), sb.toString());
+                    }
                 }
 
                 if (_checkedAction != null)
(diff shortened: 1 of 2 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
```diff
@@ -3386,6 +3386,19 @@ private final boolean checkChildren (boolean isCommit)
                     tsLogger.i18NLogger.warn_coordinator_BasicAction_57(get_uid());
                 } else {
                     tsLogger.i18NLogger.warn_coordinator_BasicAction_58(get_uid());
+
```
mmusgrov: Shouldn't this be bracketed by an if (isWarnEnabled) check

tomjenkinson: Yeah, that would be useful. I have just checked in the IDE and the method doesn't seem to exist though?

gytis: @mmusgrov comment reminded me of this: https://dmlloyd.github.io/presentations/coding-standards/index.html#/27. I'm just not sure what does he mean by saying "most constructs".

tomjenkinson: Its a good observation but this is warn, rather than trace. In terms of string contatenation I am not really sure how I can avoid it here?

gytis: I'd say rule applies to both isTraceEnabled and isWarnEnabled don't you think?
That isTraceEnabled point was the only one I was meaning to ask about. Not about concatenation.

tomjenkinson: Well, this information needs to be printed when the level includes warn (i.e most times) - if I guard the concat with a trace check then the info won't be printed unless trace is on (unlikely).

Maybe I am speaking to another area of the code as your comment currently applies to the blank 3389 - are you talking about the code in lines 3394 - 3399?

gytis: Sorry for confusing. I might have read his slide wrong. I've missed the point that he suggests to use tracef and others so that isTraceEnabled wouldn't be needed. I thought jboss logging handles that checking in the background.

Anyway, I don't remember us using isWarnEnabled before.

mmusgrov: Actually if you step through to the actual log call it does

    if (isEnabled(level))

so I withdraw my comment.

tomjenkinson: Yeah, I can't see it in the IDE :(

mmusgrov:
I wonder if later versions of jboss-logging observe Davids recommendation regarding the need to not check the level? If so we should upgrade and get a (small) speed bump)

PR #924, opened 2015-10-05
JBTM-2492 Updated to build and deploy the Java BlackTie components du…

tomjenkinson:
…ring a release

NO_TEST

https://issues.jboss.org/browse/JBTM-2492

### DIFF
--- a/build-release-pkgs.xml
+++ b/build-release-pkgs.xml
@@ -115,7 +115,13 @@
     <zip basedir="${workdir}" destfile="${workdir}/narayana-full-${tag}-src.zip" includes="${tag}/**"/>
     <!-- build the binary releases -->
     <exec dir="${workdir}/${tag}" executable="${mvn.executable}" failonerror="true">
-      <arg line="deploy -DskipTests -gs tools/maven/conf/settings.xml -Dorson.jar.location=${user.dir}/ext/ -Prelease"/>
+      <arg line="clean deploy -DskipTests -gs tools/maven/conf/settings.xml -Dorson.jar.location=${user.dir}/ext/ -Prelease"/>
+    </exec>
+    <exec dir="${workdir}/${tag}" executable="${mvn.executable}" failonerror="true">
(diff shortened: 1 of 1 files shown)


build-release-pkgs.xml
```diff
@@ -115,7 +115,13 @@
...
     <!-- build the binary releases -->
     <exec dir="${workdir}/${tag}" executable="${mvn.executable}" failonerror="true">
-      <arg line="deploy -DskipTests -gs tools/maven/conf/settings.xml -Dorson.jar.location=${user.dir}/ext/ -Prelease"/>
+      <arg line="clean deploy -DskipTests -gs tools/maven/conf/settings.xml -Dorson.jar.location=${user.dir}/ext/ -Prelease"/>
+    </exec>
+    <exec dir="${workdir}/${tag}" executable="${mvn.executable}" failonerror="true">
+      <arg line="clean deploy -DskipTests -gs tools/maven/conf/settings.xml -Prelease -f blacktie/utils/cpp-plugin/pom.xml"/>
+    </exec>
+    <exec dir="${workdir}/${tag}" executable="${mvn.executable}" failonerror="true">
+      <arg line="clean deploy -DskipTests -gs tools/maven/conf/settings.xml -Prelease  -f blacktie/pom.xml -pl :blacktie-jatmibroker-nbf -am"/>
```
gytis: wouldn't it be more efficient to build both cpp-plugin and blacktie-jatmibroker-nbf together with -pl?

tomjenkinson: cpp-plugin lives outside the lifecycle of the main build - it should really be in a separate repo tbh but it was a decision we made a while back. It basically is a plugin used by the main build to build the C++ components under maven in a platform neutral manner. You can't actually call it directly from the maven build as it is cycling during maven initialization.

gytis: Thanks for the explanation. I think it's good to merge.

PR #927, opened 2015-10-09
[JBTM-2481] Create WildFly component upgrade issue with JIRA update script

gytis:
https://issues.jboss.org/browse/JBTM-2529
NO_TESTS


gytis:
I did test it on Blacktie JIRA. Nothing automated for now. But I'll start
looking into introducing normal automated tests for them.

gytis:
https://issues.jboss.org/browse/BLACKTIE-457
It doesn't have fix version on purpose. I wasn't sure if we want to add
that, since they change quite frequently.

Also, I've just noticed that component is not there, but it should be. I'll
add that before merging.

gytis:
It does add the link to the new jira

PR #929, opened 2015-10-15
JBTM-2530 Updated to use the build.sh script instead of assuming mvn …

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2530

NO_TEST


mmusgrov:
I kicked of a build 30 minutes ago and the build takes 3 hours to fail. If you merge you fix pronto I will kill it and start a new one going with your fix

PR #932, opened 2015-10-26
JBTM-2318 Updated to make some of the messages that commonly appear a…

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2318

!BLACKTIE !QA_JTA !QA_JTS_JACORB !XTS !PERF !QA_JTS_JDKORB

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/TransactionStatusManager.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/TransactionStatusManager.java
@@ -85,8 +85,7 @@ public class TransactionStatusManager
          _listener = new Listener( serverSocket, service );
          _listener.setDaemon(true);
 
-	    tsLogger.i18NLogger.info_recovery_TransactionStatusManager_1(service.getClass().getName(),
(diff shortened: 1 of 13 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/TransactionStatusManager.java
```diff
@@ -85,8 +85,7 @@ private void addService( Service service, ServerSocket serverSocket )
          _listener = new Listener( serverSocket, service );
          _listener.setDaemon(true);
 
-	    tsLogger.i18NLogger.info_recovery_TransactionStatusManager_1(service.getClass().getName(),
-							  Integer.toString(serverSocket.getLocalPort()));
+         tsLogger.logger.debug(service.getClass().getName() + " starting");
```
mmusgrov: You have lost the service name in the debug message - the port number gets used in the thread name but I cannot see where we log this service name.

mmusgrov:
I have a general question about removing public log messages - do our deprecation rules apply here? If not then the changes look good, if not then we need to deprecate first.

PR #935, opened 2015-11-04
JBTM-2550 ClassNotFoundException when restore_state from the object store during recovering

zhfeng:
https://issues.jboss.org/browse/JBTM-2550

!BLACKTIE !PERF !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
@@ -966,13 +966,18 @@ public class XAResourceRecord extends AbstractRecord implements ExceptionDeferre
 
 	                    // Give it a go ourselves
 	                    if (!deserialized) {
-	                        _theXAResource = (XAResource) o.readObject();
+	                        try {
+	                            _theXAResource = (XAResource) o.readObject();
+	                            if (jtaLogger.logger.isTraceEnabled()) {
+	                                jtaLogger.logger.trace("XAResourceRecord.restore_state - XAResource de-serialized");
+	                            }
+	                        } catch (ClassNotFoundException e) {
+	                            // JBTM-2550 if we fail to deserialize the object, we treat it as haveXAResource is false
+	                            jtaLogger.i18NLogger.warn_resources_arjunacore_classnotfound(className);
+	                            haveXAResource = false;
+	                        }
 	                    }
 	                    o.close();
-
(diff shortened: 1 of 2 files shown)


ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
```diff
@@ -966,13 +966,17 @@ public boolean restore_state(InputObjectState os, int t)
...
 	                    if (!deserialized) {
-	                        _theXAResource = (XAResource) o.readObject();
+	                        try {
+	                            _theXAResource = (XAResource) o.readObject();
+	                            if (jtaLogger.logger.isTraceEnabled()) {
+	                                jtaLogger.logger.trace("XAResourceRecord.restore_state - XAResource de-serialized");
+	                            }
+	                        } catch (ClassNotFoundException e) {
+	                            jtaLogger.i18NLogger.warn_resources_arjunacore_classnotfound(className);
+	                            _theXAResource = getNewXAResource();
```
mmusgrov: You need to check for if (_theXAResource == null) just like we do below in the case where we do not have a serialized resource (in the else on line 990). But rather than duplicate code you should just remove the else arm on line 990.

mmusgrov:
@zhfeng Does Ondrejs test pass with this fix?

mmusgrov:
The extra commit looks good. If you assign me as reviewer I will mark it as ready for merge. Also did you retest with the test @ochaloup gave you?

zhfeng:
JMSMdbCrashRecoveryRemoteTestCase#commitHaltAtPhaseBeginningMdb pass
JMSMdbCrashRecoveryRemoteTestCase#commitHaltAtPhaseBeginning is still failing with
``` 
ARJUNA016037: Could not find new XAResource to use for recovering non-serializable XAResource XAResourceRecord
```
Ondra think that [ARJUNA016037] is expected as the ActiveMQ and IBMMQ do not implement the XAResourceWrapper

The Mdb crash test scenario
```
enlist inbound MDB JMS resource
enlist TestXA resource
enlist XA resource
prepare OK on all three resources
TM save the log correctly
commit the transaction
doCommit on inbound MDB resource returns FINISH_OK
doCommit on TestXA crash the JVM
reboot the container
recovery the inbound MDB resource returns FINISH_ERROR and shows the [ARJUNA016037]
```

The commitHaltAtPhaseBeginningMdb crash at XAResourceRecord.topLevelCommit which should killJVM when doCommit the inbound MDB resouce and the recovery works.

PR #943, opened 2015-11-15
JBTM-2560 Add generics

marschall:
Since the source level is Java 5+ the code base can be properly
generified.

Issue: JBTM-2560
https://issues.jboss.org/browse/JBTM-2560


marschall:
There you go:
https://developer.jboss.org/message/946052#946052

PR #944, opened 2015-11-17
JBTM-2537 Explicitly se the idlj compiler

mmusgrov:
!XTS !PERF !BLACKTIE !QA_JTA !QA_JTS_JDKORB !QA_JTS_JACORB !QA_JTS_OPENJDKORB

https://issues.jboss.org/browse/JBTM-2537

### DIFF
--- a/ArjunaJTS/idl/idlj/pom.xml
+++ b/ArjunaJTS/idl/idlj/pom.xml
@@ -41,107 +41,233 @@
       </plugin>
 
       <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>idlj-maven-plugin</artifactId>
+        <artifactId>maven-antrun-plugin</artifactId>
+        <version>1.7</version>
         <executions>
           <execution>
             <id>generate-sources</id>
             <phase>generate-sources</phase>
             <goals>
-              <goal>generate</goal>
+              <goal>run</goal>
             </goals>
             <configuration>
-              <compiler>idlj</compiler>
-              <sourceDirectory>${project.parent.basedir}/src/main/idl</sourceDirectory>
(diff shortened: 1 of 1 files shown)


ArjunaJTS/idl/idlj/pom.xml
```diff
@@ -41,6 +42,239 @@
...
+              <goal>run</goal>
+            </goals>
+            <configuration>
+              <tasks name="Generate idl stubs">
+                <java classname="com.sun.tools.corba.se.idl.toJavaPortable.Compile">
+
+                  <arg value="-i"/>
+                  <arg value="${project.build.directory}/../../src/main/idl"/>
+                  <arg value="-td"/>
+                  <!-- TODO figure out why the compiler does not pick up these sources
```
mmusgrov: Why doesn't generated-test-sources/idl get added to the maven compiler plugin compileSourceRoots property?

ArjunaJTS/idl/idlj/pom.xml
```diff
@@ -41,6 +42,239 @@
...
+                  </classpath>
+                </java>
+              </tasks>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+
+
+      <!-- this one does not seem to work either
```
mmusgrov: Why doesn't generated-test-sources/idl get added to the maven compiler plugin compileSourceRoots property?

ArjunaJTS/idl/idlj/pom.xml
tomjenkinson: If you mean automatically its because maven doesn't know anything about that folder unless you tell it. I think you can use something called build-helper-plugin to add source paths.

Maybe it was a "memo-to-self" in which case sorry for intruding :) If not, please can you add a little more context?

PR #946, opened 2015-11-19
JBTM-2566 Updated so that errors in the browsing of the objectstore a…

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2566

!PERF !BLACKTIE !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/objectstore/ObjectStoreIterator.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/objectstore/ObjectStoreIterator.java
@@ -46,34 +46,23 @@ import com.arjuna.ats.internal.arjuna.common.UidHelper;
 public class ObjectStoreIterator
 {
 
-    public ObjectStoreIterator(RecoveryStore recoveryStore, String tName)
+    public ObjectStoreIterator(RecoveryStore recoveryStore, String tName) throws ObjectStoreException
     {
-        try
-        {
-            recoveryStore.allObjUids(tName, uidList);
-        }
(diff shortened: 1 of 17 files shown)


mmusgrov:
(18:01:57) tomjenkinson: ok, so I created https://issues.jboss.org/browse/JBTM-2566 and linked a PR already for it but I don't think we should merge it just yet unless you are confident its the right change
(18:04:49) mmusgrov: Agreed. Also deprecation rules apply
(18:12:38) mmusgrov: the wildfly subsystem uses JMX so the new exception would be wrapped by JMX and the subsystem code catches MBeanException so we won't need to change subsystem code which is good

mmusgrov:
The MBean interfaces should not contain application specific classes since they need to be usable by generic management tools so you would need to wrap the exceptions coming from MBeans in a javax.management.MBeanException. The two interface methods affected by this are ObjStoreBrowserMBean#probe and OSEntryBeanMBean#remove

And with this change any user code out there that implements OSEntryBeanMBean will now fail.

Remark: I have a long running task to provide replacements for all of these deprecated classes.

mmusgrov:
Other that the lack of tests showing the behaviour of the iterator I am OK with the changes.

mmusgrov:
OK perhaps asking for a test was overkill since we already have one. I was thrown by the javadoc for the iterate method which is confusing and does not reflect what the method actually does.

mmusgrov:
In particular it never returns null and the phrase "Use the latter to return 0" is odd.

mmusgrov:
Since you didn't add those in this PR I am marking as ready for merge.

PR #951, opened 2015-11-26
JBTM-2569 Treat a failure to find a log as if the log existed

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2569

!PERF

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/JTATransactionLogXAResourceOrphanFilter.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/JTATransactionLogXAResourceOrphanFilter.java
@@ -28,7 +28,6 @@ import com.arjuna.ats.arjuna.objectstore.RecoveryStore;
 import com.arjuna.ats.arjuna.objectstore.StateStatus;
 import com.arjuna.ats.arjuna.objectstore.StoreManager;
 import com.arjuna.ats.arjuna.state.InputObjectState;
-import com.arjuna.ats.internal.jta.resources.arjunacore.CommitMarkableResourceRecord;
 import com.arjuna.ats.internal.jta.transaction.arjunacore.AtomicAction;
 import com.arjuna.ats.jta.logging.jtaLogger;
 import com.arjuna.ats.jta.recovery.XAResourceOrphanFilter;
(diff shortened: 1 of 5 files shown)


ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/JTATransactionLogXAResourceOrphanFilter.java
```diff
@@ -127,6 +127,7 @@ private boolean transactionLog(Xid xid)
             catch (Exception ex)
             {
                 ex.printStackTrace();
+                return true;
```
mmusgrov: It is a shame that we cannot return one of "yes/no/maybe" - why did you discount letting the exception leak out of the method for the caller to handle?

mmusgrov:
I agree that the PERF failure seems unrelated and can probably be ignored.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/JTATransactionLogXAResourceOrphanFilter.java
tomjenkinson: Good catch I think the maybe is the exception so I changed it and added to jdoc. The code is only used in one place (and that is private) but its worth the change.

mmusgrov:
> I have made the changes recommended and will mark for !PERF

Cunning (using the Vote datatype for the ternary result).

There is something I missed in the initial review though - why do you ignore the ObjectStoreException and IOExceptions in the method containsCommitMarkableResourceRecord()? This means there are two distinct behaviours if exceptions are thrown:
   
   if containsCommitMarkableResourceRecord(u) throws exceptions then the final result is Vote.ABSTAIN

   if recoveryStore.currentState(u, transactionType) throws exceptions then the final result is Vote.LEAVE_ALONE

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/JTATransactionLogXAResourceOrphanFilter.java
```diff
@@ -101,32 +103,24 @@ private boolean transactionLog(Xid xid)
...
-                }
+                return true;
             }
-            catch (Exception ex)
+            else
             {
-                ex.printStackTrace();
+                if (jtaLogger.logger.isDebugEnabled()) {
+                    jtaLogger.logger.debug("No record found for " + theXid);
+                }
```
mmusgrov: Looking good. But I do have a quibble, you are attempting to log whether or you find a record but in the exception case no debug output is produced and this would be of interest (unless you know something inside the calls to containsCommitMarkableResourceRecord and recoveryStore is logging the exceptions.

tomjenkinson: Its a fair point. Because my test throws ISE the exception is actually caught (and logged) in XARM: https://github.com/jbosstm/narayana/blob/master/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java#L738

I will add a line of debug incase OSE or IOE are thrown though

PR #952, opened 2015-11-27
[JBTM-2561] XTS code coverage failures

gytis:
https://issues.jboss.org/browse/JBTM-2561

!BLACKTIE, !QA_JTA, !QA_JTS_JACORB, !QA_JTS_JDKORB, !PERF

I've added a new version of Maven, because I needed "-pl !code-coverage" option in narayana.sh to postpone code coverage report generation.
Also, we'll need to remove "-PcodeCoverage -Pfindbugs" from code coverage job, since I've added them to narayana.sh.


scripts/hudson/narayana.sh
```diff
@@ -81,7 +83,8 @@ function init_test_options {
     elif [[ $PROFILE == "JACOCO" ]]; then
         export COMMENT_ON_PULL=""
         export AS_BUILD=1 NARAYANA_BUILD=1 NARAYANA_TESTS=1 BLACKTIE=0 XTS_AS_TESTS=0 XTS_TESTS=1 TXF_TESTS=1 txbridge=1
-        export RTS_AS_TESTS=0 RTS_TESTS=1 JTA_CDI_TESTS=1 QA_TESTS=1 SUN_ORB=1 JAC_ORB=0 JTA_AS_TESTS=0 OSGI_TESTS=0
```
mmusgrov: Why no coverage for RTS and CDI?

tomjenkinson: I am guessing they didn't work - please can you raise a Jira if they don't already exist?

gytis: Since this particular JIRA was about XTS, I didn't add them. But they would work without much problem.

mmusgrov: My comment was motivated by the fact that you disabled them. The rest of the changes are fine.

tomjenkinson: good catch! I expect that was a testing change and should be reverted?

gytis: Yeah that's a mistake. I'll change it back.

gytis: But it only applies for RTS. CDI doesn't generate coverage report now, so it's tests are not necessary here.

tomjenkinson: Sure, please can you Jira the CDI one then?

PR #953, opened 2015-11-27
[JBTM-2570] Remove unused private methods and fields in XTS

gytis:
https://issues.jboss.org/browse/JBTM-2570

!BLACKTIE !PERF !MAIN !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB


XTS/localjunit/WSTX11-interop/src/main/java/com/jboss/transaction/txinterop/interop/states/BAInteropUnsolicitedCompleteState.java
```diff
@@ -54,7 +49,7 @@ public BAInteropUnsolicitedCompleteState()
      */
     public ContentHandler getHandler(final ContentHandler nextHandler)
     {
-	return (replaced ? null : new RegistrationHandler(nextHandler)) ;
+	return new RegistrationHandler(nextHandler);
```
tomjenkinson: I agree with the change in principal but it would be good to understand what replaced might have been intended to do.

gytis: It was there since the move from SVN. It is a test class, so my guess is, that it wasn't that important, since nobody ever came across it.

nmcl:
Talk with Kevin Conner and Andrew Dinn, since it's possible one of them added it back then.

gytis:
I've emailed both Kevin and Andrew couple of weeks ago. I still haven't received reply from Kevin. However, Andrew doesn't remember that field ever being used and thinks that it's most likely a left over from an old version. He agrees that there should be no need to return null from getHandler.

PR #954, opened 2015-12-02
Update to fix the ArjunaJTA/jta tests to work under the codeCoverage

zhfeng:
https://issues.jboss.org/browse/JBTM-2115
https://issues.jboss.org/browse/JBTM-2573
https://issues.jboss.org/browse/JBTM-2574

PROFILE=JACOCO

### DIFF
--- a/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/TestCommitMarkableResourceBase.java
+++ b/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/TestCommitMarkableResourceBase.java
@@ -61,10 +61,9 @@ public class TestCommitMarkableResourceBase {
 
 		resetPropertiesFile = System
 				.getProperty("com.arjuna.ats.arjuna.common.propertiesFile");
-		if (resetPropertiesFile == null) {
-			System.setProperty("com.arjuna.ats.arjuna.common.propertiesFile",
+		System.setProperty("com.arjuna.ats.arjuna.common.propertiesFile",
 					"commitmarkableresourcejbossts-properties.xml");
-		}
+
 		manager = RecoveryManager.manager(RecoveryManager.DIRECT_MANAGEMENT);
 	}
 
--- a/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/TestCommitMarkableResourceReturnUnknownOutcomeFrom1PCCommit.java
+++ b/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/TestCommitMarkableResourceReturnUnknownOutcomeFrom1PCCommit.java
@@ -78,9 +78,8 @@ public class TestCommitMarkableResourceReturnUnknownOutcomeFrom1PCCommit {
         namingBeanImpl.start();
 
         resetPropertiesFile = System.getProperty("com.arjuna.ats.arjuna.common.propertiesFile");
-        if (resetPropertiesFile == null) {
-            System.setProperty("com.arjuna.ats.arjuna.common.propertiesFile", "commitmarkableresourcejbossts-properties.xml");
-        }
+        System.setProperty("com.arjuna.ats.arjuna.common.propertiesFile", "commitmarkableresourcejbossts-properties.xml");
+
         manager = RecoveryManager.manager(RecoveryManager.DIRECT_MANAGEMENT);
     }
 
--- a/ArjunaCore/arjuna/pom.xml
+++ b/ArjunaCore/arjuna/pom.xml
@@ -491,8 +491,6 @@
                     <artifactId>maven-surefire-plugin</artifactId>
                     <configuration>
                         <excludes>
-						  <!-- TODO breaks running with JaCoCo -->
-                          <exclude>**/ObjStoreBrowserTest.java</exclude>
 						  <!-- stress tests are not unit tests -->
                           <exclude>**/LogStressTest2.java</exclude>
                         </excludes>
--- a/ArjunaJTA/jta/pom.xml
+++ b/ArjunaJTA/jta/pom.xml
@@ -430,10 +430,7 @@
                     <configuration>
                       <excludes>
                         <exclude>**/io/narayana/perf/product/*</exclude>
-                        <exclude>**/commitmarkable/**</exclude>
-                        <!-- byteman and Jacoco seem to be interfering with each other -->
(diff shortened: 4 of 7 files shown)


zhfeng:
The windows build fails due to the new maven changes the "tools/maven/bin/mvn.bat" to "mvn.cmd". This causes  the build.bat does not work

gytis:
Hi Amos, I've raised https://issues.jboss.org/browse/JBTM-2576 for this.

gytis:
Should be all good now.

zhfeng:
http://albany.eng.hst.ams2.redhat.com/job/btny-pulls-narayana-codeCoverage/3/

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/TestCommitMarkableResourceBase.java
```diff
@@ -61,7 +61,7 @@ public final void setup() throws Exception {
 
 		resetPropertiesFile = System
 				.getProperty("com.arjuna.ats.arjuna.common.propertiesFile");
-		if (resetPropertiesFile == null) {
+		if (resetPropertiesFile == null || !resetPropertiesFile.equals("commitmarkableresourcejbossts-properties.xml")) {
```
mmusgrov: The name and location of the properties file is no longer configurable. If that is the intention then just hard code it directly and don't do any getProperty calls or tests:

`System.setProperty("com.arjuna.ats.arjuna.common.propertiesFile", 					"commitmarkableresourcejbossts-properties.xml"); `

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/TestCommitMarkableResourceReturnUnknownOutcomeFrom1PCCommit.java
```diff
@@ -78,7 +78,7 @@ public final void setup() throws Exception {
         namingBeanImpl.start();
 
         resetPropertiesFile = System.getProperty("com.arjuna.ats.arjuna.common.propertiesFile");
-        if (resetPropertiesFile == null) {
+        if (resetPropertiesFile == null || !resetPropertiesFile.equals("commitmarkableresourcejbossts-properties.xml")) {
```
mmusgrov: Same comment as above

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/TestCommitMarkableResourceBase.java
zhfeng: So that means in the tearDown() it should just clear this property 
`System.clearProperty("com.arjuna.ats.arjuna.common.propertiesFile");`

mmusgrov: The updated code in the PR that deals with this property does not seem to do anything other than ensure that the property com.arjuna.ats.arjuna.common.propertiesFile takes the value "commitmarkableresourcejbossts-properties.xml" and this can be achieved by setting the property without any tests.

It is good practice to clean up in tear down but that seems unrelated to my comment.

zhfeng: It looks like in the pom.xml the maven-surefire-plugin sets the system property
```
<systemProperties>
       <property>
                <name>com.arjuna.ats.arjuna.common.propertiesFile</name>
                <value>jbossts-properties.xml</value>
      </property>
 </systemProperties>
```
It would be better to reset to "jbossts-properties.xml" or this is the default value now ?

mmusgrov: No it wouldn't be better to reset to "jbossts-properties.xml" because this is just a test.

All I am saying is that after your changes the new code is semantically identical to:

`System.setProperty("com.arjuna.ats.arjuna.common.propertiesFile", "commitmarkableresourcejbossts-properties.xml");`

in which case you should use that since the changes you have introduced add needless complexity.

On the other hand, if you think your changes are not semantically identical to a single System.setProperty then please explain why (I may not have understood how your changes work)

tomjenkinson: Just to comment to the origin of this property manipulation. The properties file is set to jbossts-properties.xml in the surefire config and as this test needs to set it to commitmarkableresourcejbossts-properties.xml, it caches the original value and then is able to revert it to the value from the surefire config at the end of the test. I can't quite see why the diff added a check for resetPropertiesFile.equals("commitmarkableresourcejbossts-properties.xml") to maintain that  behaviour but perhaps I missed something and if so please can you explain why the proposed change to this line required?

mmusgrov: Ah right, thanks for the clarification Tom. So, Amos, I think it is safe to remove the resetPropertiesFile.equals ... check, do you agree?

zhfeng: Thanks Tom, and Mike, I agree with you that it should only set the property without any check.
The original check is `if (resetPropertiesFile == null)` that could have a problem when running the codeCoverage as the surefire config set it as "jbossts-peroperties.xml". That's why I added the equals check but it looks like redundant.

zhfeng:
Mike, I just update the fix and can you review it again ?

mmusgrov:
The changes for JBTM-2573 are still wrong. The bug report does not explain what the issue is, in fact the description is empty (this is a mandatory field) so it would help me if you could update the JIRA explaining what the issue is about and why the code fails (since it looked like the code you changed already did what @tomjenkinson  intended). Maybe the JIRA relates to a specific test but that is not described anywhere.

zhfeng:
Oh, sorry that it's my mistake and I will update the description soon.

zhfeng:
yeah, exactly.  My mistake to remove the line of getProperty.

mmusgrov:
The fix looks good, thanks

PR #957, opened 2015-12-03
JBTM-2583 Query the action status service in orphan detection

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2583
!BLACKTIE !PERF !QA_JTA !QA_JTS_JDKORB !XTS

Needs: https://github.com/jbosstm/jboss-as/pull/44

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/JTAActionStatusServiceXAResourceOrphanFilter.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/JTAActionStatusServiceXAResourceOrphanFilter.java
@@ -1,0 +1,88 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2010, Red Hat, Inc. and/or its affiliates,
+ * and individual contributors as indicated by the @author tags.
+ * See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ * This copyrighted material is made available to anyone wishing to use,
+ * modify, copy, or redistribute it subject to the terms and conditions
+ * of the GNU Lesser General Public License, v. 2.1.
+ * This program is distributed in the hope that it will be useful, but WITHOUT A
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+ * PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more details.
+ * You should have received a copy of the GNU Lesser General Public License,
+ * v.2.1 along with this distribution; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA  02110-1301, USA.
+ *
+ * (C) 2016,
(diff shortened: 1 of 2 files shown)


mmusgrov:
@tomjenkinson Do you still want me to review this?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/JTATransactionStatusManagerXAResourceOrphanFilter.java
```diff
@@ -0,0 +1,74 @@
...
+		String nodeName = XAUtils.getXANodeName(xid);
+		if (jtaLogger.logger.isDebugEnabled()) {
+			jtaLogger.logger.debug("node name of " + xid + " is " + nodeName);
+		}
+		if (xaRecoveryNodes == null || xaRecoveryNodes.size() == 0 || (!xaRecoveryNodes.contains(nodeName) && !xaRecoveryNodes.contains(NodeNameXAResourceOrphanFilter.RECOVER_ALL_NODES))) {
+			return Vote.ABSTAIN;
+		}
+
+		TransactionStatusConnectionManager transactionStatusConnectionMgr = new TransactionStatusConnectionManager();
+		int transactionStatus = transactionStatusConnectionMgr.getTransactionStatus(u);
```
mmusgrov: The associated JIRA states:
> In the case where the recovery manager and transaction manager are not in the same JVM process, the current behaviour of timeout based orphan detection MUST still be employed

Does this change guarantee this because xaRecoveryNodes will not contain remote nodes and if so I think it should be documented in the javadoc

tomjenkinson: I agree - maybe I should try to use the ASS instead of TSM

mmusgrov: Yes I believe the ASS only does an inVM check so if that is the intention then it would be preferable. I should go back to the thread to see why we cannot do the same check for remote nodes.

tomjenkinson: We can't rely on it for remote nodes. But we could use it. I think it would like be better to just use the ASS for now for clarity

mmusgrov:
What did you change with that last commit - it is not using ASS instead of TSM

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/JTATransactionStatusManagerXAResourceOrphanFilter.java
tomjenkinson: I have changed it to use the ASS as per your recommendation

mmusgrov:
Well it is now ;-)

PR #959, opened 2015-12-04
JBTM-2579 Throw XAException in XATerminator::commit if a wrapped resource fails transiently

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2579

!BLACKTIE !PERF !XTS !QA_JTS_JDKORB !QA_JTS_JACORB !QA_JTA

### DIFF
--- a/ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/orbspecific/interposition/coordinator/ServerTransaction.java
+++ b/ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/orbspecific/interposition/coordinator/ServerTransaction.java
@@ -241,6 +241,9 @@ public class ServerTransaction extends ArjunaTransactionImple
         }
 
 		super.phase2Commit(true);
+		if (super.failedList != null && super.failedList.size() > 0) {
+		    return ActionStatus.COMMITTING;
+		}
 
 		/*
 		 * Now do after completion stuff.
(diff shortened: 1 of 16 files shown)


ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/orbspecific/interposition/coordinator/ServerTransaction.java
```diff
@@ -241,6 +241,9 @@ public final int doPhase2Commit ()
         }
 
 		super.phase2Commit(true);
+		if (super.failedList != null && super.failedList.size() > 0) {
+		    return ActionStatus.COMMITTING;
+		}
 
 		/*
 		 * Now do after completion stuff.
```
mmusgrov: A remark: if users have been relying on the old behaviour then their after completions won't run so would it be worthwhile highlighting the fix in the release notes? But I guess this is a corner case you are fixing in which case it probably doesn't need to be specifically brought out?

mmusgrov:
I only reviewed the main changes and did not look at any of the tests. If you do want me to check those too let me know.

ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/orbspecific/interposition/coordinator/ServerTransaction.java
tomjenkinson: Sure, I can update the Jira issue with that note.

PR #964, opened 2015-12-16
JBTM-2584 commit LRCO and a readonly resource

mmusgrov:
https://issues.jboss.org/browse/JBTM-2584

!BLACKTIE !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_IBMORB !PERF
!XTS

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
@@ -1498,7 +1498,10 @@ public class BasicAction extends StateManager
             }
             else
             {
-                if (prepare(reportHeuristics) == TwoPhaseOutcome.PREPARE_NOTOK) {
+                int prepareStatus = prepare(reportHeuristics);
+
+                if (prepareStatus == TwoPhaseOutcome.PREPARE_NOTOK
+                        || prepareStatus == TwoPhaseOutcome.ONE_PHASE_ERROR) {
                     tsLogger.i18NLogger.warn_coordinator_BasicAction_36(get_uid());
 
                     if (heuristicDecision != TwoPhaseOutcome.PREPARE_OK) {
@@ -1801,6 +1804,7 @@ public class BasicAction extends StateManager
 
     /**
      * Second phase of the two-phase commit protocol for committing actions.
+     * Phase 2 should only be called if phase 1 has completed successfully.
      * This operation first invokes the doCommit operation on the preparedList.
      * This ensures that the appropriate commit operation is performed on each
      * entry which is then either deleted (top_level) or merged into the
@@ -2147,7 +2151,8 @@ public class BasicAction extends StateManager
 
             ActionManager.manager().remove(get_uid());
 
-            return TwoPhaseOutcome.PREPARE_ONE_PHASE_COMMITTED;
+            return actionStatus == ActionStatus.ABORTED
+                    ? TwoPhaseOutcome.ONE_PHASE_ERROR : TwoPhaseOutcome.PREPARE_ONE_PHASE_COMMITTED;
         }
         
         if ((p != TwoPhaseOutcome.PREPARE_OK)
--- a/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/lastresource/JTATest2.java
+++ b/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/lastresource/JTATest2.java
@@ -1,0 +1,191 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2006, Red Hat Middleware LLC, and individual contributors
+ * as indicated by the @author tags.
+ * See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ * This copyrighted material is made available to anyone wishing to use,
+ * modify, copy, or redistribute it subject to the terms and conditions
+ * of the GNU Lesser General Public License, v. 2.1.
+ * This program is distributed in the hope that it will be useful, but WITHOUT A
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+ * PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more details.
+ * You should have received a copy of the GNU Lesser General Public License,
+ * v.2.1 along with this distribution; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA  02110-1301, USA.
+ *
+ * (C) 2005-2006,
+ * @author JBoss Inc.
+ */
+/*
+ * Copyright (C) 2001, 2002,
+ *
+ * Hewlett-Packard Arjuna Labs,
+ * Newcastle upon Tyne,
+ * Tyne and Wear,
+ * UK.
+ */
+
+package com.hp.mwtests.ts.jta.lastresource;
+
+import com.arjuna.ats.jta.resources.LastResourceCommitOptimisation;
+import org.junit.Test;
+
+import javax.transaction.HeuristicMixedException;
+import javax.transaction.HeuristicRollbackException;
+import javax.transaction.NotSupportedException;
+import javax.transaction.RollbackException;
+import javax.transaction.SystemException;
+import javax.transaction.xa.XAException;
+import javax.transaction.xa.XAResource;
+import javax.transaction.xa.Xid;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public class JTATest2 {
+    /**
+     * Test the following scenario:
+     * 2 XA resources, one makes no changes during prepare (ie returns XA_RDONLY),
+     * and the other is a LRCO resource (so is processed after normal XA resources) which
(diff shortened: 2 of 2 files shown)


ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/lastresource/JTATest2.java
```diff
@@ -0,0 +1,193 @@
+/*
```
tomjenkinson: Maybe try to clean up this header in case we copy it again :)

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
```diff
@@ -1831,6 +1831,10 @@ protected synchronized final void phase2Commit (boolean reportHeuristics) throws
 
             phase2Abort(reportHeuristics);
         }
+        else if ((preparedList == null || preparedList.size() == 0) && actionStatus == ActionStatus.ABORTED) {
+            // the prepared list was completely processed but some resource rolled back
+            phase2Abort(reportHeuristics);
```
tomjenkinson: Would be good to get a warning (like the pending list check prior)

mmusgrov: This is a scenario where one of the resources rolls back during the prepare phase but there are no heuristics which I think is just part of the normal behaviour of committing resources. A similar situation occurs when the first resource aborts during commit so we abort the remaining participants but we don't log a warning in this case. [Here is where we abort the remaining participants](https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java#L2829)

If you still think such a scenario is sufficiently noteworthy then I will add a WARN message.

On the other hand the pending list means that there are resources for which prepare was never called and this is  unusual so is reported with a warning.

nmcl:
OK, I get that but as the comment on line 1524 shows ...

                        phase2Commit(reportHeuristics); /* first phase succeeded */

phase2Commit should only be called if prepare has succeeded. In this example prepare clearly hasn't succeeded so we shouldn't be calling phase2Commit in the first place. Seems to me that the test above ...

if (prepare(reportHeuristics) == TwoPhaseOutcome.PREPARE_NOTOK) 

needs changing instead so it also checks for this situation and we fall into the first block and call phase2Abort immediately.

BTW the test at the start of phase2Commit for the non-null pendingList is there more as a pre/post condition check in case a derived class calls this method when it really shouldn't, i.e., something breaks the protocol (is invalid). What you're describing is valid within the 2PC protocol but we simply have not checked for it in the first place.

mmusgrov:
OK that make good sense. I will move the check to where you propose.

nmcl:
Maybe it's worth adding something to the javadoc/comments for phase2Commit to call out this requirement to, i.e., that you should only call it if prepare succeeded? It's kind of implied in the comment already but probably not explicit enough.

mmusgrov:
I have moved the check to immediately after the prepare call.

@nmcl please can you take a look
@tomjenkinson putting the check here will output the existing warn_coordinator_BasicAction_36 message

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
```diff
@@ -1498,7 +1498,11 @@ protected synchronized int End (boolean reportHeuristics)
             }
             else
             {
-                if (prepare(reportHeuristics) == TwoPhaseOutcome.PREPARE_NOTOK) {
+                int prepareStatus = prepare(reportHeuristics);
+
+                if (prepareStatus == TwoPhaseOutcome.PREPARE_NOTOK
+                        || (prepareStatus == TwoPhaseOutcome.PREPARE_ONE_PHASE_COMMITTED
```
mmusgrov: I think I should be changing the prepare instead to return TwoPhaseOutcome.ONE_PHASE_ERROR instead of PREPARE_ONE_PHASE_COMMITTED if the onePhaseCommit aborts.

tomjenkinson: One option is to change onePhasePrepare API to return the right status https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java#L2322. That would be a public API change so not advisible for this release but might be something to consider for future.

How about returning TwoPhaseOutcome.PREPARE_NOTOK if the 1PC fails (determined by actionStatus in the place you are currently proposing)? We lose resolution as to the cause of the failure but it minimizes the changeset and the new return code would not actually be used anyway except in your changeset.

mmusgrov: You are looking at an outdated diff (sorry). I changed prepare to return
the correct status, as in:

`return actionStatus == ActionStatus.ABORTED ?
TwoPhaseOutcome.ONE_PHASE_ERROR :
TwoPhaseOutcome.PREPARE_ONE_PHASE_COMMITTED;`

On Thu, Dec 17, 2015 at 10:54 AM, Tom Jenkinson <notifications@github.com>
wrote:

> In
> ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
> <https://github.com/jbosstm/narayana/pull/964#discussion_r47892593>:
>
> > @@ -1498,7 +1498,11 @@ protected synchronized int End (boolean reportHeuristics)
> >              }
> >              else
> >              {
> > -                if (prepare(reportHeuristics) == TwoPhaseOutcome.PREPARE_NOTOK) {
> > +                int prepareStatus = prepare(reportHeuristics);
> > +
> > +                if (prepareStatus == TwoPhaseOutcome.PREPARE_NOTOK
> > +                        || (prepareStatus == TwoPhaseOutcome.PREPARE_ONE_PHASE_COMMITTED
>
> One option is to change onePhasePrepare API to return the right status
> https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java#L2322.
> That would be a public API change so not advisible for this release but
> might be something to consider for future.
>
> How about returning TwoPhaseOutcome.PREPARE_NOTOK if the 1PC fails
> (determined by actionStatus in the place you are currently proposing)? We
> lose resolution as to the cause of the failure but it minimizes the
> changeset and the new return code would not actually be used anyway except
> in your changeset.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jbosstm/narayana/pull/964/files#r47892593>.
>



-- 
Michael Musgrove
Transactions Team
e: mmusgrov@redhat.com
t: +44 191 243 0870

Registered in England and Wales under Company Registration No. 03798903
Directors: Michael Cunningham (US), Paul Hickey (Ireland), Matt Parson
(US), Charles Peters (US)

Michael Cunningham (US), Charles Peters (US), Matt Parson (US), Michael
O'Neill(Ireland)

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
```diff
@@ -2147,7 +2151,8 @@ protected synchronized final int prepare (boolean reportHeuristics)
 
             ActionManager.manager().remove(get_uid());
 
-            return TwoPhaseOutcome.PREPARE_ONE_PHASE_COMMITTED;
+            return actionStatus == ActionStatus.ABORTED
```
tomjenkinson: On the current diff now :)

One option is to change onePhasePrepare API to return the right status https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java#L2322. That would be a public API change so not advisible for this release but might be something to consider for future.

How about returning TwoPhaseOutcome.PREPARE_NOTOK if the 1PC fails so keep the code on line 1501 as is?

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
```diff
@@ -2147,7 +2151,8 @@ protected synchronized final int prepare (boolean reportHeuristics)
...
+                    ? TwoPhaseOutcome.ONE_PHASE_ERROR : TwoPhaseOutcome.PREPARE_ONE_PHASE_COMMITTED;
```
tomjenkinson: How about returning TwoPhaseOutcome.PREPARE_NOTOK if the 1PC fails so keep the code on line 1501 as is?

tomjenkinson: Actually, I think I prefer your approach its more accurate

mmusgrov: I disagree since the subclass can call BasicAction#status () to determine the result of calling onePhaseCommit

tomjenkinson: Disagreeing with preferring your approach?

mmusgrov: lol I should have put it in parenthesis

PR #965, opened 2015-12-21
JBTM-2550 Updated to try to locate an XAResource locally if the deser…

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2550
!BLACKTIE !XTS !QA_JTA !PERF !QA_JTS_JDKORB

### DIFF
--- a/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
+++ b/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
@@ -1008,16 +1008,19 @@ public class XAResourceRecord extends com.arjuna.ArjunaOTS.OTSAbstractRecordPOA
 	                        }
 	                    }
 
-	                    // Give it a go ourselves
-	                    if (!deserialized) {
-	                        _theXAResource = (XAResource) o.readObject();
-	                    }
-
-	                    o.close();
-
-	                    if (jtaxLogger.logger.isTraceEnabled()) {
-	                        jtaxLogger.logger.trace("XAResourceRecord.restore_state - XAResource de-serialized");
-	                    }
+                        try {
+                            // Give it a go ourselves
+                            if (!deserialized) {
+                                _theXAResource = (XAResource) o.readObject();
+                            }
+                            if (jtaxLogger.logger.isTraceEnabled()) {
+                                jtaxLogger.logger.trace("XAResourceRecord.restore_state - XAResource de-serialized");
+                            }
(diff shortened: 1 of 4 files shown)


ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -1010,7 +1010,11 @@ public boolean restoreState(InputObjectState os)
 
 	                    // Give it a go ourselves
 	                    if (!deserialized) {
-	                        _theXAResource = (XAResource) o.readObject();
+	                        try {
+	                            _theXAResource = (XAResource) o.readObject();
+	                        } catch (ClassNotFoundException cnfe) {
```
zhfeng: It looks like better to move the below jtaxLogger.logger.trace("...") into the try-catch clause.

tomjenkinson: Agreed - I will update the PR as you suggested

zhfeng:
It looks good to me. Thanks Tom !

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -1008,16 +1008,19 @@ public boolean restoreState(InputObjectState os)
...
+                        try {
+                            // Give it a go ourselves
+                            if (!deserialized) {
+                                _theXAResource = (XAResource) o.readObject();
+                            }
+                            if (jtaxLogger.logger.isTraceEnabled()) {
+                                jtaxLogger.logger.trace("XAResourceRecord.restore_state - XAResource de-serialized");
+                            }
+                        } catch (ClassNotFoundException cnfe) {
+                            jtaxLogger.i18NLogger.warn_could_not_load_class_will_wait_for_bottom_up(cnfe);
```
mmusgrov: Just a remark: am I correct in assuming the other exception types thrown by o.readObject are unrecoverable and therefore we must not fallback to using bottom up recover?  Do the other exception types indicate either mis-configuration or object store corruption?

tomjenkinson: I think it is fair to target CNFE for now https://docs.oracle.com/javase/7/docs/api/java/io/ObjectInputStream.html#readObject() - its distinct from the other ones and the only enhancement request we have is for the CNFE.

PR #966, opened 2016-01-05
JBTM-2599 - Update activemq-artemis to 1.1.0.wildfly-010

barreiro:
https://issues.jboss.org/browse/JBTM-2599

!PERF !QA_JTA !QA_JTS_JDKORB !QA_JTS_JACORB !XTS !BLACKTIE

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/hornetq/HornetqJournalEnvironmentBean.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/hornetq/HornetqJournalEnvironmentBean.java
@@ -40,6 +40,8 @@ public class HornetqJournalEnvironmentBean implements HornetqJournalEnvironmentB
 
     private volatile int minFiles = 4;
 
+    private volatile int poolSize = -1;
+
     private volatile int compactMinFiles = 10;
 
     private volatile int compactPercentage = 30;
@@ -80,7 +82,7 @@ public class HornetqJournalEnvironmentBean implements HornetqJournalEnvironmentB
 
     /**
      * Sets the desired size in bytes for each log file.
-     * 
+     *
      * @param fileSize the individual log file size, in bytes.
      */
     public void setFileSize(int fileSize)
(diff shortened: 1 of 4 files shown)


zhfeng:
I think we need to upgrade the wildfly to 10.0.0.Final-SNAPSHOT at first. I will create a Jira for this.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/hornetq/HornetqJournalEnvironmentBean.java
```diff
@@ -40,6 +40,8 @@
 
     private volatile int minFiles = 4;
 
+    private volatile int poolSize = -1;
```
zhfeng: I just wonder How the poolSize uses in the artemis journal store and why it set to -1 ?

Thanks,
Amos

barreiro: The poolSize featured was add with https://issues.apache.org/jira/browse/ARTEMIS-319

From the documentation: 
```
The upper theshold of the journal file pool,-1 (default) means no Limit. 
The system will create as many files as needed however when reclaiming files
it will shrink back to the `journal-pool-files`
```
(I guess they meant `journal-min-files` there) and also:
```
Set `journal-pool-files` to a number that represents
something near your maximum expected load.  The journal will spill over
the pool should it need to, but will shrink back to the upper threshold, 
when possible.  This allows reuse of files, without taking up more disk
space than required.  If you see new files being created on the journal
data directory too often, i.e. lots of data is being persisted, 
you need to increase the journal-pool-size, this way the journal would
reuse more files instead of creating new data files, increasing performance
```

barreiro:
You're welcome. Signed the CLA (although I don't think I needed to, as I work for Red Hat).

zhfeng:
I open the https://github.com/jbosstm/narayana/pull/967 to upgrade the wildfly to 10.0.0.Final-SNAPSHOT

PR #968, opened 2016-01-07
[JBTM-2604] Use servlet protocol in arquillian tests

gytis:
!QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_IBMORB !BLACKTIE !PERF !MAIN


gytis:
Based on stackoverflow post [1] Weld container does not support servlet protocol. The failure was on compensations test on Weld, so will have to remove that setting from this particular container.

[1] http://stackoverflow.com/questions/17824754/arquillian-no-httpcontext-found-when-launching-in-embedded-weld-ee-container

mmusgrov:
The changes look good, sorry for the delay.

PR #969, opened 2016-01-21
JBTM-2609 Updated to handle all exceptions (including IllegalStateExc…

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2609

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/jdbc/JDBCImple_driver.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/jdbc/JDBCImple_driver.java
@@ -102,13 +102,13 @@ public abstract class JDBCImple_driver {
 
 			int rowcount2 = pstmt2.executeUpdate();
 			if (rowcount2 > 0) {
-				result = true;
 				connection.commit();
+				result = true;
 			} else {
 				connection.rollback();
 			}
 
-		} catch (SQLException e) {
+		} catch (Exception e) {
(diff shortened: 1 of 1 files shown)


mmusgrov:
Sure (my notification filter seems to have stopped working).

I do notice that in some places connections are left open in the event of failure, for example:

https://github.com/tomjenkinson/narayana/blob/JBTM-2609/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/jdbc/JDBCImple_driver.java#L432

Previously the exception would have leaked out but now you are catching them and logging a warning so I think it is valid to close the connection in a finally block (I did not do an exhaustive check for this error handling, for example I didn't check other places and the cleaning up of statement handles).

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/jdbc/JDBCImple_driver.java
```diff
@@ -102,13 +102,13 @@ public boolean commit_state(Uid objUid, String typeName)
 
 			int rowcount2 = pstmt2.executeUpdate();
 			if (rowcount2 > 0) {
+                connection.commit();
```
mmusgrov: Use of spaces instead of tabs (ie not conforming to the existing formatting of the file)

tomjenkinson: :) ok will update that

PR #970, opened 2016-01-22
Make JCA txn import thread safe

mmusgrov:
!QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_IBMORB !BLACKTIE !PERF !XTS !MAIN

https://issues.jboss.org/browse/JBTM-2614

My solution is, instead of adding the TransactionImple directly to the list of known imported transactions I add a holder object using putIfAbsent (keyed by Xid). If the put returns null then I know this is the first time we have seen the imported transaction and I construct the transaction and add it to the holder object using a setter method. The getter and setter are protected from each other using the wait/notify idiom.

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
@@ -35,13 +35,17 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicReference;
 
 import javax.transaction.SystemException;
 import javax.transaction.xa.XAException;
 import javax.transaction.xa.Xid;
 
 import com.arjuna.ats.arjuna.common.Uid;
+import com.arjuna.ats.arjuna.coordinator.TxControl;
 import com.arjuna.ats.internal.jta.transaction.arjunacore.subordinate.jca.TransactionImple;
+import com.arjuna.ats.jta.xa.XATxConverter;
+import com.arjuna.ats.jta.xa.XidImple;
 
 public class TransactionImporterImple implements TransactionImporter
 {
@@ -87,19 +91,11 @@ public class TransactionImporterImple implements TransactionImporter
 			throw new IllegalArgumentException();
 
 		/*
-		 * Check to see if we haven't already imported this thing.
+		 * the imported transaction map is keyed by xid and the xid used is the one created inside
+		 * the TransactionImple ctor (it encodes the node name of this transaction manager) and is
+		 * the one returned by TransactionImple#baseXid() so pass in the converted value (using convertXid).
 		 */
-
-		TransactionImple imported = (TransactionImple) getImportedTransaction(xid);
-
-		if (imported == null)
-		{
-			imported = new TransactionImple(timeout, xid);
-			
-			_transactions.put(new SubordinateXidImple(imported.baseXid()), imported);
-		}
-
-		return imported;
+		return addImportedTransaction(null, convertXid(xid), xid, timeout);
 	}
 
 	/**
@@ -123,28 +119,16 @@ public class TransactionImporterImple implements TransactionImporter
 		    throw new IllegalArgumentException();
 		
 		/*
-		 * Is the transaction already in the list? This may be the case because
+		 * Is the transaction already in the map? This may be the case because
 		 * we scan the object store periodically and may get Uids to recover for
 		 * transactions that are progressing normally, i.e., do not need
-		 * recovery. In which case, we need to ignore them.
+		 * recovery. In which case, we need to ignore them:
+		 *
+		 * ie calling addImportedTransaction with a non null value for recovered will
+		 * call recovered.recordTransaction()
 		 */
 
-		TransactionImple tx = (TransactionImple) _transactions.get(recovered
-				.baseXid());
-
-		if (tx == null)
-		{
-
-			Xid baseXid = recovered.baseXid();
-			_transactions.put(new SubordinateXidImple(baseXid), recovered);
-			recovered.recordTransaction();
-
-			return recovered;
-		}
-		else
-		{
-			return tx;
-		}
+		return addImportedTransaction(recovered, recovered.baseXid(), null, 0);
 	}
 
 	/**
@@ -167,11 +151,19 @@ public class TransactionImporterImple implements TransactionImporter
 		if (xid == null)
 			throw new IllegalArgumentException();
 
-		SubordinateTransaction tx = _transactions.get(new SubordinateXidImple(xid));
+		AtomicReference<TransactionImple> holder = _transactions.get(new SubordinateXidImple(xid));
+		TransactionImple tx = holder == null ? null : holder.get();
 
-		if (tx == null)
+		if (tx == null) {
+			/*
+			 * Remark: if holder != null and holder.get() == null then the setter is about to
+			 * import the transaction but has not yet updated the holder. We implement the getter
+			 * (the thing that is trying to terminate the imported transaction) as though the imported
+			 * transaction only becomes observable when it has been fully imported.
+			 */
 			return null;
-		
+		}
+
 		// https://issues.jboss.org/browse/JBTM-927
 		try {
 			if (tx.getStatus() == javax.transaction.Status.STATUS_ROLLEDBACK) {
@@ -211,17 +203,65 @@ public class TransactionImporterImple implements TransactionImporter
 	}
 	
 	public Set<Xid> getInflightXids(String parentNodeName) {
-		Iterator<TransactionImple> iterator = _transactions.values().iterator();
+		Iterator<AtomicReference<TransactionImple>> iterator = _transactions.values().iterator();
 		Set<Xid> toReturn = new HashSet<Xid>();
 		while (iterator.hasNext()) {
-			TransactionImple next = iterator.next();
-			if (next.getParentNodeName().equals(parentNodeName)) {
-				toReturn.add(next.baseXid());
+			AtomicReference<TransactionImple> holder = iterator.next();
+			TransactionImple imported = holder.get();
+
+			if (imported != null && imported.getParentNodeName().equals(parentNodeName)) {
+				toReturn.add(imported.baseXid());
 			}
 		}
 		return toReturn;
 	}
 
-	private static ConcurrentHashMap<SubordinateXidImple, TransactionImple> _transactions = new ConcurrentHashMap<SubordinateXidImple, TransactionImple>();
+	private TransactionImple addImportedTransaction(TransactionImple importedTransaction, Xid mapKey, Xid xid, int timeout) {
+		SubordinateXidImple importedXid = new SubordinateXidImple(mapKey);
+		// We need to store the imported transaction in a volatile field holder so that it can be shared between threads
+		AtomicReference<TransactionImple> holder = new AtomicReference<>();
+		AtomicReference<TransactionImple> existing;
+
+		if ((existing = _transactions.putIfAbsent(importedXid, holder)) != null) {
+			holder = existing;
+		}
+
+		TransactionImple txn = holder.get();
+
+		if (txn == null) {
+			// retry the get under a lock - this double check idiom is safe because AtomicReference is effectively
+			// a volatile so can be concurrently accessed by multiple threads
+			synchronized (holder) {
+				txn = holder.get();
+				if (txn == null) {
+					// now it's safe to add the imported transaction to the holder
+					if (importedTransaction != null) {
+						importedTransaction.recordTransaction();
+						txn = importedTransaction;
+					} else {
+						txn = new TransactionImple(timeout, xid);
+					}
+
+					holder.set(txn);
+				}
+			}
+		}
+
+		return txn;
+	}
+
+	private XidImple convertXid(Xid xid)
+	{
+		if (xid != null && xid.getFormatId() == XATxConverter.FORMAT_ID) {
+			XidImple toImport = new XidImple(xid);
+			XATxConverter.setSubordinateNodeName(toImport.getXID(), TxControl.getXANodeName());
+			return new XidImple(toImport);
+		} else {
+			return new XidImple(xid);
+		}
+	}
+
+	private static ConcurrentHashMap<SubordinateXidImple, AtomicReference<TransactionImple>> _transactions =
+			new ConcurrentHashMap<>();
 }
 
--- a/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/jca/XATerminatorUnitTest.java
+++ b/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/jca/XATerminatorUnitTest.java
@@ -50,6 +50,14 @@ import com.hp.mwtests.ts.jta.common.FailureXAResource;
 import com.hp.mwtests.ts.jta.common.FailureXAResource.FailLocation;
 import com.hp.mwtests.ts.jta.common.FailureXAResource.FailType;
 
+import java.util.ArrayList;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CyclicBarrier;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Supplier;
+
 public class XATerminatorUnitTest
 {
     @Test
@@ -373,5 +381,61 @@ public class XATerminatorUnitTest
         catch (IllegalArgumentException ex)
         {
         }
+    }
+
+    @Test
+    public void testConcurrentImport () throws Exception {
+        AtomicInteger completionCount = new AtomicInteger(0);
+        XidImple xid = new XidImple(new Uid());
+
+        final int TASK_COUNT = 400;
+        final int THREAD_COUNT = 200;
+        final CyclicBarrier gate = new CyclicBarrier(THREAD_COUNT + 1);
+
+        ArrayList<CompletableFuture<SubordinateTransaction>> futures = new ArrayList<>();
+        ExecutorService executor = Executors.newFixedThreadPool(THREAD_COUNT);
+
+        for (int i = 0; i < TASK_COUNT; i++)
+            futures.add(doAsync(completionCount, gate, i < THREAD_COUNT, executor, xid));
+
+        gate.await();
+
+        SubordinateTransaction prevStx = null;
+
+        for (CompletableFuture<SubordinateTransaction> future : futures) {
+            SubordinateTransaction stx = future.get();
+            if (stx == null) {
+                fail("transaction import returned null for future ");
+            } else {
+                if (prevStx != null)
+                    assertEquals("transaction import for same xid returned a different instance", stx, prevStx);
+                else
+                    prevStx = stx;
+            }
+        }
+
+        assertEquals("some transaction import futures did not complete", completionCount.get(), TASK_COUNT);
+    }
+
+    /*
+     * import a transaction asynchronously to maximise the opportunity for concurrency errors in TransactionImporterImple
+     */
+    private CompletableFuture<SubordinateTransaction> doAsync(
+            final AtomicInteger completionCount, final CyclicBarrier gate, final boolean wait, ExecutorService executor, final XidImple xid) {
+        return CompletableFuture.supplyAsync(new Supplier<SubordinateTransaction>() {
+            @Override
+            public SubordinateTransaction get() {
+                try {
+                    if (wait)
+                        gate.await();
+                    SubordinateTransaction stx = SubordinationManager.getTransactionImporter().importTransaction(xid);
+                    completionCount.incrementAndGet();
+
+                    return stx;
+                } catch (Exception e) {
+                    throw new AssertionError(e);
+                }
+            }
+        }, executor);
     }
 }
--- a/ArjunaJTS/jtax/tests/classes/com/hp/mwtests/ts/jta/jts/jca/XATerminatorImpleUnitTest.java
+++ b/ArjunaJTS/jtax/tests/classes/com/hp/mwtests/ts/jta/jts/jca/XATerminatorImpleUnitTest.java
@@ -38,6 +38,7 @@ import static org.junit.Assert.fail;
 import javax.transaction.xa.XAException;
 import javax.transaction.xa.XAResource;
 
+import com.arjuna.ats.internal.jta.transaction.arjunacore.jca.SubordinateTransaction;
 import org.junit.Test;
 
 import com.arjuna.ats.arjuna.common.Uid;
@@ -46,6 +47,14 @@ import com.arjuna.ats.internal.jta.transaction.arjunacore.jca.TransactionImporte
 import com.arjuna.ats.internal.jta.transaction.jts.jca.XATerminatorImple;
 import com.arjuna.ats.jta.xa.XidImple;
 import com.hp.mwtests.ts.jta.jts.common.TestBase;
+
+import java.util.ArrayList;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CyclicBarrier;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Supplier;
 
 public class XATerminatorImpleUnitTest extends TestBase
 {
@@ -219,5 +228,63 @@ public class XATerminatorImpleUnitTest extends TestBase
         catch (final Exception ex)
         {
         }
+    }
+
+    @Test
+    public void testConcurrentImport () throws Exception {
+        AtomicInteger completionCount = new AtomicInteger(0);
+        XidImple xid = new XidImple(new Uid());
+
+        final int TASK_COUNT = 400;
+        final int THREAD_COUNT = 200;
+        final CyclicBarrier gate = new CyclicBarrier(THREAD_COUNT + 1);
+
+        ArrayList<CompletableFuture<SubordinateTransaction>> futures = new ArrayList<>();
+        ExecutorService executor = Executors.newFixedThreadPool(THREAD_COUNT);
+
+        for (int i = 0; i < TASK_COUNT; i++)
+            futures.add(doAsync(completionCount, gate, i < THREAD_COUNT, executor, xid));
+
+        gate.await();
+
+        SubordinateTransaction prevStx = null;
+
+        for (CompletableFuture<SubordinateTransaction> future : futures) {
+            SubordinateTransaction stx = future.get();
+            if (stx == null) {
+                fail("transaction import returned null for future ");
+            } else {
+                if (prevStx != null)
+                    assertEquals("transaction import for same xid returned a different instance", stx, prevStx);
+                else
+                    prevStx = stx;
+            }
+        }
+
+        assertEquals("some transaction import futures did not complete", completionCount.get(), TASK_COUNT);
+    }
+
+    /*
+     * import a transaction asynchronously to maximise the opportunity for concurrency errors in TransactionImporterImple
+     */
+    private CompletableFuture<SubordinateTransaction> doAsync(
+            final AtomicInteger completionCount, final CyclicBarrier gate, final boolean wait, ExecutorService executor, final XidImple xid) {
+        return CompletableFuture.supplyAsync(new Supplier<SubordinateTransaction>() {
+            @Override
+            public SubordinateTransaction get() {
+                try {
+                    if (wait)
+                        gate.await();
+
+                    TransactionImporter imp = SubordinationManager.getTransactionImporter();
+                    SubordinateTransaction stx = imp.importTransaction(xid);
+                    completionCount.incrementAndGet();
+
+                    return stx;
+                } catch (Exception e) {
+                    throw new AssertionError(e);
+                }
+            }
+        }, executor);
     }
 }
--- a/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/TransactionImporterImple.java
+++ b/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/TransactionImporterImple.java
@@ -32,6 +32,7 @@
 package com.arjuna.ats.internal.jta.transaction.jts.jca;
 
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicReference;
 
 import javax.transaction.xa.XAException;
 import javax.transaction.xa.Xid;
@@ -78,21 +79,8 @@ public class TransactionImporterImple implements TransactionImporter
 	{
 		if (xid == null)
 			throw new IllegalArgumentException();
-		
-		/*
-		 * Check to see if we haven't already imported this thing.
-		 */
-		
-		SubordinateTransaction imported = getImportedTransaction(xid);
-		
-		if (imported == null)
-		{	
-			imported = new TransactionImple(timeout, xid);
-			
-			_transactions.put(new XidImple(xid), imported);
-		}
-		
-		return imported;
+
+		return addImportedTransaction(null, new XidImple(xid), xid, timeout);
 	}
 
 	public SubordinateTransaction recoverTransaction (Uid actId) throws XAException
@@ -104,19 +92,8 @@ public class TransactionImporterImple implements TransactionImporter
 		
 		if (recovered.baseXid() == null)
 		    throw new IllegalArgumentException();
-		
-		TransactionImple tx = (TransactionImple) _transactions.get(recovered.baseXid());
 
-		if (tx == null)
-		{
-			recovered.recordTransaction();
-
-			_transactions.put(recovered.baseXid(), recovered);
-			
-			return recovered;
-		}
-		else
-			return tx;
+		return addImportedTransaction(recovered, recovered.baseXid(), null, 0);
 	}
     
 	/**
@@ -135,11 +112,19 @@ public class TransactionImporterImple implements TransactionImporter
 	{
 		if (xid == null)
 			throw new IllegalArgumentException();
-		
-		SubordinateTransaction tx = _transactions.get(new XidImple(xid));
-		
-		if (tx == null)
+
+		AtomicReference<SubordinateTransaction> holder = _transactions.get(new XidImple(xid));
+		SubordinateTransaction tx = holder == null ? null : holder.get();
+
+		if (tx == null) {
+			/*
+			 * Remark: if holder != null and holder.get() == null then the setter is about to
+			 * import the transaction but has not yet updated the holder. We implement the getter
+			 * (the thing that is trying to terminate the imported transaction) as though the imported
+			 * transaction only becomes observable when it has been fully imported.
+			 */
 			return null;
+		}
 
 		if (tx.baseXid() == null)
 		{
@@ -170,7 +155,41 @@ public class TransactionImporterImple implements TransactionImporter
 
 		_transactions.remove(new XidImple(xid));
 	}
-	
-	private static ConcurrentHashMap<Xid, SubordinateTransaction> _transactions = new ConcurrentHashMap<Xid, SubordinateTransaction>();
-	
+
+	private SubordinateTransaction addImportedTransaction(
+			TransactionImple importedTransaction, Xid importedXid, Xid xid, int timeout)
+	{
+		// We need to store the imported transaction in a volatile field holder so that it can be shared between threads
+		AtomicReference<SubordinateTransaction> holder = new AtomicReference<>();
+		AtomicReference<SubordinateTransaction> existing;
+
+		if ((existing = _transactions.putIfAbsent(importedXid, holder)) != null) {
+			holder = existing;
+		}
+
+		SubordinateTransaction txn = holder.get();
+
+		if (txn == null) {
+			// retry the get under a lock - this double check idiom is safe because AtomicReference is effectively
+			// a volatile so can be concurrently accessed by multiple threads
+			synchronized (holder) {
+				txn = holder.get();
+				if (txn == null) {
+					// now it's safe to add the imported transaction to the holder
+					if (importedTransaction != null) {
+						importedTransaction.recordTransaction();
+						txn = importedTransaction;
+					} else {
+						txn = new TransactionImple(timeout, xid);
+					}
+
+					holder.set(txn);
+				}
+			}
+		}
+
+		return txn;
+	}
+
+	private static ConcurrentHashMap<Xid, AtomicReference<SubordinateTransaction>> _transactions = new ConcurrentHashMap<>();
 }


ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImpleHolder.java
```diff
@@ -0,0 +1,42 @@
...
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package com.arjuna.ats.internal.jta.transaction.arjunacore.jca;
+
+import com.arjuna.ats.internal.jta.transaction.arjunacore.subordinate.jca.TransactionImple;
+
+import java.util.concurrent.CountDownLatch;
+
+public class TransactionImpleHolder {
```
dmlloyd: This is a relatively heavy/slow idiom to use for this class.  I'd recommend using an init-once pattern like this, which doesn't require CountDownLatch's heavy internal machinery:

```
    private volatile TransactionImple imported;

    public TransactionImple getImported() throws InterruptedException {
        TransactionImple imported = this.imported;
        if (imported != null) {
            return imported;
        }
        synchronized (this) {
            for (;;) {
                imported = this.imported;
                if (imported != null) {
                    return imported;
                }
                wait();
            }
        }
    }

    public void setImported(TransactionImple imported) {
        synchronized (this) {
            this.imported = imported;
            notifyAll();
        }
    }
````
Note that without the "volatile", this is unsafe "double-checked locking".

File: ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java [hunk 2]
```diff
@@ -211,17 +238,33 @@ public void removeImportedTransaction(Xid xid) throws XAException
...
 		while (iterator.hasNext()) {
-			TransactionImple next = iterator.next();
-			if (next.getParentNodeName().equals(parentNodeName)) {
-				toReturn.add(next.baseXid());
+			TransactionImpleHolder next = iterator.next();
+			TransactionImple imported = null;
+			try {
+				imported = next.getImported();
+			} catch (InterruptedException e) {
+				throw new RuntimeException(e);
```
dmlloyd: You need to re-set the interrupted status of the thread here by calling ```Thread.currentThread().interrupt()```.  Also, wrapping InterruptedException is generally not a good idea; better to just have a message that says "operation was interrupted" or something.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
```diff
@@ -167,11 +187,18 @@ public SubordinateTransaction getImportedTransaction(Xid xid)
...
-		if (tx == null)
+		if (holder == null)
 			return null;
-		
+
+		SubordinateTransaction tx = null;
+		try {
+			tx = holder.getImported();
+		} catch (InterruptedException e) {
+			throw new RuntimeException(e);  // TODO
```
dmlloyd: See below comment.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
```diff
@@ -129,21 +122,48 @@ public TransactionImple recoverTransaction(Uid actId)
...
+			// this imported transaction has not been seen before
+			TransactionImple importedTransaction = new TransactionImple(timeout, xid);
+			holder.setImported(importedTransaction);
+			return importedTransaction;
+		} else {
+			try {
+				return prevHolder.getImported();
+			} catch (InterruptedException e) {
+				e.printStackTrace();
+				throw new RuntimeException(e);  // TODO
```
dmlloyd: See below comment.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
```diff
@@ -129,21 +122,48 @@ public TransactionImple recoverTransaction(Uid actId)
...
+			holder.setImported(importedTransaction);
+			importedTransaction.recordTransaction();
+
+			return importedTransaction;
+		} else {
+			try {
+				return prevHolder.getImported();
+			} catch (InterruptedException e) {
+				e.printStackTrace();
+				throw new RuntimeException(e);  // TODO
```
dmlloyd: See below comment.

dmlloyd:
Change looks good to me, give or take a few comments on implementation details.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
mmusgrov: The contract for importTransaction (and recoverTransaction) is to either return a valid TransactionImple or to throw an XAException. Therefore if the wait() call is interrupted the only option is to wrap the InterruptedException as either a RuntimeException of an XA_RETRY exception.

mmusgrov:
@dmlloyd that last commit includes your recommendations except the one about the InterruptedException. I went for the following (since the contract for importTransaction() returns either a valid transaction or an XAException):

                try {
                    wait();
                } catch (InterruptedException e) {
                    jtaLogger.i18NLogger.warn_transaction_import_interrupted(e);
                    Thread.currentThread().interrupt();
                    throw new XAException(XAException.XA_RETRY);
                }

dmlloyd:
That looks good to me in terms of concurrency and interrupt handling.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
```diff
@@ -80,7 +83,7 @@ public SubordinateTransaction importTransaction(Xid xid)
 	 *             thrown if there are any errors.
 	 */
 
-	public SubordinateTransaction importTransaction(Xid xid, int timeout)
+ 	public SubordinateTransaction importTransaction(Xid xid, int timeout)
```
tomjenkinson: superfluous formatting change?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
```diff
@@ -121,7 +114,7 @@ public TransactionImple recoverTransaction(Uid actId)
 
 		if (recovered.baseXid() == null)
 		    throw new IllegalArgumentException();
-		
+
```
tomjenkinson: superfluous formatting change?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
```diff
@@ -167,11 +176,16 @@ public SubordinateTransaction getImportedTransaction(Xid xid)
...
+		if (holder == null)
+			return null;
+
+		SubordinateTransaction tx = holder.getImported();
 
 		if (tx == null)
 			return null;
-		
+
```
tomjenkinson: superfluous formatting change? but main point is can tx be null now? I would say not

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImpleHolder.java
```diff
@@ -0,0 +1,65 @@
...
+
+    public TransactionImple getImported() throws XAException {
+        TransactionImple imported = this.imported;
+
+        if (imported != null) {
+            return imported;
+        }
+
+        synchronized (this) {
+            for (;;) {
```
tomjenkinson: don't like the for loop, looks dangerous

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImpleHolder.java
```diff
@@ -0,0 +1,65 @@
...
+                if (imported != null) {
+                    return imported;
+                }
+
+                try {
+                    wait();
+                } catch (InterruptedException e) {
+                    jtaLogger.i18NLogger.warn_transaction_import_interrupted(e);
+                    Thread.currentThread().interrupt();
+                    throw new XAException(XAException.XA_RETRY);
```
tomjenkinson: how about just rethrowing the IE? Seems artificial to wrap in XAE

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImpleHolder.java
dmlloyd: There's an exit condition, it's just in mid-loop.  It could be stylistically mitigated in various ways.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImpleHolder.java
mmusgrov: The contract for importTransaction says either return a valid transaction or throw an XAE.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
mmusgrov: Fair point (I was attempting to minimise the number of changes) I will delete the check

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
mmusgrov: I will sort these out when I add the unit test

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImpleHolder.java
mmusgrov: I think Tom is right. The code that does the set is:

	TransactionImple importedTransaction = new TransactionImple(timeout, xid);
	holder.setImported(importedTransaction);

If the new TransactionImple call throws a runtime exception then setImported will never get called so the for loop will not terminate.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImpleHolder.java
tomjenkinson: Mike looks to be erring on acceding to my observation (based on his comment below). I would just add return imported after the wait and drop the loop then you catch all cases.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImpleHolder.java
tomjenkinson: Thanks, and re XAE OK I see now. It is used elsewhere in that class. Are you sure RETRY is the right code? How about fail or something? Maybe a comment to why you selected RETRY at least.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
```diff
@@ -211,17 +225,35 @@ public void removeImportedTransaction(Xid xid) throws XAException
...
 			}
 		}
 		return toReturn;
 	}
 
-	private static ConcurrentHashMap<SubordinateXidImple, TransactionImple> _transactions = new ConcurrentHashMap<SubordinateXidImple, TransactionImple>();
+	private XidImple convertXid(Xid xid) {
+		if (xid != null && xid.getFormatId() == XATxConverter.FORMAT_ID) {
+			XidImple toImport = new XidImple(xid);
+			XATxConverter.setSubordinateNodeName(toImport.getXID(), TxControl.getXANodeName());
```
tomjenkinson: I am interested because setSNN() was not used in the original file. Please can you explain why its required now?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImpleHolder.java
dmlloyd: wait() always must be retried, as it is subject to spurious wakeups.  It is almost never correct to wait without a loop.  But it sounds like there needs to be a failure mode for the loop as well.

Making the "imported" field be of type Object is one way to accomplish this.  On success it will be a TransactionImple, and on failure it will be something else - an exception object perhaps, or a special singleton Object instance (sentinel) that represents the failure case and causes the method to throw an exception.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
mmusgrov: The TransactionImple constructor calls setSNN(). I need that xid to put the (initially) empty TransactionImpleHolder into the map. When it's in the map I then construct the TransactionImple instance and add it to the holder.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImpleHolder.java
tomjenkinson: OK, that might be a solution if we want to handle RTE in the ctor of transactionimple

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
tomjenkinson: Ah, OK it could be worth a comment then as it was strange

mmusgrov:
@dmlloyd I have updated the wait/notify code to avoid unbounded waits for the setter to be called (and I also added a test the fails without the fix)

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImpleHolder.java
mmusgrov: I opted for reporting failures via XA_RETRY since failures can occur if TransactionImple ctor fails or if we time out on a busy system. In both cases a retry of the operation may succeed. I added a comment to the code where I throw the exception.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
mmusgrov: I added the comment at the place I make the call:

    SubordinateXidImple importedXid = new SubordinateXidImple(convertXid(xid));

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImpleHolder.java
```diff
@@ -0,0 +1,87 @@
...
+     * milliseconds or if the wait is interrupted.
+     */
+    public TransactionImple getImported() throws XAException {
+        TransactionImple imported = this.imported;
+
+        if (imported != null) {
+            return imported;
+        }
+
+        long maxMillis = System.currentTimeMillis() + MAX_MS_WAIT_TIME;
```
dmlloyd: Timeouts should always be based on System.nanoTime() because that clock is not affected by changes to the system clock via NTP or whatever.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImpleHolder.java
```diff
@@ -0,0 +1,87 @@
...
+                }
+
+                if (maxMillis < System.currentTimeMillis()) {
+                    jtaLogger.i18NLogger.warn_transaction_import_timedout();
+                    // throw XA_RETRY since the wait may have timed out because the system is loaded
+                    throw new XAException(XAException.XA_RETRY);
+                }
+
+                try {
+                    wait(1000); // timed wait to protect against the setter never getting called
```
dmlloyd: Arbitrary timeouts like this, if hit, indicate a either bug in the code because a precondition isn't met, or a bug in the code because the timeout expired even though an illegitimate result was produced.  Overall I wouldn't use this approach in any case.  Just ensure that all code paths result in the condition being signalled.

mmusgrov: Well that make sense. @tomjenkinson are you okay with just making sure that our code will allways invoke the setter (I ask because it means keeping the unbounded for (;;) loop)?

tomjenkinson: wait(MAX_MS_WAIT_TIME)

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImpleHolder.java
tomjenkinson: Lets check what we do with other timeouts in Narayana. If they are all millis lets add this as millis but raise an issue to update them all - at least then the code is consistent.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
tomjenkinson: OK, thanks

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImpleHolder.java
mmusgrov: Well David does not like the timed waits (and I am inclined to agree) so I have gone back to the for (;;) loop but I ensure that the setter is called for all code paths to guarantee the loop terminates.

If there is a problem then the setter is called with a null value and when the getter is subsequently called I return an XA_RETRY exception. I use XA_RETRY because I still think the failure is a transient condition so if the caller wishes he can retry the op.

ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/jtaI18NLogger.java
```diff
@@ -470,6 +470,14 @@
     @LogMessage(level = WARN)
     public void warn_could_not_access_object_store(@Cause() Exception e);
 
+    @Message(id = 16116, value = "Could not set imported transaction", format = MESSAGE_FORMAT)
+    @LogMessage(level = WARN)
+    void warn_transaction_import_interrupted(@Cause() InterruptedException e);
+
+	@Message(id = 16117, value = "Timed out waiting imported transaction", format = MESSAGE_FORMAT)
+	@LogMessage(level = WARN)
+	void warn_transaction_import_timedout();
```
tomjenkinson: No longer used

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/jca/XATerminatorUnitTest.java
```diff
@@ -374,4 +383,67 @@ public void testInvalid () throws Exception
...
+     */
+    private CompletableFuture<SubordinateTransaction> doAsync(
+            final CyclicBarrier gate, final boolean wait, ExecutorService executor, final XidImple xid) {
+        return CompletableFuture.supplyAsync(new Supplier<SubordinateTransaction>() {
+            @Override
+            public SubordinateTransaction get() {
+                try {
+                    if (wait)
+                        gate.await();
+                    SubordinateTransaction stx = SubordinationManager.getTransactionImporter().importTransaction(xid);
```
tomjenkinson: I wonder if some byteman rule atEntry on _transactions.putIfAbsent would be useful? That said then we are unit testing the putIfAbsent code.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
```diff
@@ -122,6 +112,14 @@ public TransactionImple recoverTransaction(Uid actId)
...
 		    throw new IllegalArgumentException();
 		
+		return addImportedTransaction(recovered);
+	}
+
+	private TransactionImple addImportedTransaction(TransactionImple importedTransaction) throws XAException {
+		SubordinateXidImple importedXid = new SubordinateXidImple(importedTransaction.baseXid());
+		TransactionImpleHolder holder = new TransactionImpleHolder();
+		TransactionImpleHolder prevHolder = _transactions.putIfAbsent(importedXid, holder);
+
```
mmusgrov: I think that if the thread is interrupted after putting the holder in the map but before the try block on line 131 is entered then the setter will not be called so I think I should be protecting against this. A similar comment applies to the other addImportedTransaction method.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/jca/XATerminatorUnitTest.java
mmusgrov: That would have worked too but my test does highlight the problem and it does demonstrate that the fix is good.

tomjenkinson: OK thanks

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
tomjenkinson: Does putIfAbsent do a wait/sleep though? There are some exceptions that could be thrown though: https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentMap.html#putIfAbsent(K,%20V)

dmlloyd: In practice, none of these exceptions apply to ConcurrentHashMap with this code.  Basically, they all relate to preconditions which are met; you don't have to worry about environmental factors causing sporadic behavior changes in this case.

dmlloyd:
I was examining this again and I wonder if the whole approach isn't overdone.  We don't really need a condition; just having a single point of synchronized initialization should be enough.

Pseudo-code:
```
    AtomicReference<Txn> holder = new AtomicReference<>(); // this class is an already-existing volatile field holder, might as well use it
    AtomicReference<Txn> existing;
    if ((existing = map.putIfAbsent(key, holder)) != null) {
        holder = existing;
    }
    Txn txn = holder.get();
    if (txn == null) {
        synchronized (holder) {
            txn = holder.get();
            if (txn == null) {
               txn = // do creation work under this lock; don't care if exception is thrown
               holder.set(txn);
            }
        }
    }
```
This eliminates the condition and greatly simplifies the concurrency reasoning, and allows each entry point to have their own attempt at creating the transaction.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
mmusgrov: @tomjenkinson I reasoned that none of those putIfAbsent exceptions applied - besides if the put failed then it would show up as a RTE (ie no concurrency issues).

@dmlloyd I mean't that if some external agent interrupted the thread *after* the empty holder was added but before the setter was called then the setter would never get called.

But let me take a look at the suggestion that @dmlloyd gave regarding using a single point of synchronized initialization.

dmlloyd: Interruption doesn't actually affect the program flow; it only comes into play once something either checks the flag manually or uses a construct (like Object.wait()) which is directly sensitive to interruption.

I can see no statements between the add and the set in the finally block which can trigger interruption in any way that is detectable before the setter is called.  The risk point would probably be line 140, but in this case the caller isn't the setting thread so again it doesn't influence the final set operation as far as I can see.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
```diff
@@ -167,11 +181,19 @@ public SubordinateTransaction getImportedTransaction(Xid xid)
...
-		if (tx == null)
+		if (holder == null)
 			return null;
-		
+
+		SubordinateTransaction tx = holder.get();
+
+		if (tx == null) {
+			// another thread must be in the process of importing this transaction
+			// TODO the solution does not provide a mechanism to wait for the other thread to update the holder
```
mmusgrov: This solution does not provide a mechanism to wait for the other thread to update the holder

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
```diff
@@ -211,17 +233,35 @@ public void removeImportedTransaction(Xid xid) throws XAException
...
 		while (iterator.hasNext()) {
-			TransactionImple next = iterator.next();
-			if (next.getParentNodeName().equals(parentNodeName)) {
-				toReturn.add(next.baseXid());
+			AtomicReference<TransactionImple> next = iterator.next();
+			TransactionImple imported = next.get();
+
+			if (imported == null) {
+				// another thread must be in the process of importing this transaction
+				// TODO the solution does not provide a mechanism to wait for the other thread to update the holder
```
mmusgrov: This solution does not provide a mechanism to wait for the other thread to update the holder

mmusgrov:
@dmlloyd I implemented your recommendation about holding the volatile in an AtomicReference. But how can I efficiently implement the getter. I added two TODO notes above to illustrate where the getter would need to wait for the setter.

dmlloyd:
The getter should just return the value; if the value isn't set, then the lock should be acquired and the get retried; if the value is still null then the transaction does not exist and needs to be created.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
tomjenkinson: I added https://github.com/mmusgrov/narayana/pull/2 but it requires public API change

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
```diff
@@ -121,30 +117,48 @@ public TransactionImple recoverTransaction(Uid actId)
...
+			holder = existing;
 		}
-		else
-		{
-			return tx;
+
+		TransactionImple txn = holder.get();
+
+		if (txn == null) {
+			synchronized (holder) {
```
tomjenkinson: I changed this a bit in https://github.com/mmusgrov/narayana/pull/2 - I think we can use the existing == null check here to reduce the number of holder.get calls

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
tomjenkinson: Actually its a public method in an a .internal. file so not breaking but I think david may be calling this method

mmusgrov:
@dmlloyd the getter needs to return null if the transaction does not exist, it should not create one - ie the getter needs to wait for the thread doing the set to finish so it needs some kind of wait/notify logic like we had in the previous commit.

@tomjenkinson your modification to wait for the setter to complete is going back to the solution where I introduced the TransactionImpleHolder object that did the wait/notify logic. Also changing the signature of importTransaction is changing a public interface so has knock on effects.

mmusgrov:
I was looking at

    public Set<Xid> getInflightXids(String parentNodeName) throws InterruptedException 

and made an assumption about the other calls. But this method is only used by our test code so I withdraw my second comment.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
dmlloyd: You cannot; the value of holder may change outside of the synchronization.  Also holder.get() is very very cheap.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
dmlloyd: It shouldn't matter because the act of starting the transaction is atomic - that is, it is either observed to have happened, or observed to not have happened.  If the operation is in progress, then it hasn't happened yet.

However you can work around this by adding another synch block in here that re-gets the value and re-checks for null, if you want to guarantee that there is no transaction *and* there is no operation in progress to produce one.  This preserves the atomicity of the operation while also biasing this operation a little bit towards success.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
tomjenkinson: OK, lets go back to a new class then with a holder reference that isn't an atomicreference if you are saying the underlying object that the reference points to can change dynamically and not be relied on as you can't ever get a sync on it.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
tomjenkinson: I can agree with that actually if we are returning XARETRY we can make the work on the caller.

dmlloyd:
The wait is simply unnecessary.  Look at all the worrying over interruption it has caused. :)

In terms of the getter returning the in-progress transaction, see my previous comment about how to work around that, but more broadly it's about establishing a contract.

Specifically, in a concurrent system, it must be determined when a transaction becomes visible to the world outside of the creating thread.

The existing approach effectively has the transaction be invisible until the create operation returns; that is, you are not guaranteed to observe the new transaction until/unless you have observed (or indirectly have inferred the observation) that the create method has returned.

Using the trick I showed gives a different guarantee: that the transaction is visible as soon as an attempt (which will ultimately be successful) to create the transaction has been observed (or is able to be observed).  If the attempt is ultimately unsuccessful, then (apart from a hiccup/time delay, which generally should have no effect in a well-designed program) the getter thread does not observe it at all.

Either way is a valid approach; it really comes down to what guarantees you want to provide in your API contracts.  Neither approach requires a condition.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
dmlloyd: I am not following you.  The way that AtomicReference is used here is as a holder for a volatile field that also serves as a synch target; by convention, we only change its value (and only from null to non-null) while the synch is held.  Therefore you can read it non-synch; if the result is not null then it will never change.  If the result is null then you have to get the lock and re-try the get.  If the get is still null, then no operation is in progress to set it, and we are free to try ourselves or to admit defeat (depending on what method is in play); if it is not null, then the result will (again) never change and can be returned as is.

mmusgrov:
Thanks for the rationale @dmlloyd and I agree with your analysis - it is our choice as to when the imported transaction becomes visible. NB if I used your trick (of creating the transaction if the getter finds some other thread is about to import it) then we would also need to store the timeout value for the transaction in the holder which adds complexity to the setter logic (because the new TransactionImple(timeout, xid) requires both the xid and timeout).

@tomjenkinson I will clean up the TODOs and mark the PR as ready for review - I think the only contentious issue is when we make the imported transaction visible:

1. As soon as the holder object is created;
2. After the point at which the setter is called on the holder object 

2) is what's currently in the PR and I propose to use that.

dmlloyd:
@mmusgrov  the trick I was referring to was the one from https://github.com/jbosstm/narayana/pull/970/files#r50977768 where you simply retry the get under the lock before you give up and return null, which has the effect of changing when you observe the transaction from "any time after the transaction creation returns" to "any time after the transaction creation was initiated (iff it is successful)".  So you don't have to try and create the transaction in the getter to realize this change.

mmusgrov:
@dmlloyd The setter thread does:

1 _transactions.putIfAbsent(importedXid, holder)
2 some statements
3 synchronized (holder) {
   create the txn

whilst the getter thread does:

4 holder = _transactions.get(importedXid,);
5 if (holder != null && holder.get() == null)
  apply your trick

But the trick of retrying under the lock does not provide any benefit if the setter thread is at step 2 (but I agree it would work if the setter is in step 3). Is your point that step 3 takes much longer than step 2 so the getter is more likely to see the new transaction with this extra check under the lock?

dmlloyd:
This distinction does not make a difference because (from the getter thread) it is otherwise impossible to distinguish between the setter thread being anywhere from before step 1 to before step 3.  The only reason it would matter would be if there was some other event in the setter thread which guarantees that the transaction would be available, which is observable from the getter thread.  That's also why the synchronization trick is not required - its effects are unobservable from any well-defined program.

Concurrency is all about what you can guarantee and what you can observe.

dmlloyd:
Consider that a fickle OS scheduler might decide to freeze the setter thread before step 1 for an indefinite amount of time.  How would the getter thread know this happened, and why would it care?

mmusgrov:
But my point is that the getter thread can know for certain that the transaction is about to be added if the setter is in step 2 since the getter finds that the holder exists in the map (step 1 has happened) but does not yet contain the transaction and rechecking under the lock does not guarantee the getter will see the new transaction.

I agree that it is our choice as to when we deem the transaction to be visible to the getter but I do not see the benefit of the getter retrying under the lock (unless its to increase the probability that it will see the transaction the setter is about to create).

dmlloyd:
Sure, you can do it either way; like I said, the only affect it has is to slightly bias the getter towards success when there are no other factors in play.  But either way it's essentially undefinable as to whether (and when) the getter sees a transaction that is about to be (or is in the process of) being created, so from a user's perspective it probably doesn't matter: any user program whose behavior depends on this is already not well-defined (i.e. it's defective).

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
tomjenkinson: bump

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
```diff
@@ -167,11 +151,12 @@ public SubordinateTransaction getImportedTransaction(Xid xid)
...
 
-		SubordinateTransaction tx = _transactions.get(new SubordinateXidImple(xid));
+		AtomicReference<TransactionImple> holder = _transactions.get(new SubordinateXidImple(xid));
+		TransactionImple tx = extractImportedTransaction(holder);
 
 		if (tx == null)
-			return null;
-		
+		    return null;
+
```
tomjenkinson: formatting changes?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
```diff
@@ -211,17 +196,85 @@ public void removeImportedTransaction(Xid xid) throws XAException
...
+		}
+
+		TransactionImple txn = holder.get();
+
+		if (txn == null) {
+			// retry the get under a lock - this double check idiom is safe because AtomicReference is effectively
+			// a volatile so can be concurrently accessed by multiple threads
+			synchronized (holder) {
+				txn = holder.get();
+				if (txn == null) {
```
tomjenkinson: This still seems totally wierd and broken to me - why are we not just checking if "existing != null"? Isn't that the whole point of putIfAbsent. Does a double check of holder.get() == null somehow guarantee we put the holder in the map but even then why isn't "existing == null" clearer?

dmlloyd: Because adding the reference to the map is a distinct atomic action compared to initializing the reference.  If existing is null, all that means is, we were the one to add the reference to the map.  If two threads run at the same time, it is legitimate for the other thread to add the reference and this thread to initialize it; the identity of the reference being added is unimportant as long as it is only added once and all threads recognize the same instance.

What you suggest is possible, but to do it, the reference would have to have been synchronized by the put thread before it is added the map, preventing readers from seeing empty, unlocked reference.  This is also a legitimate solution to the problem, though it might add some complexity to the code due to the strict lexical scoping of the synchronization.

In either case, you could add an optimistic _transactions.get() first, before taking the lock, in order to avoid the expense of the instantiation and (now) lock acquisition if possible (not perfectly of course, but in all but the raciest cases it would yield the optimal path: if the get fails, the putIfAbsent is likely to succeed and thus all the extra work (such as it is) likely to be worthwhile).

tomjenkinson: OK, looks fine to me then. I thought you were saying previously that the holder atomicreference could point to different objects at any point (which it can, but I read it as though it applied in this specific code) so I was thinking that synchronizing on it would be pointless. If I follow you correctly you are saying by the time the synchronized (holder) is called the object that the atomicreference points at is fixed. The wording in the comment of "effectively a volatile" actually applies in a beneficial sense - an initial reading of it makes the code look scarier but actually it means that synchronizing on the holder (declared as a local variable) will sync two threads.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
```diff
@@ -211,17 +196,85 @@ public void removeImportedTransaction(Xid xid) throws XAException
...
+
+		TransactionImple tx = holder.get();
+
+		if (tx == null) {
+			/*
+			 * The setter is about to import the transaction but has not yet updated the holder.
+			 * To increase the probability that this thread sees the transaction we retry the get under the
+			 * synchronization lock
+			 */
+			synchronized (holder) {
```
tomjenkinson: does this really increase the probability in a statistically significant way?

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/jca/XATerminatorUnitTest.java
```diff
@@ -374,4 +383,61 @@ public void testInvalid () throws Exception
...
+                    completionCount.incrementAndGet();
+
+                    return stx;
+                } catch (Exception e) {
+                    e.printStackTrace();
```
tomjenkinson: printStackTrace

ArjunaJTS/jtax/tests/classes/com/hp/mwtests/ts/jta/jts/jca/XATerminatorImpleUnitTest.java
```diff
@@ -220,4 +229,63 @@ public void testNull () throws Exception
...
+                    if (wait)
+                        gate.await();
+
+                    TransactionImporter imp = SubordinationManager.getTransactionImporter();
+                    SubordinateTransaction stx = imp.importTransaction(xid);
+                    completionCount.incrementAndGet();
+
+                    return stx;
+                } catch (Exception e) {
+                    e.printStackTrace();
```
tomjenkinson: pST()

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
tomjenkinson: If you look at uses of getImportedTransaction they might need updating if it is at all possible that holder.get can be null. They don't return XARETRY - they generally assume that a protocol error happened

tomjenkinson: ok, I see what you mean about importTransaction being concurrently executing

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
tomjenkinson: bump

PR #973, opened 2016-02-02
JBTM-2618 Improve concurrent behavior of HornetqJournalStore

barreiro:
https://issues.jboss.org/browse/JBTM-2618

!XTS !BLACKTIE !QA_JTA !QA_JTS_JDKORB

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/hornetq/HornetqJournalStore.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/hornetq/HornetqJournalStore.java
@@ -28,6 +28,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicLong;
 
 import org.apache.activemq.artemis.core.journal.Journal;
 import org.apache.activemq.artemis.core.journal.JournalLoadInformation;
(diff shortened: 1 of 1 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/hornetq/HornetqJournalStore.java
```diff
@@ -182,15 +182,15 @@ public boolean write_committed(Uid uid, String typeName, OutputObjectState txDat
             long id = getId(uid, typeName);
             byte[] data = outputBuffer.buffer();
 
-            // yup, there is a race condition here.
-            if(getContentForType(typeName).containsKey(uid)) {
+            RecordInfo record = new RecordInfo(id, RECORD_TYPE, data, false, (short)0);
+            RecordInfo previousRecord = getContentForType(typeName).putIfAbsent(uid, record);
```
tomjenkinson: We would need to remove this if the append failed, happy - can be added to the existing exception handler (line 195)

mmusgrov:
Yes it looks like a useful change. When jenkins comes back on line it will be interesting to see what impact it has on performance.

mmusgrov:
I ran it manually on my laptop and found a marginal improvement. But do bear in mind that getting consistent results out of JMH is tricky:

runs against pr/973

threads maxIO   bf/s    AIO     tput 
100     500     2000    true    45411.191
100     500     2000    true    45511.204

run against master

threads maxIO   bf/s    AIO     tput 
100     500     2000    true    45390.476
100     500     2000    true    45317.320

mmusgrov:
So if people do not think my comment has merit then this PR would look good to merge.

PR #976, opened 2016-02-08
JBTM-2409 Updated to wait while first pass is complete then only bloc…

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2409

!BLACKTIE !QA_JTA !QA_JTS_JDKORB !XTS !PERF


zhfeng:
It looks good to me.

PR #979, opened 2016-02-18
[JBTM-2622] Implement possible new compensations api from the meeting

gytis:
https://issues.jboss.org/browse/JBTM-2622

!MAIN !BLACKTIE !PERF !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_IBMORB

### DIFF
--- a/compensations/src/main/java/org/jboss/narayana/compensations/api/Action.java
+++ b/compensations/src/main/java/org/jboss/narayana/compensations/api/Action.java
@@ -1,0 +1,32 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.jboss.narayana.compensations.api;
+
+/**
+ * @author <a href="mailto:gytis@redhat.com">Gytis Trikleris</a>
+ */
+public interface Action {
+
+    void execute();
+
+}
--- a/compensations/src/main/java/org/jboss/narayana/compensations/api/Transaction.java
+++ b/compensations/src/main/java/org/jboss/narayana/compensations/api/Transaction.java
@@ -1,0 +1,41 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.jboss.narayana.compensations.api;
+
+/**
+ * @author <a href="mailto:gytis@redhat.com">Gytis Trikleris</a>
+ */
+public interface Transaction {
+
+    Transaction doAction(Action action);
+
+    Transaction doAction(Action action, CompensationHandler compensationHandler);
+
+    Transaction doAction(Action action, ConfirmationHandler confirmationHandler);
+
+    Transaction doAction(Action action, CompensationHandler compensationHandler,
+            ConfirmationHandler confirmationHandler);
+
+    void execute();
+
+}
--- a/compensations/src/main/java/org/jboss/narayana/compensations/impl/TransactionImpl.java
+++ b/compensations/src/main/java/org/jboss/narayana/compensations/impl/TransactionImpl.java
@@ -1,0 +1,157 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.jboss.narayana.compensations.impl;
+
+import org.jboss.narayana.compensations.api.Action;
+import org.jboss.narayana.compensations.api.Compensatable;
+import org.jboss.narayana.compensations.api.CompensationHandler;
+import org.jboss.narayana.compensations.api.ConfirmationHandler;
+import org.jboss.narayana.compensations.api.Transaction;
+
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * @author <a href="mailto:gytis@redhat.com">Gytis Trikleris</a>
+ */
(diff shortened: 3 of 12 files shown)


gytis:
I've raised a PR, so that it would be easier to comment on the code. Also, will see if the tests pass on CI.
"[JBTM-2622] Possible new compensations api from the meeting" is the commit demonstrating the suggested API. Other commits were to prepare the current implementation for the change.

gytis:
I guess you're right. I though that would be the fastest way to get some feedback. Of course I'll post that on the forum too.

compensations/src/main/java/org/jboss/narayana/compensations/api/Transaction.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package org.jboss.narayana.compensations.api;
+
+/**
+ * @author <a href="mailto:gytis@redhat.com">Gytis Trikleris</a>
+ */
+public interface Transaction {
+
+    Transaction doAction(Action action);
```
tomjenkinson: Is there any point of an action that does not have any kind of handler?

compensations/src/main/java/org/jboss/narayana/compensations/api/Action.java
```diff
@@ -0,0 +1,32 @@
...
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.jboss.narayana.compensations.api;
+
+/**
+ * @author <a href="mailto:gytis@redhat.com">Gytis Trikleris</a>
+ */
+public interface Action {
```
tomjenkinson: Is this some kind of well know lambda thing? I can't see what implements this interface but can see you passing in a lambda in the test.

tomjenkinson: I guess it should be something like this: https://docs.oracle.com/javase/tutorial/java/javaOO/methodreferences.html

compensations/src/main/java/org/jboss/narayana/compensations/api/Action.java
```diff
@@ -0,0 +1,32 @@
...
+
+    void execute();
```
tomjenkinson: Again, I just don't follow what implements this sorry but I imagine something does

compensations/src/main/java/org/jboss/narayana/compensations/api/Transaction.java
```diff
@@ -0,0 +1,41 @@
...
+
+    Transaction doAction(Action action, CompensationHandler compensationHandler);
```
tomjenkinson: Can these be lambda references somehow?

compensations/src/main/java/org/jboss/narayana/compensations/api/Transaction.java
```diff
@@ -0,0 +1,41 @@
...
+
+    Transaction doAction(Action action, ConfirmationHandler confirmationHandler);
+
+    Transaction doAction(Action action, CompensationHandler compensationHandler,
+            ConfirmationHandler confirmationHandler);
+
+    void execute();
```
tomjenkinson: executeAndCommit() or commit() maybe better name?

compensations/src/test/java/org/jboss/narayana/compensations/functional/transaction/TransactionLocalTest.java
```diff
@@ -0,0 +1,128 @@
...
+        return archive;
+    }
+
+    @Test
+    public void successfulScenario() {
+        final Counter actionsCounter = new Counter();
+        final Counter compensationsCounter = new Counter();
+        final Counter confirmationsCounter = new Counter();
+
+        transaction
```
tomjenkinson: looks good - its why my questions about Action are ambiguous I don't see how these map to "Action" class

compensations/src/main/java/org/jboss/narayana/compensations/impl/remote/RemoteBAControler.java
```diff
@@ -105,4 +105,10 @@ public ParticipantManager enlist(Class<? extends CompensationHandler> compensati
         return new RemoteParticipantManager(pm);
     }
 
+    @Override
+    public ParticipantManager enlist(CompensationHandler compensationHandler, ConfirmationHandler confirmationHandler, TransactionLoggedHandler transactionLoggedHandler) throws Exception {
+
+        throw new RuntimeException("Not implemented.");
```
tomjenkinson: I don't understand why this is done

compensations/src/test/java/org/jboss/narayana/compensations/functional/transaction/TransactionLocalTest.java
```diff
@@ -0,0 +1,128 @@
...
+                            (ConfirmationHandler) confirmationsCounter::increment)
+                    .doAction(() -> {
+                                throw new RuntimeException("Test");
+                            },
+                            compensationsCounter::increment,
+                            confirmationsCounter::increment)
+                    .execute();
+            fail("RuntimeException expected");
+        } catch (RuntimeException e) {
+            // Expected
```
tomjenkinson: what is it that raises the RTE?

compensations/src/main/java/org/jboss/narayana/compensations/impl/TransactionImpl.java
```diff
@@ -0,0 +1,157 @@
...
+    @Compensatable
+    public void execute() {
+        actions.forEach(actionInfo -> {
+            Set<ParticipantManager> managers = enlist(
+                    actionInfo.compensationHandler,
+                    actionInfo.confirmationHandler);
+            try {
+                actionInfo.action.execute();
+            } catch (RuntimeException e) {
+                exit(managers);
```
tomjenkinson: I don't get all these RTE handlers I think you need an exception type somewhere

compensations/src/main/java/org/jboss/narayana/compensations/impl/TransactionImpl.java
```diff
@@ -0,0 +1,157 @@
...
+            }
+        });
+    }
+
+    private void exit(Set<ParticipantManager> managers) {
+        managers.forEach(pm -> {
+            try {
+                pm.exit();
+            } catch (Exception e) {
+                e.printStackTrace();
```
tomjenkinson: Not good, needs logging

compensations/src/main/java/org/jboss/narayana/compensations/impl/TransactionImpl.java
```diff
@@ -0,0 +1,157 @@
...
+
+        return managers;
+    }
+
+    private void complete(Set<ParticipantManager> managers) {
+        managers.forEach(pm -> {
+            try {
+                pm.completed();
+            } catch (Exception e) {
+                e.printStackTrace();
```
tomjenkinson: Not good, needs logging

compensations/src/main/java/org/jboss/narayana/compensations/api/Action.java
gytis: Lambda is an implementation of an interface which has only one method. When you pass in a lambda, you actually are passing in an instance of the object (lambda at the end of the day is a syntactic sugar). While writing lambda you don't need to specify specific interface you're implementing, it still needs to be defined in the method which takes lambda as an argument.

compensations/src/main/java/org/jboss/narayana/compensations/api/Transaction.java
gytis: Probably not, I've added it here to cover all scenarios, but we can easily remove it.

compensations/src/main/java/org/jboss/narayana/compensations/api/Transaction.java
gytis: When using this Transaction interface you can implement Action and CompensationHandler using Lambdas.

compensations/src/main/java/org/jboss/narayana/compensations/api/Transaction.java
gytis: Agreed.

compensations/src/main/java/org/jboss/narayana/compensations/impl/TransactionImpl.java
gytis: Since Compensations API is based on JTA 1.2, they have the same exception handling semantics. Here's the extract from the spec:

> By default checked exceptions do not result in the transactional interceptor
> marking the transaction for rollback and instances of RuntimeException and its
> subclasses do.

compensations/src/main/java/org/jboss/narayana/compensations/impl/TransactionImpl.java
gytis: Agreed, it's only for the POC.

compensations/src/main/java/org/jboss/narayana/compensations/impl/TransactionImpl.java
gytis: Agreed, it's only for the POC.

compensations/src/main/java/org/jboss/narayana/compensations/impl/remote/RemoteBAControler.java
gytis: It's not used in the example, so I left it unimplemented. In the real implementation this will me implemented properly.

compensations/src/test/java/org/jboss/narayana/compensations/functional/transaction/TransactionLocalTest.java
gytis: "actionsCounter::increment" is the same as "actionsCounter -> actionCounter.increment()" or the same as "(actionsCounter) -> actionCounter.increment()". First option is suggested by the IDE, since implementation is only one line and calls method without arguments.
All three above examples are implementations of Action interface which has one method "execute()". You don't need to specify that explicitly, but since Transaction.doAction(Action action) is defined, compiler does the checking.

compensations/src/test/java/org/jboss/narayana/compensations/functional/transaction/TransactionLocalTest.java
```diff
@@ -0,0 +1,128 @@
...
+        try {
+            transaction
+                    .doAction(actionsCounter::increment)
+                    .doAction(actionsCounter::increment,
+                            compensationsCounter::increment,
+                            confirmationsCounter::increment)
+                    .doAction(actionsCounter::increment,
+                            (CompensationHandler) compensationsCounter::increment)
+                    .doAction(actionsCounter::increment,
+                            (ConfirmationHandler) confirmationsCounter::increment)
```
gytis: Here example from above with implied interface implementation does not work. Unspecified implementation is ambiguous for the compiler, because there are two Transaction.doAction methods which take two arguments. Therefore, we need to specifically cast lambda to tell compiler which actual method are we using.

compensations/src/test/java/org/jboss/narayana/compensations/functional/transaction/TransactionLocalTest.java
gytis: RuntimeException causes compensation to be canceled as per explanation about JTA 1.2 above.

PR #981, opened 2016-02-23
JBTM-2625 If the object store cannot be contacted return INVALID state

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2625
!BLACKTIE !XTS !PERF

### DIFF
--- a/ArjunaCore/txoj/classes/com/arjuna/ats/internal/txoj/recovery/RecoveredTransactionalObject.java
+++ b/ArjunaCore/txoj/classes/com/arjuna/ats/internal/txoj/recovery/RecoveredTransactionalObject.java
@@ -104,32 +104,31 @@ public class RecoveredTransactionalObject extends StateManager
                 txojLogger.logger.debug("RecoveredTransactionalObject - transaction status "+ActionStatus.stringForm(tranStatus));
             }
 
-            boolean inactive = false;
-
-            if (tranStatus == ActionStatus.INVALID) // should be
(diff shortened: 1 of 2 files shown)


ArjunaCore/txoj/classes/com/arjuna/ats/internal/txoj/recovery/RecoveredTransactionalObject.java
```diff
@@ -104,32 +104,31 @@ protected final void replayPhase2 ()
...
+				(tranStatus == ActionStatus.H_COMMIT) ||
+				(tranStatus == ActionStatus.H_MIXED) ||
+				(tranStatus == ActionStatus.H_HAZARD))
+			{
+				commit();
+			}
+			else if ((tranStatus == ActionStatus.ABORTED) ||
+				(tranStatus == ActionStatus.H_ROLLBACK) ||
+				(tranStatus == ActionStatus.ABORTING) ||
+				(tranStatus == ActionStatus.ABORT_ONLY))
```
mmusgrov: One could also make a case for calling abort if the ActionStatus is NO_ACTION. And also for CLEANUP, although I cannot find any code that uses the CLEANUP status.

This is my only comment.

tomjenkinson: I took this code from other replayPhase2 - I think we should go with the argument of precedence

PR #983, opened 2016-02-24
JBTM-2630 Updated to remove the zip as its tgz on linux (and zip no longer work…

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2630

!MAIN !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !PERF

### DIFF
--- a/blacktie/blacktie/pom.xml
+++ b/blacktie/blacktie/pom.xml
@@ -101,6 +101,8 @@
             </plugin>
             <plugin>
                 <artifactId>maven-assembly-plugin</artifactId>
+                <!-- this is required for EL5 as the newer versions of the assembly plugin do not allow us to update the zip -->
+                <version>2.4</version>
                 <executions>
                     <execution>
                         <id>make-assembly</id>


zhfeng:
I check the console log and it looks like that zip can not recognize the file struct
```
[mkdir] Created dir: /home/hudson/workspace/btny-pulls-narayana-quickstart/narayana/blacktie/blacktie/target/blacktie-5.3.0.Final-SNAPSHOT/lib/cxx
     [echo] copying symlinks from xatmi runtime library
     [echo] update symlinks to zip
     [exec] 
     [exec] zip error: Zip file structure invalid (blacktie-5.3.0.Final-SNAPSHOT-centos55x32-bin.zip)
```

did it only happen on the el5 ?

zhfeng:
Ah, I find that we missed the maven-assembly-plugin version in the blacktie/blacktie/pom.xml  and it used the lastest 2.5.5 now. It might be the problem that the zip command in el5 can not handle this zip file.
I added the maven-assembly-plugin version with 2.4 and it works.

So I think we need to stick with the maven-assembly-plugin with 2.4 version.
hope it is helpful !

zhfeng:
It looks good to me and there is Jira for this issue ?

PR #989, opened 2016-03-10
[JBTM-2638] Initiate connection in JmsXAResourceRecoveryHelper.getXAResources instead of recover

gytis:
https://issues.jboss.org/browse/JBTM-2638

!BLACKTIE !PERF !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_IBMORB

### DIFF
--- a/ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/JmsXAResourceRecoveryHelper.java
+++ b/ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/JmsXAResourceRecoveryHelper.java
@@ -61,6 +61,12 @@ public class JmsXAResourceRecoveryHelper implements XAResourceRecoveryHelper, XA
         this.pass = pass;
     }
 
+    /**
+     * Nothing to initialise.
+     *
+     * @param properties
+     * @return Always returns true
+     */
     @Override
     public boolean initialise(String properties) {
         if (LOGGER.isTraceEnabled()) {
@@ -70,23 +76,34 @@ public class JmsXAResourceRecoveryHelper implements XAResourceRecoveryHelper, XA
         return true;
     }
 
+    /**
+     * If JMS connection was created successfully, returns an array with one instance of JmsXAResourceRecoveryHelper. Otherwise,
+     * returns an empty array.
+     *
+     * @return Array with one instance of JmsXAResourceRecoveryHelper or an empty array
+     */
     @Override
     public XAResource[] getXAResources() {
-        if (LOGGER.isTraceEnabled()) {
-            LOGGER.trace("Returning XA resource: " + this);
+        if (connect()) {
+            if (LOGGER.isTraceEnabled()) {
+                LOGGER.trace("Returning XA resource: " + this);
+            }
+
+            return new XAResource[] { this };
         }
 
-        return new XAResource[] { this };
+        return new XAResource[0];
     }
 
+    /**
+     * Delegates XAResource#recover call to the connected JMS resource. If provided argument is XAResource.TMENDRSCAN, then JMS
+     * connection will be closed at the end of the call.
+     *
+     * @param i
+     * @throws XAException
+     */
     @Override
     public Xid[] recover(int i) throws XAException {
-        if (i == XAResource.TMSTARTRSCAN) {
-            connect();
-        }
-
-        assert delegate != null : "Recovery scan has to be started";
-
         try {
             return delegate.recover(i);
         } finally {
@@ -96,72 +113,114 @@ public class JmsXAResourceRecoveryHelper implements XAResourceRecoveryHelper, XA
         }
     }
 
+    /**
+     * Delegates XAResource#start call to the connected JMS resource.
+     *
+     * @param xid
+     * @param i
+     * @throws XAException
+     */
     @Override
     public void start(Xid xid, int i) throws XAException {
-        assert delegate != null : "Recovery scan has to be started";
-
         delegate.start(xid, i);
     }
 
+    /**
+     * Delegates XAResource#end call to the connected JMS resource.
+     *
+     * @param xid
+     * @param i
+     * @throws XAException
+     */
     @Override
     public void end(Xid xid, int i) throws XAException {
-        assert delegate != null : "Recovery scan has to be started";
-
         delegate.end(xid, i);
     }
 
+    /**
+     * Delegates XAResource#prepare call to the connected JMS resource.
+     *
+     * @param xid
+     * @return Prepare outcome
+     * @throws XAException
+     */
     @Override
     public int prepare(Xid xid) throws XAException {
-        assert delegate != null : "Recovery scan has to be started";
-
         return delegate.prepare(xid);
     }
 
+    /**
+     * Delegates XAResource#commit call to the connected JMS resource.
+     *
+     * @param xid
+     * @param b
+     * @throws XAException
+     */
     @Override
     public void commit(Xid xid, boolean b) throws XAException {
-        assert delegate != null : "Recovery scan has to be started";
-
         delegate.commit(xid, b);
     }
 
+    /**
+     * Delegates XAResource#rollback call to the connected JMS resource.
+     *
+     * @param xid
+     * @throws XAException
+     */
     @Override
     public void rollback(Xid xid) throws XAException {
-        assert delegate != null : "Recovery scan has to be started";
-
         delegate.rollback(xid);
     }
 
+    /**
+     * Delegates XAResource#isSameRM call to the connected JMS resource.
+     *
+     * @param xaResource
+     * @return True if is same resource manager or false if not.
+     * @throws XAException
+     */
     @Override
     public boolean isSameRM(XAResource xaResource) throws XAException {
-        assert delegate != null : "Recovery scan has to be started";
-
         return delegate.isSameRM(xaResource);
     }
 
+    /**
+     * Delegates XAResource#forget call to the connected JMS resource.
+     *
+     * @param xid
+     * @throws XAException
+     */
     @Override
     public void forget(Xid xid) throws XAException {
-        assert delegate != null : "Recovery scan has to be started";
-
         delegate.forget(xid);
     }
 
+    /**
+     * Delegates XAResource#getTransactionTimeout call to the connected JMS resource.
+     *
+     * @return Transaction timeout value.
+     * @throws XAException
+     */
     @Override
     public int getTransactionTimeout() throws XAException {
-        assert delegate != null : "Recovery scan has to be started";
-
         return delegate.getTransactionTimeout();
     }
 
+    /**
+     * Delegates XAResource#setTransactionTimeout call to the connected JMS resource.
+     *
+     * @param i
+     * @return True if transaction timeout was set, or false if wasn't.
+     * @throws XAException
+     */
     @Override
     public boolean setTransactionTimeout(int i) throws XAException {
-        assert delegate != null : "Recovery scan has to be started";
-
         return delegate.setTransactionTimeout(i);
     }
 
-    private void connect() throws XAException {
+    private boolean connect() {
         if (delegate != null) {
-            return;
+            return true;
         }
 
         try {
@@ -170,16 +229,17 @@ public class JmsXAResourceRecoveryHelper implements XAResourceRecoveryHelper, XA
             delegate = xaSession.getXAResource();
         } catch (JMSException e) {
             LOGGER.warn("Failed to create connection", e);
-            throw new XAException(e.getMessage());
+            return false;
         }
+
+        return true;
     }
 
-    private void disconnect() throws XAException {
+    private void disconnect() {
         try {
             xaConnection.close();
         } catch (JMSException e) {
             LOGGER.warn("Failed to close connection", e);
-            throw new XAException(e.getMessage());
         } finally {
             xaConnection = null;
             xaSession = null;
--- a/ArjunaJTA/jms/src/test/java/org/jboss/narayana/jta/jms/JmsXAResourceRecoveryHelperTests.java
+++ b/ArjunaJTA/jms/src/test/java/org/jboss/narayana/jta/jms/JmsXAResourceRecoveryHelperTests.java
@@ -26,6 +26,7 @@ import org.junit.Test;
 import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
 
+import javax.jms.JMSException;
 import javax.jms.XAConnection;
 import javax.jms.XAConnectionFactory;
 import javax.jms.XASession;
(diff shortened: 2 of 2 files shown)


ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/JmsXAResourceRecoveryHelper.java
```diff
@@ -177,4 +168,10 @@ private void disconnect() throws XAException {
         }
     }
 
+    private void assertIfRecoveryScanWasStarted() throws XAException {
+        if (delegate == null) {
+            throw new XAException("Recovery scan has to be started");
```
mmusgrov: The spec does say what code new XAException("") uses so you need to be throwing one of the valid XA codes, I would guess you either need XAER_RMERR or one of the protocol violation errors if the error is recoverable.

mmusgrov: Also "Recovery scan has to be started" is not meaningful and it should be localizable (ie appear in our JTA log messages file).

gytis: Do we have to add localization for the modules that are not productized?

mmusgrov: I do not know what the official stance is but we definitely have precedence in various community only modules (txojLogger for example)

gytis: In that case I will create a separate JIRA for this. Because there are other places that need changing outside of JBTM-2638 scope.

gytis: Here it is https://issues.jboss.org/browse/JBTM-2641

mmusgrov: OK. What about the use of XA codes instead of using the (unspecified) default code in the XAException constructor?

gytis:
@mmusgrov I've added XAER_RMERR to XAExceptions

ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/JmsXAResourceRecoveryHelper.java
```diff
@@ -177,4 +168,11 @@ private void disconnect() throws XAException {
         }
     }
 
+    private void assertIfRecoveryScanWasStarted() throws XAException {
+        if (delegate == null) {
+            LOGGER.warn("XAResource.recover has to be called first with XAResource.TMSTARTRSCAN argument");
```
mmusgrov: Remark: You should link JBTM-2638 with the one you raised for the localisation JBTM-2641

ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/JmsXAResourceRecoveryHelper.java
```diff
@@ -88,64 +88,55 @@ public boolean initialise(String properties) {
...
-
+        assertIfRecoveryScanWasStarted();
         return delegate.prepare(xid);
     }
 
     @Override
     public void commit(Xid xid, boolean b) throws XAException {
-        assert delegate != null : "Recovery scan has to be started";
-
+        assertIfRecoveryScanWasStarted();
```
tomjenkinson: RMERR out of this is not good - check the spec

ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/JmsXAResourceRecoveryHelper.java
```diff
@@ -88,64 +88,55 @@ public boolean initialise(String properties) {
...
         delegate.commit(xid, b);
     }
 
     @Override
     public void rollback(Xid xid) throws XAException {
-        assert delegate != null : "Recovery scan has to be started";
-
+        assertIfRecoveryScanWasStarted();
```
tomjenkinson: RMERR Is not great here - not as bad as commit but not compatible either

ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/JmsXAResourceRecoveryHelper.java
mmusgrov: So maybe change the signature of assertIfRecoveryScanWasStarted(int xacode) so that you can tailor the error code based on the method where the error was detected.

gytis:
I agree. XAER_RMERR also says that resource was rolled back, which is not correct in this case. XAER_PROTO is closer to the true reason.

gytis:
@tomjenkinson I've added XAER_PROTO and javadocs

ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/JmsXAResourceRecoveryHelper.java
```diff
@@ -69,13 +80,24 @@ public boolean initialise(String properties) {
         return new XAResource[] { this };
     }
 
+    /**
+     * Delegates XAResource#recover call to the connected JMS resource. Additionally, if provided argument is
+     * XAResource.TMSTARTRSCAN, then JMS connection will be initiated for all further communications. Or if provided argument is
+     * XAResource.TMENDRSCAN, then JMS connection will be closed at the end of the call.
+     *
```
mmusgrov: Remark: I assume that the calls are under the control of periodic recovery and therefore no synchronization is required to protect against recover calls interleaving.

ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/JmsXAResourceRecoveryHelper.java
```diff
@@ -177,4 +268,11 @@ private void disconnect() throws XAException {
         }
     }
 
+    private void assertIfRecoveryScanWasStarted() throws XAException {
+        if (delegate == null) {
+            LOGGER.warn("XAResource.recover has to be called first with XAResource.TMSTARTRSCAN argument");
+            throw new XAException(XAException.XAER_PROTO);
```
mmusgrov: Are we sure XAER_PROTO is correct.  Calls to this class are completely under the control of the recovery module so it would be a coding error on our behalf if we do not call recover correctly or fail to handle the error if the start scan call throws an exception. There is an argument that the correct code is XAER_INVAL since the definition says that XAER_PROTO should only be used if one of the transitions in the state table is violated:

"An RM must ensure that each calling thread makes xa_ calls in a legal sequence,
and must return [XAER_PROTO] or other suitable error if the caller violates the
state tables (see Chapter 6)."

But since calls to this class are private to our recovery system perhaps we are being too picky.

gytis: Whichever error code it is, I think we're too picky. Initially I had AssertionError. But I decided to remove it, because XARecoveryModule only catches XAException. So AssertionError used to crash the recovery module.
This scenario happened for me a few times on Spring during the shutdown when JMS connection is closed before RecoveryManager is stopped.

mmusgrov: Assertions are generally removed from production code so you are right to move to XAExceptions.

I am happy with the changes. Should I move the assignee back to @tomjenkinson.

tomjenkinson: potentially as we are under the control of Arjuna recovery manager we could completely ignore the check and report an NPE back as connect would be equal to null.

I think so long as the Javadoc expresses it correctly, going with PROTO and explaining its expectations will be enough. INVAL is defined as "invalid arguments were given" and in particular to xa_recover has some defined semantics. I think PROTO is closest.

Options:
1. A new special exception
2. PROTO or INVAL (or a mix)
3. No protection (as we know we are running under Narayana)

I think we should move the discussion to the forum such that people can contribute. But IMO we should go with proto or no protection.

mmusgrov: Actually I just looked at what we do for XAER_PROTO and XAER_INVAL during recovery and we return HEURISTIC_HAZARD in both cases. I think it should be safe to assume that we have coded up the recovery module to correctly handle the recover calls so I do think we should be ok.

tomjenkinson: By that evidence though I think I would prefer (if possible) to remove the protection and report the NPE. Reporting the outcome to a HEURISTIC is not right. Potentially we could in the future want to run this by some external third party recovery manager.

@gytis without any protection it wouldn't hurt from what I understand? We should never get in the state where TMSTARTRSCAN was not called AFAICT?

ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/JmsXAResourceRecoveryHelper.java
```diff
@@ -160,7 +251,7 @@ private void connect() throws XAException {
             delegate = xaSession.getXAResource();
         } catch (JMSException e) {
             LOGGER.warn("Failed to create connection", e);
-            throw new XAException(e.getMessage());
+            throw new XAException(XAException.XAER_PROTO);
```
tomjenkinson: RMFAIL

ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/JmsXAResourceRecoveryHelper.java
```diff
@@ -169,7 +260,7 @@ private void disconnect() throws XAException {
             xaConnection.close();
         } catch (JMSException e) {
             LOGGER.warn("Failed to close connection", e);
-            throw new XAException(e.getMessage());
+            throw new XAException(XAException.XAER_PROTO);
```
tomjenkinson: RMFAIL

mmusgrov: I agree:
    case XAException.XAER_RMFAIL:
	 _committed = true;  // will cause log to be rewritten

ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/JmsXAResourceRecoveryHelper.java
gytis: With XAException it's possible to happen. If connect() method fails, XARecoveryModule#xaRecoveryFirstPass will catch the exception and will then call XAResource.TMENDRSCAN which will fail because of not available connection. NPE however, should be populated to XARecoveryModule#periodicWorkFirstPass once the connect() fails and only end up being logged and no XAResource.TMENDRSCAN should be executed.

tomjenkinson: OK, so lets go with no protection if it works for Narayana. We should definitely change the exception out of connect() to RMFAIL.

tomjenkinson: And to be on the same paage what I mean by no protection is to remove assertIfRecoveryScanWasStarted

tomjenkinson: I would be prepared to discuss this further on the forum

mmusgrov: We could bypass these issues by ensuring that the call to:

    public XAResource[] getXAResources() {...}

only returns fully connected resources. The TMENDRSCAN call would then be the trigger to clean up the connection.

tomjenkinson: +1, that could be useful here. It would constrain this part of the behaviour behind the SPI and allow the rest of the calls to be unguarded. Good catch.

gytis: But how does that make it different? getXAResources always returns "this" and XARecoveryModule calls recover(XAResource.TMSTARTRSCAN) on it. Thus, in both cases connection will be always created by the same instance. And whichever method it is, it will be always called once. Same thing applies to closing.
In addition, this class is only used by XARecoveryModule and the only way that assertIfRecoveryScanWasStarted will fail is not because recover(XAResource.TMSTARTRSCAN) is not called, but because connection creation has failed (e.g. when Spring is shutting JMS before PeriodicRecovery finishes the scan).

mmusgrov: In the getXAResources() call you attempt to create the connection. If it succeeds you return this in the list but if it fails you return an empty list.

gytis: Ah ok, that sounds good! Shouldn't I try to close connection in recover(XAResource.TMENDSCAN) too (in addition to commit/rollback/forget)? It is executed in "catch" clause if recover(XAResource.TMSTARTRSCAN) fails.

mmusgrov: The connection needs to be left open until either:

1. the current periodic recovery pass has finished, or
2. when the recovery manager stops or the JVM shuts down (here you would have cached the connection to be used by all recovery passes).

gytis: I prefer to close at the end of the pass. At least based on the JMS spec, it is recommended to close it ASAP, since it's using a lot of resources.

mmusgrov: OK so that would be in the end scan call

mmusgrov:
I am happy with this PR. Not sure if @tomjenkinson wants to take a look before you merge it?

PR #1004, opened 2016-04-29
JBTM-2658 Implement JMS integration classes for MessageListener

hasalex:
https://issues.jboss.org/browse/JBTM-2658

See https://developer.jboss.org/thread/269475

!PERF !BLACKTIE !XTS !QA_JTA !QA_JTS_JDKORB

### DIFF
--- a/ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/MessageListenerProxy.java
+++ b/ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/MessageListenerProxy.java
@@ -19,6 +19,10 @@ public class MessageListenerProxy implements MessageListener {
 
     static boolean isDeclaredTransactional(MessageListener messageListener) {
         try {
+            if (messageListener == null) {
+                return false;
+            }
+
             Class<? extends MessageListener> listenerClass = messageListener.getClass();
             if (listenerClass.isAnnotationPresent(Transactional.class)) {
                 return true;
--- a/ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/MessageListenerProxy.java
+++ b/ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/MessageListenerProxy.java
@@ -1,0 +1,70 @@
+package org.jboss.narayana.jta.jms;
+
+import com.arjuna.ats.jta.logging.jtaLogger;
+
+import javax.jms.Message;
+import javax.jms.MessageListener;
+import javax.jms.XASession;
+import javax.transaction.SystemException;
+import javax.transaction.Transaction;
+import javax.transaction.TransactionManager;
+import javax.transaction.Transactional;
+import java.lang.reflect.Method;
+
+public class MessageListenerProxy implements MessageListener {
+
+    private MessageListener messageListener;
+    private TransactionManager transactionManager;
+    private XASession xaSession;
+
+    static boolean isDeclaredTransactional(MessageListener messageListener) {
+        try {
+            Class<? extends MessageListener> listenerClass = messageListener.getClass();
+            if (listenerClass.isAnnotationPresent(Transactional.class)) {
+                return true;
+            } else {
+                Method onMessageMethod = listenerClass.getDeclaredMethod("onMessage", Message.class);
+                return onMessageMethod.isAnnotationPresent(Transactional.class);
+            }
+        } catch (NoSuchMethodException e) {
+            // Should not occur, as messageListener is an instance of javax.jms.MessageListener
+        }
+        return false;
+    }
+
+    public MessageListenerProxy(MessageListener messageListener, XASession session) {
+        this(messageListener, session, com.arjuna.ats.jta.TransactionManager.transactionManager());
+    }
+
+    /**
+     * Mainly for tests
+     */
+    MessageListenerProxy(MessageListener messageListener, XASession session, TransactionManager transactionManager ) {
+        this.messageListener = messageListener;
+        this.transactionManager = transactionManager;
+        this.xaSession = session;
+    }
+
+    @Override
+    public void onMessage(Message jmsMessage) {
+        try {
+            transactionManager.begin();
+            Transaction transaction = transactionManager.getTransaction();
+            transaction.enlistResource(xaSession.getXAResource());
+
+            messageListener.onMessage(jmsMessage);
+
+            transactionManager.commit();
+        } catch (Exception e) {
+            jtaLogger.logger.error(e);
+            try {
+                transactionManager.rollback();
+            } catch (SystemException se) {
+                jtaLogger.logger.warn("Problem on transaction rollback", e);
+            }
+        }
+
+    }
+
+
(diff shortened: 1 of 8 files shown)


ArjunaJTA/jms/src/test/java/org/jboss/narayana/jta/jms/integration/JmsIntegrationTests.java
```diff
@@ -141,19 +137,30 @@ public void testAsyncRollback() throws Exception {
             @Override @Transactional
             public void onMessage(Message message) {
                 try {
+                    System.out.println("RECEIVED");
```
tomjenkinson: We should definitely remove the printlns if we go ahead with this

ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/MessageListenerProxy.java
```diff
@@ -0,0 +1,74 @@
...
+     * Mainly for tests
+     */
+    MessageListenerProxy(MessageListener messageListener, XASession session, TransactionManager transactionManager ) {
+        this.messageListener = messageListener;
+        this.transactionManager = transactionManager;
+        this.xaSession = session;
+    }
+
+    @Override
+    public void onMessage(Message jmsMessage) {
```
tomjenkinson: IMO This doesn't do what you think. By the time you have got to this call the message has already been dequeued from the message broker. It is why I was saying you would need to do:
begin()
message = XAmessageConsumer.receive() // synchronous I know but what can you do?
onMessage(message)
commit()

You basically have to write the container side of things.

hasalex: It's JMS client's job to deal with the queue. The message is dequeued and requeued at rollback. And redelivered depending to the queue configuration.

tomjenkinson: What I mean is the rollback will not rollback the message as it is already dequeued from the queue before you reach this point as the enlist happened after the message has left the broker.

I was a developer on HP-MS (an old JMS implementation from Bluestone/HP) and so have some familiarity with the requirements.

hasalex: OK, I can understand. But in the integration test, it's working : the message is back to the queue. A similar test is working as well with Joram. 
Do you have an idea of test showing that it's broken ?

tomjenkinson: Where is the verification that the message is back in the queue, I missed seeing it in the code sorry.

hasalex: I didn't check it, but the redelivery annoyed me in the integration test. I've changed the tests (commit and rollback) with this verification.

tomjenkinson: I asked @gytis  to look into this one. It could be a quirk of the message broker being used. Conceptually it does not seem possible for it to work (even given your experience) because the message has been removed from the queue before the enlist happened. It needs looking at in the debugger.

We could ask the Artemis team for some input if its still unclear why it is working.

hasalex: I've sent a question on ActiveMQ's ML.

tomjenkinson: Perfect - thanks Alexis. I did my own research and found this: 
https://github.com/wildfly/wildfly/blob/master/ejb3/src/main/java/org/jboss/as/ejb3/inflow/MessageEndpointInvocationHandler.java#L95
https://github.com/apache/activemq-artemis/blob/master/artemis-ra/src/main/java/org/apache/activemq/artemis/ra/inflow/ActiveMQMessageHandler.java#L301

It seems to suggest that for Artemis it would work unless lines 302-310 were critical (as your PR does not do those lines).

I am concerned that this would only work with certain JMS implementations. Conceptually it does not seem logical to enlist the connection after the message is dequeued, perhaps the JMS/JCA spec mandates that it is possible for the inflow use case. In which case we are back on with this :)

gytis: FYI this is a Martyn's reply on the mailing list http://activemq-users.markmail.org/thread/zqm34i32nl3uqmqb#query:+page:1+mid:ez67zqkj7lbn2lgj+state:results.

I had a chat with him yesterday on IRC too. He said that most likely this scenario will not work with other vendors since it's not in the spec. The reason it works now is because of the way Artemis is implemented.

tomjenkinson: Thanks, if its not spec covered we shouldn't go this route. We could investigate moving to a thread pool that does synchronous receives but calls back to message listeners. It's likely not as efficient as the broker doing that for us though.

Another option would be to look at how to integrate with the resource adapter outside of a container. The links I have provided before show the code we would need (beforeDelivery/afterDelivery). Possibly we would need to try to include IronJacamar standalone to get that working.

You can call createSession within a dummy transaction to go through the route you needed in ConnectionProxy. I don't think you should rely on the MessageListenerProxy behaviour if you aren't covered by spec as although it may work with todays message brokers they are free to change in the future and you would likely not spot the change in behaviour until it was too late and you had data integrity issues.

hasalex: Martyn's reply is interesting. If there's no opposition, I'll close the PR.

On my side, I'll follow the container idea. It's probably possible to make such a container class independantly of the JTA provider and the JMS broker.

tomjenkinson: Sure, if you need help from the JTA integration code we can change things (e.g. make it so you always have an XASession). The alternative I mentioned a few times was to create a thread pool that are effectively blocked on MessageConsumer::receive() and call the MessageListener. It's async from the application layer but will of course consume threads in the JVM so really a workaround.

The alternative would be to look at IronJacamar which supports resource adapters and integrates with Narayana for the transaction manager. It can be ran standalone too.

We don't have a resource adapter example but https://github.com/jbosstm/quickstart/tree/master/jca-and-hibernate https://github.com/jbosstm/quickstart/tree/master/jca-and-tomcat

hasalex: Yes, it could be helpful. up to now, I'm starting a transaction just for the session. 

I'll come back later for this one.

PR #1006, opened 2016-05-03
[JBTM-2648] Add 500ms delay after starting the client in CrashRecovery12_Test03

gytis:
https://issues.jboss.org/browse/JBTM-2648

!MAIN !BLACKTIE !XTS !PERF

### DIFF
--- a/qa/tests/src/org/jboss/jbossts/qa/junit/testgroup/TestGroup_crashrecovery12.java
+++ b/qa/tests/src/org/jboss/jbossts/qa/junit/testgroup/TestGroup_crashrecovery12.java
@@ -75,6 +75,7 @@ public class TestGroup_crashrecovery12 extends TestGroupBase
 		setTestName("Test03");
 		Task client0 = createTask("client0", org.jboss.jbossts.qa.CrashRecovery12Clients.Client01.class, Task.TaskType.EXPECT_READY_PASS_FAIL, 240, "store");
 		client0.start("commit", "CR12_03.log");
+		suspend(500); // Workaround for JBTM-2648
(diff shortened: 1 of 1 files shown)


mmusgrov:
The associated JIRA does not give any analysis of why the fix is required or why it fixes the issue.

gytis:
Click through to the cloned issue on EAP

mmusgrov:
OK I clicked through. The same symptom should have been effecting many tests so I guess it is a spurious failure. If your fix resolves Ondrejs test runs then I agree we should localise the fix to the one test.

gytis:
I agree that it should technically fail in other cases, but that's the only
one failing AFAIK. Until other failures appear, I've decided to put the
workaround only in this specific test case.

qa/tests/src/org/jboss/jbossts/qa/junit/testgroup/TestGroup_crashrecovery12.java
```diff
@@ -124,4 +125,12 @@ public String getTestGroupName()
 		outcome0.start("CR12_07.log", "no");
 		outcome0.waitFor();
 	}
+
+	private void suspend(long millis) {
+		try {
+			Thread.sleep(millis);
+		} catch (InterruptedException e) {
+			System.out.println("Suspend interrupted: " + e.getMessage());
```
mmusgrov: Normally there is a reason the thread has been interrupted so you should either re-throw the  InterruptedException or re-enable the interrupt status with:

`Thread.currentThread().interrupt() `

mmusgrov: There is good write up of best practice here: http://www.ibm.com/developerworks/java/library/j-jtp05236/index.html

gytis:
@mmusgrov I've added Thread.currentThread().interrupt()

PR #1014, opened 2016-05-25
[JBTM-2679] narayna-full to contain bits of jbossxts

ochaloup:
https://issues.jboss.org/browse/JBTM-2679

!QA_JTA !QA_JTS_JDKORB !BLACKTIE !PERF !MAIN

### DIFF
--- a/narayana-full/pom.xml
+++ b/narayana-full/pom.xml
@@ -89,12 +89,12 @@
       <groupId>org.jboss.narayana.xts</groupId>
       <artifactId>jbossxts</artifactId>
       <version>${project.version}</version>
-      <classifier>api</classifier>
     </dependency>
     <dependency>
       <groupId>org.jboss.narayana.xts</groupId>
       <artifactId>jbossxts</artifactId>
       <version>${project.version}</version>
+      <classifier>api</classifier>
     </dependency>
     <dependency>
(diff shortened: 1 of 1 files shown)


ochaloup:
hm... I'm a bit lost here. What influence could have change in narayana-full for compensations module?

I've rerun the compilation and the tests on my machine and I can't see the reffered exception.
My investigation on the stack trace was not much informative either. The error which seems caused the test failure is `com.arjuna.wst.SystemException: com.arjuna.wsc.InvalidStateException: ARJUNA042083: Invalid coordination context state`. Some kind of test instability?

I would like to left this being retested and then would see.

ochaloup:
retest this please

ochaloup:
I see. Just I thought that `narayna-full` is a module independent on other modules. It's built as last one and means like "create a distro".

ochaloup:
Sure, I've created PR #1017 that targets master. I hope I'll be more lucky with the new one ;)

ochaloup:
Test for this change passed on master branch (see #1017)

ochaloup:
Nope, from my point this small change is perfectly fine without JBEAP. But depends. I don't know if this would not break productization process or something if there is not JBEAP.

For me this is fine as JBTM.

PR #1021, opened 2016-06-13
[JBTM-2687] Fix of typos in TxBridge tests.

dsimko:


txbridge/src/test/java/org/jboss/jbossts/txbridge/tests/outbound/junit/DisabledContextPropagationTests.java
```diff
@@ -209,13 +210,16 @@ public void testNoTransactionWithEnabledJTAOverWSATFeature(@ArquillianResource U
...
         List<NameValuePair> parameters = new ArrayList<NameValuePair>();
         parameters.add(new BasicNameValuePair("isCommit", "true"));
-        parameters.add(new BasicNameValuePair("isTransaction", "ture"));
+        parameters.add(new BasicNameValuePair("isTransaction", "true"));
         parameters.add(new BasicNameValuePair("isWSATService", "false"));
         parameters.add(new BasicNameValuePair("isJTAOverWSATFeatureEnabled", "true"));
         parameters.add(new BasicNameValuePair("clientType", String.valueOf(TestATClient.CLIENT_WITH_JTA_FEATURE)));
-
-        List<String> invocations = makeRequest(baseURL, parameters);
-        assertInvocations(invocations);
```
gytis: Is there a particular reason to remove this assertion?

txbridge/src/test/java/org/jboss/jbossts/txbridge/tests/outbound/junit/EnabledContextPropagationTests.java
```diff
@@ -205,13 +206,16 @@ public void testNoTransactionWithEnabledJTAOverWSATFeature(@ArquillianResource U
...
         List<NameValuePair> parameters = new ArrayList<NameValuePair>();
         parameters.add(new BasicNameValuePair("isCommit", "true"));
-        parameters.add(new BasicNameValuePair("isTransaction", "ture"));
+        parameters.add(new BasicNameValuePair("isTransaction", "true"));
         parameters.add(new BasicNameValuePair("isWSATService", "false"));
         parameters.add(new BasicNameValuePair("isJTAOverWSATFeatureEnabled", "true"));
         parameters.add(new BasicNameValuePair("clientType", String.valueOf(TestATClient.CLIENT_WITH_JTA_FEATURE)));
-
-        List<String> invocations = makeRequest(baseURL, parameters);
-        assertInvocations(invocations);
```
gytis: Is there a particular reason to remove this assertion?

txbridge/src/test/java/org/jboss/jbossts/txbridge/tests/outbound/junit/DisabledContextPropagationTests.java
dsimko: It would be dead code because exception is thrown here:
makeRequest(baseURL, parameters); 
Otherwise test fails:
Assert.fail("'MustUnderstand headers' exception should be thrown.");

dsimko: It would be dead code because exception is thrown here:
makeRequest(baseURL, parameters); 
Otherwise test fails:
Assert.fail("'MustUnderstand headers' exception should be thrown.");

gytis: Was the test failing before? Because it didn't/doesn't have @Test(expected = ...) annotation, so it should have been failing every time when exception was thrown.

dsimko: Test wasn't fail before because there was a typo in the test configuration:
parameters.add(new BasicNameValuePair("isTransaction", "ture")); // ture instead of true
After fixing this typo, an exception is thrown from the line:
makeRequest(baseURL, parameters); 
But the test pass thanks to old school try-catch. If you want I can replace try-catch to newer @Test(expected = ...)

gytis: Got it. Try-catch is ok with me. Thank's for fixing the typo.

PR #1022, opened 2016-06-15
Do not null the actual connection when delaying close. That means whe…

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2676

!BLACKTIE !XTS !PERF

### DIFF
--- a/ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/drivers/modifiers/SupportsMultipleConnectionsModifier.java
+++ b/ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/drivers/modifiers/SupportsMultipleConnectionsModifier.java
@@ -1,0 +1,71 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2006, Red Hat Middleware LLC, and individual contributors 
+ * as 2016 by the @author tags.
+ * See the copyright.txt in the distribution for a
+ * full listing of individual contributors. 
+ * This copyrighted material is made available to anyone wishing to use,
+ * modify, copy, or redistribute it subject to the terms and conditions
+ * of the GNU Lesser General Public License, v. 2.1.
+ * This program is distributed in the hope that it will be useful, but WITHOUT A 
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A 
+ * PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more details.
+ * You should have received a copy of the GNU Lesser General Public License,
+ * v.2.1 along with this distribution; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, 
+ * MA  02110-1301, USA.
+ * 
+ * (C) 2016,
+ * @author JBoss Inc.
+ */
+package com.arjuna.ats.internal.jdbc.drivers.modifiers;
+
+import com.arjuna.ats.jta.exceptions.NotImplementedException;
+import com.arjuna.ats.jta.xa.XAModifier;
+
+import javax.sql.XAConnection;
+import javax.transaction.xa.Xid;
+import java.sql.Connection;
+import java.sql.SQLException;
+
+public class SupportsMultipleConnectionsModifier implements XAModifier, ConnectionModifier {
+
+    @Override
+    public String initialise(String dbName) {
+        return dbName;
+    }
+
+    @Override
+    public Xid createXid(Xid xid) throws SQLException, NotImplementedException {
+        return xid;
+    }
+
+    @Override
+    public XAConnection getConnection(XAConnection conn) throws SQLException,
+            NotImplementedException {
+        throw new NotImplementedException(); // NEVER CALLED
+    }
+
+    @Override
+    public boolean supportsMultipleConnections() throws SQLException,
+            NotImplementedException {
+        return true; // This ensures connection close is delayed
+    }
+
+    @Override
+    public void setIsolationLevel(Connection conn, int level)
+            throws SQLException, NotImplementedException {
+        // Non-modifier path does not call this
+    }
+
+    @Override
+    public int xaStartParameters(int level) throws SQLException,
+            NotImplementedException {
+        return level;
+    }
+
+    @Override
+    public boolean requiresSameRMOverride() {
+        return false;
+    }
+}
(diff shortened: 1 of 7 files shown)


nmcl:
I've noticed a few commits like this which don't appear to have associated JIRAs. Am I missing something?

ArjunaJTA/jdbc/tests/classes/com/hp/mwtests/ts/jdbc/basic/ConnectionImpleTest.java
```diff
@@ -125,13 +124,14 @@ public void beforeCompletion() { }
             public void afterCompletion(int status) {
                 try {
                     connectionToTest.isClosed();
+                    connectionToTest.close();
```
gytis: Not sure about that. This test is specifically for isClosed method. Close method is tested by closeConnectionInAfterCompletion.
I did that on purpose to make test units as small as possible.

ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/drivers/modifiers/SupportsMultipleConnectionsModifier.java
```diff
@@ -0,0 +1,89 @@
...
+import java.sql.SQLException;
+
+/*
+ * This is a stateless class to allow us to get round
+ * problems in Oracle. For example, they can't work with
+ * an arbitrary implementation of Xid - it has to be their
+ * own implementation!
+ */
+
+public class SupportsMultipleConnectionsModifier implements XAModifier, ConnectionModifier {
```
gytis: Here and in some other places tab is used to indent rather than a space.

tomjenkinson: Well, we don't normally worry about the code format but I will reformat with IntelliJ and see how that comes out

ArjunaJTA/jdbc/tests/classes/com/hp/mwtests/ts/jdbc/basic/ConnectionImpleTest.java
tomjenkinson: Sure, but its part of spec that you have to call connection.close if you call getConnection. I can move it out of the verify I think

tomjenkinson: Yeah. I can. I will move it down so it doesn't look wrong for the test

gytis: But connection wasn't really opened. xaDataSource is a mock, so instead of creating a connection it returns a mock connection.
Also message in the catch block is a bit misleading since you've removed close statement.

tomjenkinson: I updated the message. 
RE connection.close() I am not sure if you are saying it doesn't matter because we are using a mock in which case I would disagree because someone might change it from a mock to a real connection in the future and then it would break. OTOH aybe you are saying even with a real connection you would not need the .close in which case I can remove it.
I do see from the rest of your tests that you do have the .close()

gytis: Why would somebody change it? It's a unit test designed to test one specific test and not the dependencies. That specific test purpose is to verify that when isClose is called on ConnectionImpl call is delegated to the underlying connection (which is also mocked). Before the test used to fail because of the issue you're fixing.

tomjenkinson: The test is not spec compliant because it does not close the connection is the point

tomjenkinson: What if we moved from using mockito to some custom JDBC mock framework that detected connection.close was not called.

tomjenkinson: One compromise I can see is to comment that the connection is not being closed because of the mock framework not requiring it? My point is that without commenting the close is not done it looks wrong to have not closed the connection.

gytis: But unit test is there to test specific units of code. It doesn't matter if it complies with the spec or not. Integration and functional tests should make sure that the functionality works according to all specs. 
As for moving to the different framework, we should worry about that if the time comes IMO.

tomjenkinson: OK, I will just add the comment then

tomjenkinson: Hmm, as you can see from the diff I have made the change you requested to not close connection however in doing so there is an issue that the close functionality is not executed:
https://github.com/tomjenkinson/narayana/blob/bcc7c01599e46fcd5e5412ce9277bda62146ddb0/ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/ConnectionImple.java#L388

tomjenkinson: Did you have change to review this yet? I wonder about if adding the .close would also be useful if we in the future added pooling to ConnectionImpl. By not calling .close we would leak the connection not just from ConnectionManager but also this hypothetical pool.

Thanks!

gytis: Sorry, I didn't get a notification for some reason...
We would leak in the real application. But this is a unit test which doesn't use real connections. There is nothing to leak.
I'm not sure what you mean by close not being executed. Why would isClose execute it? You should check for closeImpl execution in closeConnectionInAfterCompletion test.

tomjenkinson: Hi gytis, I don't think I am being clear here. The connection we would leak would be a ConnectionImpl.

But lets not talk hypothetical. If this test does not call connection.close() then the following line is not called:  https://github.com/tomjenkinson/narayana/blob/bcc7c01599e46fcd5e5412ce9277bda62146ddb0/ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/ConnectionImple.java#L388

PR #1025, opened 2016-06-17
Only call xa_end if the XAResource is associated

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2753
!PERF !BLACKTIE !XTS


ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
```diff
@@ -211,6 +211,7 @@ public int topLevelPrepare()
 			if (endAssociation())
 			{
 				_theXAResource.end(_tranID, XAResource.TMSUCCESS);
```
tomjenkinson: No problem - I have added these changes

PR #1033, opened 2016-07-11
JBTM-2689 narayana-osgi-jta integration test does not work with the karaf 4.x

zhfeng:
!BLACKTIE !XTS !PERF !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB


zhfeng:
re-test

PR #1034, opened 2016-07-22
master JBTM-2703

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2703

…ed in parallel we do not try to use state we cannot load

!PERF !XTS !BLACKTIE !QA_JTA !QA_JTS_JDKORB

Docs: https://github.com/jbosstm/documentation/pull/37

### DIFF
--- a/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/jca/SubordinateTxUnitTest.java
+++ b/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/jca/SubordinateTxUnitTest.java
@@ -31,17 +31,22 @@
 
 package com.hp.mwtests.ts.jta.jca;
 
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
-
-import org.junit.Test;
-
 import com.arjuna.ats.arjuna.ObjectType;
 import com.arjuna.ats.arjuna.common.Uid;
 import com.arjuna.ats.arjuna.state.InputObjectState;
 import com.arjuna.ats.arjuna.state.OutputObjectState;
+import com.arjuna.ats.internal.jta.resources.arjunacore.XAResourceRecord;
 import com.arjuna.ats.internal.jta.transaction.arjunacore.subordinate.jca.SubordinateAtomicAction;
 import com.arjuna.ats.internal.jta.transaction.arjunacore.subordinate.jca.TransactionImple;
+import com.arjuna.ats.jta.xa.XidImple;
+import org.junit.Test;
+
+import javax.transaction.xa.XAException;
+import javax.transaction.xa.XAResource;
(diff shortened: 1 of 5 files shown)


sfikes:
I could see a request arising for a way to suppress the warning (since this has been seen to happen somewhat frequently in some scenarios). I normally recommend *not* suppressing WARN level logging as that can lead to serious issues being overlooked. I guess a filter could be used to just suppress this specific WARN (after verifying the root cause) ... or is there a better option?

sfikes:
I think INFO is fine ... but should the message be changed to reflect that the record may have been removed as the result of successful completion of a commit on another thread or else as the result of user deletion of something (hopefully less likely)? That way someone viewing the message (WARN is usually the threshold for logging ... but the message would show up if the threshold were changed to INFO or TRACE) would have a better sense of what might be happening?

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/jca/SubordinateTxUnitTest.java
```diff
@@ -73,4 +78,85 @@ public void testAtomicAction () throws Exception
...
+            }
+        }), null));
+
+        saa.doPrepare();
+        SubordinateAtomicAction saa2 = new SubordinateAtomicAction(saa.get_uid(), true);
+        assertTrue(saa2.getXid() != null);
+
+        saa.doCommit();
+        SubordinateAtomicAction saa3 = new SubordinateAtomicAction(saa.get_uid(), true);
+        assertTrue(saa3.getXid() == null);
```
tomjenkinson: Done

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/jca/SubordinateTxUnitTest.java
```diff
@@ -73,4 +78,85 @@ public void testAtomicAction () throws Exception
...
+            }
+        }), null));
+
+        saa.doPrepare();
+        SubordinateAtomicAction saa2 = new SubordinateAtomicAction(saa.get_uid(), true);
+        assertTrue(saa2.getXid() != null);
+
+        saa.doCommit();
+        SubordinateAtomicAction saa3 = new SubordinateAtomicAction(saa.get_uid(), true);
+        assertTrue(saa3.getXid() == null); // Since the SAA was committed the transaction log record will have been removed so the xid returned from getXid() should no longer be available and the intention is the SAA creator would disregard this instance
```
mmusgrov: I will still mark the PR as ready to merge but a 254 character line length is rather long!

tomjenkinson: I was going for the record :)

PR #1038, opened 2016-07-28
master JBTM-2710

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2710

!PERF !BLACKTIE !XTS !QA_JTA !XTS

### DIFF
--- a/ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/orbspecific/interposition/coordinator/ServerTransaction.java
+++ b/ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/orbspecific/interposition/coordinator/ServerTransaction.java
@@ -284,7 +284,7 @@ public class ServerTransaction extends ArjunaTransactionImple
 
 		org.omg.CosTransactions.Status s = get_status();
 
-		if ((s == org.omg.CosTransactions.Status.StatusCommitted)
+		if ((s == org.omg.CosTransactions.Status.StatusCommitted && preparedList.size() == 0)
 				|| (s == org.omg.CosTransactions.Status.StatusRolledBack))
 		{
 		    int status = finalStatus();
--- a/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/jca/XATerminatorUnitTest.java
+++ b/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/jca/XATerminatorUnitTest.java
@@ -36,6 +36,8 @@ import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import javax.transaction.Transaction;
+import javax.transaction.RollbackException;
+import javax.transaction.SystemException;
 import javax.transaction.xa.XAException;
 import javax.transaction.xa.XAResource;
 import javax.transaction.xa.Xid;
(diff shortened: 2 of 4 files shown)


mmusgrov:
Why do we have this JIRA (2710) and JBTM-2677 (XATerminator.rollback does not invoke XAResource.rollback for failed resources when JTS is used) - was this an oversight and one of them needs closing as a duplicate of the other?

ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/orbspecific/interposition/coordinator/ServerTransaction.java
```diff
@@ -284,7 +284,7 @@ public final int doPhase2Abort ()
 
 		org.omg.CosTransactions.Status s = get_status();
 
-		if ((s == org.omg.CosTransactions.Status.StatusCommitted)
+		if ((s == org.omg.CosTransactions.Status.StatusCommitted && preparedList.size() == 0)
```
mmusgrov: This kind of makes sense - if there was something in the intentions list but the transaction has already committed then have another go at completing it to resolve the intentions list (the else part of this test) by changing the state back to aborting (instead of committed). But this kind of implies we lied the first time around when we reported the transaction as committed. So my question is, are we certain that it is safe to morph the state back to a prior state (ie COMMITTED -> ABORTING)?

tomjenkinson: Its about massaging (the comment in the code before). Though this issue 2710 demonstrated that the massage should only be done if there was nothing in the prepared list.

tomjenkinson: For more background I designed this to match the behavior in JTA now: https://github.com/jbosstm/narayana/blob/master/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/subordinate/TransactionImple.java#L219 however I still wanted to allow the existing massage for cases where it was still safe.

mmusgrov: Okay so it about completing/implementing what was originally intended in which case fine.
And my question about changing the state from committed to aborting - so that is OK because the [ActionStatus] state model in internal to our implementation and provided our test suite is not throwing up any issues we are fine to introduce new state transitions?

tomjenkinson: Well, the new state transaction is to run from COMMITTED through the ABORTING code rather than just return ABORTED.

As you will see, COMMITTED -> ABORTED was actually already possible (at least to an external caller): https://github.com/jbosstm/narayana/blob/master/ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/orbspecific/interposition/coordinator/ServerTransaction.java#L298

mmusgrov: Okay, thanks for the explanation.

PR #1044, opened 2016-08-08
AtomicTransaction equals fix

tadamski:
Related to: https://developer.jboss.org/message/956793#956793

### DIFF
--- a/ArjunaJTS/jts/classes/com/arjuna/ats/jts/extensions/AtomicTransaction.java
+++ b/ArjunaJTS/jts/classes/com/arjuna/ats/jts/extensions/AtomicTransaction.java
@@ -611,12 +611,14 @@ public class AtomicTransaction
 				AtomicTransaction tx = (AtomicTransaction) obj;
 				ControlWrapper txControl = tx._theAction;
 
-				if ((_theAction == null) && (txControl == null))
+				if ((_theAction == null) && (txControl == null)) {
 					return true;
-				else if (_theAction != null) {
+				}
+				else if (_theAction == null) {
 					return false;
-				} else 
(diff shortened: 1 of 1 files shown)


gytis:
@tomjenkinson , I didn't understand who should create a JIRA?

tadamski:
@tomjenkinson @gytis  Will do.

tadamski:
@tomjenkinson @gytis What was the rationale behind JBTM-1038 change? I want to make sure that I didn't brake anything but to my understanding after this change we would always obtains false if we are dealing with non-null objects.

ArjunaJTS/jts/classes/com/arjuna/ats/jts/extensions/AtomicTransaction.java
```diff
@@ -611,12 +611,14 @@ public boolean equals (Object obj)
 				AtomicTransaction tx = (AtomicTransaction) obj;
 				ControlWrapper txControl = tx._theAction;
 
-				if ((_theAction == null) && (txControl == null))
+				if ((_theAction == null) && (txControl == null)) {
 					return true;
-				else if (_theAction != null) {
+				}
+				else if ((_theAction != null) && (txControl==null)) {
```
scottmarlow: What is the problem that your addressing here?  Should this instead check for null _theAction, since the below _theAction.equals(txControl) check already handles the null txControl case:
else if (_theAction == null)
    return false;

tadamski: Yeah, you are right. Fixing.

PR #1047, opened 2016-08-10
[JBTM-2730] Static code analysis issue: default encoding

ochaloup:
This PR fixes a bug came from static code analysis:
21994 Dm: Dubious method used In org.jboss.narayana.osgi.jta.internal.ObjStoreBrowserImpl.ObjStoreBrowserImpl (com.arjuna.ats.arjuna.tools.osb.mbean.ObjStoreBrowser): Found a call to a method which will perform a byte to String (or String to byte) conversion, and will assume that the default platform encoding is suitable)

The reason why the using default encoding could cause problems is that code behaves differently depending on the platform it runs on. Especially this is harmful if the data flows from one platform to another.

System default encoding could differ based on OS and geographical location.
Java uses parameter `-Dfile.encoding` to change default encoding to be different from system one. Value of this parameter is used whenever default encoding is needed to be used for program calls.

One example which come to my mind specifically with TM is if JVM crash occurs and object store is moved from one machine to another. Base on how node identifier is set it occurs that is decoded differently on each machine and that way transactions won't be recovered as the newly started node will not identify that is responsible for in-doubt ones.

It is considered bad practice to rely on the default platform encoding.

 !BLACKTIE !PERF !XTS !QA_JTS !QA_JTS_JDKORB

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/CoreEnvironmentBean.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/CoreEnvironmentBean.java
@@ -116,8 +116,8 @@ public class CoreEnvironmentBean implements CoreEnvironmentBeanMBean
 
         if (nodeIdentifier.getBytes().length > NODE_NAME_SIZE)
         {
-            tsLogger.i18NLogger.fatal_nodename_too_long(nodeIdentifier);
-            throw new CoreEnvironmentBeanException(tsLogger.i18NLogger.get_fatal_nodename_too_long(nodeIdentifier));
+            tsLogger.i18NLogger.fatal_nodename_too_long(nodeIdentifier, NODE_NAME_SIZE);
+            throw new CoreEnvironmentBeanException(tsLogger.i18NLogger.get_fatal_nodename_too_long(nodeIdentifier, NODE_NAME_SIZE));
         }
 
     	this.nodeIdentifier = nodeIdentifier;
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/CoreEnvironmentBean.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/CoreEnvironmentBean.java
@@ -21,7 +21,7 @@
 package com.arjuna.ats.arjuna.common;
 
 import java.io.File;
-
+import java.nio.charset.StandardCharsets;
 import com.arjuna.ats.arjuna.logging.tsLogger;
 import com.arjuna.ats.arjuna.utils.Process;
 import com.arjuna.ats.arjuna.utils.Utility;
@@ -114,7 +114,7 @@ public class CoreEnvironmentBean implements CoreEnvironmentBeanMBean
             throw new CoreEnvironmentBeanException(tsLogger.i18NLogger.get_fatal_nodename_null());
         }
 
-        if (nodeIdentifier.getBytes().length > NODE_NAME_SIZE)
+        if (nodeIdentifier.getBytes(StandardCharsets.UTF_8).length > NODE_NAME_SIZE)
         {
             tsLogger.i18NLogger.fatal_nodename_too_long(nodeIdentifier, NODE_NAME_SIZE);
             throw new CoreEnvironmentBeanException(tsLogger.i18NLogger.get_fatal_nodename_too_long(nodeIdentifier, NODE_NAME_SIZE));
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TxControl.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TxControl.java
@@ -34,7 +34,6 @@ package com.arjuna.ats.arjuna.coordinator;
 import com.arjuna.ats.arjuna.common.arjPropertyManager;
 import com.arjuna.ats.arjuna.logging.tsLogger;
 import com.arjuna.ats.arjuna.recovery.TransactionStatusManager;
-import com.arjuna.ats.arjuna.utils.Utility;
 
 /**
  * Transaction configuration object. We have a separate object for this so that
@@ -168,7 +167,7 @@ public class TxControl
 	public static void setXANodeName(String name)
 	{
 	    if (name.getBytes().length > NODE_NAME_SIZE) {
-            tsLogger.i18NLogger.warn_coordinator_toolong();
+            tsLogger.i18NLogger.warn_coordinator_toolong(NODE_NAME_SIZE);
 
             throw new IllegalArgumentException();
         }
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TxControl.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TxControl.java
@@ -31,6 +31,7 @@
 
 package com.arjuna.ats.arjuna.coordinator;
 
+import java.nio.charset.StandardCharsets;
 import com.arjuna.ats.arjuna.common.arjPropertyManager;
 import com.arjuna.ats.arjuna.logging.tsLogger;
 import com.arjuna.ats.arjuna.recovery.TransactionStatusManager;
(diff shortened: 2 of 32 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/CoreEnvironmentBean.java
```diff
@@ -114,10 +115,16 @@ public void setNodeIdentifier(String nodeIdentifier) throws CoreEnvironmentBeanE
...
+            {
+                tsLogger.i18NLogger.fatal_nodename_too_long(nodeIdentifier, NODE_NAME_SIZE);
+                throw new CoreEnvironmentBeanException(tsLogger.i18NLogger.get_fatal_nodename_too_long(nodeIdentifier, NODE_NAME_SIZE));
+            }
+        }
+        catch (UnsupportedEncodingException e)
         {
-            tsLogger.i18NLogger.fatal_nodename_too_long(nodeIdentifier);
-            throw new CoreEnvironmentBeanException(tsLogger.i18NLogger.get_fatal_nodename_too_long(nodeIdentifier));
+            throw new AssertionError("UTF-8 is not supported");
```
tomjenkinson: Lets throw a CoreEnvironmentBeanException

ochaloup: fixed

ochaloup:
> Can the other exceptions match existing clauses? Is there a precedent for AssertionError?

I used the AssertionError as, by my opinion, the best fitting exception. There could be used any unchecked exception.
The point here is the java design where UnsupportedEncodingException is checked one. Then it needs to be handled somehow.
Other point here is that if you do not touch rt.jar or do something similar there is no chance to get the exception as UTF-8 is standard and just supported.


> Also, please can you provide a reference to where it is defined "It is considered bad practice to rely on the default platform encoding."

The main reference I would say is the find bugs report.
From documentation I have to say I can't find any such strict statement which I've used. To general idea differences are described here
https://docs.oracle.com/javase/tutorial/i18n/text/string.html
If you check internet, not using default encoding is the general recommendation. I can list down stackoverflow discussions or blog posts here.

ochaloup:
I've played with this a little bit more and I have a case where using default encoding could be a trouble. It needs to be a place where some non-ascii characters are put in play (default encoding which is used by java could be changed by jvm param `file.encoding`)

* node id set as `𐍈𐍈𐍈𐍈𐍈𐍈𐍈𐍈𐍈𐍈𐍈𐍈` will let container being started for encoding ISO-8859-1 and won't be started for encoding UTF-8. That's when configuration moved from ISO-8859-1 machine to UTF-8 machine when the another machine will be expected to finish recovery of some failed txn the second machine fails to start. (sure, workaround is to set jvm default encoding via parameter)
* node id set to `kulaťoučký`, whole eap installation moved from one machine with ISO-8859-1 to machine with default encoding set to UTF-8. Txn object store is copied too for second machine finishing in-doubt transaction by recovery. The second machine won't finish transactions which depends on node identifer check (bottom-up recovery) as node id will be decoded differently than it was understood at the first machine.

ochaloup:
@tomjenkinson ok, I've created the issue and make the squash

plus I realized that Narayana is restricted to be built in JDK8 and higher. I've done a change to use charset constants and not added too

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TxControl.java
```diff
@@ -161,20 +161,20 @@ public static boolean isReadonlyOptimisation()
     }
 
     public static final String getXANodeName()
-	{
-		return xaNodeName;
-	}
+    {
```
mmusgrov: We do not normally change the whitespace because it makes diffs with previous versions more difficult. You have done this in a few places in this PR

osgi/jta/src/main/java/org/jboss/narayana/osgi/jta/internal/ObjStoreBrowserImpl.java
```diff
@@ -54,13 +56,18 @@
...
-    private PrintStream printStream = new PrintStream(System.out, true);
+    private PrintStream printStream;
     private List<String> recordTypes = new ArrayList<String>();
     private String currentType = null;
     private String currentLog = "";
     private boolean attached = false;
 
     public ObjStoreBrowserImpl(ObjStoreBrowser osb) {
+        try {
+            printStream = new PrintStream(System.out, true, "UTF-8");
```
mmusgrov: StandardCharsets.UTF_8

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/utils/ExecProcessId.java
```diff
@@ -141,7 +143,12 @@ public int getpid ()
                     if (tempFile != null)
                         tempFile.delete();
 
-                    StringTokenizer theTokenizer = new StringTokenizer(bstream.toString());
+                    StringTokenizer theTokenizer;
+                    try {
+                        theTokenizer = new StringTokenizer(bstream.toString(StandardCharsets.UTF_8.name()));
+                    } catch (UnsupportedEncodingException e) {
+                        throw new AssertionError("UTF-8 is not supported");
```
mmusgrov: AssertionErrors can be disabled from the command line so you probably need some other exception type.  The error should also be logged. Also use StandardCharsets.UTF_8.name() for consistency.

osgi/jta/src/main/java/org/jboss/narayana/osgi/jta/internal/ObjStoreBrowserImpl.java
```diff
@@ -54,13 +56,18 @@
...
+        } catch (UnsupportedEncodingException e) {
+            throw new AssertionError("UTF-8 is not supported");
```
mmusgrov: Same comment as above

ochaloup:
@mmusgrov updated, thanks for your review. I hope I addressed all reproaches you mentioned.

I have got a little bit trouble with indentation as it is defined differently at each file and at each method :)

mmusgrov:
@ochaloup Yes the variety of styles in our code base is a problem - the key is to avoid letting your editor auto-format the code.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/utils/ExecProcessId.java
```diff
@@ -141,7 +143,13 @@ public int getpid ()
                     if (tempFile != null)
                         tempFile.delete();
 
-                    StringTokenizer theTokenizer = new StringTokenizer(bstream.toString());
+                    StringTokenizer theTokenizer;
+                    try {
+                        theTokenizer = new StringTokenizer(bstream.toString(StandardCharsets.UTF_8.name()));
+                    } catch (UnsupportedEncodingException e) {
+                        tsLogger.i18NLogger.fatal_encoding_not_supported(StandardCharsets.UTF_8.name());
+                        throw new IllegalStateException(StandardCharsets.UTF_8.name() + " is not supported");
```
gytis: Exception message should be in com.arjuna.ats.arjuna.logging.arjunaI18NLogger

mmusgrov:
The failure@

> MAIN profile tests failed (http://albany.eng.hst.ams2.redhat.com/job/btny-pulls-narayana jdk8/PROFILE=MAIN,jdk=jdk8.latest,label=linux/1607/): REST-AT Integration Test failed

is something Gytis and I fixed a couple of weeks ago. Is your master fully up to date?

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/utils/ExecProcessId.java
ochaloup: ah, ok. sorry for my over-sighting this. it should be fixed now

ochaloup:
@mmusgrov I've rechecked the source before fixing Gytis's remark and I can say that's rebased to the most recent version of master branch. Moreover I thought rebase/merge is done up to master for particular branch before running the test, right?

mmusgrov:
@ochaloup OK in that case either myself or @gytis will need to investigate, but the failure is almost certainly unrelated to your changes.

gytis:
@ochaloup , could you please squash the commits?

ochaloup:
@gytis I can but I have intentionally left them separated as each solves a different thing. But if you say I can squash them to one?

gytis:
@ochaloup then we should create a JIRA and link to it in the commit

ochaloup:
@gytis ok, I see. I've created an issue. Is it fine this way?

PR #1049, opened 2016-08-11
JBTM-2725 Fix race condition issue in ControlWrapper#rollback

tadamski:
https://issues.jboss.org/browse/JBTM-2575

When the thread rolls back the transaction immediately after the reaper they both end in ControlWrapper#rollback method which is unexpected and leads to error in ArjunaTransactionImple class. It should be synchronized so that only one thread performs the ArjunaTransactionImple#rollback method. Synchronizing ControlWrapper#rollback method doesn't work as there are lot of those objects created for one transaction. I have synchronized the invocation on ControlImple object. As I understand the code this object is shared across the whole transaction. I have run the tests with the update and no error occurs.

### DIFF
--- a/ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/ControlWrapper.java
+++ b/ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/ControlWrapper.java
@@ -240,8 +240,11 @@ public class ControlWrapper implements Reapable
 	{
 		try
 		{
-			if (_controlImpl != null)
-				_controlImpl.getImplHandle().commit(report_heuristics);
+			if (_controlImpl != null) {
+				synchronized(_controlImpl) {
+					_controlImpl.getImplHandle().commit(report_heuristics);
+				}
+			}
 			else
 			{
 				Terminator t = null;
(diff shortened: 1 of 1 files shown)


mmusgrov:
@tadamski Wouldn't you see the same problem if the application tries to commit the transaction at the same time the reaper is rolling it back? I am also wondering about the other methods such as rollback_only, for example.

tadamski:
@mmusgrov Current version of the test is unable to reproduce commit scenario, but as I understand the code there is a possibility that the same scenario would happen - I have already updated the PR with the commit method fix.

mmusgrov:
Our tests are showing up a deadlock with this modification. The following sequence happens on a single thread:

1) The commit request takes the lock on the ControlImpl instance (the change you added);
2) The commit then calls the beforeCompleton synchs (which are CORBA calls);
3) The interceptors for these calls need to see if there is a transaction context and so they need to invoke ControlImple.getControl but this method is synchronised so needs the lock taken in step 1

May I take a look at the stack trace output of your tests that originally highlighted the race condition to further investigate the problem?

mmusgrov:
@tadamski I have rejected this PR since it has been replaced by https://github.com/jbosstm/narayana/pull/1057

mmusgrov:
@tadamski I am rejecting this PR since it has been replaced by https://github.com/jbosstm/narayana/pull/1057

PR #1055, opened 2016-08-16
JBTM-2725 Throw TransactionRolledbackException in ArjunaTransactionIm…

tadamski:
…ple#commit,rollback methods if transaction already aborted


mmusgrov:
@tadamski I have rejected this PR since it has been replaced by #1057

PR #1057, opened 2016-08-18
JBTM-2725 

tadamski:
!PERF !BLACKTIE !XTS

### DIFF
--- a/ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/ControlWrapper.java
+++ b/ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/ControlWrapper.java
@@ -306,7 +306,7 @@ public class ControlWrapper implements Reapable
 		}
 		catch (NullPointerException ex)
 		{
-			throw new TRANSACTION_ROLLEDBACK();
+			return;
 		}
 	}
 
--- a/ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/orbspecific/coordinator/ArjunaTransactionImple.java
+++ b/ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/orbspecific/coordinator/ArjunaTransactionImple.java
@@ -502,6 +502,9 @@ public class ArjunaTransactionImple extends
 			status = super.status();
 		} else if ((status == ActionStatus.ABORTING)
 				|| (status == ActionStatus.ABORTED) || (status == ActionStatus.H_ROLLBACK)) {
+			if (jtsLogger.logger.isInfoEnabled()) {
+				jtsLogger.i18NLogger.info_rollback_aborted_transaction(get_transaction_name());
+			}
 			return;
 		}
 		else
--- a/ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/orbspecific/coordinator/ArjunaTransactionImple.java
+++ b/ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/orbspecific/coordinator/ArjunaTransactionImple.java
@@ -500,6 +500,9 @@ public class ArjunaTransactionImple extends
 			destroyAction();
 
 			status = super.status();
+		} else if ((status == ActionStatus.ABORTING)
+				|| (status == ActionStatus.ABORTED) || (status == ActionStatus.H_ROLLBACK)) {
+			return;
 		}
 		else
 		{
--- a/ArjunaJTS/jts/classes/com/arjuna/ats/jts/logging/jtsI18NLogger.java
+++ b/ArjunaJTS/jts/classes/com/arjuna/ats/jts/logging/jtsI18NLogger.java
@@ -1070,6 +1070,10 @@ public interface jtsI18NLogger {
     @Message(id = 22262, value = "The node identifier cannot be null", format = MESSAGE_FORMAT)
     public String get_nodename_null();
 
+	@Message(id = 22263, value = "rollback for {0} was already rolled back", format = MESSAGE_FORMAT)
+	@LogMessage(level = INFO)
+	public void info_rollback_aborted_transaction(String arg0);
+
     /*
         Allocate new messages directly above this notice.
           - id: use the next id number in sequence. Don't reuse ids.


ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/ControlWrapper.java
```diff
@@ -270,7 +270,7 @@ public final void commit (boolean report_heuristics) throws Unavailable,
 		    // check termination status
 		    
 		    if (_controlImpl.getFinalStatus() != org.omg.CosTransactions.Status.StatusCommitted)
-                        throw new TRANSACTION_ROLLEDBACK();
+                        return;
```
mmusgrov: Are you sure about this one? If we tried to commit but ended up rolling back then we need the TRANSACTION_ROLLEDBACK exception.

ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/ControlWrapper.java
```diff
@@ -306,7 +306,7 @@ public final void rollback () throws Unavailable, NoTransaction,
 		}
 		catch (NullPointerException ex)
 		{
-			throw new TRANSACTION_ROLLEDBACK();
+			return;
```
mmusgrov: But this one looks good. We tried to rollback but got NPE because there is no control (aka transaction) and we use presumed abort semantics so therefore we just return without any exception.

ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/orbspecific/coordinator/ArjunaTransactionImple.java
```diff
@@ -390,16 +390,6 @@ public void commit (boolean report_heuristics) throws HeuristicMixed,
 			}
 
 			destroyAction();
-		} else if (outcome == ActionStatus.ABORTED || outcome == ActionStatus.H_ROLLBACK) {
-			throw new TRANSACTION_ROLLEDBACK(ExceptionCodes.FAILED_TO_COMMIT,
```
mmusgrov: You do not need to change anything in the commit method since we are already checking for the ABORTED state and we correctly throw the TRANSACTION_ROLLEDBACK exception.

ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/ControlWrapper.java
tadamski: You are right - this is incorrect if we want to be consistent with ArjunaTransactionImple code which throws TRANSACTION_ROLLEDBACK in it's commit method - fixing.

ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/orbspecific/coordinator/ArjunaTransactionImple.java
tadamski: This is what I wanted to ask as the commit is a bit different that we discussed on forums. I noticed that we have duplicated code in both commit and rollback methods - it caused problems in rollback, but not in commit - anyway I removed the last else from commit too, because to my understanding this check is already performed in switch below with identical exceptions beeing thrown.

BTW: In rollback there were no difference between different statuses in if - those differences are present in switch though, so I removed last else instead of adding another if.

mmusgrov: We are normally very conservative about changing behaviour in the JTS code since it has evolved over many years to address many edge cases.

You have not demonstrated any actual issue with the commit code and in fact I see that you had a test failure in  the commit and had to modify two tests (TransactionTest1 and TerminationTest) to make it work. So unless there are good reasons behind a change we normally reject them.

ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/orbspecific/coordinator/ArjunaTransactionImple.java
```diff
@@ -501,18 +491,6 @@ public void rollback () throws SystemException
 
 			status = super.status();
 		}
-		else
```
mmusgrov: But I agree that there is an issue in the rollback case, namely that trying to rollback an already aborted transaction should not throw an exception and instead should quietly succeed.

However, I am not convinced your code only addresses this requirement because the change introduces more changes than was called for:

* INVALID and CLEANUP now throw UNKNOWN whereas before we threw INVALID_TRANSACTION
* DISABLED, NO_ACTION and CREATED now return silently whereas before we threw INVALID_TRANSACTION

So I have the same comment, we err on the side of caution and usually reject behaviour change unless there are solid reasons to do so.

We can get a second opinion if required but my advice as reviewer is to only address the stated problem and no more and do something along the lines that I suggested earlier.

tadamski:
I have removed the change in ArjunaTransactionImple#commit.

ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/orbspecific/coordinator/ArjunaTransactionImple.java
tomjenkinson: +1

mmusgrov:
@tadamski I have put this PR on hold and removed the assignee label since I am awaiting input from you.

tadamski:
@mmsugrov I know - I agree with your comment - I have returned to "if" version of fix and I'm currently testing it - should push it within minutes

ArjunaJTS/jts/tests/classes/com/hp/mwtests/ts/jts/orbspecific/local/timeout/TerminationTest.java
```diff
@@ -124,8 +124,6 @@ public void test()
                 else
                     handle.rollback();
 
-                assertFalse(commit);
```
mmusgrov: This line should *not* be removed because if commit is true then we expect an exception so this assert will not be triggered. On the other hand if commit is false then we do not expect an exception so this assertFalse(commit) statement will succeed.

ArjunaJTS/jts/tests/classes/com/hp/mwtests/ts/jts/orbspecific/local/timeout/TerminationTest.java
```diff
@@ -192,8 +190,6 @@ public void test()
                 current.commit(true);
             else
                 current.rollback();
-
-            assertFalse(commit);
```
mmusgrov: This line should *not* be removed:

If commit is true then we try to commit a timed out transaction. Therefore current.commit(true) will throw an exception bypassing this assertFalse(commit). If the exception is not thrown then we want the assert to fail.

On the other hand if we try to rollback a timed out transaction then then it will (quietly) succeed and assertFalse(commit) will be true (since commit is false).

ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/orbspecific/coordinator/ArjunaTransactionImple.java
```diff
@@ -500,6 +500,9 @@ public void rollback () throws SystemException
 			destroyAction();
 
 			status = super.status();
+		} else if ((status == ActionStatus.ABORTING)
+				|| (status == ActionStatus.ABORTED) || (status == ActionStatus.H_ROLLBACK)) {
+			return;
```
mmusgrov: Since you will need to fix the test below what do you think about adding an info message informing the user that we have hit this race condition:

```java
			if (jtsLogger.logger.isInfoEnabled()) {
				jtsLogger.i18NLogger.info_rollback_aborted_transaction(get_transaction_name());
			}
```

Where you will need to add the the info_rollback_aborted_transaction method to our logger com.arjuna.ats.jts.logging.jtsI18NLogger by adding to the end of the file (just before the final comment):

```java
	@Message(id = 22263, value = "rollback for {0} was already rolled back", format = MESSAGE_FORMAT)
	@LogMessage(level = INFO)
	public String info_rollback_aborted_transaction(String arg0);
```

mmusgrov: If you agree can you check that the TerminationTest below outputs the info message.

PR #1059, opened 2016-08-25
JBTM-2734 Make sure that when we recover the transactions we refresh …

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2734

!PERF !BLACKTIE !QA_JTA !QA_JTS_JDKORB

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
@@ -259,8 +259,10 @@ public class XARecoveryModule implements RecoveryModule
 
 				// JBTM-1255 moved stale check back to bottomUpRecovery
 				if (xids.contains(xid)) {
-					// This Xid is going to be recovered by the AtomicAction
-					xids.remove(xid);
+					// This Xid will hopefully be recovered by the AtomicAction
+					// We do not remove it from the map as there is garbage collection in RecoveryXids already
+					// and it is possible that the Xid is recovered by both txbridge and XATerminator - the second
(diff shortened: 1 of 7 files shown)


ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
```diff
@@ -259,8 +259,10 @@ private XAResource getNewXAResource(Xid xid)
 
 				// JBTM-1255 moved stale check back to bottomUpRecovery
 				if (xids.contains(xid)) {
-					// This Xid is going to be recovered by the AtomicAction
-					xids.remove(xid);
+					// This Xid will hopefully be recovered by the AtomicAction
+					// We do not remove it from the map as there is garbage collection in RecoveryXids already
```
mmusgrov: Okay this makes sense, I see that the garbage collection is in bottomUpRecovery()

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
```diff
@@ -228,20 +237,23 @@ private TransactionImple addImportedTransaction(TransactionImple importedTransac
...
+		if (recoveredTransaction!= null) {
+			synchronized (holder) {
+				// now it's safe to add the imported transaction to the holder
+				recoveredTransaction.recordTransaction();
+				txn = recoveredTransaction;
+				holder.set(txn);
+			}
+		}
+
 		if (txn == null) {
```
mmusgrov: I would prefer
```java
} else if (txn == null) {
```
here since you have just checked that recoveredTransaction != null

PR #1060, opened 2016-08-25
JBTM-2701 Updated to allow JCA recover call to refresh the list of XI…

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2701

!PERF

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
@@ -53,6 +53,7 @@ import com.arjuna.ats.arjuna.logging.tsLogger;
 import com.arjuna.ats.arjuna.objectstore.RecoveryStore;
 import com.arjuna.ats.arjuna.objectstore.StateStatus;
 import com.arjuna.ats.arjuna.objectstore.StoreManager;
+import com.arjuna.ats.arjuna.recovery.RecoveryManager;
 import com.arjuna.ats.arjuna.recovery.RecoveryModule;
 import com.arjuna.ats.arjuna.state.InputObjectState;
 import com.arjuna.ats.internal.arjuna.common.UidHelper;
@@ -137,13 +138,15 @@ public class XARecoveryModule implements RecoveryModule
 
 	public synchronized void periodicWorkFirstPass()
 	{
-		// JBTM-1354 allow a second thread to execute the first pass but make sure it is only done once per scan (TMSTART/ENDSCAN)
-        synchronized (scanState) {
-            if (!getScanState().equals(ScanStates.IDLE))
-                return;
+		// JBTM-1354 JCA needs to be able to recover XAResources associated with a subordinate transaction so we have to do at least
+		// the start scan to make sure that we have loaded all the XAResources we possibly can to assist subordinate transactions recovering
+		// the reason we can't do bottom up recovery is if this server has an XAResource which tries to recover a remote server (e.g. distributed JTA)
+		// then we get deadlock on the secondpass
+		if (getScanState() == ScanStates.BETWEEN_PASSES) {
+			periodicWorkSecondPass();
+		}
 
-            setScanState(ScanStates.FIRST_PASS); // synchronized uses a reentrant lock
-        }
+		setScanState(ScanStates.FIRST_PASS); // synchronized uses a reentrant lock
 
         if(jtaLogger.logger.isDebugEnabled()) {
             jtaLogger.logger.debugv("{0} - first pass", _logName);
@@ -192,7 +195,7 @@ public class XARecoveryModule implements RecoveryModule
         setScanState(ScanStates.BETWEEN_PASSES);
 	}
 
-	public void periodicWorkSecondPass()
+	public synchronized void periodicWorkSecondPass()
 	{
         setScanState(ScanStates.SECOND_PASS);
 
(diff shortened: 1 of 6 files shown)


ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
```diff
@@ -229,6 +232,26 @@ public void periodicWorkSecondPass()
...
+			Vector recoveryModules = recMan.getModules();
+
+			if (recoveryModules != null) {
+				Enumeration modules = recoveryModules.elements();
+
+				while (modules.hasMoreElements()) {
+					RecoveryModule m = (RecoveryModule) modules.nextElement();
+
+					if (m instanceof XARecoveryModule) {
+						registeredXARecoveryModule = (XARecoveryModule) m;
```
mmusgrov: you should break out of the loop early after finding the target module

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
```diff
@@ -968,9 +995,9 @@ private boolean isHelperInUse(XAResourceRecoveryHelper xaResourceRecoveryHelper)
      */
     private boolean waitForScanState(ScanStates state) {
         try {
-            do {
+			while (!getScanState().equals(state)) {
```
mmusgrov: why did you change the do ... while into a while do loop. The caller already knows the scan state is not equal to state and since this is a private method it is safe to assume that the caller has already checked (under the lock) that he has to wait for a state change.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/XATerminatorImple.java
```diff
@@ -319,12 +324,20 @@ public int prepare (Xid xid) throws XAException
         case XAResource.TMSTARTRSCAN: // check the object store
             if (_recoveryStarted)
                 throw new XAException(XAException.XAER_PROTO);
-            else
+            else {
                 _recoveryStarted = true;
+                if (XARecoveryModule.getRegisteredXARecoveryModule() != null) {
+                    XARecoveryModule.getRegisteredXARecoveryModule().periodicWorkFirstPass();
```
mmusgrov: Remark: if periodicWorkFirstPass() is already running this call will wait and then trigger another first pass which is wasteful of CPU cycles. Prior to your changes repeating the first pass call would have been skipped since we had the following check in first pass (which you removed):
```java
if (!getScanState().equals(ScanStates.IDLE))
```

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/TestCommitMarkableResourceFailActivate.java
```diff
@@ -154,7 +154,6 @@ public void run() {
 
 		// Run the scan to clear the content
 		manager.scan();
-		manager.scan();
```
mmusgrov: Why did you delete the second recovery scan?

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/TestXAResourceRecovery.java
```diff
@@ -30,21 +30,32 @@
 public class TestXAResourceRecovery implements XAResourceRecovery {
 
 	private static Stack<XAResource> resources = new Stack<XAResource>();
+	private static int count = 0;
 
 	public XAResource getXAResource() throws SQLException {
-		return resources.pop();
+		count--;
+		XAResource toReturn = resources.remove(0);
+		resources.push(toReturn);
```
mmusgrov: OK so you removed the item from the bottom of the stack and put it on the top of the stack. I don't know why you are doing that so some comments are in order.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/TestXAResourceRecovery.java
```diff
@@ -30,21 +30,32 @@
 public class TestXAResourceRecovery implements XAResourceRecovery {
 
 	private static Stack<XAResource> resources = new Stack<XAResource>();
+	private static int count = 0;
 
 	public XAResource getXAResource() throws SQLException {
-		return resources.pop();
+		count--;
```
mmusgrov: We are testing that our recovery system is calling XAResourceRecovery correctly. What if if we called it more that three times so count would equal -1 and then the hasMoreResources() call would always return true after that.

If this scenario cannot happen in the test then you should provide more documentation.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
tomjenkinson: +1

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
tomjenkinson: It looks like a relic of interim changes I made - I will revert this

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/TestCommitMarkableResourceFailActivate.java
tomjenkinson: Prior to this change the list of XAResources would need refreshing from getNewXAResource function. This is no longer required to the CMR records will be garbage collected a cycle earlier

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/TestXAResourceRecovery.java
tomjenkinson: +1

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/XATerminatorImple.java
tomjenkinson: If periodicWorkFirstPass is very slow it is possible that the Xids would need refreshing though so I favoured correctness here over optimizing the recovery process

mmusgrov: Ok, agreed

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/TestCommitMarkableResourceFailActivate.java
mmusgrov: OK

PR #1066, opened 2016-09-05
JBTM-2748 Added a new recovery module to load in-doubt JCA transactio…

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2748
!BLACKTIE !XTS !QA_JTA !QA_JTS_JDKORB !PERF

Enables: https://github.com/jbosstm/jboss-as/pull/51

### DIFF
--- a/ArjunaJTS/jtax/tests/classes/com/arjuna/ats/internal/jta/recovery/jts/JCAServerTransactionRecoveryModuleTest.java
+++ b/ArjunaJTS/jtax/tests/classes/com/arjuna/ats/internal/jta/recovery/jts/JCAServerTransactionRecoveryModuleTest.java
@@ -72,7 +72,7 @@ public class JCAServerTransactionRecoveryModuleTest {
     }
 
     @Test
-    public void testIdempotentLoad() throws XAException, RollbackException, SystemException {
+    public void testReloadStateAcceptable() throws XAException, RollbackException, SystemException {
         JCAServerTransactionRecoveryModule module = new JCAServerTransactionRecoveryModule();
         XATerminator terminator = new XATerminatorImple();
         Xid xid = new Xid() {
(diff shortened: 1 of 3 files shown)


ArjunaJTS/jtax/tests/classes/com/arjuna/ats/internal/jta/recovery/jts/JCAServerTransactionRecoveryModuleTest.java
```diff
@@ -0,0 +1,155 @@
...
+            }
+
+            @Override
+            public void start(Xid xid, int i) throws XAException {
+
+            }
+        });
+        terminator.prepare(xid);
+        module.periodicWorkFirstPass();
+        module.periodicWorkFirstPass();
```
mmusgrov: I think you intended periodicWorkSecondPass() even though for this module it does nothing.

tomjenkinson: No, it is to test it is idempotent and doesn't cause an issue with reload although to be fair I could have added a periodicWorkSecondPass() to make it clearer

tomjenkinson: Pushed a change with the secondpasses - as you say they do nothing but might look less confusing. Renamed the test too

mmusgrov: Aha, hence the name of the test. I am not sure why you feel that warrants a test but I will mark the pr as ready to merge anyway.

PR #1068, opened 2016-09-13
JBTM-2749 Thread safe method to import (or create) an Xid

mmusgrov:
https://issues.jboss.org/browse/JBTM-2749

!PERF !BLACKTIE !XTS !PERF !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporter.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporter.java
@@ -59,7 +59,9 @@ public interface TransactionImporter
 	 * @throws javax.transaction.xa.XAException
 	 *             thrown if there are any errors.
 	 */
-    public TransactionImportResult importTransaction(Xid xid, int timeout) throws XAException;
+    public SubordinateTransaction importTransaction(Xid xid, int timeout) throws XAException;
+
+    public TransactionImportResult importRemoteTransaction(Xid xid, int timeout) throws XAException;
 
 	/**
 	 * Used to recover an imported transaction.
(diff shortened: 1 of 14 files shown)


mmusgrov:
@tomjenkinson  This PR and jboss-transaction-spi#14 are now ready for review

mmusgrov:
We should not be changing the ejb subsystem and instead leave that to the owner. The issue is that we changed the signature of an existing method that they are using. I will revert that change add a new method to TransactionImporter that returns the TransactionImportResult and leave the existing importTransaction methoid as is. I will do that tomorrow.

mmusgrov:
@tomjenkinson These are the changes for the new SPI methods I added in jbosstm/jboss-transaction-spi#15 (because I found that ejb3 remoting was using other internal SPI methods)

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporter.java
```diff
@@ -60,6 +61,8 @@
 	 */
     public SubordinateTransaction importTransaction(Xid xid, int timeout) throws XAException;
 
+    public TransactionImportResult importForeignTransaction(Xid xid, int timeout) throws XAException;
```
tomjenkinson: I don't like the name "foreign" it implies its not a Narayana one. Maybe importRemoteTransaction?

mmusgrov: Agreed

mmusgrov: If that is the only change request I will push the update otherwise I will hold back until you have finished the review?

tomjenkinson: Happy with your analysis and changes

mmusgrov:
Retest

PR #1071, opened 2016-09-21
[JBTM-2758][JBTM-2759] Locking on interned string

ochaloup:
Based on static code analysis provided by Coverity tool proposing two fixes for locks on strings which is not recommended approach. String literals are centrally interned and could also be locked on by a library, which could lead to potentially have deadlocks or lock collisions with other code.

!QA_JTA !BLACKTIE !XTS !PERF

### DIFF
--- a/ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/BaseTransactionManagerDelegate.java
+++ b/ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/BaseTransactionManagerDelegate.java
@@ -336,6 +336,11 @@ public abstract class BaseTransactionManagerDelegate implements TransactionManag
     // Bad things will probably happen if users ever use this key themselves
     private final String LOCKS_MAP = "__LOCKS_MAP";
 
+    // Using string LOCKS_MAP directly as object of synchronization is not recommended
+    // String literals are centrally interned and could also be locked on by a library,
+    // potentially having deadlocks or lock collisions
+    private static final Object locksMapLock = new Object();
+
     // locate and return the lock for a given TransactionLocal+Transaction tuple.
     // create it if it does not exist.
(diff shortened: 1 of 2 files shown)


ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/BaseTransactionManagerDelegate.java
```diff
@@ -336,6 +336,11 @@ public void unlock(final TransactionLocal local, final Transaction transaction)
     // Bad things will probably happen if users ever use this key themselves
     private final String LOCKS_MAP = "__LOCKS_MAP";
 
+    // Using string LOCKS_MAP directly as object of synchronization is not recommended
+    // String literals are centrally interned and could also be locked on by a library,
+    // potentially having deadlocks or lock collisions
+    private final Object locksMapLock = new Object();
```
ochaloup: Hi @tomjenkinson , I would like check with you my doubt of this change. I'm not sure if it could be that straightforward as I suggest here. Could not happen `locks` being filled with an empty map from two threads when `locks` variable is empty? It seems that synchronization block starts after `locks` is searched and after is checked to be null. Thanks.

tomjenkinson: +1 
You need locksMapLock to be a static final Object, because of the reason the static analysis tool gave ("_LOCKS_MAP" as a final string is interned so same object used by other threads).

ochaloup: I see. Changed. Thanks!

tomjenkinson: Please can you also confirm that the change has removed the findbugs issue? Thanks Ondra!

PR #1072, opened 2016-09-27
JBTM-2749 SPI implementation of ExtendedJBossXATerminator

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2749

!PERF !BLACKTIE !QA_JTA !QA_JTS_JDKORB !XTS


mmusgrov:
I think we need to retest the quickstarts with these changes.

PR #1076, opened 2016-10-03
[JBTM-2334] Narayana Tomcat module

gytis:
https://issues.jboss.org/browse/JBTM-2334

!XTS !BLACKTIE !PERF !QA_JTA !QA_JTS_JDKORB

### DIFF
--- a/tomcat/tomcat-jta/src/main/java/org/jboss/narayana/tomcat/jta/NarayanaJtaServletContextListener.java
+++ b/tomcat/tomcat-jta/src/main/java/org/jboss/narayana/tomcat/jta/NarayanaJtaServletContextListener.java
@@ -1,0 +1,135 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2016, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.jboss.narayana.tomcat.jta;
+
+import com.arjuna.ats.arjuna.common.CoreEnvironmentBeanException;
+import com.arjuna.ats.arjuna.common.arjPropertyManager;
+import com.arjuna.ats.arjuna.common.recoveryPropertyManager;
+import com.arjuna.ats.arjuna.coordinator.TransactionReaper;
+import com.arjuna.ats.arjuna.coordinator.TxControl;
+import com.arjuna.ats.arjuna.recovery.RecoveryManager;
+import com.arjuna.ats.internal.arjuna.recovery.AtomicActionRecoveryModule;
+import com.arjuna.ats.internal.arjuna.recovery.ExpiredTransactionStatusManagerScanner;
+import com.arjuna.ats.internal.jta.recovery.arjunacore.JTANodeNameXAResourceOrphanFilter;
+import com.arjuna.ats.internal.jta.recovery.arjunacore.JTATransactionLogXAResourceOrphanFilter;
+import com.arjuna.ats.internal.jta.recovery.arjunacore.XARecoveryModule;
+import com.arjuna.ats.jdbc.TransactionalDriver;
+import com.arjuna.ats.jta.common.jtaPropertyManager;
+
+import javax.servlet.ServletContextEvent;
+import javax.servlet.ServletContextListener;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * @author <a href="mailto:gytis@redhat.com">Gytis Trikleris</a>
+ */
+public class NarayanaJtaServletContextListener implements ServletContextListener {
+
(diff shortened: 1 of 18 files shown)


ArjunaJTA/tomcat/src/test/java/org/jboss/narayana/jta/tomcat/integration/TestApplication.java
```diff
@@ -0,0 +1,34 @@
...
+package org.jboss.narayana.jta.tomcat.integration;
+
+import javax.ws.rs.ApplicationPath;
+import javax.ws.rs.core.Application;
+
+/**
+ * @author <a href="mailto:gytis@redhat.com">Gytis Trikleris</a>
+ */
+@ApplicationPath("/")
+public class TestApplication extends Application {
```
tomjenkinson: Out of interest - what does this do?

gytis:
Well the only other option is to move the module to the root level. Possibly creating a wrapper module first.

gytis:
jbossts-properties.xml is not added to the archive. User needs to add it to his application

ArjunaJTA/tomcat/src/main/java/org/jboss/narayana/jta/tomcat/NarayanaJtaLifecycleListener.java
```diff
@@ -0,0 +1,135 @@
...
+
+    private static final List<String> DEFAULT_ORPHAN_FILTERS = Arrays
+            .asList(JTATransactionLogXAResourceOrphanFilter.class.getName(), JTANodeNameXAResourceOrphanFilter.class.getName());
+
+    private static final List<String> DEFAULT_EXPIRY_SCANNERS = Arrays
+            .asList(ExpiredTransactionStatusManagerScanner.class.getName());
+
+    @Override
+    public void contextInitialized(ServletContextEvent servletContextEvent) {
+        LOGGER.info("Initializing Narayana");
```
tomjenkinson: Is there a precedent? We don't want to be the only subsystem that does this. If there is a precedent (either to info or trace) lets go with that. Otherwise we can use the precedent in WFLY.

ArjunaJTA/tomcat/src/main/java/org/jboss/narayana/jta/tomcat/NarayanaJtaLifecycleListener.java
```diff
@@ -0,0 +1,135 @@
...
+                DriverManager.deregisterDriver(d);
+            } catch (SQLException e) {
+                LOGGER.log(Level.WARNING, e.getMessage(), e);
+            }
+        });
+    }
+
+    private void initNodeIdentifier() {
+        if (arjPropertyManager.getCoreEnvironmentBean().getNodeIdentifier() == null) {
+            LOGGER.info("Node identifier was not set. Setting it to the default value: " + DEFAULT_NODE_IDENTIFIER);
```
tomjenkinson: Thanks!

gytis:
I think it would make things more complicated for the user. Especially when we only ever got requests for JTA and not JTS.
Also would make things more complicated for us too by adding a necessity for more testing and support.

gytis:
I'm ok with moving it to a different place by the way. I'm just agains taking JTA out

gytis:
Also, in case we need to provide JTS for Tomcat, we could provide a different package.
Now I've tried to make it as easy as possible and currently user only needs to add one extra dependency to his application. Otherwise he'd have to pick all extras such as JTA/JTS, JDBC etc.

gytis:
There will be in Java 9 once they introduce modules.
With maven they only need to add one dependency to this tomcat module. If we wouldn't provide fat jar, user would need to add dependencies to the following: JTA spec, Narayana Common, Arjuna Core, Arjuna JTA, Narayana JDBC, Narayana SPI, and JBoss Logging.
I don't see a problem with fat jar however. It's quite common now and that's how Spring Boot and WildFly Swarm work.

gytis:
We could name this module tomcat-jta and if another user comes up who asks for JTS, we can introduce another module tomcat-jts.

gytis:
Well all in all it's the testing what took the most time. Code wise it's not a lot of stuff going on. For JTS the context listener should be rewritten to add required recovery bits.
As for the dependencies that module could depend on this one and exclude unnecessary bits.

gytis:
It only provides default values. If for example no recovery modules are provided by the config file, then it adds the default ones. Same applies for orphan filters, expiry scanners, and node id.

gytis:
So where do you want me to move the module? to /tomcat/tomcat-jta?

gytis:
There are three classes which would be common between JTA and JTS, so I could move them to /tomcat/tomcat-common (/tomcat would be a pom module). But again they wouldn't be needed by any other possible tomcat modules e.g. rest, stm, compensations. Those classes would be factories for TransactionManager, UserTransaction, and SynchronizationRegistry.
I really think that we're overthinking it at the moment. Do you think that there will be a requirement for tomcat-jts module any time soon?

gytis:
Fair enough, then we can plan to have JTS one in the future. And abstract this one where possible
I still believe that we should make it as easy as possible for the user to deploy it. I.e. deploy only one jar and provide sensible defaults. Majority of the users wouldn't need to change any properties anyway, so it would be just an additional hassle for them to research how to properly configure jbossts-properties.xml

gytis:
Tomcat is a bit different from WildFly in that sense, and config should go in the application for this. So we would need to provide a custom configuration file for that.

How about we shade all bits for jta and jts (basically copy narayana-jts-idlj) but don't add properties file into the fat jar.  Then if user doesn't provide his jbossts-properites.xml we configure default options (jta) programatically.
Also, user can provide any options he wants with his own jbossts-properites.xml which will be taken instead of defaults.
And in addition, we add a new config field for the user to enable jts via jbossts-properites.xml.

gytis:
@tomjenkinson I've moved all the stuff to /tomcat/tomcat-jta

tomcat/tomcat-jta/src/main/java/org/jboss/narayana/tomcat/jta/NarayanaJtaServletContextListener.java
```diff
@@ -0,0 +1,135 @@
...
+                DriverManager.deregisterDriver(d);
+            } catch (SQLException e) {
+                LOGGER.log(Level.WARNING, e.getMessage(), e);
+            }
+        });
+    }
+
+    private void initNodeIdentifier() {
+        if (arjPropertyManager.getCoreEnvironmentBean().getNodeIdentifier() == null) {
+            LOGGER.info("Node identifier was not set. Setting it to the default value: " + DEFAULT_NODE_IDENTIFIER);
```
tomjenkinson: We did say this would be a WARN

gytis:
Yes sorry forgot about the logging. Will add that

gytis:
Updated the logging and created a JIRA for JTS

PR #1078, opened 2016-10-18
JBTM-2774 master

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2774

!PERF !BLACKTIE !XTS !QA_JTS_JDKORB !QA_JTA

PR #1082, opened 2016-10-21
JBTM-2774 5.2

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2774

!PERF !BLACKTIE !XTS !QA_JTS_JDKORB !QA_JTA

Backport to 5.2 of https://github.com/jbosstm/narayana/pull/1078

### DIFF
--- a/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/TransactionImporterImple.java
+++ b/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/TransactionImporterImple.java
@@ -34,6 +34,7 @@ package com.arjuna.ats.internal.jta.transaction.jts.jca;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicReference;
 
+import javax.transaction.SystemException;
 import javax.transaction.xa.XAException;
 import javax.transaction.xa.Xid;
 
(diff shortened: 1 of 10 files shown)


ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/TransactionImporterImple.java
```diff
@@ -159,7 +169,10 @@ public void removeImportedTransaction (Xid xid) throws XAException
 		if (xid == null)
 			throw new IllegalArgumentException();
 
-		_transactions.remove(new SubordinateXidImple(xid));
+		AtomicReference<SubordinateTransaction> remove = _transactions.remove(new SubordinateXidImple(xid));
+		if (remove != null) {
+			TransactionImple.removeTransaction((com.arjuna.ats.internal.jta.transaction.jts.TransactionImple) remove.get());
+		}
```
mmusgrov: I know I have already reviewed the other two PRs but I am wondering if there is a race here. When we first import a new xid we initially put an AtomicReference in the map with a null value:

https://github.com/jbosstm/narayana/blob/master/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/TransactionImporterImple.java#L180

Therefore it can happen that this remove.get() call could be null and if this is the case then the TransactionImple.removeTransaction() call will blow up.

PR #1084, opened 2016-10-27
WFLY-6966 Updated to shutdown the ORB after the recovery manager is c…

tomjenkinson:
https://issues.jboss.org/browse/WFLY-6966

!PERF !BLACKTIE !XTS !QA_JTA


gytis:
I might be missing something, but this seems like simply moving problem to the different service. What if recovery manager is stopped before transaction manager? Then transaction manager will get the warning.

gytis:
That doesn't sound right. If recovery manager depends on the transaction manager, shouldn't transaction manager run while recovery manager is running, because otherwise it will not be able to function?

gytis:
I've checked the code. It's the opposite, transaction manager service depends on recovery manager service

gytis:
So it should work ok then if this code is only used by the application server.

PR #1087, opened 2016-11-08
JBTM-2673 JBTM-2786-7 JBTM-2800-4

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2786
https://issues.jboss.org/browse/JBTM-2787

!MAIN !XTS !QA_JTA !QA_JTS_JDKORB !PERF !BLACKTIE

JDBCSTORE

Added hold as I will need to raise JBTM for some of these issues that are related to errors running the test harness

### DIFF
--- a/ArjunaJTS/jts/pom.xml
+++ b/ArjunaJTS/jts/pom.xml
@@ -196,6 +196,12 @@
       <scope>test</scope>
       <version>${version.org.jboss.byteman}</version>
     </dependency>
+    <dependency>
+      <groupId>log4j</groupId>
+      <artifactId>log4j</artifactId>
+      <scope>test</scope>
+      <version>${version.log4j}</version>
+    </dependency>
   </dependencies>
 
   <profiles>
--- a/ArjunaJTS/jts/tests/classes/com/hp/mwtests/ts/jts/orbspecific/resources/AtomicWorker3.java
+++ b/ArjunaJTS/jts/tests/classes/com/hp/mwtests/ts/jts/orbspecific/resources/AtomicWorker3.java
@@ -35,9 +35,11 @@ import com.arjuna.ats.internal.jts.OTSImpleManager;
 import com.arjuna.ats.internal.jts.orbspecific.CurrentImple;
 import com.hp.mwtests.ts.jts.exceptions.TestException;
 import com.hp.mwtests.ts.jts.utils.Util;
+import org.jboss.logging.Logger;
 
 public class AtomicWorker3
 {
+	public static final Logger logger = Logger.getLogger("AtomicWorker3");
 
 public static void randomOperation (int thr, int level)
     {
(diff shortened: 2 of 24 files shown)


scripts/hudson/narayana.sh
```diff
@@ -603,7 +603,7 @@ function qa_tests_once {
     [ $CODE_COVERAGE = 1 ] && codeCoverage=true
   done
 
-  git checkout TaskImpl.properties
+  cp TaskImpl.properties.template TaskImpl.properties
```
mmusgrov: This is a handy fix.

Note that you should also update qa/README.txt

tomjenkinson: Thanks Mike - I have updated the PR as you recommended

ArjunaJTS/jts/pom.xml
```diff
@@ -196,6 +196,12 @@
       <scope>test</scope>
       <version>${version.org.jboss.byteman}</version>
     </dependency>
+    <dependency>
+      <groupId>log4j</groupId>
+      <artifactId>log4j</artifactId>
+      <scope>test</scope>
+      <version>1.2.17</version>
```
mmusgrov: Use ${version.log4j}

ArjunaJTS/jts/tests/classes/com/hp/mwtests/ts/jts/orbspecific/resources/AtomicWorker3.java
```diff
@@ -74,20 +76,20 @@ public static void randomOperation (int thr, int level)
 		{
 		    current.begin();
 
-		    Util.indent(thr, level);
-		    System.out.println("begin");
+		    logger.info(""+level);
```
mmusgrov: Reading the word level rather then looking at an actual text indentation is not so visually appealing. But if you really think we should be avoiding System.out then that's fine (and you probably need to remove the Util.indent methods to make sure they aren't used anywhere else).

ArjunaJTS/jts/tests/classes/com/hp/mwtests/ts/jts/orbspecific/resources/AtomicWorker3.java
```diff
@@ -476,30 +478,20 @@ public static int get1() throws Exception
...
             throw e;
+        } finally {
+            if (res)
+                current.commit(false);
+            else
+                current.rollback();
         }
 
-        if (!res)
-            throw new Exception("Get1: Failed to retrieve value");
```
mmusgrov: The test would previously have failed if this exception was thrown - please can you indicate why it is safe to ignore failed calls to atomicObject_1.get() etc

tomjenkinson: I think if (!res) to throw an exception should be handled in the "catch throw e" above as res should never be false by this point now

ArjunaJTS/jts/pom.xml
tomjenkinson: +1 thanks

ArjunaJTS/jts/tests/classes/com/hp/mwtests/ts/jts/orbspecific/resources/AtomicWorker3.java
tomjenkinson: the interleaving of the threads just did not work for me as the resulting out.print just interleaved. I purposefully didn't touch the other tests that do use Util.java as I didn't have an issue with those tests and wanted to minimize the change set.

I would therefore prefer to not make further changes for this one unless there is something you spotted?

mmusgrov:
An epic fix.

Also did you discount the rawresources02_2b RawResources02_2b_Test113 failure as unrelated to your changes.

PR #1090, opened 2016-11-08
JBTM-2673 Updated to use a map for storing the thread IDs rather than…

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2673

!QA_JTA !QA_JTS_JDKORB !XTS !PERF !BLACKTIE

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/utils/ThreadUtil.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/utils/ThreadUtil.java
@@ -20,6 +20,10 @@
  */
 package com.arjuna.ats.arjuna.utils;
 
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicLong;
+
 /**
  * Provides utilities to manage thread ids.
  */
@@ -28,11 +32,11 @@ public class ThreadUtil
     /**
      * The ID associated with the thread.
      */
-    private static final ThreadLocal THREAD_ID = new ThreadLocal() ;
+    private static final Map<Thread,String> THREAD_ID = new HashMap<Thread,String>() ;
     /**
      * The thread id counter.
      */
-    private static long id ;
+    private static AtomicLong id = new AtomicLong();
     
     /**
      * Get the string ID for the current thread.
@@ -50,23 +54,33 @@ public class ThreadUtil
      */
     public static String getThreadId(final Thread thread)
     {
-	final Object id = THREAD_ID.get() ;
+	final String id = THREAD_ID.get(thread) ;
 	if (id != null)
 	{
-	    return (String)id ;
+	    return id ;
 	}
 	
 	final String newId = getNextId() ;
-	THREAD_ID.set(newId) ;
+	THREAD_ID.put(thread,newId) ;
 	return newId ;
+    }
+    
+    /**
+     * As the association is stored in a map callers need to ensure they remove from the thread when the threadID is no longer required.
+     * 
+     * @param thread The thread to remove from the map
+     * @return The ID that matched the thread
+     */
+    public static String removeThreadId(final Thread thread) {
+        return THREAD_ID.remove(thread);
     }
     
     /**
      * Get the next thread id to use.
      * @return The next thread id.
      */
-    private static synchronized String getNextId()
+    private static String getNextId()
     {
-	return "TSThread:" + Long.toHexString(++id) ;
+	return Long.toHexString(id.incrementAndGet()) ;
     }
 }
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/thread/ThreadActionData.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/thread/ThreadActionData.java
@@ -242,7 +242,7 @@ public class ThreadActionData
 
 					if (act != null)
                     {
-						act.removeChildThread(ThreadUtil.getThreadId(t));
+						act.removeChildThread(ThreadUtil.removeThreadId(t));
                     }
 				}
 			}
--- a/ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/context/ContextManager.java
+++ b/ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/context/ContextManager.java
@@ -194,7 +194,7 @@ public class ContextManager
     }
 
 	ControlWrapper action = null;
-	Object arg = otsCurrent.get(threadId);
+	Object arg = threadId == null ? null : otsCurrent.get(threadId);
 
 	if (arg != null)
 	{
@@ -296,7 +296,7 @@ public class ContextManager
 
     public final void purgeActions ()
     {
-        purgeActions(ThreadUtil.getThreadId());
+        purgeActions(ThreadUtil.removeThreadId(Thread.currentThread()));
     }
 
     /**
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
@@ -634,7 +634,7 @@ public class BasicAction extends StateManager
 
     public final boolean removeChildThread () // current thread
     {
-        return removeChildThread(ThreadUtil.getThreadId());
+        return removeChildThread(ThreadUtil.removeThreadId(Thread.currentThread()));
     }
 
     /**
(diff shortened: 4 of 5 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/utils/ThreadUtil.java
```diff
@@ -50,18 +53,28 @@ public static String getThreadId()
      */
     public static String getThreadId(final Thread thread)
     {
-	final Object id = THREAD_ID.get() ;
+	final Object id = THREAD_ID.get(thread) ;
```
mmusgrov: This is a String so use final String id = THREAD_ID.get(thread) ; and then you don't need to do the cast on the next line.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/utils/ThreadUtil.java
```diff
@@ -28,7 +31,7 @@
     /**
      * The ID associated with the thread.
      */
-    private static final ThreadLocal THREAD_ID = new ThreadLocal() ;
+    private static final Map<Thread,String> THREAD_ID = new HashMap<Thread,String>() ;
```
mmusgrov: I would internally use Map<Thread,Long> since longs are more efficient than Strings

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/utils/ThreadUtil.java
```diff
@@ -50,18 +53,28 @@ public static String getThreadId()
...
     public static String getThreadId(final Thread thread)
     {
-	final Object id = THREAD_ID.get() ;
+	final Object id = THREAD_ID.get(thread) ;
 	if (id != null)
 	{
 	    return (String)id ;
 	}
 	
 	final String newId = getNextId() ;
```
mmusgrov: I think an AtomicLong will be more performant

ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/context/ContextManager.java
```diff
@@ -296,7 +296,7 @@ public final void purgeActions (String threadId)
 
     public final void purgeActions ()
     {
-        purgeActions(ThreadUtil.getThreadId());
+        purgeActions(ThreadUtil.removeThreadId(Thread.currentThread()));
```
mmusgrov: If Thread.currentThread() is not in the THREAD_ID map then this statement will be equivalent to
```java
purgeActions(null)
```
and this will call
```java
 public final ControlWrapper popAction (String threadId)
```
which blows up when it calls
```java
Object arg = otsCurrent.get(threadId);
```
because otsCurrent is a Hashtable which 
```java
@throws NullPointerException if the specified key is null
```

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/utils/ThreadUtil.java
tomjenkinson: well, I can acknowledge that but if I moved to using a long for threadId then we would end up changing the public interface of some classes. Initially I found BasicAction::removeChildThread(String threadId) - there could be more....

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/utils/ThreadUtil.java
tomjenkinson: sure, good catch

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/utils/ThreadUtil.java
tomjenkinson: I think that is a separate enhancement due to the API change required?

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/thread/ThreadActionData.java
```diff
@@ -242,7 +242,7 @@ public static void purgeActions (Thread t, boolean unregister)
 
 					if (act != null)
                     {
-						act.removeChildThread(ThreadUtil.getThreadId(t));
+						act.removeChildThread(ThreadUtil.removeThreadId(t));
                     }
```
mmusgrov: Same comment as below about the possibility that ThreadUtil.removeThreadId(t) can return a null which will cause removeChildThread(null) to throw an NPE. The old code guaranteed that ThreadUtil.getThreadId(t) would always return a non null value. So I think you need an if null here:

```java
String tid = ThreadUtil.removeThreadId(t)

if (tid != null)
  act.removeChildThread(tid);
```

ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/context/ContextManager.java
tomjenkinson: I think I already mitigated that with: Object arg = threadId == null ? null : otsCurrent.get(threadId);

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/thread/ThreadActionData.java
tomjenkinson: Can it return null at this point of the execution though? From what I can see it should not be possible.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/utils/ThreadUtil.java
mmusgrov: I meant use a long inside the ThreadUtil class but convert them to Strings in the interface.

ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/context/ContextManager.java
mmusgrov: That test is in a different method. My comment relates to:
```java
    public final void purgeActions ()
    {
        purgeActions(ThreadUtil.removeThreadId(Thread.currentThread()));
    }
```

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/utils/ThreadUtil.java
tomjenkinson: I wondered if you meant that but then I thought that catting the "TSThread"+getId() each time would be more expensive than storing a string in the map?

ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/context/ContextManager.java
mmusgrov: OK I see you point. Please ignore this comment.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/thread/ThreadActionData.java
mmusgrov: The call to act.removeChildThread ... occurs in the method
```java
public static void purgeActions (Thread t, boolean unregister)
```
If the passed in thread t has not be added to the map before then ThreadUtil.removeThreadId(t) will be null.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/utils/ThreadUtil.java
mmusgrov: You also need to change the implementation:
```java
private static synchronized String getNextId()
    {
	return Long.toHexString(id.incrementAndGet()) ;
    }
```

tomjenkinson: so you mean to no longer prefix our threadIds?

mmusgrov: I do. This ids are internal and are used for key stuff in private maps.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/thread/ThreadActionData.java
mmusgrov: This called method checks for null so I withdraw my comment.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/utils/ThreadUtil.java
tomjenkinson: OK, I will do that then - thanks for the suggestion

PR #1093, opened 2016-11-29
[JBTM-2799] Javadoc Tomcat module

gytis:
https://issues.jboss.org/browse/JBTM-2799

!BLACKTIE !PERF !XTS !QA_JTA !QA_JTS_JDKORB

### DIFF
--- a/narayana-full/pom.xml
+++ b/narayana-full/pom.xml
@@ -167,6 +167,14 @@
       <scope>compile</scope>
     </dependency>
 
+    <!-- Tomcat module javadocs -->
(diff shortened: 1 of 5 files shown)


gytis:
Release profile needs to be enabled in order to generate javadocs and I'm not sure that CI build does that. I did however test it locally.

gytis:
OK in that case I'll run a build

gytis:
Jenkins doesn't save javadocs in the artifacts. So cannot actually see them, but they were generated based on the logs.

PR #1096, opened 2016-12-02
JBTM-2808 Must sync weakhashmap or hang occurs

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2808

!XTS !QA_JTA !QA_JTS_JDKORB !BLACKTIE


mmusgrov:
Yes I took a look at this PR this morning and it looks like a safe fix. Did it not resolve the hang/concurrent modification. Why did you need to revert #1096?

PR #1100, opened 2016-12-06
JBTM-2812 Ensure JTS/JTA forget behaviour is consistent

mmusgrov:
https://issues.jboss.org/browse/JBTM-2812

### DIFF
--- a/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
+++ b/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
@@ -843,10 +843,16 @@ public class XAResourceRecord extends com.arjuna.ArjunaOTS.OTSAbstractRecordPOA
 
 	                    throw new org.omg.CosTransactions.HeuristicHazard();
 	                case XAException.XA_HEURCOM:
-	                    handleForget() ;
+	                    try {
+	                        handleForget() ;
+	                    } catch (SystemException ignore) {
+	                    }
 	                    break;
 	                case XAException.XA_HEURRB:
-	                       handleForget() ;
+	                       try {
+	                           handleForget() ;
+	                       } catch (SystemException ignore) {
+	                       }
 	                       throw new TRANSACTION_ROLLEDBACK();
 	                case XAException.XA_RBROLLBACK:
 	                case XAException.XA_RBCOMMFAIL:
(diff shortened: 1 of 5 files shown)


ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -909,14 +909,15 @@ public void forget() throws org.omg.CORBA.SystemException
             jtaxLogger.logger.trace("XAResourceRecord.forget for " + _tranID);
         }
 
-		handleForget() ;
+		if (!handleForget())
+		    throw new UNKNOWN(); // will cause log to be rewritten.
```
tomjenkinson: Great find - not sure how I missed that one!

mmusgrov:
@tomjenkinson I do have a couple of more changes to push. I will put those in a separate commit and squash if you are ok with them.

mmusgrov:
@tomjenkinson I added a small extra commit

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -909,14 +909,15 @@ public void forget() throws org.omg.CORBA.SystemException
...
 		destroyState();
 
 		removeConnection();
 	}
 
-	private void handleForget()
+	private boolean handleForget()
 	{
 		if ((_theXAResource != null) && (_tranID != null))
 		{
```
tomjenkinson: I think you need to move the _heuristic = TwoPhaseOutcome.FINISH_OK; to the same place as the local JTA one.

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -928,8 +929,11 @@ private void handleForget()
 			}
 			catch (Exception e)
 			{
+                return false;
```
tomjenkinson: Could add the warning as you did in local JTA

mmusgrov:
The issue title is: "JBTM-2812 Ensure failed forget calls are reported"
I am also reported the underlying cause (e.getMessage()) in the UNKNOW exception.
I am issuing a log warning if forget fails in the JTS case (just like I was in the JTA case).

What did you mean by your other point: "reflect them in the JTS _heuristic setting"?

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -909,14 +915,14 @@ public void forget() throws org.omg.CORBA.SystemException
             jtaxLogger.logger.trace("XAResourceRecord.forget for " + _tranID);
         }
 
-		handleForget() ;
+		handleForget();
```
tomjenkinson: superfluous?

mmusgrov: I meant I squashed the commits and reworded the title.

mmusgrov:
I changed the issue title and squashed the commits in response to your comment (the PR comment will not reflect that).

Good catch regarding not updating the heuristic in the failure case.

mmusgrov:
The issue title is currently: "JBTM-2812 Ensure failed forget calls are reported"
How about:
   "JBTM-2812 Ensure we retain a log if the resource forget call fails"

mmusgrov:
I will amend the commit message to reflect the new well crafted title: "Retain a log if a forget call on a resource fails" ;)

And I think that the new jira description is equally poetic too:

> When we clear a heuristic from XAResourceRecord we call forget on the resource and then delete the record even if the forget operation fails. The preferred behaviour is to retain the record if the forget fails so that it can be reported via our tooling and so that we still have the option of later retrying the forget operation (also via the tooling).

mmusgrov:
@tomjenkinson The hold is pending unit tests for the JTS version of XAResourceRecordMBean

mmusgrov:
What do you think of my new tests for the JTS variant of XAResourceRecord?

mmusgrov:
Sorry, I added and then immediately deleted a comment about my new tests. I deleted it since I noticed that you had already reviewed the new ObjStoreBrowserTest.java

PR #1103, opened 2016-12-08
Jbtm 2814 tx vertx

mmusgrov:
https://issues.jboss.org/browse/JBTM-2814


mmusgrov:
@tomjenkinson @nmcl has already reviewed this PR but I want to give you the opportunity to comment before I merge.

Note that in a future PR it may make sense to put some of the code in the quickstart repo but for now I would like it to be in the main repo so that we have an easily accessible audit trail.

Also I think the inclusion of transactional vertx in the main repo is worth a blog post before you put out the next release so I will schedule that task.

vertx/echo/EchoClient.java
```diff
@@ -1,25 +1,7 @@
-/*
```
tomjenkinson: Why copyright removed?

vertx/echo/EchoServer.java
```diff
@@ -1,34 +1,30 @@
+
+import io.vertx.core.Vertx;
+import io.vertx.core.AbstractVerticle;
+import io.vertx.core.streams.Pump;
+
 /*
- * Copyright 2011 the original author or authors.
```
tomjenkinson: Why copyright removed?

vertx/echo/build.sh
```diff
@@ -0,0 +1,29 @@
+#JAVA_HOME=/usr/local/jdk1.8.0_91
```
tomjenkinson: add copyright

vertx/echo/pom.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tomjenkinson: add copyright

vertx/module/pom.xml
```diff
@@ -59,134 +63,84 @@
...
+        <groupId>io.vertx</groupId>
+        <artifactId>vertx-dependencies</artifactId>
+        <version>${vertx.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    <dependency>
+      <groupId>org.jboss.logging</groupId>
+      <artifactId>jboss-logging</artifactId>
+      <version>3.1.4.GA</version>
```
tomjenkinson: declare version in properties

vertx/raw/pom.xml
```diff
@@ -0,0 +1,111 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.jboss.narayana.stm</groupId>
+      <artifactId>stm</artifactId>
+      <version>${narayana.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jboss.logging</groupId>
+      <artifactId>jboss-logging</artifactId>
+      <version>3.1.4.GA</version>
```
tomjenkinson: version from properties

vertx/echo/EchoClient.java
mmusgrov: The original was taken from the vertx repo and the examples do not contain copyrights.

vertx/echo/EchoServer.java
mmusgrov: ditto

vertx/echo/build.sh
mmusgrov: ditto

vertx/echo/pom.xml
mmusgrov: Although I created this pom, to be consistent with the original vert.x example, there should be no copyright.

mmusgrov:
@tomjenkinson:

1. I did not remove any copyright notices - but I did need to take a copy of the echo example from the Vert.x 3 repo and those examples no longer contain copyrights
2. I added a version property for referencing ${jboss-logging.version}
3. I added copyrights to the poms that I added

mmusgrov:
Yes I noticed that you had drilled down into the root case with Gytis. 

I will merge then

PR #1106, opened 2016-12-13
JBTM-2813 By default do not create JMX instrumentation for CosTransac…

mmusgrov:
…tion/XAResourceRecord record types
https://issues.jboss.org/browse/JBTM-2813
https://issues.jboss.org/browse/JBEAP-7735

@ochaloup I changed the tooling such that CosTransaction/XAResourceRecord records types are not instrumented by default. That fixes the first assert failure in your test suite
```java
JcaInflowTransactionTestCase#rmerrWithRecovery
Assert.assertEquals("Recovery should not touch the txn as it was created by ...);
```

but now the assert a little later in the method is failing (which should be unrelated to the tooling instrumentation changes):
```java
instrumentedTestXAResource.assertMethodCalled("commit");
```

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/ObjStoreBrowser.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/ObjStoreBrowser.java
@@ -84,6 +84,7 @@ public class ObjStoreBrowser implements ObjStoreBrowserMBean {
     private static OSBTypeHandler[] defaultJTSOsbTypes = {
             new OSBTypeHandler(
                     true,
+                    false, // by default do not probe for this type
                     "com.arjuna.ats.internal.jta.tools.osb.mbean.jts.JTSXAResourceRecordWrapper",
                     "com.arjuna.ats.internal.jta.tools.osb.mbean.jts.JTSXAResourceRecordWrapper",
                     "CosTransactions/XAResourceRecord",


mmusgrov:
On Hold pending discussion with QE

ochaloup:
Hi @mmusgrov,

I've checked the fix and it works fine for me.

The failure of the testcase is caused by the fact that cli operation `:recover()` run on particular participant for the inflow transaction does not recover the participant to `prepare` state. It's different (probably) issue. I would create a new issue on it, what do you think? Or it's rather connected with your change in this PR and it should be checked here?

What I can see
```
/subsystem=transactions/log-store=log-store/transactions=0\:ffff7f000001\:-4cecb39b\:585109e2\:28/participants=0\:ffff7f000001\:-4cecb39b\:585109e2\:31:recover()
```
ends as
```
[/subsystem=transactions/log-store=log-store/transactions=0\:ffff7f000001\:-4cecb39b\:585109e2\:28:read-resource(recursive=true, include-runtime=true)
{
    "outcome" => "success",
    "result" => {
        "age-in-seconds" => "1481706240",
        "id" => "0:ffff7f000001:-4cecb39b:585109e2:28",
        "jmx-name" => undefined,
        "type" => "StateManager/BasicAction/TwoPhaseCoordinator/ArjunaTransactionImple/ServerTransaction/JCA",
        "participants" => {"0:ffff7f000001:-4cecb39b:585109e2:31" => {
            "eis-product-name" => "unavailable",
            "eis-product-version" => "unavailable",
            "jmx-name" => undefined,
            "jndi-name" => "0:ffff7f000001:-4cecb39b:585109e2:31",
            "status" => "HEURISTIC_ROLLBACK",
            "type" => "/StateManager/AbstractRecord/ExtendedResourceRecord"
        }}
    }
}
```
instead of participant status being changed to PREPARED.

mmusgrov:
recover failing is not related to this PR.

Does the log say why the subordinate thinks the (subordinate) transaction has rolled back (is the actual test XA resource rolled back?).

mmusgrov:
@ochaloup Also did you run your full crash rec test suite to validate that no other tests were relying on the instrumentation of CosTransaction/XAResourceRecord record types.

ochaloup:
@mmusgrov th subordinate transaction is not rolled-back. The point is that the record is left in heuristic state in txn object store. Because the status is not changed to prepared the RAR commit operation has no effect.

I run some of tests of our crashrec testsuite which were failing before and it seems good from this point (running all tests is really time consuming and I would say manageable only during whole eap testing cycle)

mmusgrov:
Are you saying that the tooling :reover operation is no longer moving heuristics back onto the prepared list - I am sure I have unit tests for that.

ochaloup:
yes, that's what I can see in my test. But it's only JTS and it's only (I think) for `"type" => "StateManager/BasicAction/TwoPhaseCoordinator/ArjunaTransactionImple/ServerTransaction/JCA",`

mmusgrov:
@tomjenkinson only the tooling tests are effected by this change and they all passed so I think this PR is ready for review

mmusgrov:
I only disabled creation of MBeans for the OTS resource records. We still expose the inlined (with the trdansaction log) logs for ExtendedResourceRecord and in the CLI these appear underneath the owning transaction node. 
 
 I did ask @ochaloup which tests he had reran:

> I run some of tests of our crashrec testsuite which were failing before and it seems good from this point 

but do not know if that set included the ones for validating https://jira.jboss.org/browse/JBTM-2767.

If the QE test suite really does rely on it this MBean then it can be re-enabled using the *"expose all log records"* property. Ondrej, can you confirm whether this change breaks the fix we did for  JBTM-2767.

Finally, finally the reason why we are no longer creating MBeans for these OTS records comes from QE:

when they list transactions in the WildFly management model they get transactions + these participant records. Normally participants appear underneath their owning transaction node.But these OTS records could be remote participants and hence the store does not necessarily contain the owning transaction and therefore they are displayed in the CLI at the same level as transactions. Ondrej says that this is:

> not user friendly - when I list transactions in cli I expect [only] transactions and not some [other kind of] log record

So, to address this point I simply removed them - the alternative of setting up a distinct management model node to serve as the context for these other record types would involve a lot of effort and the resulting gain did not justify that. Besides if they are required then one can just set the *"expose all log records"* property.

mmusgrov:
@ochaloup Are you OK with me pushing this fix (see my previous comment)?

ochaloup:
@tomjenkinson I checked our test which caused raising issue JBTM-2767 and Mike's fix does not harm anything
@mmusgrov yes, I'm for pushing this fix.

I do understand your points just I would like to elaborate a bit more. We (QE) uses some MBeans to check state of TM but this issue and issues touching the behavior of object store are bound on our testing of JBoss CLI. By my point of view its entry point for any user to check state of transactions in app server. That should be consistent (and logical) between EAP/WFLY releases.

I created a new issue that points to the fact that {{recover}} does not work and it's why the test {{JcaInflowTransactionTestCase.rmerrWithRecovery}} fails even with Mike's fix.
https://issues.jboss.org/browse/JBTM-2819

mmusgrov:
Thanks Ondrej for you retest and the rationale. I will take a look at the new JIRA today.

ochaloup:
Thank you!

PR #1114, opened 2016-12-23
[JBTM-2829] Avoid NPE when comparing JDBC connections

gytis:
https://issues.jboss.org/browse/JBTM-2829
!XTS !BLACKTIE !QA_JTA !QA_JTS_JDKORB !PERF


gytis:
Done. I've tweaked my IDE settings last week and enabled auto optimisation of imports. It's off now.

gytis:
I'm thinking of the situation where two different datasources with empty username and password would used by Spring Boot. Since it doesn't provide URLs, that condition would pass for both of the datasources and the first created connection would be reused.

gytis:
Here's the test: https://github.com/jbosstm/narayana/pull/1114/files#diff-1b51a1071b8326189ddd521653da489b

PR #1117, opened 2017-01-11
[JBTM-2835] Avoid NPE when CMR cannoct connect DB

dsimko:
https://issues.jboss.org/browse/JBTM-2835

!BLACKTIE !XTS !PERF

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/objectstore/StoreManager.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/objectstore/StoreManager.java
@@ -145,20 +145,20 @@ public class StoreManager
     {
         ObjectStoreEnvironmentBean storeEnvBean = BeanPopulator.getNamedInstance(ObjectStoreEnvironmentBean.class, name);
         String storeType = storeEnvBean.getObjectStoreType();
-        ObjectStoreAPI store;
 
         try
         {
-            store = ClassloadingUtility.loadAndInstantiateClass(ObjectStoreAPI.class, storeType, name);
+            ObjectStoreAPI store = ClassloadingUtility.loadAndInstantiateClass(ObjectStoreAPI.class, storeType, name);
+
+            store.start();
+
+            return store;
+
         }
         catch (final Throwable ex)
         {
             throw new FatalError(tsLogger.i18NLogger.get_StoreManager_invalidtype() + " " + storeType, ex);
         }
-
-        store.start();
-
-        return store;
     }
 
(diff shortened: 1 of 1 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/objectstore/StoreManager.java
```diff
@@ -145,20 +145,20 @@ private static final ObjectStoreAPI initStore(String name)
...
         ObjectStoreEnvironmentBean storeEnvBean = BeanPopulator.getNamedInstance(ObjectStoreEnvironmentBean.class, name);
         String storeType = storeEnvBean.getObjectStoreType();
-        ObjectStoreAPI store;
 
         try
         {
-            store = ClassloadingUtility.loadAndInstantiateClass(ObjectStoreAPI.class, storeType, name);
+            ObjectStoreAPI store = ClassloadingUtility.loadAndInstantiateClass(ObjectStoreAPI.class, storeType, name);
+
+            store.start();
```
mmusgrov: store can still be null so I think you will have the same issue.

Why is the load failing? It sounds like it ought to be a fatal error if the load and instantiate fails.

dsimko: Thank you for review! Yes, I still get NPE but it is caught earlier and user gets better info about what happened - "ARJUNA012135: Could not create Store type: com.arjuna.ats.internal.arjuna.objectstore.jdbc.JDBCStore". You can look at log files before and after patch which I have uploaded to the JBTM-2835.

mmusgrov: So the PR has not introduced any behavioural change. I therefore reject it unless you can convince me that the change adds benefit.

tomjenkinson: As unlikely as it might be, people might be expecting the NPE now so it could constitute a public API change which is not a good compromise if its just for aesthetics.

dsimko: Yes I agree with you. Thank you for explanation. I am going to close this PR.

mmusgrov: If you really do need better info you could have logged a warning instead

PR #1118, opened 2017-01-11
[JBTM-2836] Adding support for suppressed exceptions in prepare for heuristic

ochaloup:
!BLACKTIE !PERF !QA_JTA !QA_JTS_JDKORB

This is a follow-up enhancement to enhance exception with suppressed ones in case that `prepare` of `XAResource` ends in heuristic state.

https://issues.jboss.org/browse/JBTM-2822
https://issues.jboss.org/browse/JBTM-2836

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
@@ -2724,11 +2724,12 @@ public class BasicAction extends StateManager
                         }
                     }
 
+                    addDeferredThrowables(record, deferredThrowables);
+
                     /*
                               * Prepare on this record failed - we are in trouble.
                               * Add the record back onto the pendingList and return.
                               */
-                    addDeferredThrowables(record, deferredThrowables);
 
                     record = insertRecord(pendingList, record);
 
@@ -2751,6 +2752,8 @@ public class BasicAction extends StateManager
 
                     if (reportHeuristics)
                         updateHeuristic(p, false);
+
+                    addDeferredThrowables(record, deferredThrowables);
 
                     /*
                     * Don't add to the prepared list. We process heuristics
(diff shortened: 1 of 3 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
```diff
@@ -2725,9 +2725,10 @@ protected int doPrepare (boolean reportHeuristics, AbstractRecord record)
...
 
                     /*
-                              * Prepare on this record failed - we are in trouble.
-                              * Add the record back onto the pendingList and return.
-                              */
+                     * Prepare on this record failed - we are in trouble.
+                     * Add the record back onto the pendingList and return.
+                     */
+
                     addDeferredThrowables(record, deferredThrowables);
```
mmusgrov: The comment just above is now a little unclear . Move this call to before the comment.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
```diff
@@ -2761,6 +2762,8 @@ protected int doPrepare (boolean reportHeuristics, AbstractRecord record)
                     * appear on multiple lists at the same time.
                     */
 
+                    addDeferredThrowables(record, deferredThrowables);
```
mmusgrov: The comment just above is now a little unclear. Move this call to before the comment.

ochaloup:
Hi Mike,

thank you for the review. I tried to address all your remarks and pushed changes here.

ochaloup:
To your general question. I'm not able to cover whole context I think this could be more question on @tomjenkinson and I didn't do any investigation on it. Within fast check in code I'm not sure where that top level could be situated. I tend to think that this is place where information about failure is included.

ochaloup:
@tomjenkinson , @mmusgrov because of spending some time on customer case 01596228 I did personal investigation, my two cents...

...there isn't other possibility where to put calls for `addDeferredThrowables`. This call gather exceptions from `XAResourceRecord` and put them to `BasicAction.deferredThrowables`. For this could be done the code needs to have access to record which is handled. This record is normally available on places where calls of `addDeferredThrowables` happens now. There are places in `BasicAction` where the record is set to null just few lines beneath.

When thinking about this maybe there could be approach that at time when `TransactionImple` ask for deferred exceptions at method `getDeferredThrowables` - the method could be reimplemented to go through all records which are part of transaction and gathers deferred exceptions at that time.

mmusgrov:
No further insights ;)

PR #1119, opened 2017-01-17
JBTM-2837 always run periodic recovery phase 2 if phase 1 has ran

mmusgrov:
https://issues.jboss.org/browse/JBTM-2837

!BLACKTIE !PERF !QA_JTA !QA_JTS_JDKORB !XTS

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
@@ -136,7 +136,11 @@ public class XARecoveryModule implements RecoveryModule
 		return _seriablizableXAResourceDeserializers;
 	}
 
-	public synchronized void periodicWorkFirstPass()
+	public synchronized void periodicWorkFirstPass() {
+		periodicWorkFirstPass(ScanStates.BETWEEN_PASSES);
+	}
+
+	private synchronized void periodicWorkFirstPass(ScanStates endState)
 	{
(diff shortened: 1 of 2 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/PeriodicRecovery.java
```diff
@@ -783,15 +783,9 @@ private void doWorkInternal()
             // we carry on scanning even if scanning is SUSPENDED because the suspending thread
             // might be waiting on us to complete and we don't want to risk deadlocking it by waiting
             // here for a resume.
-            // if we have been TERMINATED we bail out now
+            // if we have been TERMINATED we still run phase 2 since some modules (XARecoveryModule for example)
+            // require phase 2 if phase 1 has ran
             // n.b. if we give up here the caller is responsible for clearing the active scan
```
tomjenkinson: i think this line could be removed - did you check the history to know about whether there was a need to bail out?

tomjenkinson: Should we avoid doing the backoff?

mmusgrov:
@tomjenkinson @ochaloup  has pointed out that the problem is an "out-of-band" call to XARecoveryModule#periodicWorkFirstPass via an XAResourceRecord.getNewXAResource() call from the tooling. That call leaves the scanState in ScanStates.BETWEEN_PASSES so when the shutdown thread calls XARecoveryModule#removeXAResourceRecoveryHelper it hangs waiting for the scanState to move to IDLE.

ochaloup:
As an info - I can confirm that my integration tests passed with this new Mike's fix.

PR #1121, opened 2017-01-19
JBTM-2834 always call forget

mmusgrov:
https://issues.jboss.org/browse/JBTM-2834

NO_TEST because the last (rebased) commit only added an explanatory comment

### DIFF
--- a/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
+++ b/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
@@ -916,13 +916,20 @@ public class XAResourceRecord extends com.arjuna.ArjunaOTS.OTSAbstractRecordPOA
 		        " _forgotten=" + _forgotten);
 		}
 
-                if (!_forgotten) {
-                    handleForget();
+		// if we have not yet seen a successful forget call then tell the resource to forget
+		if (!_forgotten)
+		    handleForget(); // this call can fail silently which will leave _forgotten as false
 
-                    destroyState();
+		// if it is known that the resource has forgotten the heuristic then delete our record of it
+		if (_forgotten) {
+		    destroyState();
 
-                    removeConnection();
-                }
+		    removeConnection();
+		}
+	}
+
+	public boolean isForgotten() {
+		return _forgotten;
 	}
 
 	private void handleForget() throws org.omg.CORBA.SystemException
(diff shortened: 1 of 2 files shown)


mmusgrov:
Well as far as I am concerned the fix is good, however the QE test is failing. The test

1. - adds a test resource and a message queue resource;
2. - both generate heuristics (using byteman)
3. - the test uses the tooling to remove both resources
4. - the tooling calls forget on both. The test resource returns ok  but ActiveMQRAXAResource throws XA_RETRY and therefore the tooling does *not* remove the log

But the test is expecting both logs to be removed. So with my fix, the issue is now with the QE testsuite.

As an aside, notice also that valid exception values from forget are XAER_RMERR, XAER_RMFAIL, XAER_NOTA, XAER_INVAL, or XAER_PROTO so ActiveMQRAXAResource should not be using XA_RETRY to report problems. I will ask Daniel Simko to investigate how the messaging resource is configured.

mmusgrov:
I did another push that should fix the build failure. Also we have lots of unit tests for this functionality already (and it was 3 of those that failed the last build).

I believe the regression is because in DR9 on some code paths we were ignoring the return value from forget and have tightened up on that.

mmusgrov:
Thanks Daniel for the test update to replace ActiveMQRAXAResource with a second instance of TestXAResourceCommon. The QE test now passes so once our own tests have passed this PR should be good.

mmusgrov:
@tomjenkinson This needs reviewing prior to the release

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -916,15 +916,20 @@ public void forget() throws org.omg.CORBA.SystemException
 		        " _forgotten=" + _forgotten);
 		}
 
-                if (!_forgotten) {
+                if (!_forgotten)
                     handleForget();
 
+                if (_forgotten) {
```
tomjenkinson: (16:58:37) mmusgrov: if _theXAResource is null handleForget just returns and then we used to go ahead anyway and delete the log which is wrong
(17:02:24) mmusgrov: it starts out false and then handleForget() is called. If _theXAResource is null it stays false
(17:03:18) mmusgrov: I can change the signature of private void handleForget() to return a boolean which will make the code easier to parse.

tomjenkinson: OK, that makes sense now thanks

PR #1122, opened 2017-01-25
JBTM-2843 Updated the version of the build-tools

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2843

!MAIN !QA_JTA !QA_JTS_JDKORB !PERF !XTS

### DIFF
--- a/blacktie/pom.xml
+++ b/blacktie/pom.xml
@@ -177,7 +177,7 @@
             <dependency>
                 <groupId>org.jboss.narayana.jts</groupId>
                 <artifactId>narayana-jts-jacorb</artifactId>
-                <version>${project.version}</version>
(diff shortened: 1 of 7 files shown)


blacktie/pom.xml
```diff
@@ -177,7 +177,7 @@
             <dependency>
                 <groupId>org.jboss.narayana.jts</groupId>
                 <artifactId>narayana-jts-jacorb</artifactId>
-                <version>${project.version}</version>
+                <version>5.3.3.Final</version>
```
tomjenkinson: It's hard coded to the version in WFLY 10.1.0 so we don't lose track of the dependency. Otherwise we could build with APIs that won't be there at runtime for BT

PR #1123, opened 2017-01-27
[JBTM-2844] Move test service client instantiation to the method

gytis:
https://issues.jboss.org/browse/JBTM-2844
!BLACKTIE !QA_JTA !QA_JTS_JDKORB !PERF


mmusgrov:
The change looks OK though there is nothing in the PR or JIRA to indicate why the fix is effective, can you had a brief note.

Also there is a test hang on the job but it is unrelated to your change (org.jboss.narayana.rest.bridge.inbound.test.integration.InboundBridgeRecoveryTestCase)

gytis:
It must be something inside JDK9. Maybe FileSystemNotFoundException was thrown because something was not initialised when instance was created early. Exception disappeared once I moved object creation to the test method.

mmusgrov:
OK thanks

PR #1133, opened 2017-02-23
[JBTM-2852][JBTM-2853] integration with wildfly transaction client

ochaloup:
Changes to integrate with widlfy transaction client. The glue is done through interfaces from SPI and used here by implementation being set in JTAEnvironmentalBean and in SPI (still not sure if all the settings should not be only maintained in jta and not in SPI. maybe it would be more consistent).

https://issues.jboss.org/browse/JBTM-2852
https://issues.jboss.org/browse/WFTC-13
https://issues.jboss.org/browse/WFTC-15

!BLACKTIE !QA_JTA !QA_JTS_JDKORB !PERF

requires: https://github.com/jbosstm/jboss-transaction-spi/pull/19
requires: https://github.com/jbosstm/jboss-as/pull/65

### DIFF
--- a/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
+++ b/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
@@ -26,8 +26,6 @@ package com.arjuna.ats.jta.cdi.transactional;
 import com.arjuna.ats.jta.common.jtaPropertyManager;
 import com.arjuna.ats.jta.logging.jtaLogger;
 
-import org.jboss.tm.usertx.client.ServerVMClientUserTransaction;
-
 import javax.enterprise.context.ContextNotActiveException;
 import javax.enterprise.inject.Intercepted;
 import javax.enterprise.inject.spi.Bean;
@@ -39,6 +37,8 @@ import javax.transaction.Status;
 import javax.transaction.Transaction;
 import javax.transaction.TransactionManager;
 import javax.transaction.Transactional;
+
+import org.jboss.tm.usertx.UserTransactionOperationsProvider;
 
 import java.io.Serializable;
 import java.lang.annotation.Annotation;
@@ -178,14 +178,16 @@ public abstract class TransactionalInterceptorBase implements Serializable {
 
     protected boolean setUserTransactionAvailable(boolean available) {
 
-        boolean previousUserTransactionAvailability = ServerVMClientUserTransaction.isAvailable();
-        ServerVMClientUserTransaction.setAvailability(available);
+        UserTransactionOperationsProvider userTransactionProvider =
+            jtaPropertyManager.getJTAEnvironmentBean().getUserTransactionOperationsProvider();
+        boolean previousUserTransactionAvailability = userTransactionProvider.getAvailability();
+        userTransactionProvider.setAvailability(available);
         return previousUserTransactionAvailability;
     }
 
     protected void resetUserTransactionAvailability(boolean previousUserTransactionAvailability) {
-
-        ServerVMClientUserTransaction.setAvailability(previousUserTransactionAvailability);
+        jtaPropertyManager.getJTAEnvironmentBean().getUserTransactionOperationsProvider()
+            .setAvailability(previousUserTransactionAvailability);
     }
 
     protected TransactionManager getTransactionManager() {
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
@@ -29,7 +29,12 @@ import javax.transaction.TransactionManager;
 import javax.transaction.TransactionSynchronizationRegistry;
 import javax.transaction.UserTransaction;
 
+import org.jboss.tm.SubordinateTransactionImporter;
+import org.jboss.tm.usertx.UserTransactionOperationsProvider;
+import org.jboss.tm.usertx.client.ServerVMClientUserTransactionOperationsProvider;
+
 import com.arjuna.ats.internal.jta.resources.arjunacore.XAResourceRecordWrappingPlugin;
+import com.arjuna.ats.internal.jta.transaction.arjunacore.jca.NarayanaSubordinateTransactionImporter;
 import com.arjuna.ats.jta.recovery.XAResourceOrphanFilter;
 import com.arjuna.ats.jta.recovery.XAResourceRecovery;
 import com.arjuna.ats.jta.resources.XAResourceMap;
@@ -109,6 +114,14 @@ public class JTAEnvironmentBean implements JTAEnvironmentBeanMBean
 	private Map<String, Boolean> performImmediateCleanupOfCommitMarkableResourceBranchesMap = new HashMap<String, Boolean>();
 
 	private Map<String, Integer> commitMarkableResourceRecordDeleteBatchSizeMap = new HashMap<String, Integer>();
+
+	private static final String defaultSubordinateTransactionImporterClassName = NarayanaSubordinateTransactionImporter.class.getName();
+	private volatile String subordinateTransactionImporterClassName = defaultSubordinateTransactionImporterClassName;
+	private volatile SubordinateTransactionImporter subordinateTransactionImporter;
+
+	private static final String defaultTransactionOperationsProviderClassName = ServerVMClientUserTransactionOperationsProvider.class.getName();
+	private volatile String userTransactionOperationsProviderClassName = defaultTransactionOperationsProviderClassName;
+	private volatile UserTransactionOperationsProvider userTransactionOperationsProvider = null;
 
 	/**
      * Returns true if subtransactions are allowed.
@@ -1225,4 +1238,101 @@ public class JTAEnvironmentBean implements JTAEnvironmentBeanMBean
 			boolean notifyCommitMarkableResourceRecoveryModuleOfCompleteBranches) {
 		this.notifyCommitMarkableResourceRecoveryModuleOfCompleteBranches = notifyCommitMarkableResourceRecoveryModuleOfCompleteBranches;
 	}
+
+	/**
+	 * <p>
+     * Setting name of class which is taken to be used as transaction importer
+     * when a {@link javax.transaction.xa.Xid} needs to be imported as a transaction to currently running TM.<br>
+     * The classname which is set here has to implement {@link org.jboss.tm.SubordinateTransactionImporter} interface.
+     * <p>
+     * When null is set then default provider implementation is used which is
+     * {@code #defaultSubordinateTransactionImporterClassName}.
+     */
+    public void setSubordinateTransactionImporterClassName(String importerClassName) {
+        synchronized(this) {
+            if (!subordinateTransactionImporterClassName.equals(importerClassName)) {
+                subordinateTransactionImporter = null;
+            }
+
+            if(importerClassName == null) {
+                subordinateTransactionImporterClassName = defaultSubordinateTransactionImporterClassName;
+            } else {
+                subordinateTransactionImporterClassName = importerClassName;
+            }
+        }
+    }
+
+    public String getSubordinateTransactionImporterClassName(){
+        return this.subordinateTransactionImporterClassName;
+    }
+
+    /**
+     * Importing subordinate transaction.
+     */
+    public SubordinateTransactionImporter getSubordinateTransactionImporter() {
+        if(subordinateTransactionImporter == null && subordinateTransactionImporterClassName != null) {
+            synchronized(this) {
+                if(subordinateTransactionImporter == null && subordinateTransactionImporterClassName != null) {
+                    subordinateTransactionImporter = ClassloadingUtility.loadAndInstantiateClass(SubordinateTransactionImporter.class,
+                        subordinateTransactionImporterClassName, null);
+                }
+            }
+        }
+
+        return subordinateTransactionImporter;
+    }
+
+    /**
+     * <p>
+     * Setting of class name that defines {@link UserTransactionOperationsProvider}.
+     * The provider is later used to get more info about {@link UserTransaction}.
+     * <p>
+     * When null is set then default provider implementation is used which is
+     * {@code #defaultTransactionOperationsProviderClassName}.
+     *
+     * @param providerClassName  class name implementing {@link UserTransactionOperationsProvider}
+     */
+    public void setUserTransactionOperationsProviderClassName(String providerClassName) {
+        synchronized (this) {
+            if (!userTransactionOperationsProviderClassName.equals(providerClassName)) {
+                userTransactionOperationsProvider = null;
+            }
+
+            if(providerClassName == null) {
+                userTransactionOperationsProviderClassName = defaultTransactionOperationsProviderClassName;
+            } else {
+                userTransactionOperationsProviderClassName = providerClassName;
+            }
+        }
+    }
+
+    /**
+     * Get class name that is used as {@link UserTransactionOperationsProvider}
+     *
+     * @return class name implementing {@link UserTransactionOperationsProvider}
+     */
+    public String getUserTransactionOperationsProviderClassName() {
+        return this.userTransactionOperationsProviderClassName;
+    }
+
+    /**
+     * Returning singleton instance of {@link UserTransactionOperationsProvider} instantiated based
+     * based on name specified by {@link #setUserTransactionOperationsProviderClassName(String)}.<br>
+     * When class name is redefined during runtime there should be instantiated new provider.
+     *
+     * @return instance of class implementing {@link UserTransactionOperationsProvider}
+     */
+    public UserTransactionOperationsProvider getUserTransactionOperationsProvider() {
+
+        if(userTransactionOperationsProvider == null && userTransactionOperationsProviderClassName != null) {
+            synchronized(this) {
+                if(userTransactionOperationsProvider == null && userTransactionOperationsProviderClassName != null) {
+                    userTransactionOperationsProvider = ClassloadingUtility.loadAndInstantiateClass(UserTransactionOperationsProvider.class,
+                        userTransactionOperationsProviderClassName, null);
+                }
+            }
+        }
+
+        return userTransactionOperationsProvider;
+    }
 }
--- a/txbridge/src/main/java/org/jboss/jbossts/txbridge/inbound/InboundBridge.java
+++ b/txbridge/src/main/java/org/jboss/jbossts/txbridge/inbound/InboundBridge.java
@@ -24,13 +24,16 @@
 package org.jboss.jbossts.txbridge.inbound;
 
 import com.arjuna.ats.jta.TransactionManager;
-import com.arjuna.ats.internal.jta.transaction.arjunacore.jca.SubordinationManager;
+import com.arjuna.ats.jta.common.jtaPropertyManager;
+
 import org.jboss.jbossts.txbridge.utils.txbridgeLogger;
 
 import javax.transaction.xa.Xid;
 import javax.transaction.xa.XAException;
 import javax.transaction.Status;
 import javax.transaction.SystemException;
+import javax.naming.InitialContext;
+import javax.naming.NamingException;
 import javax.transaction.InvalidTransactionException;
 import javax.transaction.Transaction;
 
@@ -47,6 +50,8 @@ public class InboundBridge
      */
     private final Xid xid;
 
+    private final InitialContext ctx;
+
     /**
      * Create a new InboundBridge to manage the given subordinate JTA transaction.
      *
@@ -61,6 +66,12 @@ public class InboundBridge
         this.xid = xid;
 
         getTransaction(); // ensures transaction is initialized
+
+        try {
+            ctx = new InitialContext();
+        } catch (NamingException ne) {
+            throw new IllegalStateException("Can't intialize intial context");
+        }
     }
 
     /**
@@ -77,7 +88,7 @@ public class InboundBridge
 
         Transaction tx = getTransaction();
 
-        TransactionManager.transactionManager().resume(tx);
+        TransactionManager.transactionManager(ctx).resume(tx);
     }
 
     /**
@@ -92,7 +103,7 @@ public class InboundBridge
     {
         txbridgeLogger.logger.trace("InboundBridge.stop("+xid+")");
 
-        TransactionManager.transactionManager().suspend();
+        TransactionManager.transactionManager(ctx).suspend();
     }
 
     public void setRollbackOnly() throws XAException, SystemException
@@ -112,7 +123,8 @@ public class InboundBridge
     private Transaction getTransaction()
             throws XAException, SystemException
     {
-        Transaction tx = SubordinationManager.getTransactionImporter().importTransaction(xid);
+        Transaction tx = jtaPropertyManager.getJTAEnvironmentBean()
+            .getSubordinateTransactionImporter().getTransaction(xid);
 
         switch (tx.getStatus())
         {
--- a/txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/OutboundBridgeManager.java
+++ b/txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/OutboundBridgeManager.java
@@ -24,10 +24,12 @@
 package org.jboss.jbossts.txbridge.outbound;
 
 import com.arjuna.ats.jta.TransactionManager;
-import com.arjuna.ats.jta.transaction.Transaction;
 import com.arjuna.ats.arjuna.common.Uid;
 
 import javax.transaction.SystemException;
+import javax.transaction.Transaction;
+import javax.naming.InitialContext;
+import javax.naming.NamingException;
 import javax.transaction.RollbackException;
 import javax.transaction.Synchronization;
 import javax.transaction.xa.XAResource;
@@ -52,7 +54,8 @@ public class OutboundBridgeManager
     public static String BRIDGEWRAPPER_PREFIX = "txbridge_";
 
     // maps JTA Tx Id to OutboundBridge instance.
-    private static final ConcurrentMap<Uid, org.jboss.jbossts.txbridge.outbound.OutboundBridge> outboundBridgeMappings = new ConcurrentHashMap<Uid, org.jboss.jbossts.txbridge.outbound.OutboundBridge>();
+    private static final ConcurrentMap<Transaction, Uid> transctionToUidMappings = new ConcurrentHashMap<>();
+    private static final ConcurrentMap<Uid, org.jboss.jbossts.txbridge.outbound.OutboundBridge> outboundBridgeMappings = new ConcurrentHashMap<>();
 
     /**
      * Return an OutboundBridge instance that maps the current Thread's JTA transaction context
@@ -66,18 +69,21 @@ public class OutboundBridgeManager
 
         try
         {
-            Transaction transaction = (Transaction)TransactionManager.transactionManager().getTransaction();
+            Transaction transaction = (Transaction)TransactionManager.transactionManager(new InitialContext()).getTransaction();
 
-            Uid externalTxId = transaction.get_uid();
+            // generating new Uid to get it connected to incoming transaction
+            // the incoming transaction could come from different TM than Narayana is (e.g. Wildfly Transaction client)
+            transctionToUidMappings.putIfAbsent(transaction, new Uid());
+            Uid outboundMappingUid = transctionToUidMappings.get(transaction);
 
-            if(!outboundBridgeMappings.containsKey(externalTxId)) {
-                createMapping(transaction, externalTxId);
+            if(!outboundBridgeMappings.containsKey(outboundMappingUid)) {
+                createMapping(transaction, outboundMappingUid);
             }
 
-            return outboundBridgeMappings.get(externalTxId);
+            return outboundBridgeMappings.get(outboundMappingUid);
 
         }
-        catch(SystemException e)
+        catch(SystemException | NamingException e)
         {
             txbridgeLogger.logger.error(e);
         }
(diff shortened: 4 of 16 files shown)


ochaloup:
Hi @mmusgrov  and @tomjenkinson 
I can say that my fixes work - for cdi and for inbound/outbound too. Besides I'm asking for review I have one special doubt in outbound connection fix. It's this place

https://github.com/jbosstm/narayana/compare/master...ochaloup:WFTC-13-check-txn-availability-with-jta-bean-settings?expand=1#diff-9555df5227a2c558d02a36fb37cdc1fbL69

I decided to generate new Uid as the incoming transaction does not provide such information (wildfly transaction client does not have such notion). From what I can see in the code it seems to me fine but I'm really not sure about good practices.

txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/OutboundBridgeManager.java
```diff
@@ -66,18 +68,20 @@
 
         try
         {
-            Transaction transaction = (Transaction)TransactionManager.transactionManager().getTransaction();
+            Transaction transaction = (Transaction)TransactionManager.transactionManager(new InitialContext()).getTransaction();
 
-            Uid externalTxId = transaction.get_uid();
+            // generating new Uid to get it connected to incoming transaction
+            // the incoming transaction could come from different TM than Narayana is (e.g. Wildfly Transaction client)
+            Uid outboundMappingUid = new Uid();
```
tomjenkinson: I believe this is wrong because the txId is important for ensuring we only outbound bridge once. By using a new uid each time it will not find it in the map in outboundBridgeMappings.

You can get the Uid from even a WFTC tx client by doing the following:
Uid externalTxId = new Uid(((org.wildfly.transaction.client.LocalTransaction)transaction).getXid().getGlobalTransactionId())

The reason I spotted this was because I noticed you had renamed all the variable names, I think we should revert the variable name changes.

Also, we should check what are the circumstances you might get a org.wildfly.transaction.client.RemoteTransaction because that seems to have no obvious way to get the Xid however it doesn't allow enlisting resources either so I expect it is rare to get and should not happen in this code path.

ochaloup: I see. This was my main doubt around all the handling. I didn't realize the fact of importing txn several times.

I didn't do it in the way you provide previously as I wanted to stay without dependency to wfly transaction client in Narayana. But this points and discussion with David I expect it's expected. I'm going to change it.

ochaloup:
Hi @tomjenkinson , may I ask you for a help please? Even if trying I'm not able to understand what's behind failures.
http://narayanaci1.eng.hst.ams2.redhat.com/job/btny-pulls-narayana/279/PROFILE=MAIN,jdk=jdk8.latest,label=linux/console

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:2.10.3:javadoc (javadoc) on project narayana-full: An error has occurred in JavaDocs report generation:
[ERROR] Exit code: 1 - /home/jenkins/workspace/btny-pulls-narayana/PROFILE/MAIN/jdk/jdk8.latest/label/linux/ArjunaCore/arjuna/services/classes/com/arjuna/ats/arjuna/services/recovery/RecoveryManagerService.java:23: error: package org.tanukisoftware.wrapper does not exist
[ERROR] import org.tanukisoftware.wrapper.WrapperListener;
```

txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/OutboundBridgeManager.java
ochaloup: @tomjenkinson after tries to bound uid from client as you suggests I'm again rethinking the approach. When checking all dependencies in project I think it's not a good idea to depend on wildfly transaction client in Narayana directly. XTS should be 

What about the approach like this?
```
// maps JTA Tx Id to OutboundBridge instance.
private static final ConcurrentMap<Transaction, Uid> transctionToUidMappings = new ConcurrentHashMap<>();
private static final ConcurrentMap<Uid, org.jboss.jbossts.txbridge.outbound.OutboundBridge> outboundBridgeMappings = new ConcurrentHashMap<>();

...
   
transctionToUidMappings.putIfAbsent(transaction, new Uid());
Uid outboundMappingUid = transctionToUidMappings.get(transaction);

if(!outboundBridgeMappings.containsKey(outboundMappingUid)) {
    createMapping(transaction, outboundMappingUid);
}
```

tomjenkinson: How about:
private static final ConcurrentMap<Transaction, org.jboss.jbossts.txbridge.outbound.OutboundBridge> transctionToUidMappings = new ConcurrentHashMap<>();

That was not guaranteed to work with JTA < 1.2 but as per https://issues.jboss.org/browse/JBTM-2848 it should be safe now

ochaloup: I see. In the fact my version needs equal method to work too.

This was my fist try to join those approches I had got to trouble with saving transaction to log store. From my understanding I need `Uid` there to be safe.

Outbound bridge initializes the XAResource
https://github.com/jbosstm/narayana/blob/master/txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/BridgeXAResource.java#L73
and it needs to be saved for recovery purpose
https://github.com/jbosstm/narayana/blob/master/txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/BridgeXAResource.java#L93

txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/OutboundBridgeManager.java
```diff
@@ -66,18 +69,22 @@
...
         try
         {
-            Transaction transaction = (Transaction)TransactionManager.transactionManager().getTransaction();
+            Transaction transaction = (Transaction)TransactionManager.transactionManager(new InitialContext()).getTransaction();
 
-            Uid externalTxId = transaction.get_uid();
+            // generating new Uid to get it connected to incoming transaction
+            // the incoming transaction could come from different TM than Narayana is (e.g. Wildfly Transaction client)
+            transctionToUidMappings.putIfAbsent(transaction, new Uid());
+            txbridgeLogger.logger.error("ochaloup: new version " + transaction);
```
tomjenkinson: You can simplify this to:
Uid externalTxId = transctionToUidMappings.putIfAbsent(transaction, new Uid());

(and remove the ochaloup log message :)

ochaloup: ah, right, correct. thanks!

txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/OutboundBridgeManager.java
```diff
@@ -88,49 +93,51 @@
     /**
      * Remove the mapping for the given externalTxId. This should be called for gc when the tx is finished.
      *
-     * @param externalTxId The JTA transaction identifier.
+     * @param outboundMappingUid identifier bound to the JTA transaction
      */
-    public static synchronized void removeMapping(Uid externalTxId)
```
tomjenkinson: As this is a public interface we can't remove the public access to this method unless it is a serious defect

txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/OutboundBridgeManager.java
```diff
@@ -66,18 +69,20 @@
 
         try
         {
-            Transaction transaction = (Transaction)TransactionManager.transactionManager().getTransaction();
+            Transaction transaction = (Transaction)TransactionManager.transactionManager(new InitialContext()).getTransaction();
 
-            Uid externalTxId = transaction.get_uid();
+            // generating new Uid to get it connected to incoming transaction
+            // the incoming transaction could come from different TM than Narayana is (e.g. Wildfly Transaction client)
+            Uid outboundMappingUid = transctionToUidMappings.putIfAbsent(transaction, new Uid());
```
mmusgrov: I am not familiar with the outbound bridge so could you please explain why are you using new Uid() instead of transaction.get_uid(). Based on your answer I will attempt to address the "potential synchronization issue" you mentioned in the jbossts chat room.

ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
```diff
@@ -178,14 +178,16 @@ protected void endTransaction(TransactionManager tm, Transaction tx) throws Exce
 
     protected boolean setUserTransactionAvailable(boolean available) {
 
-        boolean previousUserTransactionAvailability = ServerVMClientUserTransaction.isAvailable();
-        ServerVMClientUserTransaction.setAvailability(available);
+        UserTransactionOperationsProvider userTransactionProvider =
+            jtaPropertyManager.getJTAEnvironmentBean().getUserTransactionOperationsProvider();
+        boolean previousUserTransactionAvailability = userTransactionProvider.getAvailability();
+        userTransactionProvider.setAvailability(available);
```
ochaloup: yes, WTC does not return the previous value

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
```diff
@@ -1225,4 +1236,127 @@ public void setNotifyCommitMarkableResourceRecoveryModuleOfCompleteBranches(
...
+    public void setSubordinateTransactionImporterClassName(String importerClassName) {
+        synchronized(this)
+        {
+            if(importerClassName == null)
+            {
+                this.subordinateTransactionImporter = null;
+            }
+            else if(!importerClassName.equals(this.subordinateTransactionImporterClassName))
+            {
+                this.subordinateTransactionImporter = null;
```
ochaloup: it needs to be nulled as when it's put to null the new call of getter (lazy loading getter) initializes new implementation

I don't understand your second point
```
if (!subordinateTransactionImporterClassName.equals(importerClassName))
```
you mean to change the codesttyle? you mean does not use { } and change the order?

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
```diff
@@ -1225,4 +1236,127 @@ public void setNotifyCommitMarkableResourceRecoveryModuleOfCompleteBranches(
...
+            }
+            this.subordinateTransactionImporterClassName = importerClassName;
+        }
+    }
+
+    public String getSubordinateTransactionImporterClassName(){
+        return this.subordinateTransactionImporterClassName;
+    }
+
+    public void setSubordinateTransactionImporter(SubordinateTransactionImporter importer) {
```
ochaloup: you were absolutely right, I was wrong. the setter is not used and it's not necessary what  I can say. I am going to remove it

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
```diff
@@ -1225,4 +1236,127 @@ public void setNotifyCommitMarkableResourceRecoveryModuleOfCompleteBranches(
...
+        synchronized(this) {
+            SubordinateTransactionImporter oldImporter = this.subordinateTransactionImporter;
+            this.subordinateTransactionImporter = importer;
+
+            if(importer == null)
+            {
+                this.subordinateTransactionImporterClassName = null;
+            }
+            else if(importer != oldImporter)
```
ochaloup: AtomicReference is fine when you have eager loading. I think here is not a good place for it. But I can't say that there could not be a nicer solution, just I cannot see it right now.

And secondly I really don't want to change the way how code in the class is structuralized at one place. If there is better solution I think it should be applied for all settings not only for the new one.

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
```diff
@@ -1225,4 +1236,127 @@ public void setNotifyCommitMarkableResourceRecoveryModuleOfCompleteBranches(
...
+            {
+                this.lastResourceOptimisationInterfaceClassName = ClassloadingUtility.getNameForClass(importer);
+            }
+            this.subordinateTransactionImporter = importer;
```
ochaloup: correct, I'm dumb. just I'm going to remove the whole method as mentioned above

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
```diff
@@ -1225,4 +1236,127 @@ public void setNotifyCommitMarkableResourceRecoveryModuleOfCompleteBranches(
...
+        }
+    }
+    
+    /**
+     * Importing subordinate transaction.
+     */
+    public SubordinateTransactionImporter getSubordinateTransactionImporter() {
+        if(subordinateTransactionImporter == null && subordinateTransactionImporterClassName != null) {
```
ochaloup: I don't think so, already mentioning above

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
```diff
@@ -1225,4 +1236,127 @@ public void setNotifyCommitMarkableResourceRecoveryModuleOfCompleteBranches(
...
+            if(providerClassName == null)
+            {
+                this.userTransactionOperationsProvider = null;
+            }
+            else if(!providerClassName.equals(this.userTransactionOperationsProviderClassName))
+            {
+                this.userTransactionOperationsProvider = null;
+            }
+            this.userTransactionOperationsProviderClassName = providerClassName;
+        }
```
ochaloup: I don't think so. what happens if `userTransactionOperationsProviderClassName` is null?
or maybe I don't understand your point

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
```diff
@@ -1225,4 +1236,127 @@ public void setNotifyCommitMarkableResourceRecoveryModuleOfCompleteBranches(
...
+    /**
+     * Returning singleton instance of {@link UserTransactionOperationsProvider} instantiated based
+     * based on name specified by {@link #setUserTransactionOperationsProviderClassName(String)}.<br>
+     * When class name is redefined during runtime there should be instantiated new provider.
+     *
+     * @return instance of class implementing {@link UserTransactionOperationsProvider}
+     */
+    public UserTransactionOperationsProvider getUserTransactionOperationsProvider() {
+
+        if(userTransactionOperationsProvider == null && userTransactionOperationsProviderClassName != null) {
```
ochaloup: I was thinking that generic routine could be nice, just I wasn't brave enough to start refactoring this class. I can do so in case.

txbridge/src/main/java/org/jboss/jbossts/txbridge/inbound/InboundBridge.java
```diff
@@ -70,14 +73,16 @@
      * @throws XAException
      * @throws SystemException
      * @throws InvalidTransactionException
+     * @throws NamingException
+     * @throws IllegalStateException
      */
-    public void start() throws XAException, SystemException, InvalidTransactionException
+    public void start() throws XAException, SystemException, InvalidTransactionException, IllegalStateException, NamingException
```
ochaloup: yes. I've already noticed that. thanks a lot for pointing that out

txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/OutboundBridgeManager.java
ochaloup: true, there is a mistakei for `putIfAbsent` agree.

Transaction could be null and the nullpointer was thrown in prior implementation too. I don't plan to change the way how that was behaving.

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
tomjenkinson: From my research Ondra is using the same pattern as on the other setter/getters.

txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/OutboundBridgeManager.java
ochaloup: I see. I found out that it should not be a problem. I didn't notice before that remove method is synchronized on class. So there is already synchronization and my doubt was already covered in the original implementation.

The reason why I need Uid is that I don't want to depend on wfy transaction client. I need to work only with javax.transaction.Transaction and not with particular impl which gives me Uid directly. Thus I map Transaction to Uid for purpose of bridge.

tomjenkinson: The previous code didn't check for null - so maybe Ondra could take an action to investigate but I don't think Ondra is adding a new problem by not checking for null

ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
mmusgrov: OK

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
mmusgrov: Why do we need to lazy load the getter, if someone calls the setter there is good chance someone later on will call the getter. These setters and getters are too complicated and I would rather sacrifice a tiny perceived performance gain over code clarity.

mmusgrov: My second point is that
```java
if (!subordinateTransactionImporterClassName.equals(importerClassName))
    subordinateTransactionImporter = null;
```
is simpler and semantically identical to

```java
          if(importerClassName == null)
          {
                 this.subordinateTransactionImporter = null;
           }
           else if(!importerClassName.equals(this.subordinateTransactionImporterClassName))
           {
                 this.subordinateTransactionImporter = null;
           }
```

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
mmusgrov: Thanks

txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/OutboundBridgeManager.java
ochaloup: yes, thank you, my wrong here

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
tomjenkinson: I think the reason is to be consistent with the rest of the class....

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
mmusgrov: What about my first question: "Did you intend using identity as opposed importer.equals(oldImporter)?"

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
tomjenkinson: I think the reason is to be consistent with the rest of the class. I saw these kind of comments myself. Compare what Ondra is adding to (for example) the get|setTransactionManager[ClassName] and see that they are similar.

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
mmusgrov: ok

ochaloup:
@mmusgrov @tomjenkinson  hi there 

I tried to address all remarks and push my changes.

There is one trouble currently that I miss fact that rest-at tests are failing. I need to investigate what's happening.

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
mmusgrov: Justifying copying difficult to understand code on the basis that it how it is done elsewhere is not a good reason.

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
mmusgrov: It is initialized as non null. Why are you allowing callers to set it to null?

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
tomjenkinson: We do have precedent for respecting precedent in files. For example the codestyle directive we have in place.

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
ochaloup: Again I took it as a pattern used whole over the class. For this could be fixed at your way I need to change tests as they depend on current behavior - when null is set then is expected to get null. If you ask me why it's so I have no idea ;o)

ochaloup:
@tomjenkinson @mmusgrov 
to sum this up - RTS issue was solved by changing module dependencies and thank Tom's fix. Tests are (hopefully) passing

I think this PR is more or less ready t obe merged

personally, I do not plan to change the shape of methods in `JTAEnvironmentalBean` now. I will be happy to check it later. By your preference, I can create a Jira targeting refactoring effort of the class. I can take a look at it in the following weeks.

Then if this PR will be agreed I would like to know how to proceed further. SPI needs to be released, then Narayana has to be released and then PR for WFLY and EAP could be created. I will be at the computer tomorrow (on Saturday) in the evening again to address possible comments. In case you are fine with this PR in the current state, feel free to merge.

dmlloyd:
These changes seem like they might be a bit more than what is necessary for JBTM-2853: we can "join up" with an existing transaction easily enough as long as we can get a notification that such a transaction was created or associated with a thread by the provider, or even if we can predict when such associations are going to happen.

In this case the flow is, import to Narayana -> resume in Narayana -> notify a listener -> WFTC joins the transaction -> everything after proceeds normally.

dmlloyd:
Yeah I was just thinking we could save a lot of that % if you didn't need to use our import API.  Instead you could import as you do today, and just ensure that the corresponding method on LocalTransactionContext is called before anything used it.

mmusgrov:
@tomjenkinson That listener registry is for thread to transaction association changes only and the notifications are sent from TransactionManagerDelegate. If WFTC is using that then it will work.

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
mmusgrov: I think it is dangerous copying an existing pattern unless you understand why it was coded the way it was.

And if you start out with a non null default and then allow the user to set it null then you need update the javadoc to say that it is allowed and why it's a sensible value.

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
mmusgrov: OK double check locking idiom it is then (even though there are nicer alternatives).

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
mmusgrov: OK I will allow precedent to trump readability here.

mmusgrov:
OK I have approved the changes with the following caveats:
1) nowhere was I asking for a refactor of JTAEnviornmentBean, rather I was recommending that you don't propagate clunky code;
2) I would still like to see a javadoc explanation of why (in the relevant routines you added) it is okay to override the non zero property defaults with null values.

ochaloup:
@mmusgrov  ok, I see. My point on your notes
Personally I don't like inconsistency in doing things. When all methods in the bean works one way I think it's not a good to have one method works differently. I didn't write the original code but based on the tests those methods was designed in some way that I wanted to follow. I think there are other properties similar to the new ones that I would like to add where your objections count the same. That's why I was talking about refactoring.
As you are more experienced than me and you would think it should be done differently I will do so. This counts to having clunky code or setting null to method. All those rules are wired into tests currently. I can't tell you anything else than the presented way is done because of tests.
I think changing tests should be part of some investigation where those tests are necessary and where they aren't. I agree that rules of those tests are ok in some parts and in some parts they aren't. Disabling them globally would mean not covering some parts, I think.

@tomjenkinson @mmusgrov I'm really not sure what is expected for these fixes right now. Fixes provided by David are not final (at least it seems from testing) and new client will not be released today probably(?).
Maybe there could be merged and release only the part fixing trouble with CDI which is part of the tck tests and the part with importing transaction could be left for next eap release? From what I understand the David's feature fix could cover https://issues.jboss.org/browse/JBTM-2853 and https://issues.jboss.org/browse/JBEAP-8993, right?

ochaloup:
@dmlloyd I took the 18-SNAPSHOT of the client and run it with Narayana tests and I can see curently following exception

```
2017-02-26 09:03:22,932 WARN [org.jboss.narayana.rest.bridge.inbound.InboundBridge] (default task-3) WFTXN0002: Transaction is not a supported instance: TransactionImple < jca-subordinate, BasicAction: 0:ffff7f000001:-57d8db0:58b28bb2:12 status: ActionStatus.RUNNING >: javax.transaction.InvalidTransactionException: WFTXN0002: Transaction is not a supported instance: TransactionImple < jca-subordinate, BasicAction: 0:ffff7f000001:-57d8db0:58b28bb2:12 status: ActionStatus.RUNNING >
    at org.wildfly.transaction.client.ContextTransactionManager.resume(ContextTransactionManager.java:119)
    at org.jboss.narayana.rest.bridge.inbound.InboundBridge.start(InboundBridge.java:103)
    at org.jboss.narayana.rest.bridge.inbound.InboundBridgeFilter.startBridge(InboundBridgeFilter.java:84)
    at org.jboss.narayana.rest.bridge.inbound.InboundBridgeFilter.filter(InboundBridgeFilter.java:63)
    at org.jboss.resteasy.core.ResourceMethodInvoker.invokeOnTarget(ResourceMethodInvoker.java:263)
    at org.jboss.resteasy.core.ResourceMethodInvoker.invoke(ResourceMethodInvoker.java:249)
    at org.jboss.resteasy.core.ResourceMethodInvoker.invoke(ResourceMethodInvoker.java:236) 
```

mmusgrov:
@ochaloup Historically we have made our config properties available via a JMX mbean (JTAEnvironmentBeanMBean). I say historically because the WildFly app server no longer needs to expose its config via JMX. @tomjenkinson do you know if it is still a requirement to expose our config via JMX (I suspect not since there is no mention of them in the EAP docs)? I would say we do according to our own documentation:

> 2.1.4. Narayana Runtime Information
> 
> Compile-time configuration information is available via class com.arjuna.common.util.ConfigurationInfo . > Runtime configuration is embodied in the various name EnvironmentBean classes where name refers to > the particular configuration category (see the configuration section of the user guide). These beans > have corresponding MBean interfaces and may be linked to JMX for remote inspection of the configuration if desired.

mmusgrov:
@ochaloup the InboundBridge failure implies that WildFly is *not* calling your new setter method (for subordinateTransactionImporterClassName) since the bit of code in the stack implies that the InboundBridge#start method has called
```java
if (!isTransactionGoodToResume(transaction))
```
and got back true - ie its a valid transaction just not one that is wrapped by WFTC.

ochaloup:
@mmusgrov  sure, the new code is not called. But it's on purpose. As David commented the call stack would be 
```
Narayana -> resume in Narayana -> notify a listener -> WFTC joins the transaction -> everything after proceeds normally
```

that means is using the old call
`SubordinationManager.getTransactionImporter().importTransaction(xid);`

I think that code in WFTC does not count with different transaction type than the `org.wildfly.transaction.client.AbstractTransaction` and because there is listener bound the WFTC now gain the `TransactionImpl` from Narayana. My personal feeling is that the WFTC would not directly use the Narayana txn but there is need to do some wiring int JBoss Provider to merge worlds of Narayana and WFTC. But that's just my feeling and David will know much better.

mmusgrov:
@ochaloup I (and the javadoc) still need an explanation of why it is permissible to override a non null value with a null one. Note that you are assuming that it  [subordinateTransactionImporter] cannot be null in both bridges (ie there are no null checks).

ochaloup:
@mmusgrov yeap, I'm currently working on that. You are absolutely right with that. I'm going to investigate on the tests now.

ochaloup:
this was overtaken from PR #1137 and will be followed with new PR covering https://issues.jboss.org/browse/JBTM-2858 and https://issues.jboss.org/browse/JBTM-2853

PR #1134, opened 2017-02-23
JBTM-2846 Make actionStatus volatile

mmusgrov:
https://issues.jboss.org/browse/JBTM-2846

!BLACKTIE !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
@@ -259,8 +259,10 @@ public class BasicAction extends StateManager
                 * If we are active then change status. Otherwise it may be an error so check status.
                 */
 
-            if (actionStatus == ActionStatus.RUNNING)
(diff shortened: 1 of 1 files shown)


mmusgrov:
I did not include a test case since it is clear by reading the code that there is at least one path where reordering of writes to actionStatus can result in seeing the wrong status (as your test case would in fact prove).

mmusgrov:
@tomjenkinson the PERF profile did not run so that needs looking into (missing benchmark.jar). I cannot tell how long it has been broken for since for every PR I looked at (at least the last 10) people are setting !PERF lol (are they telling me/us something).

mmusgrov:
I updated the PR indicating why I think this fix is sufficient. If we foresee any other issues we can raise a separate JIRA for it.

dmlloyd:
@tomjenkinson Sorry, my Internet was out this morning after a storm.

Yes, the change is equivalent in terms of functionality, it just removes the race conditions by making read-and-assign be atomic as well as using volatile to prevent stale read situations.

PR #1138, opened 2017-02-27
JBTM-2853 + JBTM-2858: transaction importer

ochaloup:
This adds implematation of `org.jboss.tm.SubordinateTransactionImporter` from SPI. This PR then uses it in RTS and XTS bridge to import transaction from propriate (Subordinate)TM implementation.

https://issues.jboss.org/browse/JBTM-2858

!PERF !BLACKTIE !QA_JTA !QA_JTS_JDKORB !MAIN

requires: https://github.com/jbosstm/jboss-as/pull/68
requires: https://github.com/jbosstm/jboss-transaction-spi/pull/21

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
@@ -29,10 +29,12 @@ import javax.transaction.TransactionManager;
 import javax.transaction.TransactionSynchronizationRegistry;
 import javax.transaction.UserTransaction;
 
+import org.jboss.tm.SubordinateTransactionImporter;
 import org.jboss.tm.usertx.UserTransactionOperationsProvider;
 import org.jboss.tm.usertx.client.ServerVMClientUserTransactionOperationsProvider;
 
 import com.arjuna.ats.internal.jta.resources.arjunacore.XAResourceRecordWrappingPlugin;
+import com.arjuna.ats.internal.jta.transaction.arjunacore.jca.NarayanaSubordinateTransactionImporter;
 import com.arjuna.ats.jta.recovery.XAResourceOrphanFilter;
 import com.arjuna.ats.jta.recovery.XAResourceRecovery;
 import com.arjuna.ats.jta.resources.XAResourceMap;
@@ -113,7 +115,11 @@ public class JTAEnvironmentBean implements JTAEnvironmentBeanMBean
 
 	private Map<String, Integer> commitMarkableResourceRecordDeleteBatchSizeMap = new HashMap<String, Integer>();
 
-	private static final String defaultTransactionOperationsProviderClassName = ServerVMClientUserTransactionOperationsProvider.class.getName(); 
+	private static final String defaultSubordinateTransactionImporterClassName = NarayanaSubordinateTransactionImporter.class.getName();
+	private volatile String subordinateTransactionImporterClassName = defaultSubordinateTransactionImporterClassName;
+	private volatile SubordinateTransactionImporter subordinateTransactionImporter;
+
+	private static final String defaultTransactionOperationsProviderClassName = ServerVMClientUserTransactionOperationsProvider.class.getName();
 	private volatile String userTransactionOperationsProviderClassName = defaultTransactionOperationsProviderClassName;
 	private volatile UserTransactionOperationsProvider userTransactionOperationsProvider = null;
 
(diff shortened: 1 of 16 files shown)


ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
```diff
@@ -113,7 +115,11 @@
 
 	private Map<String, Integer> commitMarkableResourceRecordDeleteBatchSizeMap = new HashMap<String, Integer>();
 
-	private static final String defaultTransactionOperationsProviderClassName = ServerVMClientUserTransactionOperationsProvider.class.getName(); 
+	private static final String defaultSubordinateTransactionImporterClassName = NarayanaSubordinateTransactionImporter.class.getName();
```
mmusgrov: How do the changes to JTAEnvironmentBean.java relate to jbosstm/narayana#1137 - they look very similar.

txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/OutboundBridgeManager.java
```diff
@@ -66,18 +69,21 @@
 
         try
         {
-            Transaction transaction = (Transaction)TransactionManager.transactionManager().getTransaction();
+            Transaction transaction = (Transaction)TransactionManager.transactionManager(new InitialContext()).getTransaction();
 
-            Uid externalTxId = transaction.get_uid();
+            // generating new Uid to get it connected to incoming transaction
+            // the incoming transaction could come from different TM than Narayana is (e.g. Wildfly Transaction client)
+            transctionToUidMappings.putIfAbsent(transaction, new Uid());
```
mmusgrov: Why don't you use the hash code of the transaction since that does not change no matter how the transaction gets wrapped (see section 3.3.4 Transaction Equality and Hash Code of the JTA spec). That way you can avoid having to maintain these two maps (and any race conditions you were worried about)?

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
ochaloup: they relates indirectly. They use the same way how params are defined in the `JTAEnvironmentBean` but they are two different settings

`defaultTransactionOperationsProviderClassName` and `defaultSubordinateTransactionImporterClassName`

just I'm not sure if I respond to your question or not.

txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/OutboundBridgeManager.java
ochaloup: the trouble is that I'm not able to get `transaction.get_uid()` as there is used `javax.transaction.Transaction` interface which is returned from wfly transaction client. That way I can't take the uid from transaction directly. Because of it there is generated new `Uid` which is bound to incoming transaction where I rely on the transaction equality (as you mentioned) and generating new `Uid` which is needed for later recovery usage (Uid is saved to txn log).

tomjenkinson: I think you should just rely on the transaction.equals() method - would there be a problem with that?

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
tomjenkinson: I imagine it looks strange because of the way the diff looks like @ochaloup removed and added other stuff but I believe it is correct how eb8239b477cb1fb32e55c2cf489fbed1c84cbb9d renders. Is that right? You didn't actually remove anything in eb8239b477cb1fb32e55c2cf489fbed1c84cbb9d?

txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/OutboundBridgeManager.java
mmusgrov: @tomjenkinson The only problem with relying on equality is that we need a second map and then both maps need to be updated atomically.

ochaloup: @tomjenkinson I think that usage of `Map` and getting records from it is based on `Transaction.equals()` implementation. I hope that current code relies on the `transaction.equal()`. Am I wrong here?

txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/OutboundBridgeManager.java
```diff
@@ -88,49 +94,51 @@
...
     /**
      * Create a WS-AT transaction mapping and support objects for a given JTA transaction context.
      *
-     * @param externalTxId The JTA transaction identifier.
+     * @param externalTransaction  transaction where bridge resource will be registered to
+     * @param outboundMappingUid identifier bound to JTA transaction
      * @throws SystemException
      */
-    private static synchronized void createMapping(Transaction transaction, Uid externalTxId) throws SystemException
+    private static synchronized void createMapping(Transaction externalTransaction, Uid outboundMappingUid) throws SystemException
```
tomjenkinson: I think we can still call the "externalTransaction" parameter "transaction"

txbridge/src/test/java/org/jboss/jbossts/txbridge/tests/inbound/service/TestServiceImpl.java
```diff
@@ -78,4 +80,13 @@ public void enlistXAResource(int count) {
             log.error("could not enlist", e);
         }
     }
+
+    private TransactionManager getTransactionManager() {
+        try {
+            return com.arjuna.ats.jta.TransactionManager.transactionManager(new InitialContext());
+        } catch (NamingException ne) {
+            throw new IllegalStateException("Not possible to find transaction manager, "
```
tomjenkinson: As this is a test I think throw the NamingException out of the test.

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
ochaloup: right, I didn't remove anything. it's how git does merging. I'm really not sure if it's possible to hack it somehow.

mmusgrov: OK

txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/BridgeXAResource.java
```diff
@@ -87,10 +88,9 @@ public BridgeXAResource(Uid externalTxId, BridgeWrapper bridgeWrapper)
      */
     private void writeObject(ObjectOutputStream out) throws IOException
     {
-        txbridgeLogger.logger.trace("BridgeXAResource.writeObject() for tx id="+externalTxId);
```
mmusgrov: Have a trace message for the transaction is still useful.

As an aside (I am not asking you to make a change, it is just a comment) these log messages are not conditional on whether trace is enabled.

txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/OutboundBridgeManager.java
```diff
@@ -66,18 +68,19 @@
 
         try
         {
-            Transaction transaction = (Transaction)TransactionManager.transactionManager().getTransaction();
+            Transaction transaction = (Transaction)TransactionManager.transactionManager(new InitialContext()).getTransaction();
 
-            Uid externalTxId = transaction.get_uid();
+            // generating new Uid to get it connected to incoming transaction
+            // the incoming transaction could come from different TM than Narayana is (e.g. Wildfly Transaction client)
```
mmusgrov: This comment is now superfluous.

txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/BridgeXAResource.java
```diff
@@ -70,7 +71,7 @@
      * @param externalTxId the parent JTA transaction identifier.
      * @param bridgeWrapper the control for the subordinate WS-AT transaction.
      */
-    public BridgeXAResource(Uid externalTxId, BridgeWrapper bridgeWrapper)
+    public BridgeXAResource(Transaction externalTxId, BridgeWrapper bridgeWrapper)
```
tomjenkinson: Ah, you are right. Perhaps mark the class as deprecated at this time too? We can then move it later. Also, we could provide the two param ctor for compatability and pass through to the one-arg ctor?

rts/at/bridge/src/main/java/org/jboss/narayana/rest/bridge/inbound/InboundBridge.java
```diff
@@ -212,11 +222,15 @@ private void enlist(final InboundBridge inboundBridge) {
      *
      * @return
      */
-    private Transaction getTransaction() {
+    private Transaction getTransaction(boolean useContainerToLookup) {
```
tomjenkinson: I was wondering about that myself. I wonder if we can revert that now and always use the container lookup'er

tomjenkinson: I just raised this to check if we do really need the resume handling: https://github.com/jbosstm/narayana/pull/1139

ochaloup: @mmusgrov : my understanding is that there is no scenario where txbridge is not running in WFLY

@tomjenkinson : from my testing yes, there is need your handling. There is some trouble with synchronization as you talked with David before. But maybe David's changes in wfly txn client could fix that ...?

txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/BridgeXAResource.java
ochaloup: @tomjenkinson I don't understand exactly what you mean by two param constructor. I can add there the old constructor
```
public BridgeXAResource(Uid externalTxId, BridgeWrapper bridgeWrapper)
```
but still I would need the `Transaction externalTxId` which is not provided. It's the same with one-arg constructor. What I miss here?

txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/BridgeXAResource.java
ochaloup: @mmusgrov ok, just I'm unsure - should I change it back to trace logging (which I think is good idea) + add the check if trace is enabled?

mmusgrov: There are too many places where logging does not check the level first so no I am not asking you to go and change them all. My comment was so that we don't forget to do it sometime in the future.

As to including the transaction in the trace output you may as well if you are making other changes anyway.

txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/BridgeXAResource.java
mmusgrov: Not changing the old constructor will mean any client code (that is not under our control) will still compile. However, unless you can somehow map from a Uid to a transaction then client code will fail at run time.

ochaloup: @mmusgrov so the back-ward compatible constructor should throw some `IllegalStateException` when is called? Or how to proceed?

I mean it would be good to fail fast than leave the runtime fail to some further times where tracking could be harder. What do you think?

mmusgrov: If you cannot map from Uid to Transaction then you could deprecate the old constructor and throw a java.lang.UnsupportedOperationException with a message that says the method is now deprecated in favour of the other constructor.

txbridge/src/main/java/org/jboss/jbossts/txbridge/outbound/BridgeXAResource.java
ochaloup: ok, I see. I created a new jira to check it later
https://issues.jboss.org/browse/JBTM-2862
please, feel free to add or change the description

ochaloup:
Closing as this PR was replaced by https://github.com/jbosstm/narayana/pull/1145

PR #1144, opened 2017-03-06
[WFLY-8275][JBEAP-8993] jca inflow integration with wildfly transaction client

ochaloup:
This change is part of fix of jca inflow integration with new transaction client. The fix is mainly done in WFLY integration (see https://github.com/jbosstm/jboss-as/pull/69) and change is needed in wfly client as well (see https://github.com/wildfly/wildfly-transaction-client/pull/17#issuecomment-283989487).

The point why the change is needed in Narayana (and what I would kindly ask for review here) is that wfly client checks if underlaying TM (for JBoss it's Narayana) knows about transaction. And the `XATerminator` implementation, as it was written, consider knowledge only of top level transactions but not subordinate. That's why I'm adding check for subordinate ones and those one is then returned to wfly transaction client which then works with it.

!BLACKTIE !PERF

https://issues.jboss.org/browse/JBEAP-8993
https://issues.jboss.org/browse/WFLY-8275

requires: https://github.com/jbosstm/jboss-as/pull/69
requires: https://github.com/jbosstm/jboss-transaction-spi/pull/23

### DIFF
--- a/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/TransactionImple.java
+++ b/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/TransactionImple.java
@@ -1153,7 +1153,7 @@ public class TransactionImple implements javax.transaction.Transaction,
 	 * Creates if does not exist and adds to our internal mapping table.
 	 */
 
-	static final TransactionImple getTransaction ()
+	public static final TransactionImple getTransaction ()
 	{
 		TransactionImple tx = null;
 
@@ -1208,6 +1208,21 @@ public class TransactionImple implements javax.transaction.Transaction,
 
 		return tx;
 	}
+
+       public static final TransactionImple getTransaction(Uid id)
+       {
+            try
+            {
+                if (id != null)
+                    return (TransactionImple) _transactions.get(id);
+                else
+                    return null;
+            }
+            catch (Exception e)
+            {
+                return new TransactionImple(null);
+            }
+        }
 
 	public final void shutdown ()
 	{
--- a/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/XATerminatorImple.java
+++ b/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/XATerminatorImple.java
@@ -61,6 +61,7 @@ import com.arjuna.ats.internal.jta.transaction.arjunacore.jca.SubordinationManag
 import com.arjuna.ats.internal.jta.transaction.arjunacore.jca.TransactionImporter;
 import com.arjuna.ats.internal.jta.transaction.jts.subordinate.jca.TransactionImple;
 import com.arjuna.ats.internal.jta.transaction.jts.subordinate.jca.coordinator.ServerTransaction;
+import com.arjuna.ats.internal.jta.utils.jtaxLogger;
 import com.arjuna.ats.jta.exceptions.UnexpectedConditionException;
 import com.arjuna.ats.jta.xa.XidImple;
 import org.jboss.tm.ExtendedJBossXATerminator;
@@ -452,7 +453,12 @@ public class XATerminatorImple implements javax.resource.spi.XATerminator, XATer
 
     public Transaction getTransaction(Xid xid) throws XAException {
         // first see if the xid is a root coordinator
-        return com.arjuna.ats.internal.jta.transaction.arjunacore.TransactionImple.getTransaction(new XidImple(xid).getTransactionUid());
+        Transaction transaction = TransactionImple.getTransaction(new XidImple(xid).getTransactionUid());
+        // second see if the xid is a subordinate txn
+        if(transaction == null) {
+            transaction = SubordinationManager.getTransactionImporter().getImportedTransaction(xid);
+        }
+        return transaction;
     }
 
     public TransactionImportResult importTransaction(Xid xid, int timeoutIfNew) throws XAException {
@@ -465,12 +471,11 @@ public class XATerminatorImple implements javax.resource.spi.XATerminator, XATer
     }
 
     public Transaction getTransactionById(Object id) {
-        return com.arjuna.ats.internal.jta.transaction.arjunacore.TransactionImple.getTransaction((Uid) id);
+        return TransactionImple.getTransaction((Uid) id);
     }
 
     public Object getCurrentTransactionId() {
-        com.arjuna.ats.internal.jta.transaction.arjunacore.TransactionImple transaction =
-            com.arjuna.ats.internal.jta.transaction.arjunacore.TransactionImple.getTransaction();
+        com.arjuna.ats.internal.jta.transaction.jts.TransactionImple transaction = TransactionImple.getTransaction();
         if (transaction == null)
             return null;
 
@@ -485,19 +490,13 @@ public class XATerminatorImple implements javax.resource.spi.XATerminator, XATer
         final Set<Xid> xidsToRecover = new HashSet<Xid>();
         if (recoverInFlight) {
             final TransactionImporter transactionImporter = SubordinationManager.getTransactionImporter();
-            if (transactionImporter instanceof com.arjuna.ats.internal.jta.transaction.arjunacore.jca.TransactionImporterImple) {
-                final Set<Xid> inFlightXids = ((com.arjuna.ats.internal.jta.transaction.arjunacore.jca.TransactionImporterImple) transactionImporter).getInflightXids(parentNodeName);
-                if (inFlightXids != null) {
-                    xidsToRecover.addAll(inFlightXids);
-                }
+            if (transactionImporter instanceof TransactionImporterImple) {
+                throw new UnsupportedOperationException(jtaxLogger.i18NLogger.get_not_supported());
             }
         }
         final javax.resource.spi.XATerminator xaTerminator = SubordinationManager.getXATerminator();
-        if (xaTerminator instanceof com.arjuna.ats.internal.jta.transaction.arjunacore.jca.XATerminatorImple) {
-            final Xid[] inDoubtTransactions = ((com.arjuna.ats.internal.jta.transaction.arjunacore.jca.XATerminatorImple) xaTerminator).doRecover(null, parentNodeName);
-            if (inDoubtTransactions != null) {
-                xidsToRecover.addAll(Arrays.asList(inDoubtTransactions));
-            }
+        if (xaTerminator instanceof XATerminatorImple) {
+            throw new UnsupportedOperationException(jtaxLogger.i18NLogger.get_not_supported());
         } else {
             final Xid[] inDoubtTransactions = xaTerminator.recover(recoveryFlags);
             if (inDoubtTransactions != null) {
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/XATerminatorImple.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/XATerminatorImple.java
@@ -592,7 +592,12 @@ public class XATerminatorImple implements javax.resource.spi.XATerminator, XATer
 
 	public Transaction getTransaction(Xid xid) throws XAException {
 		// first see if the xid is a root coordinator
-		return TransactionImple.getTransaction(new XidImple(xid).getTransactionUid());
+		Transaction transaction = TransactionImple.getTransaction(new XidImple(xid).getTransactionUid());
+		// second see if the xid is a subordinate txn
+		if(transaction == null) {
+		    transaction = SubordinationManager.getTransactionImporter().getImportedTransaction(xid);
+		}
+		return transaction;
 	}
 
     public TransactionImportResult importTransaction(Xid xid, int timeoutIfNew) throws XAException {
--- a/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/utils/jtaxI18NLogger.java
+++ b/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/utils/jtaxI18NLogger.java
@@ -260,6 +260,9 @@ public interface jtaxI18NLogger {
     @LogMessage(level = WARN)
     public void warn_could_not_load_class_will_wait_for_bottom_up(@Cause() ClassNotFoundException cnfe);
 
+	@Message(id = 24059, value = "Inflow recovery is not supported for JTS mode", format = MESSAGE_FORMAT)
+	String get_not_supported();
+
     /*
         Allocate new messages directly above this notice.
           - id: use the next id number in sequence. Don't reuse ids.


ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/XATerminatorImple.java
```diff
@@ -452,7 +452,13 @@ public boolean beforeCompletion (Xid xid) throws javax.transaction.SystemExcepti
 
     public Transaction getTransaction(Xid xid) throws XAException {
         // first see if the xid is a root coordinator
-        return com.arjuna.ats.internal.jta.transaction.arjunacore.TransactionImple.getTransaction(new XidImple(xid).getTransactionUid());
+        Transaction transaction = com.arjuna.ats.internal.jta.transaction.arjunacore.TransactionImple
```
tomjenkinson: I think there is a good chance there is an existing bug here. I think it is wrong to try to get an ArjunaCore TransactionImple here, it should be a JTS one and could be a copy/paste error here: https://github.com/jbosstm/narayana/commit/db5e42340d636c89d348ae2bfec761809b8e0b70#diff-ea4bafab11ba2836771979e805769d8bR445 what do you think @mmusgrov

tomjenkinson: @mmusgrov it looks like this is where getTransactionUid was introduced to JTS side: https://github.com/jbosstm/narayana/commit/9f47df920514255c612c9d9c1cb9c7ad5ad5963a#diff-938c9dfac82d12a192ec11ae14cacf09

ochaloup: @mmusgrov I can see more places in jts `XATerminatorImple` where jta `TransactionImple` is used. I'm not going to propose change till Mike review it
(https://github.com/jbosstm/narayana/blob/master/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/XATerminatorImple.java)

mmusgrov: I agree, this was a cut and paste error. The original implementation was for JTA and the JTS version was only added because it was in the interface. It needs changing to 
```java
com.arjuna.ats.internal.jta.transaction.jts.subordinate.jca.TransactionImple
```
Just change *all* usages (in this file) of com.arjuna.ats.internal.jta.transaction.arjunacore.TransactionImple to TransactionImple (the one in the java import statement).

scripts/hudson/narayana.sh
```diff
@@ -293,8 +293,11 @@ function build_as {
   git pull --rebase --ff-only upstream master
   [ $? = 0 ] || fatal "git rebase failed"
 
+  git clone -b rollback-bit-check https://github.com/ochaloup/wildfly-transaction-client.git
+  ./build.sh clean install -f wildfly-transaction-client/pom.xml
+
   export MAVEN_OPTS="-XX:MaxPermSize=512m -XX:+UseConcMarkSweepGC $MAVEN_OPTS"
-  JAVA_OPTS="-Xms1303m -Xmx1303m -XX:MaxPermSize=512m $JAVA_OPTS" ./build.sh clean install -DskipTests -Dts.smoke=false -Dlicense.skipDownloadLicenses=true $IPV6_OPTS -Drelease=true -Dversion.org.jboss.narayana=5.5.4.Final-SNAPSHOT
+  JAVA_OPTS="-Xms1303m -Xmx1303m -XX:MaxPermSize=512m $JAVA_OPTS" ./build.sh clean install -DskipTests -Dts.smoke=false -Dlicense.skipDownloadLicenses=true $IPV6_OPTS -Drelease=true -Dversion.org.jboss.narayana=5.5.4.Final-SNAPSHOT -Dversion.org.wildfly.transaction.client=1.0.0.Beta19-SNAPSHOT
```
mmusgrov: Maybe add a
```java
// TODO change WFTC to a final version once released
```

ochaloup: Thanks for pointing this out.

I do not plan to commit this particular change.

It's part of the commit which defined SNAPSHOT versions to be used against the latest snapshot built.

mmusgrov: ok

ochaloup:
@mmusgrov @tomjenkinson 

hi guys,

I would need a little bit more help on the proposed changes about `TransactionImple`. There are few troubles.

The jta `TransactionImple` is type of `com.arjuna.ats.internal.jta.transaction.arjunacore.TransactionImple` but the imported jts `TransactionImple` is type of `com.arjuna.ats.internal.jta.transaction.jts.subordinate.jca.TransactionImple`. That is a first clash where one assignment does not work when change is done that way.

Thus I tried to change from `com.arjuna.ats.internal.jta.transaction.arjunacore.TransactionImple` to `com.arjuna.ats.internal.jta.transaction.jts.TransactionImple` but the jts `TransactionImple` does not implement `getTransaction(Uid)` so I added that there.

There is a similar issue with `TransactionImporter` which I changed to `com.arjuna.ats.internal.jta.transaction.jts.jca.TransactionImporterImple` but there is not existing method `getInflightXids(nodeName)` and I'm not sure if I should add it there.

What should be done here, please?

Changes are here
https://github.com/jbosstm/narayana/pull/1144/commits/1205ee915462ae9be5a3b0615d0dcd35bba9430c

*THE LAST COMMIT IS NOT COMPILABLE*

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/TransactionImple.java
```diff
@@ -1153,7 +1153,7 @@ public void setXAResourceState (XAResource xaRes, int state) {
 	 * Creates if does not exist and adds to our internal mapping table.
 	 */
 
-	static final TransactionImple getTransaction ()
+	public static final TransactionImple getTransaction ()
```
tomjenkinson: LGTM

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/TransactionImple.java
```diff
@@ -1209,6 +1209,21 @@ static final TransactionImple getTransaction ()
 		return tx;
 	}
 
+       public static final TransactionImple getTransaction(Uid id)
```
tomjenkinson: LGTM

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/XATerminatorImple.java
```diff
@@ -452,7 +452,7 @@ public boolean beforeCompletion (Xid xid) throws javax.transaction.SystemExcepti
 
     public Transaction getTransaction(Xid xid) throws XAException {
         // first see if the xid is a root coordinator
-        Transaction transaction = com.arjuna.ats.internal.jta.transaction.arjunacore.TransactionImple
+        Transaction transaction = com.arjuna.ats.internal.jta.transaction.jts.TransactionImple
```
tomjenkinson: Try remove package

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/XATerminatorImple.java
```diff
@@ -471,12 +471,12 @@ public SubordinateTransaction getImportedTransaction(Xid xid) throws XAException
     }
 
     public Transaction getTransactionById(Object id) {
-        return com.arjuna.ats.internal.jta.transaction.arjunacore.TransactionImple.getTransaction((Uid) id);
+        return com.arjuna.ats.internal.jta.transaction.jts.TransactionImple.getTransaction((Uid) id);
```
tomjenkinson: Try remove package

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/XATerminatorImple.java
```diff
@@ -471,12 +471,12 @@ public SubordinateTransaction getImportedTransaction(Xid xid) throws XAException
...
     }
 
     public Object getCurrentTransactionId() {
-        com.arjuna.ats.internal.jta.transaction.arjunacore.TransactionImple transaction =
-            com.arjuna.ats.internal.jta.transaction.arjunacore.TransactionImple.getTransaction();
+        com.arjuna.ats.internal.jta.transaction.jts.TransactionImple transaction =
```
tomjenkinson: Try remove package

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/XATerminatorImple.java
```diff
@@ -491,16 +491,16 @@ public void removeImportedTransaction(Xid xid) throws XAException {
         final Set<Xid> xidsToRecover = new HashSet<Xid>();
         if (recoverInFlight) {
             final TransactionImporter transactionImporter = SubordinationManager.getTransactionImporter();
-            if (transactionImporter instanceof com.arjuna.ats.internal.jta.transaction.arjunacore.jca.TransactionImporterImple) {
-                final Set<Xid> inFlightXids = ((com.arjuna.ats.internal.jta.transaction.arjunacore.jca.TransactionImporterImple) transactionImporter).getInflightXids(parentNodeName);
+            if (transactionImporter instanceof com.arjuna.ats.internal.jta.transaction.jts.jca.TransactionImporterImple) {
```
tomjenkinson: investigating....

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/XATerminatorImple.java
tomjenkinson: sorry, I meant:

com.arjuna.ats.internal.jta.transaction.jts.TransactionImple transaction = TransactionImple.getTransaction();

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/XATerminatorImple.java
tomjenkinson: OK, I think we should throw an exception because it just isn't supported, I will PR you
/cc @mmusgrov

tomjenkinson: @ochaloup https://github.com/ochaloup/narayana/pull/4

mmusgrov:
@ochaloup In response to your question about the method public Object getCurrentTransactionId(). Change it to:
```java
com.arjuna.ats.internal.jta.transaction.jts.TransactionImple transaction = TransactionImple.getTransaction();
```
As to your second question about getInflightXids: I did not follow which classes you are asking about.

mmusgrov:
@tomjenkinson those changes you pushed look ok. But can we document somewhere that it isn't meant to work in JTS mode?

ochaloup:
@tomjenkinson thanks Tom for your fixes. I wouldn't have idea that that it should not work for JTS. Your changes were squashed and pushed here. Waiting for CI to run tests.

mmusgrov:
I think just documenting ExtendedJBossXATerminator would be good enough.

ochaloup:
@tomjenkinson JTS mode and EJB remoting is not compatible but could not happen that anybody from outer space (e.g. WFLY integration or JCA) calls the method `getXidsToRecoverForParentNode` and it just wants to get Xids and do not care if Narayana is running in JTS/JTA mode?

Maybe I'm again out of context just wanted to point that my `XATerminator` fixes are mentioned for jca inflow transactions in general (jts and jta). They do not touch any ejb remoting issues.

Regarding documentation I'm not sure about jira but it's mentioned e.g. here (see the _NOTE_)
https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/7.0/html-single/developing_ejb_applications/#iiop_remote_call

ochaloup:
@tomjenkinson hm... I'm sorry, I'm afraid I don't follow here.

You mean that SPI `ExtendedJBossXATerminator` should get information that it could throw `UnsupportedOperationException`? I mean in javadoc.

Or that the method `getXidsToRecoverForParentNode(boolean recoverInFlight, String parentNodeName, int recoveryFlags)` should work as it worked till now (it means your changes will be reverted) and there should be added some overridden variant of `getXidsToRecoverForParentNode`?

mmusgrov:
Tom means to change the signature of
```java
Xid[] getXidsToRecoverForParentNode(boolean recoverInFlight, final String parentNodeName, int recoveryFlags) throws XAException;
```
to 
```java
Xid[] getXidsToRecoverForParentNode(boolean recoverInFlight, final String parentNodeName, int recoveryFlags) throws XAException, java.lang.UnsupportedOperationException;
```

ochaloup:
@mmusgrov ah, I see. I got it, I will do so. Thank you!

mmusgrov:
@ochaloup @tomjenkinson In fact this ExtendedJBossXATerminator interface was added to support EJB3 on request from Flavia and possibly @dmlloyd as well. In light of the new WFTC we should revisit ExtendedJBossXATerminator to decide which bits (if any) of the interface is still required.

ochaloup:
@mmusgrov @tomjenkinson what I can say - some methods from `ExtendedJBossXATerminator` are used in WFTC. Not all defined in interface but still some of them are needed.

ochaloup:
@mmusgrov yes, fully agree. EJB3 subsystem was refactored and replaced by handling in WFTC. WFTC uses just some methods from `ExtendedJBossXATerminator`.

Do you prefer I checked which are those methods and deprecated other ones? Or do you think removal would be better option?

dmlloyd:
One thing I think we need to clarify is how JCA inflow differs from our distributed-JTA protocol, in terms of the transaction import (what parts of the XID are used and how?), recovery (in distributed-JTA we recover by node name, but JCA probably doesn't because you don't know what the upstream coordinator is), and interaction with JTS (anything allowed for JCA *should* be allowed with distributed-JTA, right?).

ochaloup:
@tomjenkinson removing means removing from interface or from interface and implementation from Narayana code too? I'm just asking if it's ok to fail when somebody uses old SPI against new Narayana.

I can change the jbosstm/jboss-transaction-spi#23 as this PR depends on it.

ochaloup:
@tomjenkinson I changed the PR https://github.com/jbosstm/jboss-transaction-spi/pull/23 to deprecated methods unused in WFTC. I'm throwing `UnsupportedOperationException` as implemented before.
Can you take a look at it, please?

ochaloup:
@mmusgrov hi Mike, can you please review this PR and let me know what should be changed or if it's ok to be merged? I mean when I left out those changes in pom.xml and narayana.sh which won't be in the final merge but it's here to get CI pass. Thanks a lot.

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/XATerminatorImple.java
```diff
@@ -452,7 +453,12 @@ public boolean beforeCompletion (Xid xid) throws javax.transaction.SystemExcepti
 
     public Transaction getTransaction(Xid xid) throws XAException {
         // first see if the xid is a root coordinator
-        return com.arjuna.ats.internal.jta.transaction.arjunacore.TransactionImple.getTransaction(new XidImple(xid).getTransactionUid());
+        Transaction transaction = TransactionImple.getTransaction(new XidImple(xid).getTransactionUid());
```
tomjenkinson: I do not think getTransactionUid works for JTS but this method will be deprecated and is not currently used - should schedule a follow up action to remove this code

tomjenkinson: you could add throws UnsupportedOperation to the SPI and throw that here it would be better

ochaloup:
@tomjenkinson I commented on the deprecation of `getTransaction(Xid xid)` at jbosstm/jboss-transaction-spi#23. Not sure about changing the exception being thrown. I see the point but it will break source code compatibility.

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/XATerminatorImple.java
ochaloup: As mentioned on spi comment it's not possible as WFTC uses this to integrate with Narayana. I tested that with JTS and JTA on the failing cases and it worked.
If the functionality should be removed we need to think about different way for WFTC to integrate.

tomjenkinson: OK, I see why it works. That first call will always be null but it will drop into  SubordinationManager.getTransactionImporter().getImportedTransaction(xid); which should pick it out. It would be clearer to implement (for the JTS version) just as the  SubordinationManager.getTransactionImporter().getImportedTransaction(xid); call

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/XATerminatorImple.java
```diff
@@ -481,23 +486,17 @@ public void removeImportedTransaction(Xid xid) throws XAException {
         SubordinationManager.getTransactionImporter().removeImportedTransaction(xid);
     }
 
-    public Xid[] getXidsToRecoverForParentNode(boolean recoverInFlight, String parentNodeName, int recoveryFlags) throws XAException {
+    public Xid[] getXidsToRecoverForParentNode(boolean recoverInFlight, String parentNodeName, int recoveryFlags) throws XAException, UnsupportedOperationException {
```
mmusgrov: You have changed the signature of the interface ExtendedJBossXATerminator#getXidsToRecoverForParentNode and we normally deprecate first. What I would suggest is that we do not declare unchecked exceptions in the interface since the caller is not meant to handle them - add it to the javadoc instead.

scripts/hudson/narayana.sh
mmusgrov: So when will you update it?

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/XATerminatorImple.java
ochaloup: this change was asked by Tom, but I agree to revert it back (the same way in SPI) and leave it as it was

PR #1145, opened 2017-03-07
JBTM-2853 + JBTM-2858: based on XTS handler and RTS filter

ochaloup:
This is an alternative fix for https://github.com/jbosstm/narayana/pull/1138. In fact there are fixes only at WFLY side and fix in test here in Narayana.

The point is that transaction client is hooked after transaction is imported by txn bridge in Narayana. The RTS/XTS hok then import (pull) the transaction from Narayana to wildfly transaction client global context.

For the fix would wok it needs the fix of WFLY-8275 which is discussed at https://github.com/jbosstm/narayana/pull/1144. I mean this PR is dependent on PR#1144. That's why one of the commits refers to  	WFLY-8275 but it won't be part of this PR at the end.

requires: https://github.com/jbosstm/jboss-as/pull/70

!BLACKTIE !PERF NO_WIN

https://issues.jboss.org/browse/JBTM-2858
https://issues.jboss.org/browse/WFLY-8275

### DIFF
--- a/rts/at/bridge/pom.xml
+++ b/rts/at/bridge/pom.xml
@@ -225,9 +225,6 @@
                                 <server.jvm.args>${server.jvm.args}</server.jvm.args>
                                 <node.address>127.0.0.1</node.address>
                             </systemPropertyVariables>
-                            <excludes>
-                              <exclude>AnnotationsTestCase.java</exclude>
-                            </excludes> 
                         </configuration>
                     </plugin>
                 </plugins>
--- a/txbridge/pom.xml
+++ b/txbridge/pom.xml
@@ -306,10 +306,6 @@
 							<includes>
 								<include>**/*Tests.java</include>
 							</includes>
-              <excludes>
-								<exclude>InboundBasicTests.java</exclude>
-								<exclude>InboundCrashRecoveryTests.java</exclude>
-              </excludes>              
 							<!-- System properties passed to test cases -->
 							<systemPropertyVariables>
 								<!-- Used in arquillian.xml - arguments for all JBoss AS instances. -->


ochaloup:
@tomjenkinson agree, that's the last thing that I fight with right now. It was the issue what I was asking Gytis in the email. I would like to avoid the need of adding new item under `Dependencies` to user's deployment. But maybe it won't be possible to hack that. But still I have some time to find a way :)

ochaloup:
@tomjenkinson that's a good point, thank you! That could be the way. The `export` in `module.xml` could provide that. I will look at it.

Then I got another idea as it seems that there is a trouble for `cxf` not seeing the new provider. I'm thinking to investigate on implicit dependencies in that subsystem. Maybe adding the dependency there could work.

ochaloup:
At the end I changed the dependencies handling to wfly rts extension being exported with the narayana rts package. Thanks Tom for the proposal. 

It works on my local machine. If you have time can you please review?  https://github.com/jbosstm/jboss-as/pull/70

scripts/hudson/narayana.sh
```diff
@@ -294,7 +294,7 @@ function build_as {
   [ $? = 0 ] || fatal "git rebase failed"
 
   export MAVEN_OPTS="-XX:MaxPermSize=512m -XX:+UseConcMarkSweepGC $MAVEN_OPTS"
-  JAVA_OPTS="-Xms1303m -Xmx1303m -XX:MaxPermSize=512m $JAVA_OPTS" ./build.sh clean install -DskipTests -Dts.smoke=false -Dlicense.skipDownloadLicenses=true $IPV6_OPTS -Drelease=true -Dversion.org.jboss.narayana=5.5.4.Final-SNAPSHOT
+  JAVA_OPTS="-Xms1303m -Xmx1303m -XX:MaxPermSize=512m $JAVA_OPTS" ./build.sh clean install -DskipTests -Dts.smoke=false -Dlicense.skipDownloadLicenses=true $IPV6_OPTS -Drelease=true -Dversion.org.jboss.narayana=5.5.4.Final-SNAPSHOT -Dversion.org.wildfly.transaction.client=1.0.0.Beta19
```
tomjenkinson: I updated 5_BRANCH now so this should not be required

mmusgrov:
In that case I  approve the PR

PR #1146, opened 2017-03-10
JBTM-2850

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2850

!XTS !RTS !QA_JTA !PERF !BLACKTIE !QA_JTS_JDKORB NO_WIN


ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/TransactionImple.java
```diff
@@ -1052,10 +1052,13 @@ public void endAssociation(Xid _tranID, XAResource _theXAResource, int xaState,
 				try {
```
mmusgrov: The initial if (toThrow == null) on the non duplicate resource on line 1042 will always be true (I probably missed this on the previous review). It is mostly harmless but looks messy.

tomjenkinson: +1 - I made the change

PR #1147, opened 2017-03-14
Removing unused variable declaration in xa JTA/JTSTest

ochaloup:
Based on check of static code analysis got warning of unused variables at
ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/xa/JTATest.java
ArjunaJTS/jtax/tests/classes/com/hp/mwtests/ts/jta/jts/xa/JTSTest.java

I'm proposing here update to remove those unused ones.

This is only a minor update and I'm not sure how do you them. If you think it's not good/relevant to propose such adjustments, please, close this PR. thanks.

!QA_JTA !QA_JTS_JDKORB !BLACKTIE !XTS !PERF NO_WIN

### DIFF
--- a/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/xa/JTATest.java
+++ b/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/xa/JTATest.java
@@ -46,8 +46,6 @@ import javax.transaction.xa.Xid;
 
 import org.junit.Test;
 
-import com.arjuna.ats.jta.common.jtaPropertyManager;
-
 import java.lang.reflect.Field;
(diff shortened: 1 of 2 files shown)


mmusgrov:
Do you mean to say you only found this one in the whole of our code base. There must be others.

ochaloup:
@mmusgrov @tomjenkinson 
this one is new one with 5.5.5.Final release. That was reason why I create this PR. I expected it's still remembered so it's reason to fix it.

I plan to check more about from static code analysis but not for tests. I normally ignore them. As said before, I thought there is a reason as this is a new one and it could help quality of the code. If I understand right still it's not a good reason to provide such adjustments.

PR #1148, opened 2017-03-15
JBTM-2865 write to _work inside _workList lock

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2865

!PERF !QA_JTA !QA_JTS_JDKORB !XTS !RTS NO_WIN !BLACKTIE !AS_TESTS

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/CacheStore.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/CacheStore.java
@@ -736,7 +736,10 @@ class AsyncStore extends Thread // keep priority same as app. threads
             }
         }
 
-        _work = null;
+        synchronized (_workList) {
+            // Must have the lock to write to _work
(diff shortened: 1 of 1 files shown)


mmusgrov:
i was in the middle of the review - you should take me of the reviewers if you plan to merge

mmusgrov:
@tomjenkinson  I would have vetoed it since there are other accesses to _work that aren't protected by a synchronized block.

mmusgrov:
That means that the more diligent the reviewer the greater the chance that he will be ignored ;)

Anyway I am wondering if it is worth the effort cleaning up CacheStore (like I said there are other opportunities for synchronization issues that we missed) since it is only ever used for demo/testing purposes? As a matter of interest what was the circumstance that triggered the JIRA?

mmusgrov:
It also wastes engineering time so the policy should be:

- only assign one reviewer if that's all you feel it requires
- if you assign multiple reviewers (so as to benefit from better coverage) then you must wait for all reviewers to approve the PR

PR #1152, opened 2017-03-17
JBTM-2868 invoke getContextClassLoader within privileged block

istudens:
https://issues.jboss.org/browse/JBTM-2868


mmusgrov:
Thanks for the contribution Ivo

PR #1154, opened 2017-03-24
[JBTM-2870] api for deferred suppressed throwables

ochaloup:
Adding a API to SubordinateTransaction for ejb/WFTC layer could consume deferred throwables and fill them to suppressed part of thrown Exception.

The following fix for eap 6.4 code will be based on
https://github.com/ochaloup/jboss-eap/tree/bz1435549-suppressed-exceptions

!BLACKTIE !PERF

### DIFF
--- a/ArjunaCore/arjuna/pom.xml
+++ b/ArjunaCore/arjuna/pom.xml
@@ -16,7 +16,7 @@
   <parent>
     <groupId>org.jboss.jbossts.arjunacore</groupId>
     <artifactId>arjunacore-all</artifactId>
-    <version>4.17.39.Final</version>
+    <version>4.17.40.Final-SNAPSHOT</version>
     <relativePath>../pom.xml</relativePath>
   </parent>
   <modelVersion>4.0.0</modelVersion>
--- a/ArjunaCore/arjuna/services/pom.xml
+++ b/ArjunaCore/arjuna/services/pom.xml
@@ -15,7 +15,7 @@
   <parent>
     <groupId>org.jboss.jbossts.arjunacore</groupId>
     <artifactId>arjunacore-all</artifactId>
-    <version>4.17.39.Final</version>
+    <version>4.17.40.Final-SNAPSHOT</version>
     <relativePath>../../pom.xml</relativePath>
   </parent>
   <modelVersion>4.0.0</modelVersion>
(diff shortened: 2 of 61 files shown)


ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/SubordinateTransaction.java
```diff
@@ -113,4 +116,28 @@ public void doOnePhaseCommit () throws IllegalStateException,
...
+     *
+     * <p>
+     * JTS implementation of the {@link SubordinateTransaction} interface does not support the deferred throwable
+     * based on the limitation of ORB protocol. This method called on JTS implementation causes {@link UnsupportedOperationException}
+     * being thrown.
+     *
+     * @return list of throwables causing the failure state of the transaction
+     * @throws UnsupportedOperationException for JTS implementation
+     */
+    public List<Throwable> getDeferredThrowables();
```
tomjenkinson: you could add a supportsDeferredThrowables() boolean to the IFC so that the caller knows what to expect?

This is internal interface, you would need to add it to ImportedTransaction instead I expect.

ochaloup: I added the `supportsDeferredThrowables` and used in EAP 6.4 ejb subsystem fix.

The ejb subsystem runs directly with the `SubordinateTransaction` as call `SubordinationManager.getTransactionImporter()..getImportedTransaction(Xid)`. From EAP 6.4 point I think it's not needed this methods being added under `ImportedTransaction` but if you think it should be there I can try to investigate. Should I?

By quick checking the WFTC I think it uses the simlar approach and direct calls to `SubordinateTransaction` too. But it needs to be confirmed by @dmlloyd .

tomjenkinson: Understood - thanks

ochaloup: I checked with @dmlloyd in HipChat conversation and he is fine with this api change. It should be sufficient for WFTC to get information for suppressed exceptions.

PR #1155, opened 2017-03-24
[JBTM-2793] intermittent test failures on xts happening

ochaloup:
After investigation and discussion with Andrew I propose this change.

The `Rule` could be loaded from different class loader which could cause errors of the already loaded rule wouldn't been visible - different class loader means different `Rule` for the Byteman. The `boot:` option is not needed if we rule does not transform some classes from `rt.jar` which seems not being the case here.

This change/fix causes Byteman `-agent:` being loaded only by system classloader.

!QA_JTA !QA_JTS_JDKORB !BLACKTIE !PERF

### DIFF
--- a/XTS/localjunit/crash-recovery-tests/src/main/java/com/arjuna/qa/extension/BaseServerKillProcessor.java
+++ b/XTS/localjunit/crash-recovery-tests/src/main/java/com/arjuna/qa/extension/BaseServerKillProcessor.java
@@ -122,7 +122,9 @@ public abstract class BaseServerKillProcessor implements ServerKillProcessor {
         String scriptName = config.get("scriptName");
 
         String dir = "target" + File.separator + "surefire-reports" + File.separator + "processes";
-        runShellCommand("mkdir " + dir);
+        if(!new File(dir).exists()) {
+            runShellCommand("mkdir " + dir);
+        }
 
         String logFile = dir + File.separator + scriptName + ":" + testClass + "_" + processLogId++ + ".txt";
         runShellCommand(getProcessesCmd() + " > " + logFile);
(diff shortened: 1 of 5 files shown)


XTS/localjunit/crash-recovery-tests/src/main/java/com/arjuna/qa/extension/BaseServerKillProcessor.java
```diff
@@ -122,7 +122,9 @@ protected void dumpProcesses(Container container) throws Exception {
         String scriptName = config.get("scriptName");
 
         String dir = "target" + File.separator + "surefire-reports" + File.separator + "processes";
-        runShellCommand("mkdir " + dir);
+        if(!new File(dir).exists()) {
+            runShellCommand("mkdir " + dir);
```
tomjenkinson: What if the second run of the command also fails? How about just throw an exception?

ochaloup: I think there is no purpose to throw an exception.

This part only tries to create directory. If fails then prints a warning to log. If pass ok. Next test does the same. It tries to create the directory and if fails it just prints warning etc. The subsequent commands then put data to that directory which is expected to exist.

This change only ensures that there won't be attempt to create a new directory if it exists. It means the subsequent code still uses the same directory, the directory is already existing but there won't be obtrusive warning in test log.

ochaloup:
@tomjenkinson : if you have time can you please re-check this PR? I would promot this to being merge, if you are fine with it.
Thank you.

XTS/localjunit/crash-recovery-tests/src/main/java/com/arjuna/qa/extension/BaseServerKillProcessor.java
tomjenkinson: ok, fine by me thanks

PR #1163, opened 2017-04-07
[JBTM-2879] before completition swallow exception

ochaloup:
I felt this being a quick fix, decided to do so.

https://issues.jboss.org/browse/JBTM-2879


ochaloup:
@tomjenkinson I see, I changed the PR to contain only the fix belonging to the issue. Btw. as this is enhancement it is not expected to be put to 5.5 branch, or could be?

I just checked that `e.printStackTrace()` call is used on some more places in different modules than just in `jtax` module. I will create an issue which will talk about refactoring to remove or minimize the use of the direct exception printing to `stderr` but rather using `logger`.

ochaloup:
AS_TESTS fails with some failure seems not being connected with this PR at any way

```
ResourceAdapterCapacityPoliciesTestCase.testNonDefaultDecrementerAndIncrementer:169->checkStatistics:201 Unexpected DestroyedCount expected:<0> but was:<1>
```

PR #1109, opened 2016-12-19
[JBTM-1107] get back: Recovery Support in Compensation API

gytis:
Getting back merge with support of recovery for compensation transactions.

https://issues.jboss.org/browse/JBTM-1107
https://issues.jboss.org/browse/JBTM-2784

Requires https://github.com/jbosstm/jboss-as/pull/61

!PERF !MAIN !QA_JTA !QA_JTS_JDKORB !BLACKTIE

PR #1165, opened 2017-04-12
[JBTM-2799][JBTM-2806] javadoc for jms and tomcat modules

ochaloup:
Splitting the pull request of adding recovery for compensation to two parts (this and https://github.com/jbosstm/narayana/pull/1166). This one rip of the part where javadoc for jms and tomcat modules were added. The original PR#1109 then consist only from changes for compensation.

https://issues.jboss.org/browse/JBTM-2799
https://issues.jboss.org/browse/JBTM-2806

!PERF !QA_JTA !QA_JTS_JDKORB !BLACKTIE !XTS NO_WIN !RTS

Split of the PR: https://github.com/jbosstm/narayana/pull/1109
The other part of split PR: https://github.com/jbosstm/narayana/pull/1166

### DIFF
--- a/ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/ConnectionClosingSynchronization.java
+++ b/ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/ConnectionClosingSynchronization.java
@@ -28,12 +28,17 @@ import javax.jms.JMSException;
 import javax.transaction.Synchronization;
 
 /**
+ * Synchronization to close JMS connection at the end of the transaction.
+ *
  * @author <a href="mailto:gytis@redhat.com">Gytis Trikleris</a>
  */
 public class ConnectionClosingSynchronization implements Synchronization {
 
     private final Connection connection;
 
+    /**
+     * @param connection connection to be closed.
+     */
     public ConnectionClosingSynchronization(Connection connection) {
         this.connection = connection;
     }
@@ -43,6 +48,11 @@ public class ConnectionClosingSynchronization implements Synchronization {
         // Nothing to do
     }
 
+    /**
+     * Close the connection no matter what the status of the transaction is.
+     *
+     * @param status the status of the completed transaction
+     */
     @Override
     public void afterCompletion(int status) {
         if (jtaLogger.logger.isTraceEnabled()) {
--- a/ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/ConnectionFactoryProxy.java
+++ b/ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/ConnectionFactoryProxy.java
@@ -39,11 +39,21 @@ public class ConnectionFactoryProxy implements ConnectionFactory {
 
     private final TransactionHelper transactionHelper;
 
+    /**
+     * @param xaConnectionFactory factory to get XA connection instances, not null.
+     * @param transactionHelper utility to make transaction resources registration easier.
+     */
     public ConnectionFactoryProxy(XAConnectionFactory xaConnectionFactory, TransactionHelper transactionHelper) {
         this.xaConnectionFactory = xaConnectionFactory;
         this.transactionHelper = transactionHelper;
     }
 
+    /**
+     * Get XA connection from the provided factory and wrap it with {@link ConnectionProxy}.
+     *
+     * @return XA connection wrapped with {@link ConnectionProxy}.
+     * @throws JMSException if failure occurred creating XA connection.
+     */
     @Override
     public Connection createConnection() throws JMSException {
         Connection connection = new ConnectionProxy(xaConnectionFactory.createXAConnection(), transactionHelper);
@@ -55,6 +65,14 @@ public class ConnectionFactoryProxy implements ConnectionFactory {
         return connection;
     }
 
+    /**
+     * Get XA connection from the provided factory with credentials and wrap it with {@link ConnectionProxy}.
+     * 
+     * @param userName
+     * @param password
+     * @return XA connection wrapped with {@link ConnectionProxy}.
+     * @throws JMSException if failure occurred creating XA connection.
+     */
     @Override
     public Connection createConnection(String userName, String password) throws JMSException {
         Connection connection = new ConnectionProxy(xaConnectionFactory.createXAConnection(userName, password),
--- a/ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/ConnectionProxy.java
+++ b/ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/ConnectionProxy.java
@@ -47,11 +47,22 @@ public class ConnectionProxy implements Connection {
 
     private final TransactionHelper transactionHelper;
 
+    /**
+     * @param xaConnection XA connection which needs to be proxied.
+     * @param transactionHelper utility to make transaction resources registration easier.
+     */
     public ConnectionProxy(XAConnection xaConnection, TransactionHelper transactionHelper) {
         this.xaConnection = xaConnection;
         this.transactionHelper = transactionHelper;
     }
 
+    /**
+     * Simply create a session with an XA connection if there is no active transaction. Or create a proxied session and register
+     * it with an active transaction.
+     *
+     * @see SessionProxy
+     * @see Connection#createSession(boolean, int)
+     */
     @Override
     public Session createSession(boolean transacted, int acknowledgeMode) throws JMSException {
         if (transactionHelper.isTransactionAvailable()) {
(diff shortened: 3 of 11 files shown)


ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/ConnectionFactoryProxy.java
```diff
@@ -39,11 +39,21 @@
 
     private final TransactionHelper transactionHelper;
 
+    /**
+     * @param xaConnectionFactory factory to get XA connection instances.
```
mmusgrov: Do you need a NotNull on the xaConnectionFactory - it is used elsewhere without any null checks

ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/ConnectionProxy.java
```diff
@@ -61,6 +72,12 @@ public Session createSession(boolean transacted, int acknowledgeMode) throws JMS
         return xaConnection.createSession(transacted, acknowledgeMode);
     }
 
+    /**
+     * Simply close the proxied connection if there is no active transaction. Or register a
+     * {@link ConnectionClosingSynchronization} if active transaction exists.
+     * 
+     * @throws JMSException if transaction status check, synchronization registration, or connection closing has failed.
```
mmusgrov: What is a "a failed transaction status check". The javadoc needs to say which status values result in an exception being thrown.

ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/ConnectionProxy.java
```diff
@@ -75,53 +92,106 @@ public void close() throws JMSException {
         }
     }
 
+    /**
+     * Delegate to {@link #xaConnection}
```
mmusgrov: This javadoc comment is describing how the method is implemented. I would just duplicate the javadoc for whatever xaConnection.getClientID() says. Actually just saying @see Connection#getClientID() should be good enough. But it is up to you if you think that the text adds something.

Similarly for the other delegation methods below.

ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/ConnectionClosingSynchronization.java
```diff
@@ -43,6 +48,11 @@ public void beforeCompletion() {
         // Nothing to do
     }
 
+    /**
+     * Close the connection despite the status of the transaction.
```
mmusgrov: Instead of using the word despite I would use any of the following:
"Close the connection."
"Close the connection regardless of what the status of the transaction is."
"Close the connection no matter what the status of the transaction is."

Similarly for other usages of the word despite below.

ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/SessionProxy.java
```diff
@@ -54,11 +56,21 @@
...
+     * @param transactionHelper utility to make transaction resources registration easier.
+     */
     public SessionProxy(XASession xaSession, TransactionHelper transactionHelper) {
         this.xaSession = xaSession;
         this.transactionHelper = transactionHelper;
     }
 
+    /**
+     * Simply close proxied session if there is no active transaction. Or if transaction exists, delist session's XA resource
+     * and register {@link SessionClosingSynchronization}.
```
mmusgrov: how about explicitly saying what the registration is for, ie:
"and register a {@link SessionClosingSynchronization} to close the proxied session."

ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/TransactionHelper.java
```diff
@@ -26,16 +26,43 @@
...
+ *
  * @author <a href="mailto:gytis@redhat.com">Gytis Trikleris</a>
  */
 public interface TransactionHelper {
 
+    /**
+     * Check if transaction is active. If error occurs wrap an original exception with {@link JMSException}.
+     *
+     * @return whether transaction is active or not.
+     * @throws JMSException if error occurred getting transaction status.
```
mmusgrov: Are the reasons for not being able to get the transaction status known? If so it would be useful to state those reasons here.

Similarly for the other methods below. There seems to be a few methods that can throw a generic JMSException so the text could go into the javadoc for the class instead of each method.

tomcat/tomcat-jta/src/main/java/org/jboss/narayana/tomcat/jta/NarayanaJtaServletContextListener.java
```diff
@@ -64,6 +66,23 @@
     private static final List<String> DEFAULT_EXPIRY_SCANNERS = Arrays
             .asList(ExpiredTransactionStatusManagerScanner.class.getName());
 
+    /**
+     * Initialize and start Narayana JTA services.
+     * <p>
+     * During the setup node identifier, recovery modules, orphan filters, and expiry scanners are setup. Configuration file
```
mmusgrov: Is there a javadoc link we can use to say where this "Configuration" is?

{@link JTAEnvironmentBean} probably has most of what you are intending.

tomcat/tomcat-jta/src/main/java/org/jboss/narayana/tomcat/jta/NarayanaJtaServletContextListener.java
```diff
@@ -91,6 +116,9 @@ public void contextDestroyed(ServletContextEvent servletContextEvent) {
         });
     }
 
+    /**
+     * If node identifier wasn't set byt property manager, then set default.
```
mmusgrov: typo (byt). Can you {@link DEFAULT_NODE_IDENTIFIER} for the default value.

Similarly for the other default values below.

ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/ConnectionFactoryProxy.java
ochaloup: do I understand right that it should be changed to

```
@param xaConnectionFactory factory to get XA connection instances, not null.
```
?

mmusgrov: Yes (since we do not have a standard @NotNull annotation)

ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/ConnectionProxy.java
ochaloup: I see, I agree with you. Just I think that explicitly mentioning that method just delegates call to somewhere else brings information value. I would personally leave it here.

mmusgrov: OK leave it here and I withdraw my comment.

ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/TransactionHelper.java
ochaloup: the JMSException is thrown in case when some unexpected failure occurs in transaction manager, when SystemException is thrown from there: https://docs.oracle.com/javaee/7/api/javax/transaction/TransactionManager.html#getStatus--

I changed the sentence to: `if transaction service has failed in unexpected way to obtain transaction status`

I think the information should be put to each occurence of `@throws JMSException` rather than to the javadoc of the whole class. When I want to use some method want to see info directly at the method not to search somewhere at class level. Especially this is true when using standard IDE which shows hints of javadoc when completing a method.

ochaloup:
@mmusgrov thanks for your objections. I tried to address them. To be honest all the appreciation for the work belongs to @gytis. I'm just trying to arrange the state and to get merged this not merged bits

tomcat/tomcat-jta/src/main/java/org/jboss/narayana/tomcat/jta/NarayanaJtaServletContextListener.java
mmusgrov: @ochaloup Ah I was waiting for you to respond to this question but I see you added another commit that includes the link. In future if you ack/nack each comment or include a catch all comment then that will let the reviewer know that the PR is ready to be reviewed again.

ochaloup: I see, my wrong. I forgot to summarize that I finished with changes and you can look again.

Thank you for the review and your comments!

PR #1166, opened 2017-04-12
[JBTM-1107] compensations recovery revert revert

ochaloup:
Getting back merge with support of recovery for compensation transactions.

https://issues.jboss.org/browse/JBTM-1107
https://issues.jboss.org/browse/JBTM-2784

Requires: https://github.com/jbosstm/jboss-as/pull/72

Replacing PR: https://github.com/jbosstm/narayana/pull/1109
The original PR#1109 was split to two parts - to this PR and https://github.com/jbosstm/narayana/issues/1165 which consists update of jms/tomcat javadoc.

!MAIN !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !PERF NO_WIN !RTS !AS_TESTS !postgres !oracle !db2 !mysql
NO_TEST


ochaloup:
@tomjenkinson sure, I plan to and I look at it. I just didn't get to it last week. I will try to do so this one.

ochaloup:
@tomjenkinson thanks Tom!

I haven't got to this yet today. I'm going to create new PR for jboss-as.

btw. you mean that I don't need to change the wildfly/pom.xml to Narayana snapshot version, like `<version.org.jboss.narayana>5.5.1.Final-SNAPSHOT</version.org.jboss.narayana>`, right?

I didn't know that you fixed the scripts. Thanks, that's really nice!

ochaloup:
I forgot the change of Narayana version in jboss as PR when I was testing it on my local machine. It seems that rebase failed because of it, thus leaving the original narayana version in the PR and leaving the script to update it for WFLY using snapshot version of Narayana used in this PR. I hope it will compile now.

ochaloup:
The problem was (and I'm not sure why) that using the property `-Dversion.org.jboss.narayana` with version `5.6.0.Final-SNAPSHOT` was not reflected during the build. When I changed the Wildfly PR with hardcoding the version to pom.xml it seems to work fine.

I'm going to run the tests once again to have MAIN and AS_TESTS included.

ochaloup:
Summary of this PR

* I propose merging this PR to the Narayana master upstream
* I solved the documentation issues and fixed the troubles with recovery
* The parallel PR for WFLY is ready for creating a PR to WFLY 11.x+. The WFLY PR would depend on this PR would be merged and when WFLY adopt Narayana 5.6.x version.  This won't happen before WFLY 11 release.
* My last troubles which causing AS TEST permanent failures were caused by not having enough entropy for generating token in on of the WS tests (see https://issues.jboss.org/browse/WFLY-5507). After struggling configure docker image to support that I found that all that depends on host configuration. The solution was to install package `haveged` at `narayanaci1` and start the deamon (`systemctl start haveged`). `cat /proc/sys/kernel/random/entropy_avail` before configuration returns about 150, after `haveged` being configured the value goes around 1100.

@tomjenkinson what do you think about this being merged? what is the process?

ochaloup:
@tomjenkinson I understand your point. I responded on the as PR here:
https://github.com/jbosstm/jboss-as/pull/72#pullrequestreview-53740837

I can't see a simple solution when compensations should be offered as part of the transactions subystem. I can imagine to split them for transactions and compensations - thus having the compensations subsystem. Or move the compensations capability under XTS subsystem as they are kind of bound together in a way.

When I'm thinking about it - if the dependency on xts susbystem is wrong then I would prefer to move the whole compensation stuff under XTS subsystem (thus using compensations would need XTS subsystem being active).

Would you have some better or different stand point on this? Thanks.

ochaloup:
@tomjenkinson kind of. There are two "modes" how the compesation framework can work. It can work with local participants - these are participants local to the TM - like MongoDB jdbc driver. And the remote participant - which is for the compensation framework the participant communicates over WS, thus in fact WS-BA. The compensation framework offers a kind of similar functionality as `@LRA` (aka CDI interceptors/REST filters) do for the LRA transactions.

I haven't dived much deep to the matter of the dependency. As we started to talk here about it. Compensation framework is dependent on XTS (compilation time). But if you use only local participants you don't need to have the xts code on your classpath. If you want to use it for managing WS calls and having WS-BA being part of the compensating transaction then yes you will probably need it on the classpath.

ochaloup:
@tomjenkinson I'm planning it just I haven't got time to do so yet, unfortunately. I'll do my best to provide the the fix as soon as possible.

ochaloup:
closing as this will be worked on in the future in different way

PR #1167, opened 2017-04-18
[WFLY-8379] setAvailability of UserTransaction should be done in doPrivileged section

ochaloup:
WildFly 11 and EAP 7.1 switch to WFTC (https://github.com/wildfly/wildfly-transaction-client) which defines it's own security permissions of security manager for some of the actions.

When UserTransaction setting availability (https://github.com/wildfly/wildfly-transaction-client/blob/master/src/main/java/org/wildfly/transaction/client/LocalUserTransaction.java#L79) it requires `modifyUserTransactionAvailability`. This makes troubles e.g. for MDB invocations (see Jeff's comment at JBEAP-9599).

I think Narayana should process the call in doPrivileged section for not forcing user to set permission `modifyUserTransactionAvailability` when he just want to send a jms message (even it's under security manager).

This PR wraps the call in the privileged section.

!QA_JTA !QA_JTS_JDKORB !BLACKTIE !PERF

https://issues.jboss.org/browse/WFLY-8379
https://issues.jboss.org/browse/JBEAP-9599
5.5 PR: https://github.com/jbosstm/narayana/pull/1168

### DIFF
--- a/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
+++ b/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
@@ -23,6 +23,8 @@
 package com.arjuna.ats.jta.cdi.transactional;
 
 
+import static java.security.AccessController.doPrivileged;
+
 import com.arjuna.ats.jta.common.jtaPropertyManager;
 import com.arjuna.ats.jta.logging.jtaLogger;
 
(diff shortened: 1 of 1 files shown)


ochaloup:
Failure on AS testsuite

```
org.jboss.as.test.integration.jca.capacitypolicies.DatasourceCapacityPoliciesTestCase
testNonDefaultDecrementerAndIncrementer(org.jboss.as.test.integration.jca.capacitypolicies.DatasourceCapacityPoliciesTestCase)  Time elapsed: 0.281 sec  <<< FAILURE!
java.lang.AssertionError: Unexpected DestroyedCount expected:<0> but was:<1>
```

is known and should be fixed by https://github.com/wildfly/wildfly/pull/9941

The failure is not a trouble of this change. I'm going to merge it.

PR #1171, opened 2017-04-20
JBTM-2885 If an XAR returns an XAE in XAR::recover() it is possible t…

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2885

### DIFF
--- a/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
+++ b/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
@@ -41,12 +41,15 @@ import java.io.File;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 import javax.transaction.xa.XAException;
 import javax.transaction.xa.XAResource;
 import javax.transaction.xa.Xid;
 
+import com.arjuna.ats.internal.jta.recovery.arjunacore.JTANodeNameXAResourceOrphanFilter;
+import org.jboss.tm.XAResourceWrapper;
 import org.junit.Test;
 
 import com.arjuna.ats.arjuna.common.Uid;
@@ -73,6 +76,8 @@ import com.hp.mwtests.ts.jta.common.RecoveryXAResource;
 
 public class XARecoveryModuleUnitTest
 {
+    private boolean rolledback;
+
     @Test
     public void testNull ()
     {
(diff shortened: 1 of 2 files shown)


ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
```diff
@@ -374,6 +378,106 @@ public void start(Xid xid, int i) throws XAException {
         assertEquals(recoverCalled, 6);
     }
 
+    @Test
```
mmusgrov: Why don't we need a JTS test for this too?

tomjenkinson: We don't actually have orphan detection for JTS: https://issues.jboss.org/browse/JBTM-2124

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
```diff
@@ -374,6 +378,106 @@ public void start(Xid xid, int i) throws XAException {
...
+
+        xarm.periodicWorkFirstPass();
+        xarm.periodicWorkSecondPass();
+        assertFalse(rolledback);
+        xarm.periodicWorkFirstPass();
+        xarm.periodicWorkSecondPass();
+        assertFalse(rolledback);
+        xarm.periodicWorkFirstPass();
+        xarm.periodicWorkSecondPass();
+        assertTrue(rolledback);
```
mmusgrov: A comment would be useful here (since the logic requires some insider knowledge), something like

> * 1st pass: returns one xid (count is 1)
> * 2nd pass: throws an exception (count is 2)
> * 1st pass: throws an exception (count is 3)
> * 2nd pass: throws an exception (count is 4)
> * 1st pass: returns an empty list of xids (count is 5)
> * 2nd pass: returns an empty list of xids (count is 6)

> So the first two recovery cycles do nothing with the resource (because phase two is getting the exception)
> When count reaches 6 it sees that the xid has gone and presumes abort so calls rollback and hence assertTrue(rolledback) passes

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
mmusgrov: OK

ochaloup:
I think this is a safe way how to proceed in this case, just I wonder if it's necessary to do such change or if it could work as it is fine too. Just I would like to understand if you don't mind.

I mean - orphan detection works when there is no `xid` in narayana object store but there is some `xid` in artemis txn store. It would mean that Narayana already removes the `xid` from its object store, right? But it should not happen when XAException is returned (if fixed in artemis) from the call of `recover()` function, or am I understanding it wrong? Or this fix is self defence against artemis not returning the exception?

ochaloup:
I see. And I can see that Andy will just shutdown the Artemis when connection to store is lost. But... I got another case when investigating the code, again I'm not fully sure here, I have several blank spots here.

Does have the `_xidScans` dome influence on how assumed completion (JBTM-860) works? I mean the `XAResource.recover()` throws exception, it's 'ignored' and `_xidScans` is cleared. But it could happen that the recover phase passed in some previous recovery run fine and the `xares` is already saved in map of contected jndis (https://github.com/jbosstm/narayana/blob/master/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java#L688). Now the disruption occurs and next `recover()` clean the `_xidScans` and the method `getContactedJndiNames()` will returns the `xares`.
Could that mean that the `xid` will be removed from the narayana store?

ochaloup:
@tomjenkinson thank you for the test update

PR #1172, opened 2017-04-26
JBTM-2890 Allow a user to set async for afterCompletion without befor…

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2890

!AS_TESTS !BLACKTIE !QA_JTA !QA_JTS_JDKORB !RTS NO_WIN !XTS !PERF !QA_JTS_JACORB  !QA_JTS_OPENJDKORB

### DIFF
--- a/ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/AtomicActionAsyncTest.java
+++ b/ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/AtomicActionAsyncTest.java
@@ -49,6 +49,13 @@ public class AtomicActionAsyncTest extends AtomicActionTestBase
     }
 
     @Test
+    public void testCompletionWithFailuresAsyncAfterCompletion() throws Exception {
+        AtomicActionTestBase.init(false);
+        arjPropertyManager.getCoordinatorEnvironmentBean().setAsyncAfterSynchronization(true);
+        super.testCompletionWithFailures();
+    }
+
+    @Test
     @Override
     public void testPrepareWithLRRSuccess() {
         super.testPrepareWithLRRSuccess();
--- a/ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/AtomicActionTestBase.java
+++ b/ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/AtomicActionTestBase.java
@@ -50,7 +50,7 @@ public class AtomicActionTestBase
 
     protected void testAbort () throws Exception
     {
-        executeTest(true, ActionStatus.COMMITTED, null, new BasicRecord(), new BasicRecord());
+        executeTest(false, ActionStatus.ABORTED, null, new BasicRecord(), new BasicRecord());
     }
 
     protected void testPrepareWithLRRSuccess()
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TwoPhaseCoordinator.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TwoPhaseCoordinator.java
@@ -406,6 +406,13 @@ public class TwoPhaseCoordinator extends BasicAction implements Reapable
     protected boolean asyncAfterCompletion(int myStatus, boolean report_heuristics) {
         boolean problem = false;
 
+		if (synchronizationCompletionService == null) {
+			synchronizationCompletionService = TwoPhaseCommitThreadPool.getNewCompletionService();
+		}
+		if (runningSynchronizations == null) {
+			runningSynchronizations = new ArrayList<Future<Boolean>>(_synchs.size());
+		}
+
         // note there is no need to synchronize on _synchs since synchronizations cannot be registered once
         // the action has started to commit
         for (Iterator<SynchronizationRecord> i =_synchs.iterator(); i.hasNext(); ) {


ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/AtomicActionAsyncTest.java
```diff
@@ -49,6 +56,28 @@ public void testAbort() throws Exception {
...
+                return true;
+            }
+        }, new SyncRecord(true, SyncRecord.FailureMode.NONE) {
+            @Override
+            public boolean afterCompletion(int status) {
+                testAbortAsyncAC = testAbortAsyncAC && (Thread.currentThread() != thread);
+                return true;
+            }
+        }}, new BasicRecord(), new BasicRecord());
+        assertTrue(testAbortAsyncAC);
```
mmusgrov: testAbortAsyncAC starts out true so if there is a bug which results in the synchronizations never getting called then this assert will be vacuously true.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/AtomicActionAsyncTest.java
```diff
@@ -49,6 +56,28 @@ public void testAbort() throws Exception {
     }
 
     @Test
+    public void testAbortAsyncAC() throws Exception {
```
mmusgrov: Prior to your fix would this new test have produced the exception reported in the JIRA:

> When tx abort, and we have asyncAfterSynchronization option, there's exception

If it did not then would like to see a test that acts a reproducer.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/AtomicActionAsyncTest.java
```diff
@@ -21,15 +21,22 @@
...
 import org.junit.Assert;
 import org.junit.BeforeClass;
 
 import com.arjuna.ats.arjuna.common.arjPropertyManager;
 
+import static org.junit.Assert.assertTrue;
+
 public class AtomicActionAsyncTest extends AtomicActionTestBase
 {
+    private boolean testAbortAsyncAC = true;
```
tomjenkinson: it can;t be there as it would need to be final as the inner classes mutate it

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/AtomicActionAsyncTest.java
tomjenkinson: it should do yes

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/AtomicActionAsyncTest.java
```diff
@@ -49,6 +56,28 @@ public void testAbort() throws Exception {
...
     @Test
+    public void testAbortAsyncAC() throws Exception {
+        AtomicActionTestBase.init(false);
+        arjPropertyManager.getCoordinatorEnvironmentBean().setAsyncAfterSynchronization(true);
+
+        final Thread thread = Thread.currentThread();
+        executeTest(true, ActionStatus.COMMITTED, new SyncRecord[] { new SyncRecord() {
+            @Override
+            public boolean afterCompletion(int status) {
+                testAbortAsyncAC = testAbortAsyncAC && (Thread.currentThread() != thread);
```
tomjenkinson: I don't believe it was designed with the executor to do that. the test is running in main thread and an executor pool executes the syncs

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/AtomicActionAsyncTest.java
tomjenkinson: I see, I can add a wasCalled bool to the class too.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/AtomicActionAsyncTest.java
mmusgrov: Make it an array then the individual elements can be mutated.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/AtomicActionAsyncTest.java
mmusgrov: It means that if someone were to rework the implementation he would need to puzzle over why an existing test is failing. Can you add more commentary about why asserting that the callbacks run in different threads from the caller make the exception disappear.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/AtomicActionAsyncTest.java
mmusgrov: It is not obvious why the test validates that abort no longer produces an exception. Particularly so because the test commits the transaction (since you pass ActionStatus.COMMITTED to the executeTest method).

tomjenkinson: you are right, the true/ActionStatus.COMMITTED I was relying on the behaviour in the existing test to be correct which it is not. I have improved this test and corrected the previous test

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/AtomicActionTestBase.java
```diff
@@ -50,7 +50,7 @@ protected void testCommit () throws Exception
 
     protected void testAbort () throws Exception
     {
-        executeTest(true, ActionStatus.COMMITTED, null, new BasicRecord(), new BasicRecord());
+        executeTest(false, ActionStatus.ABORTED, null, new BasicRecord(), new BasicRecord());
```
mmusgrov: Good catch. Mind, you also need to set the second parameter (expectedResult) to ActionStatus.ABORTED.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/AtomicActionAsyncTest.java
```diff
@@ -49,6 +55,42 @@ public void testAbort() throws Exception {
...
+        arjPropertyManager.getCoordinatorEnvironmentBean().setAsyncAfterSynchronization(true);
+
+        final Thread thread = Thread.currentThread();
+        executeTest(false, ActionStatus.ABORTED, new SyncRecord[]{new SyncRecord() {
+            @Override
+            public boolean afterCompletion(int status) {
+                if (status == ActionStatus.ABORTED) {
+                    rollback[0]++;
+                }
+                testAbortAsyncAC[0] = testAbortAsyncAC[0] && (Thread.currentThread() != thread);
```
mmusgrov: It looks like this test only validates that after completion callbacks run in new threads. If that is the case, I have three questions:

1. why not just say assertNotEquals(Thread.currentThread(),  thread) here and get rid of the testAbortAsyncAC and rollback variables, it makes the test clear.
2. why do you need to test for rollback, we have other tests that to that or is that the bug you have created the test for? But if you do need the check you could do it here: if (status != ABORTED) then fail("should have aborted")
3. it not clear how the test validates that the bug is fixed - please can add a comment to explain what the test is testing for

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/AtomicActionTestBase.java
mmusgrov: Good catch

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/AtomicActionAsyncTest.java
tomjenkinson: Good point, I could remove the thread checks to be fair. The main thing is that without the patch you get an exception

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/AtomicActionAsyncTest.java
tomjenkinson: I have updated the test to reflect the issue being raised (i.e. rollback causes NPE)

mmusgrov:
@tomjenkinson The easiest way to test this is to add the test
```java
    public void testCompletionWithAbort() throws Exception {
        super.testCompletionWithFailures(false);
    }
```

and then just update AtomicActionTestBase#testCompletionWithFailures to accept the boolean to tell it whether or not to commit or abort when calling executeTest

mmusgrov:
FYI calling any of the following (and passing abort to executeTest) would test this bug (they are effectively the same test since before completion isn't called for abort):

testCompletionWithFailures returns false in before completion
testCompletionWithoutFailures return true in both synchs

PR #1173, opened 2017-05-02
[JBTM-2892] Not IndexOutOfBound exception when enlist on subordinate txn

ochaloup:
https://issues.jboss.org/browse/JBTM-2892

When method TransactionImple.enlistResource is called with empty params array `IndexOutOfBound` is emitted.

The check for array length does not correspond by one with index hit.

### DIFF
--- a/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
+++ b/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
@@ -120,7 +120,7 @@ public class XAResourceRecord extends com.arjuna.ArjunaOTS.OTSAbstractRecordPOA
 
 		if (params != null)
 		{
-			if (params.length >= XACONNECTION)
+			if (params.length > XACONNECTION)
 			{
 				if (params[XACONNECTION] instanceof RecoverableXAConnection)
 					_recoveryObject = (RecoverableXAConnection) params[XACONNECTION];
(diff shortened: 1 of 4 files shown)


ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -120,7 +120,7 @@ public XAResourceRecord(TransactionImple tx, XAResource res, Xid xid,
 
 		if (params != null)
 		{
-			if (params.length >= XACONNECTION)
+			if (params.length >= XACONNECTION + 1)
```
mmusgrov:
```java
if (params.length > XACONNECTION)
```
would be better

ochaloup: @mmusgrov I see, but I wanted to follow how is done here
https://github.com/jbosstm/narayana/blob/master/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/TransactionImple.java#L549

should I change the both places in way you suggest?

mmusgrov: Well I am not sure how clever the compiler but I think that it will perform the addition and then do the compare which is a few more instructions that a single compare. So yes change it in both places.

ArjunaJTS/jtax/tests/classes/com/hp/mwtests/ts/jta/jts/recovery/XARecoveryModuleUnitTest.java
```diff
@@ -106,7 +137,7 @@ public void tearDown () throws Exception
         myOA.destroy();
         myORB.shutdown();
     }
-    
+
```
tomjenkinson: This part seems to be superfluous formatting change?

ochaloup:
I cleaned indentation issues, the CI passes, I'm going to merge it.

Thank you for the review!

PR #1174, opened 2017-05-03
[JBTM-2124] enable orphan detection for JTS

ochaloup:
Moving handling of orphan detection from JTA land only to JTS too.

I will be happy for you comments. I'm unsure if some "design" decision was made correctly.

https://issues.jboss.org/browse/JBTM-2124
Discussion on the topic: https://developer.jboss.org/message/971565#971565


ochaloup:
@tomjenkinson the forum post created: https://developer.jboss.org/message/971565

currently looking on reason of instabilities

PR #1175, opened 2017-05-05
JBTM-2881 Do not hide application exceptions

mmusgrov:
https://issues.jboss.org/browse/JBTM-2881

!PERF !QA_JTA !QA_JTS_JDKORB !BLACKTIE !QA_JTS_JACORB !QA_JTS_OPENJDKORB !XTS !RTS 
 NO_WIN !RTS

### DIFF
--- a/STM/src/main/java/org/jboss/stm/internal/reflect/InvocationHandler.java
+++ b/STM/src/main/java/org/jboss/stm/internal/reflect/InvocationHandler.java
@@ -21,10 +21,12 @@
 
 package org.jboss.stm.internal.reflect;
 
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.HashMap;
 
+import com.arjuna.ats.txoj.logging.txojLogger;
 import org.jboss.stm.LockException;
 import org.jboss.stm.TransactionException;
 import org.jboss.stm.annotations.LockFree;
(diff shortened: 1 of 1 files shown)


STM/src/main/java/org/jboss/stm/internal/reflect/InvocationHandler.java
```diff
@@ -400,8 +401,12 @@ public Object invoke (Object proxy, java.lang.reflect.Method method, Object[] ar
                             }
                         }
                     }
-    
-                    return method.invoke(_theObject, args);
+
+                    try {
+                        return method.invoke(_theObject, args);
+                    } catch (InvocationTargetException e) {
+                        throw e.getCause() != null ? e.getCause() : e;
```
ochaloup: would not be good to log the exception with some contextual information (what was invoked with what parameters) before the exception is thrown?

mmusgrov: No, this is application code throwing an exception. It is up to the developer to catch his own exception and log accordingly.

ochaloup: I meant log,  e.g. to debug, for easier further investigation. But it was just a question to think about.

mmusgrov: OK good point, done.

PR #1177, opened 2017-05-08
JBTM-2896 Do not produce invalid warnings when committing OptimisticL…

mmusgrov:
…ockRecords

https://issues.jboss.org/browse/JBTM-2896

!PERF !QA_JTA !QA_JTS_JDKORB !BLACKTIE !QA_JTS_JACORB !QA_JTS_OPENJDKORB !XTS !RTS
NO_WIN !RTS

### DIFF
--- a/STM/src/main/java/org/jboss/stm/internal/optimistic/OptimisticLockRecord.java
+++ b/STM/src/main/java/org/jboss/stm/internal/optimistic/OptimisticLockRecord.java
@@ -135,7 +135,7 @@ class OptimisticLockRecord extends LockRecord
     {       
         boolean stateOK = checkState();
         
-        if (stateOK)
+        if (!stateOK)
         {
             txojLogger.i18NLogger.warn_OptimisticLockRecord_2((LockManager) value());
         }


STM/src/main/java/org/jboss/stm/internal/optimistic/OptimisticLockRecord.java
```diff
@@ -135,7 +135,7 @@ public int topLevelCommit ()
     {       
         boolean stateOK = checkState();
         
-        if (stateOK)
+        if (!stateOK)
```
tomjenkinson: LGTM - did you verify that the topLevelCommit does affect a rollback in this case?

mmusgrov: I did not but if it didn't then that would be a separate bug.

mmusgrov: I did not but if it didn't then that would be a separate bug.

mmusgrov: Are you wanting more?

tomjenkinson: If you have a test case that is triggering this behaviour I guessed you would have seen the rollback anyway. No need to double check if not.

mmusgrov: We get the warning on the normal path where the commit will succeed. So to the casual onlooker it appears as if there is a problem when in fact all is well.

PR #1180, opened 2017-05-18
JBTM-2766 make sure that when enlistResource is called we register th…

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2766

!BLACKTIE !XTS !RTS !AS_TESTS NO_WIN !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB


mmusgrov:
@tomjenkinson The linked JIRA says "JBTM-2766 Implement pooling for transactional driver" but your fix does not seem to relate to pooling. Please can you comment on why this adds pooling support?

PR #1181, opened 2017-05-22
JBTM-2766 Updated to ensure when JPA is used with pooling that we onl…

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2766

This was additional checks that were discovered during the creation of a JPA quickstart: 
https://github.com/jbosstm/quickstart/pull/196

!BLACKTIE !XTS !RTS !AS_TESTS NO_WIN !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB

### DIFF
--- a/ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/DirectRecoverableConnection.java
+++ b/ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/DirectRecoverableConnection.java
@@ -246,7 +246,8 @@ public class DirectRecoverableConnection implements RecoverableXAConnection, Con
             if (_theConnection != null)
             {
                 _theConnection.close();
-                _theConnection = null;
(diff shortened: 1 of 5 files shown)


ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/DirectRecoverableConnection.java
```diff
@@ -246,7 +246,9 @@ public void closeCloseCurrentConnection() throws SQLException
             if (_theConnection != null)
             {
                 _theConnection.close();
-                _theConnection = null;
+				_theConnection = null;
+				_theXAResource = null;
```
mmusgrov: A call to
```java
reset ()
```
would have been neater here. And similarly for IndirectRecoverableConnection below.

PR #1182, opened 2017-05-29
Make sure empty credentials are not used to create new connection

gytis:
https://issues.jboss.org/browse/JBTM-2902


gytis:
I didn't add test for IndirectRecoverableConnection because it needs to setup JNDI server. Could add that if you agree with the fix.

mmusgrov:
@gytis Can you raise a JIRA for this. Also Tom is on leave so you won't get a review from him for a couple of weeks.

PR #1183, opened 2017-05-29
setRollbackOnly() should be called on any exception if rollbackOn() is empty. Not only RuntimeException.

Tibor17:
If rollbackOn() is empty or if pure @Transactional, the transaction is marked for rollback if RuntimeException was thrown. Any exception should mark transaction for rollback.

### DIFF
--- a/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
+++ b/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
@@ -69,7 +69,7 @@ public abstract class TransactionalInterceptorBase implements Serializable {
         this.userTransactionAvailable = userTransactionAvailable;
     }
 
-    public Object intercept(InvocationContext ic) throws Exception {
+    public Object intercept(InvocationContext ic) throws Throwable {
 
         final TransactionManager tm = getTransactionManager();
         final Transaction tx = tm.getTransaction();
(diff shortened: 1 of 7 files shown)


Tibor17:
The JMS has both checked (JMSException) and runtime exceptions (JMSRuntimeException).
So this exception handling would be nice to respect.
The JPA has PersistenceException which is runtime.

mmusgrov:
The JTA 1.2 spec says:

> By default checked exceptions do not result in the transactional interceptor
> marking the transaction for rollback and instances of RuntimeException and its
> subclasses do

ochaloup:
Hi @Tibor17,

thanks for the proposal. As Mike said the code change does not complement with the specification.

Having rollback being called only for `RuntimeException`s are intentional. Checked exceptions are expected to be handled by the developer thus it's up to him in case to define that they should cause the rollback being triggered. For `Error`s it's similar. If JVM generates the error (let's see OutOfMemory) it's expected to be in inconsistent state to follow with the rollback (in example of OOM we can expect on a try of the rollback another OOM will be generated). As you mentioned if your code needs to be rollback on any exception you can still use the `rollbackOn` with `Throwable.class`.

Tibor17:
@ochaloup 
@mmusgrov 
I see the text in `@Transactional` Javadoc but as I said this is in conflict with JMS.
From developer point of view whatever exception is application error and therefore nobody can say that the application is in consistent state especially if checked exception crossed the outer most transaction boundary. Although I understand that caught checked exceptions between nested interceptors would be quite problematic.

ochaloup:
@Tibor17
Ok, just bear in mind that Narayana implements JTA api and it has to comply with it. I don't think this PR could be merged to upstream.

To your point about JMS Exception - I don't think it's in conflict with JMS. JMS uses checked exception as java designed them (I mean it's up to developer how he handles it. He can denote the method to throw it to the caller, if he does so he has to think why he handle it and up to him to rethink how that change impacts transaction handling.). By my view it's rather question why java has checked exception and if it's good?

mmusgrov:
> checked exception crossed the outer most transaction boundary

Are you suggesting that if the application does not catch his own business exceptions then it is the application servers responsibility to clean up? The standards say 

> When an enterprise bean throws an application exception, the container does not wrap it in another exception. The client should be able to handle any application exception it receives.

and therefore it is not spec compliant for the developer to allow a checked exceptions to "cross the outer most transaction boundary"

PR #1184, opened 2017-05-31
JBTM-2901 skip recovery pass 2 if it has already ran

mmusgrov:
https://issues.jboss.org/browse/JBTM-2901

!BLACKTIE !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !PERF !RTS

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
@@ -211,6 +211,11 @@ public class XARecoveryModule implements RecoveryModule
 
 	public synchronized void periodicWorkSecondPass()
 	{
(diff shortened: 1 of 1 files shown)


gytis:
There is still something missing. XARecoveryModule second pass invoked by PeriodicRecovery comes from an IDLE state and is ignored. Let me try to put together a reproducer which is easier to run, so I could share it with you.

mmusgrov:
Yes but those other narayana changes relate to a different PR: https://github.com/jbosstm/narayana/pull/1182

I am asking you to just review the change we discussed via email.

mmusgrov:
If is ok because the following sequence of events has occurred:

pass 1 is called (by periodic recovery) and then something else calls getNewXAResource which calls pass 2. Now when periodic recovery calls pass 2 the XARecoveryModule says, oh I have just done pass 2 so I will return without doing anything.

gytis:
OK understood. I'll look into updating Spring Boot integration to function with this change.

PR #1185, opened 2017-06-05
[JBTM-2903] Update connection management in JMS integration

gytis:
!BLACKTIE !PERF !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB

https://issues.jboss.org/browse/JBTM-2903

Similar fix will be needed for JDBC side as well, but the code for that lives in Spring Boot repository.

### DIFF
--- a/ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/JmsXAResourceRecoveryHelper.java
+++ b/ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/JmsXAResourceRecoveryHelper.java
@@ -21,10 +21,7 @@
  */
 package org.jboss.narayana.jta.jms;
 
-import javax.jms.JMSException;
-import javax.jms.XAConnection;
 import javax.jms.XAConnectionFactory;
-import javax.jms.XASession;
 import javax.transaction.xa.XAException;
 import javax.transaction.xa.XAResource;
 import javax.transaction.xa.Xid;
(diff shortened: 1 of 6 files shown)


ArjunaJTA/jms/src/main/java/org/jboss/narayana/jta/jms/JmsXAResourceRecoveryHelper.java
```diff
@@ -82,31 +71,31 @@ public boolean initialise(String properties) {
...
-    public Xid[] recover(int i) throws XAException {
+    public Xid[] recover(int flag) throws XAException {
         try {
-            return delegate.recover(i);
+            return connectionManager.connectAndApply(delegate -> delegate.recover(flag));
         } finally {
-            if (i == XAResource.TMENDRSCAN) {
-                disconnect();
+            if (flag == XAResource.TMENDRSCAN) {
+                connectionManager.disconnect();
```
tomjenkinson: I wonder if there is any real benefit to closing the connection at the ENDRSCAN anyway? Reason being that an open idle connection should not take much resource compared to open/closing for the cycle. But I am OK with the change if it addresses the problem. 

I think https://issues.jboss.org/browse/JBTM-2903 should be updated with a stack trace showing an exception coming out of commit though.

tomjenkinson: @gytis there were a couple of queries outstanding on this, please can you take a look?

gytis: Sorry for a delay, I was away last week.
@mmusgrov it used to be closed at TMENDRSCAN before, so nothing has changed.
@tomjenkinson, we did discuss it the first time I've implemented this and we end up deciding to open connection on getXAResources and closing it on TMENDRSCAN. I'm fine with keeping connection open as long as it is an acceptable practice. It would make implementation easier too.

tomjenkinson: Based on your answer I am ok with the change so marking as approved. Is it possible to capture the original stacktrace in the JBTM report?

gytis: Just to keep it clear. I don't object on keeping connection open. If you think that this is an acceptable way of handling this, I would be happy with it.
I'll try to find a stack trace and will add it to the JIRA issue.

tomjenkinson: From my side, I don't object to *not* keeping the connection open either. i.e. Keeping the connection open should not be necessary IMO. If you would like to keep the connection open I don't object and if you want to close it I don't object :)

gytis: OK, let's just keep it as is if you agree then. I'll put together a similar change for JDBC.

PR #1187, opened 2017-06-15
[JBTM-2907] doRecover for JTS XATerminator implementation

ochaloup:
https://issues.jboss.org/browse/JBTM-2907
https://issues.jboss.org/browse/JBEAP-11528

!BLACKTIE !XTS !PERF !RTS

### DIFF
--- a/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/XATerminatorImple.java
+++ b/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/XATerminatorImple.java
@@ -59,6 +59,7 @@ import com.arjuna.ats.internal.jta.resources.spi.XATerminatorExtensions;
 import com.arjuna.ats.internal.jta.transaction.arjunacore.jca.SubordinateTransaction;
 import com.arjuna.ats.internal.jta.transaction.arjunacore.jca.SubordinationManager;
 import com.arjuna.ats.internal.jta.transaction.arjunacore.jca.TransactionImporter;
+import com.arjuna.ats.internal.jta.transaction.jts.subordinate.jca.SubordinateAtomicTransaction;
 import com.arjuna.ats.internal.jta.transaction.jts.subordinate.jca.TransactionImple;
 import com.arjuna.ats.internal.jta.transaction.jts.subordinate.jca.coordinator.ServerTransaction;
 import com.arjuna.ats.internal.jta.utils.jtaxLogger;
@@ -66,6 +67,8 @@ import com.arjuna.ats.jta.exceptions.UnexpectedConditionException;
 import com.arjuna.ats.jta.xa.XidImple;
 import org.jboss.tm.ExtendedJBossXATerminator;
 import org.jboss.tm.TransactionImportResult;
+import com.arjuna.ats.jts.extensions.Arjuna;
+import com.arjuna.ats.jts.logging.jtsLogger;
 
 /**
  * The XATerminator implementation.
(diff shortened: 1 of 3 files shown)


ochaloup:
@tomjenkinson : per email discussion this is intention to implement JTS XATerminator `doRecover` method, can you please check if this implemtation is ok from your point of view?

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/XATerminatorImple.java
```diff
@@ -506,9 +441,107 @@ public void removeImportedTransaction(Xid xid) throws XAException {
...
+                    if (uid.notEquals(Uid.nullUid()))
+                    {
+                        if (parentNodeName != null)
+                        {
+                            // JTS uid does not encode parent node name thus we can't filter on it, see JBTM-2124
+                            throw new IllegalArgumentException("JTS doRecover method does not handle parentNodeName, see JBTM-2124");
+                        }
+                        else if (xid == null)
+                        {
+                            Transaction tx = SubordinationManager.getTransactionImporter().recoverTransaction(uid);
```
tomjenkinson: I think this needs to IllegalArgumentException too?

tomjenkinson: hang on, I meant that for the next clause, i.e. where xid != null

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/XATerminatorImple.java
```diff
@@ -506,9 +441,107 @@ public void removeImportedTransaction(Xid xid) throws XAException {
...
+
+                            if (tx != null)
+                                values.push(tx);
+                        }
+                        else // parentNodeName == null && tx != null
+                        {
+                            SubordinateAtomicTransaction sat = new SubordinateAtomicTransaction(uid);
```
tomjenkinson: I think xid != null is an error isn't it?

ochaloup: you mean that the comment is incorrect? (agree)
or the handling of the case should not be implemented and just throwing exception?

ochaloup:
@tomjenkinson : thanks, true (#1188). I just borrowed your fix for testing this PR.

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/jca/XATerminatorImple.java
tomjenkinson: I was think it is fine to throw the exception because iiuc it would never be called except with nulls?

ochaloup: I see, I was thinking to provide functionality that  `doRecover` method expects. but I agree that's not used anywhere. I'm going to remove that part and just throw the exception

ochaloup:
@tomjenkinson after your fix applied the test passes. This PR nows permits only call of `doRecover(null, null)` otherwise `NotSupportedException` is thrown (for not breaking backward compatibility)

ochaloup:
@tomjenkinson  Yes, I would like. I planned to open PR for 5.5 when you approve this one.

The other fix that should be in the realease is fix for the https://issues.jboss.org/browse/JBEAP-11154. Then I think it would be beneficial to have there fix https://issues.jboss.org/browse/JBTM-2905, what do you think?

PR #1192, opened 2017-06-16
[JBTM-2908] fix for moving jts jca transaction to assume completed

ochaloup:
https://issues.jboss.org/browse/JBTM-2908
https://issues.jboss.org/browse/JBEAP-11154

!BLACKTIE !XTS !PERF !RTS !MAIN !QA_JTA !QA_JTS_JDKORB NO_WIN

PR #1194, opened 2017-06-20
[JBTM-2908] fix for moving jts jca transaction to assume completed

ochaloup:
https://issues.jboss.org/browse/JBTM-2908
https://issues.jboss.org/browse/JBEAP-11154
https://github.com/jbosstm/narayana/pull/1192

!BLACKTIE !XTS !PERF !RTS


ochaloup:
@tomjenkinson I was thinking that could be so, just walking around to confirm that assumption. Thank you!

PR #1199, opened 2017-07-07
Disable dynamic1PC during prepare for subordinates

tomjenkinson:
Was introduced here: 
https://issues.jboss.org/browse/JBTM-1511
https://github.com/jbosstm/narayana/commit/d213dae6dc7a657cdfe4430ebad1ffd19e9f5607


mmusgrov:
@tomjenkinson What is the JBTM for this?

PR #1201, opened 2017-07-13
dynamic1PC

tomjenkinson:
Was introduced here:
https://issues.jboss.org/browse/JBTM-1511
d213dae

The error I am protecting against is:
subordinate transaction called with doPrepare
subordinate transaction has 2 x resources
1st resource returns XARD_ONLY
must avoid the doPrepare call proceeding on and commiting the other resource

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
@@ -2605,7 +2605,7 @@ public class BasicAction extends StateManager
              * intentions list.
              */
             
-            if (keepGoing && TxControl.dynamic1PC)
+            if (!subordinate && keepGoing && TxControl.dynamic1PC)
             {
                 /*
                  * If N-1 returned read-only and 1 record left then exit prepare now and force
@@ -3732,6 +3732,8 @@ public class BasicAction extends StateManager
 
     //    private Mutex _lock = new Mutex(); // TODO
     private List<Throwable> deferredThrowables = new ArrayList<>();
+
+    protected boolean subordinate;
     
 }
 
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/XATerminatorImple.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/XATerminatorImple.java
@@ -241,6 +241,12 @@ public class XATerminatorImple implements javax.resource.spi.XATerminator, XATer
 
             switch (tx.doPrepare())
             {
+            case TwoPhaseOutcome.PREPARE_ONE_PHASE_COMMITTED:
+                // Should never happen Would be great if there was heuristic commit :(
+                SubordinationManager.getTransactionImporter()
+                        .removeImportedTransaction(xid);
+                jtaLogger.i18NLogger.fatalSubordinate1PCDuringPrepare(xid);
+                throw new XAException(XAException.XAER_RMERR);
             case TwoPhaseOutcome.PREPARE_READONLY:
                 SubordinationManager.getTransactionImporter()
                         .removeImportedTransaction(xid);
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/subordinate/SubordinateAtomicAction.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/subordinate/SubordinateAtomicAction.java
@@ -50,11 +50,13 @@ public class SubordinateAtomicAction extends
 	public SubordinateAtomicAction ()
 	{
 		this(AtomicAction.NO_TIMEOUT);
+		subordinate = true;
 	}
 
 	public SubordinateAtomicAction (int timeout)
 	{
 		super();
+		subordinate = true;
 
 		start();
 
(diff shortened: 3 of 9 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
```diff
@@ -3732,6 +3732,8 @@ else if (record.value() instanceof ExceptionDeferrer)
 
     //    private Mutex _lock = new Mutex(); // TODO
     private List<Throwable> deferredThrowables = new ArrayList<>();
+
+    protected boolean dynamic1PC = TxControl.dynamic1PC;
```
mmusgrov: I know TxControl caches the initial value of dynamic1PC but doing this is another layer of caching which makes it harder in the future to make dynamic1PC changeable at runtime.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/XATerminatorImple.java
```diff
@@ -241,6 +241,12 @@ public int prepare (Xid xid) throws XAException
 
             switch (tx.doPrepare())
             {
+            case TwoPhaseOutcome.PREPARE_ONE_PHASE_COMMITTED:
+                // Should never happen Would be great if there was heuristic commit :(
+                SubordinationManager.getTransactionImporter()
+                        .removeImportedTransaction(xid);
+                jtaLogger.i18NLogger.fatalSubordinate1PCDuringPrepare(xid);
+                throw new XAException(XAException.XAER_RMERR);
```
mmusgrov: My understanding here is that there was only one resource on the pending list and it returned TwoPhaseOutcome.PREPARE_READONLY during prepare. Hence I would not expect an error outcome.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/subordinate/SubordinateAtomicAction.java
```diff
@@ -55,6 +55,7 @@ public SubordinateAtomicAction ()
 	public SubordinateAtomicAction (int timeout)
 	{
 		super();
+		dynamic1PC = false;
```
mmusgrov: Why do you disable the optimisation for subordinates and for ServerTransaction?

ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/jtaI18NLogger.java
```diff
@@ -470,6 +473,10 @@
     @LogMessage(level = WARN)
     public void warn_could_not_access_object_store(@Cause() Exception e);
 
+	@Message(id = 16116, value = "Subordinate ransaction was committed during prepare, this will look like a rollback {0}", format = MESSAGE_FORMAT)
```
mmusgrov: typo

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/XATerminatorImple.java
tomjenkinson: I updated the description with some wording to describe this. There are two resources. If you look at the test case you can see what I am protecting against, if you removed the assignment to false of dynamic1PC from SubordinateAtomicAction  then it wouldn't pass actually

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/subordinate/SubordinateAtomicAction.java
tomjenkinson: Because ServerTransaction is a subordinateTransaction that can have doPrepare called on it the same as SubordinateAtomicAction

ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/jtaI18NLogger.java
tomjenkinson: thanks, will updated

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
tomjenkinson: The thing is, we need to make sure subordinates don't employ this facility - I guess we could introduce a boolean subordinate flag that is also checked?

tomjenkinson: I made that change

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/XATerminatorImple.java
tomjenkinson: Unfortunately RMERR is not perfect either at it suggests it couldn't prepare rather than a heuristic commit of the second resource. With my patch this outcome would hopefully never be seen anyway.

mmusgrov: OK so the normal behaviour for this optimisation is to commit the last participant immediately (if the others all returned READONLY) but that such behaviour is wrong for subordinates.

So you throw this exception is in case you failed to disable the optimisation for subordinates, ie is it just defensive programming in case the logic that disables the optimisation breaks?

tomjenkinson: Yeah, or maybe if someone erroneously implemented a new SubordinateTransaction type that overrode the behaviour?

PR #1202, opened 2017-07-17
Allow an override so we attempt to override for all connections

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2918
https://issues.jboss.org/browse/JBTM-2919
!XTS !BLACKTIE !PERF !RTS NO_WIN


sutaakar:
Looks good.

sutaakar:
@tomjenkinson The PR seems to work for Kie server tests on MSSQL 2014. I tried it just on one test, will run full DB matrix when merged and available in Kie server war.

PR #1204, opened 2017-08-03
[JBTM-2921] commit-one-phase to be invoked REST-AT

ochaloup:
https://issues.jboss.org/browse/JBTM-2921
in case of the two phase unaware participant when optional link header
with link of 'commit-one-phase' is proposed it is expected to be called
during the transaction manager commit phase

!QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !PERF


ochaloup:
Hi @mmusgrov , would you be so kind and check this PR? My point is the REST-AT does not invoke commit-one-phase link in 1PC processing for two phase unaware participant. There is call always just `commit` and the `commit-one-phase` is not invoked at all. This is a fix for follow the rest-at spec behavior. Thanks.

ochaloup:
thanks!

PR #1206, opened 2017-09-01
Added a trace statement to help identify the modifier in use

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2924 
!PERF !QA_JTA !BLACKTIE !XTS !RTS !AS_TESTS !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB

### DIFF
--- a/ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/drivers/modifiers/ModifierFactory.java
+++ b/ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/drivers/modifiers/ModifierFactory.java
@@ -35,6 +35,7 @@ import java.util.Enumeration;
 import java.util.Hashtable;
 
 import com.arjuna.ats.jdbc.common.jdbcPropertyManager;
+import com.arjuna.ats.jdbc.logging.jdbcLogger;
 import com.arjuna.common.internal.util.ClassloadingUtility;
 
 /**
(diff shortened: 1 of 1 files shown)


ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/drivers/modifiers/ModifierFactory.java
```diff
@@ -85,15 +86,19 @@ public static synchronized ConnectionModifier getModifier (String dbName, int ma
...
-		return _modifiers.get(driverMatch);
+		modifier = _modifiers.get(driverMatch);
+	if (majorMatch != null)
+		modifier = _modifiers.get(majorMatch);
+	if (exactMatch != null)
+		modifier = _modifiers.get(exactMatch);
 
-	return defaultIsSameRMOverride ? isSameRMModifier : null;
+	if (jdbcLogger.logger.isTraceEnabled()) {
+		jdbcLogger.logger.trace("ConnectionModifier for: " + modifier.getClass().getName() + " for " + dbName + " " + major + "/" + minor + " (defaultIsSameRMOverride was " + defaultIsSameRMOverride + ")");
```
ochaloup: I think it would be nicer to use method `tracef` or `tracev` here.

ochaloup: I think there is still possibility that `modifier` will be null. I mean when either `if` case match then it could happen and `NullPointerException` could be thrown from the `logger.trace` call.

Am I wrong?

tomjenkinson: You are not wrong - in fact you are right :) I will adjust

PR #1211, opened 2017-09-12
[JBTM-2927/WFLY-2485] transitive handling @Transactional on @Stereotype

ochaloup:
I finally got to a working implementation of the @Transactional @Stereotype issue. I followed the @mkouba steps:

- modify TransactionExtension - add ProcessManagedBean observer and store <Bean, AnnotatedType> map
- modify TransactionalInterceptorBase to use the AnnotatedType stored in TransactionExtension <Bean, AnnotatedType> map (the key - intercepted bean - is available)
- use BeanAttributes.getStereotypes() together with BeanManager.getStereotypeDefinition() to cover the transitive stereotype declarations
- add reflection fallback for intercepted non-contextual instances (e.g. EE components)

Now I would like ask @mkouba and @manovotn to check it. I expect the behavior needs some tuning.

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS

### DIFF
--- a/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
+++ b/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
@@ -28,8 +28,13 @@ import static java.security.AccessController.doPrivileged;
 import com.arjuna.ats.jta.common.jtaPropertyManager;
 import com.arjuna.ats.jta.logging.jtaLogger;
 
+import java.util.HashSet;
+import java.util.Set;
+
 import javax.enterprise.context.ContextNotActiveException;
 import javax.enterprise.inject.Intercepted;
+import javax.enterprise.inject.spi.AnnotatedMethod;
+import javax.enterprise.inject.spi.AnnotatedType;
 import javax.enterprise.inject.spi.Bean;
 import javax.inject.Inject;
 import javax.interceptor.InvocationContext;
@@ -45,6 +50,7 @@ import org.jboss.tm.usertx.UserTransactionOperationsProvider;
 import java.io.Serializable;
 import java.lang.annotation.Annotation;
 import java.security.PrivilegedAction;
+import com.arjuna.ats.jta.cdi.TransactionExtension;
 
 /**
  * @author paul.robinson@redhat.com 02/05/2013
@@ -56,6 +62,9 @@ public abstract class TransactionalInterceptorBase implements Serializable {
 
     @Inject
     transient javax.enterprise.inject.spi.BeanManager beanManager;
+
+    @Inject
+    private TransactionExtension extension;
 
     @Inject
     @Intercepted
@@ -84,31 +93,76 @@ public abstract class TransactionalInterceptorBase implements Serializable {
 
     protected abstract Object doIntercept(TransactionManager tm, Transaction tx, InvocationContext ic) throws Exception;
 
+    /**
+     * <p>
+     * Looking for the {@link Transactional} annotation first on the method, second on the class.
+     * <p>
+     * Method handles CDI types to cover cases where extensions are used.
+     * In case of EE container uses reflection.
+     *
+     * @param ic  invocation context of the interceptor
+     * @return instance of {@link Transactional} annotation or null
+     */
     private Transactional getTransactional(InvocationContext ic) {
-
-        Transactional transactional = ic.getMethod().getAnnotation(Transactional.class);
-        if (transactional != null) {
-            return transactional;
-        }
-
-        Class<?> targetClass = ic.getTarget().getClass();
-        transactional = targetClass.getAnnotation(Transactional.class);
-        if (transactional != null) {
-            return transactional;
-        }
-
-        // see if the target is a stereotype
-        for (Annotation annotation : interceptedBean.getBeanClass().getAnnotations()) {
-            if (beanManager.isStereotype(annotation.annotationType())) {
-                for (Annotation stereotyped : beanManager.getStereotypeDefinition(annotation.annotationType())) {
-                    if (stereotyped.annotationType().equals(Transactional.class)) {
-                        return (Transactional) stereotyped;
-                    }
+        if(interceptedBean != null) { // not-null for CDI
+            // getting annotated type and method corresponding of the intercepted bean and method
+            AnnotatedType<?> currentAnnotatedType = extension.getBeanToAnnotatedTypeMapping().get(interceptedBean);
+            AnnotatedMethod<?> currentAnnotatedMethod = null;
+            for(AnnotatedMethod<?> methodInSearch: currentAnnotatedType.getMethods()) {
+                if(methodInSearch.getJavaMember().equals(ic.getMethod())) {
+                    currentAnnotatedMethod = methodInSearch;
+                    break;
                 }
+            }
+    
+            // check existence of the stereotype on method
+            Transactional transactionalMethod = getTransactionalAnnotationRecursive(currentAnnotatedMethod.getAnnotations());
+            if(transactionalMethod != null) return transactionalMethod;
+            // stereotype recursive search, covering ones added by an extension too
+            Transactional transactionalExtension = getTransactionalAnnotationRecursive(currentAnnotatedType.getAnnotations());
+            if(transactionalExtension != null) return transactionalExtension;
+            // stereotypes already merged to one chunk by BeanAttributes.getStereotypes()
+            for(Class<? extends Annotation> stereotype: interceptedBean.getStereotypes()) {
+                Transactional transactionalAnn = stereotype.getAnnotation(Transactional.class);
+                if(transactionalAnn != null) return transactionalAnn;
+            }
+        } else { // null for EE components
+            Transactional transactional = ic.getMethod().getAnnotation(Transactional.class);
+            if (transactional != null) {
+                return transactional;
+            }
+    
+            Class<?> targetClass = ic.getTarget().getClass();
+            transactional = targetClass.getAnnotation(Transactional.class);
+            if (transactional != null) {
+                return transactional;
             }
         }
 
         throw new RuntimeException(jtaLogger.i18NLogger.get_expected_transactional_annotation());
+    }
+
+    private Transactional getTransactionalAnnotationRecursive(Annotation... annotationsOnMember) {
+        if(annotationsOnMember == null) return null;
+        Set<Class<? extends Annotation>> stereotypeAnnotations = new HashSet<>();
+
+        for(Annotation annotation: annotationsOnMember) {
+            if(annotation.annotationType().equals(Transactional.class)) {
+                return (Transactional) annotation;
+            }
+            if (beanManager.isStereotype(annotation.annotationType())) {
+                stereotypeAnnotations.add(annotation.annotationType());
+            }
+        }
+        for(Class<? extends Annotation> stereotypeAnnotation: stereotypeAnnotations) {
+            return getTransactionalAnnotationRecursive(beanManager.getStereotypeDefinition(stereotypeAnnotation));
+        }
+        return null;
+    }
+
+    private Transactional getTransactionalAnnotationRecursive(Set<Annotation> annotationsOnMember) {
+        return getTransactionalAnnotationRecursive(
+            annotationsOnMember.toArray(new Annotation[annotationsOnMember.size()]));
     }
 
     protected Object invokeInOurTx(InvocationContext ic, TransactionManager tm) throws Exception {
--- a/ArjunaJTA/cdi/tests/classes/com/hp/mwtests/ts/jta/cdi/transactional/stereotype/StereotypeWithTransactionalBean.java
+++ b/ArjunaJTA/cdi/tests/classes/com/hp/mwtests/ts/jta/cdi/transactional/stereotype/StereotypeWithTransactionalBean.java
@@ -1,0 +1,51 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2017, Red Hat Middleware LLC, and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package com.hp.mwtests.ts.jta.cdi.transactional.stereotype;
+
+import javax.annotation.Resource;
+import javax.enterprise.inject.Stereotype;
+import javax.transaction.Status;
+import javax.transaction.SystemException;
+import javax.transaction.TransactionManager;
+import javax.transaction.Transactional;
+
+@TransitiveToRequiredTransactionalNever
+public class StereotypeWithTransactionalBean {
+
+    @Resource(mappedName = "java:jboss/TransactionManager")
+    private TransactionManager txnMgr;
+
+    /**
+     * Having declared the {@link Transactional} annotation in {@link Stereotype}
+     * last in the transitive row and in the middle one too. 
+     */
+    public void process() throws SystemException, TestException {
+        // expected: @TransitiveToRequiredTransactionalNever with NEVER
+        if(Status.STATUS_NO_TRANSACTION != txnMgr.getStatus())
+            throw new TestException();
+
+        // not expected/fail:
+        throw new AssertionError("There was active transaction even NEVER is expected to be taken"
+            + " as Transactional attribute");
+    }
+}
--- a/ArjunaJTA/cdi/tests/classes/com/hp/mwtests/ts/jta/cdi/transactional/stereotype/extension/AddTransactionalAnnotationExtension.java
+++ b/ArjunaJTA/cdi/tests/classes/com/hp/mwtests/ts/jta/cdi/transactional/stereotype/extension/AddTransactionalAnnotationExtension.java
@@ -1,0 +1,41 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2017, Red Hat Middleware LLC, and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package com.hp.mwtests.ts.jta.cdi.transactional.stereotype.extension;
+
+import javax.enterprise.event.Observes;
+import javax.enterprise.inject.spi.Extension;
+import javax.enterprise.inject.spi.ProcessAnnotatedType;
+import javax.transaction.Transactional;
+
+
+/**
+ * Extension adding the {@link Transactional} at any processed bean.
+ */
+public class AddTransactionalAnnotationExtension implements Extension {
+
+    void processAnnotatedType(@Observes ProcessAnnotatedType<NoAnnotationBean> bean) {
+        AnnotatedTypeWrapper<NoAnnotationBean> wrapper = new AnnotatedTypeWrapper<NoAnnotationBean>(bean);
+        wrapper.addAnnotation(new TransactionalLiteral());
+        bean.setAnnotatedType(wrapper);
+    }
+}
--- a/ArjunaJTA/cdi/tests/classes/com/hp/mwtests/ts/jta/cdi/transactional/stereotype/extension/AnnotatedTypeWrapper.java
+++ b/ArjunaJTA/cdi/tests/classes/com/hp/mwtests/ts/jta/cdi/transactional/stereotype/extension/AnnotatedTypeWrapper.java
@@ -1,0 +1,119 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2017, Red Hat Middleware LLC, and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package com.hp.mwtests.ts.jta.cdi.transactional.stereotype.extension;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.util.HashSet;
+import java.util.Set;
+
+import javax.enterprise.inject.spi.AnnotatedConstructor;
+import javax.enterprise.inject.spi.AnnotatedField;
+import javax.enterprise.inject.spi.AnnotatedMethod;
+import javax.enterprise.inject.spi.AnnotatedType;
+import javax.enterprise.inject.spi.ProcessAnnotatedType;
+
+public class AnnotatedTypeWrapper<T> implements AnnotatedType<T> {
+
+    private final AnnotatedType<T> wrapped;
+    private final Set<Annotation> annotations;
+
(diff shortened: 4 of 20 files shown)


ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
```diff
@@ -97,18 +120,55 @@ private Transactional getTransactional(InvocationContext ic) {
...
+            if(annotation.annotationType().equals(Transactional.class)) {
+                return (Transactional) annotation;
+            }
             if (beanManager.isStereotype(annotation.annotationType())) {
-                for (Annotation stereotyped : beanManager.getStereotypeDefinition(annotation.annotationType())) {
-                    if (stereotyped.annotationType().equals(Transactional.class)) {
-                        return (Transactional) stereotyped;
-                    }
-                }
+                stereotypeAnnotation = annotation.annotationType();
```
manovotn: I think that if there is more than one stereotype on the annotation (which is allowed), then this will override the variable and you will only inspect one of them which is incorrect.

ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
```diff
@@ -97,18 +120,55 @@ private Transactional getTransactional(InvocationContext ic) {
...
+        AnnotatedType<?> annTypeBean = extension.getBeanToAnnotatedTypeMapping().get(interceptedBean);
+        Transactional transactionalExtension = getTransactionalAnnotationRecursive(annTypeBean.getAnnotations());
+        if(transactionalExtension != null) return transactionalExtension;
+        // stereotype taken from intercepted bean getStereotype
+        Transactional transactionalBasedOnStereotype = null;
+        for(Class<? extends Annotation> stereotype: interceptedBean.getStereotypes()) {
+            transactionalBasedOnStereotype = stereotype.getAnnotation(Transactional.class);
+            if(transactionalBasedOnStereotype != null) return transactionalBasedOnStereotype;
+        }
+        // stereotype taken from intercepted bean getStereotypeDefinition
```
manovotn: What's the use of this part of code? I think it basically duplicates job done in the above part. It should be sufficient to go through the `AnnotatedType`.

The only case I can think of is when the actual class is annotated by `@Transactional` but you deliberately remove this annotation from `AnnotatedType` - however, in this case I am not sure you actually want to intercept it. I'd wager you don't.

ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
```diff
@@ -97,18 +120,55 @@ private Transactional getTransactional(InvocationContext ic) {
             return transactional;
         }
 
-        // see if the target is a stereotype
-        for (Annotation annotation : interceptedBean.getBeanClass().getAnnotations()) {
+        // check existence of the stereotype on method
+        Transactional transactionalMethod = getTransactionalAnnotationRecursive(ic.getMethod().getAnnotations());
```
manovotn: You can make use of `AnnotatedMethod` here -> `AnnotatedType<T>#getMethods()` and filter it. This will make sure you have the actual annotations as people might have changed this via extensions (as they can with whole AT).

mkouba:
I think the correct approach would be:
1. If `interceptedBean != null` (would be null for EE components, such as servlets) obtain the `AnnotatedType` and if available then use the `Annotated` SPI to find relevant `@Transactional` on the method, type and stereotypes:
   - use `javax.enterprise.inject.spi.AnnotatedType.getMethods()` and `javax.enterprise.inject.spi.AnnotatedMethod.getJavaMember()` to find the relevant `AnnotatedMethod`
   - use `BeanAttributes.getStereotypes()` and `BeanManager.getStereotypeDefinition(Class<? extends Annotation>)` to obtain the stereotype annotations, and recursively if `BeanManager.isStereotype(Class<? extends Annotation>)` returns true
2. If `interceptedBean == null` use reflection to find the `@Transactional` (there is no need to handle stereotypes for EE components)
   - https://github.com/jbosstm/narayana/blob/master/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java#L89-L98

Also few more tests would be really useful.

ochaloup:
@manovotn @mkouba perfect, thanks a lot for the review, I will do my best to fix that based on your points.

As a side note: I'm aware of lack of tests 😄 and I'm planning to add them. Here I wanted to know what you would say looking at the implementation.

ochaloup:
@mkouba @manovotn : hi, I tried to make changes you talked about in you comment. Would you be so kind and recheck this PR?

There are two things which I would like to check with you

- if there is a Stereotype put at the class level and this stereotype defines the @Transactional annotation transitively where both uses it (first stereotype contains the @Transactional ann, the second too) then the container fails to deploy with error
```
org.jboss.arquillian.container.spi.client.container.DeploymentException: Cannot deploy: stereotype-test.war
Caused by: java.lang.Exception: 
{"WFLYCTL0080: Failed services" => {"jboss.deployment.unit.\"stereotype-test.war\".WeldStartService" => "Failed to start service
    Caused by: org.jboss.weld.exceptions.DefinitionException: WELD-000048: Conflicting interceptor bindings found on class com.hp.mwtests.ts.jta.cdi.transactional.stereotype.StereotypeWithTransactionalBean2"}}
```
It seems to me reasonable just wanted to check if it's how it should work?

- the second thing is not working transitive @Stereotype annotations put at the method level. I have test where there are two stereotypes defining both @Transactional annotation. The Weld does not uses the transitve stereotype at the method level. There is used one configured at class level but it should be (by my knowledge) overriden by method annotations. Do you have a similar experience how the Weld works?

Thanks

mkouba:
> if there is a Stereotype put at the class level and this stereotype defines...

I believe this is undefined. Interceptors spec only states that a method-level binding replaces an interceptor binding of the same type declared at class level (or inherited from superclass). The current behavior makes sense imho but it might be worth a CDI clarification issue. @manovotn WDYT?

> the second thing is not working transitive @stereotype annotations put at the method level...

Stereotypes may only be applied to a bean class or producer method or field. But I would rather see the test if possible ;-)

manovotn:
> if there is a Stereotype put at the class level and this stereotype defines the @transactional

Agree that this is _undefined_ - the behaviour makes sense though, if you have annotations with members (as you do here), failing is better than magically choosing one value. That being said, I think Weld doesn't fail if those annotations don't have any members. Would need to check though.
@mkouba yea, we might as well create one

> the second thing is not working transitive @stereotype annotations put at the method level...

Like Martin says, this is forbidden by spec - see [2.8.2 Declaring Stereotypes chapter](http://docs.jboss.org/cdi/spec/1.2/cdi-spec.html#declaring_stereotypes)

ochaloup:
@mkouba @manovotn : ok, perfect, thanks for your reply. I was led by the fact that `@Transactional` is expected to be at class and method level. I somehow expected the `@Stereotype` goes the same. I will change the tests to follow your words. Plus I need to fix the compilation failures.

ochaloup:
@mkouba @manovotn @mmusgrov : I made the changes as proposed, fixed compilation (aka javadoc creation) failures. What do you think about the fix now? Is it ok to be merged to Narayana codebase?

ArjunaJTA/cdi/tests/classes/com/hp/mwtests/ts/jta/cdi/transactional/stereotype/StereotypeAtMethodBean.java
```diff
@@ -0,0 +1,59 @@
...
+import javax.annotation.Resource;
+import javax.enterprise.inject.Stereotype;
+import javax.transaction.Status;
+import javax.transaction.SystemException;
+import javax.transaction.TransactionManager;
+import javax.transaction.Transactional;
+import javax.transaction.Transactional.TxType;
+
+@TransactionalRequiredStereotype
+public class StereotypeAtMethodBean {
```
manovotn: I'd probably change the name of this bean - you don't use stereotype at method level.

ArjunaJTA/cdi/tests/classes/com/hp/mwtests/ts/jta/cdi/transactional/stereotype/StereotypeWithTransactionalBean.java
```diff
@@ -0,0 +1,51 @@
...
+package com.hp.mwtests.ts.jta.cdi.transactional.stereotype;
+
+import javax.annotation.Resource;
+import javax.enterprise.inject.Stereotype;
+import javax.transaction.Status;
+import javax.transaction.SystemException;
+import javax.transaction.TransactionManager;
+import javax.transaction.Transactional;
+
+// @TransitiveToRequiredTransactionalNever
```
manovotn: Should this been commented-out even when the test is skipped?

ArjunaJTA/cdi/tests/classes/com/hp/mwtests/ts/jta/cdi/transactional/stereotype/extension/NoStereotypeBean.java
```diff
@@ -0,0 +1,27 @@
...
+    private TransactionManager txnMgr;
+
+    /**
+     * Expecting the {@link AnnotationProcessingExtension} adds the stereotype
+     * over the bean to get the {@link Transactional} defined by {@link Stereotype}
+     * added programmatically in extension.
+     */
+    public void process() throws SystemException {
+        if(Status.STATUS_ACTIVE == txnMgr.getStatus())
+        	throw new RuntimeException("expected status active");
```
manovotn: Testing this by means of forcible exception feels kind of wrong - wouldn't it be "cleaner" to just use a boolean and set it to `true` when there is `STATUS_ACTIVE`?

mkouba:
Looks good now. Do you also have a test which adds `@Transactional` using an extension (similar to `AnnotationProcessingExtension`)? I believe a similar test should be part of the JTA TCK but obviously it is not.

ArjunaJTA/cdi/tests/classes/com/hp/mwtests/ts/jta/cdi/transactional/stereotype/StereotypeAtMethodBean.java
ochaloup: ah, correct, thanks

ArjunaJTA/cdi/tests/classes/com/hp/mwtests/ts/jta/cdi/transactional/stereotype/StereotypeWithTransactionalBean.java
ochaloup: a good point, thanks

ArjunaJTA/cdi/tests/classes/com/hp/mwtests/ts/jta/cdi/transactional/stereotype/extension/NoStereotypeBean.java
ochaloup: No, exception is needed to be thrown here as then the interceptor starts its handling and I need to test the behaviour of the transactional interceptor in such case. I will add there some comment to say it's for testing purposes.

ochaloup:
@mkouba hm... I'm cycling with not being able to finish this PR aka I have a trouble with the test where `@Transactional` is added by extension. When I run with the `@Stereotype` it works fine.

I use the extension where a generic wrapper is defined:
https://github.com/jbosstm/narayana/pull/1211/files#diff-1ead56674a1b9cbc7fc821ad5719c78aR37

and with Transacitonal I  get exception
```
[0m[0m13:32:21,818 INFO  [org.jboss.as.clustering.infinispan] (ServerService Thread Pool -- 62) WFLYCLINF0002: Started client-mappings cache from ejb container
[0m[31m13:32:21,996 ERROR [org.jboss.msc.service.fail] (MSC service thread 1-1) MSC000001: Failed to start service jboss.deployment.unit."stereotype-test.war".WeldStartService: org.jboss.msc.service.StartException in service jboss.deployment.unit."stereotype-test.war".WeldStartService: Failed to start service
	at org.jboss.msc.service.ServiceControllerImpl$StartTask.run(ServiceControllerImpl.java:1978)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: org.jboss.weld.exceptions.WeldException: WELD-001602: Cannot create qualifier instance model for com.hp.mwtests.ts.jta.cdi.transactional.stereotype.extension.AnnotatedTypeWrapper$1@68695819
	at javax.transaction.Transactional.value(Transactional.java:0)
  StackTrace:
	at org.jboss.weld.resolution.QualifierInstance.createValues(QualifierInstance.java:137)
	at org.jboss.weld.resolution.QualifierInstance.of(QualifierInstance.java:94)
	at org.jboss.weld.resolution.ResolvableBuilder.addQualifier(ResolvableBuilder.java:148)
	at org.jboss.weld.resolution.ResolvableBuilder.addQualifiers(ResolvableBuilder.java:198)
	at org.jboss.weld.resolution.ResolvableBuilder.addQualifiers(ResolvableBuilder.java:193)
	at org.jboss.weld.resolution.InterceptorResolvableBuilder.addQualifiers(InterceptorResolvableBuilder.java:71)
	at org.jboss.weld.manager.BeanManagerImpl.resolveInterceptors(BeanManagerImpl.java:951)
	at org.jboss.weld.injection.producer.InterceptionModelInitializer.initLifeCycleInterceptor(InterceptionModelInitializer.java:168)
	at org.jboss.weld.injection.producer.InterceptionModelInitializer.initCdiLifecycleInterceptors(InterceptionModelInitializer.java:161)
	at org.jboss.weld.injection.producer.InterceptionModelInitializer.initCdiInterceptors(InterceptionModelInitializer.java:143)
	at org.jboss.weld.injection.producer.InterceptionModelInitializer.init(InterceptionModelInitializer.java:110)
	at org.jboss.weld.injection.producer.BeanInjectionTarget.buildInterceptionModel(BeanInjectionTarget.java:94)
	at org.jboss.weld.injection.producer.BeanInjectionTarget.initializeInterceptionModel(BeanInjectionTarget.java:89)
	at org.jboss.weld.injection.producer.BeanInjectionTarget.initializeAfterBeanDiscovery(BeanInjectionTarget.java:99)
	at org.jboss.weld.injection.producer.InjectionTargetInitializationContext.initialize(InjectionTargetInitializationContext.java:42)
	at org.jboss.weld.injection.producer.InjectionTargetService.initialize(InjectionTargetService.java:63)
	at org.jboss.weld.bootstrap.WeldStartup.deployBeans(WeldStartup.java:463)
	at org.jboss.weld.bootstrap.WeldBootstrap.deployBeans(WeldBootstrap.java:83)
	at org.jboss.as.weld.WeldStartService.start(WeldStartService.java:95)
	at org.jboss.msc.service.ServiceControllerImpl$StartTask.startService(ServiceControllerImpl.java:2032)
	at org.jboss.msc.service.ServiceControllerImpl$StartTask.run(ServiceControllerImpl.java:1955)
	... 3 more
Caused by: java.lang.IllegalArgumentException: object is not an instance of declaring class
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.jboss.weld.resolution.QualifierInstance.createValues(QualifierInstance.java:135)
	... 23 more
```

I tried to debug that a little bit and I can see that `org.jboss.weld.resolution.QualifierInstance` uses as `instance`: https://github.com/weld/core/blob/master/impl/src/main/java/org/jboss/weld/resolution/QualifierInstance.java#L135 my wrapper class (https://github.com/jbosstm/narayana/pull/1211/files#diff-f6468c72b31839a047713877a143ba21R36) but the `method` counts with instance of `Transactional` annotation object. As said when `Stereotype` is used all works fine.

Would you have some idea for me how to fix the annotation processing code, please?

Thanks in advance
Ondra

mkouba:
@ochaloup The problem is you're using your own `Annotation` instance (https://github.com/jbosstm/narayana/pull/1211/files#diff-f6468c72b31839a047713877a143ba21R55) which is not usable. You should use `javax.enterprise.util.AnnotationLiteral` instead. See for example https://github.com/weld/core/blob/master/tests-arquillian/src/test/java/org/jboss/weld/tests/interceptors/producer/InterceptorsLiteral.java.

ochaloup:
@mkouba thanks and crap :/ this is a kind of magic :) I followed the `InterceptorLiteral` but I got

```
ERROR [org.jboss.msc.service.fail] (MSC service thread 1-4) MSC000001: Failed to start service jboss.deployment.unit."transactional-test.war".WeldStartService: org.jboss.msc.service.StartException in service jboss.deployment.unit."transactional-test.war".WeldStartService: Failed to start service
	at org.jboss.msc.service.ServiceControllerImpl$StartTask.run(ServiceControllerImpl.java:1978)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: org.jboss.weld.exceptions.WeldException: WELD-001508: Cannot create an InjectionTarget from [EnhancedAnnotatedTypeImpl] public abstract interface @Inherited @InterceptorBinding @Retention @Target class javax.transaction.Transactional as it is an interface
	at org.jboss.weld.util.InjectionTargets.createNonProducibleInjectionTarget(InjectionTargets.java:82)
	at org.jboss.weld.injection.producer.WeldInjectionTargetBuilderImpl.buildInternal(WeldInjectionTargetBuilderImpl.java:107)
	at org.jboss.weld.injection.producer.WeldInjectionTargetBuilderImpl.run(WeldInjectionTargetBuilderImpl.java:98)
	at org.jboss.weld.injection.producer.WeldInjectionTargetBuilderImpl.build(WeldInjectionTargetBuilderImpl.java:93)
	at org.jboss.weld.interceptor.reader.PlainInterceptorFactory.of(PlainInterceptorFactory.java:50)
	at org.jboss.weld.interceptor.reader.InterceptorMetadataReader$1.apply(InterceptorMetadataReader.java:41)
	at org.jboss.weld.interceptor.reader.InterceptorMetadataReader$1.apply(InterceptorMetadataReader.java:36)
	at org.jboss.weld.util.cache.ReentrantMapBackedComputingCache$1.apply(ReentrantMapBackedComputingCache.java:55)
	at org.jboss.weld.util.cache.ReentrantMapBackedComputingCache$1.apply(ReentrantMapBackedComputingCache.java:51)
	at org.jboss.weld.util.cache.ReentrantMapBackedComputingCache.getValue(ReentrantMapBackedComputingCache.java:64)
	at org.jboss.weld.util.cache.ReentrantMapBackedComputingCache.getCastValue(ReentrantMapBackedComputingCache.java:80)
	at org.jboss.weld.interceptor.reader.InterceptorMetadataReader.getPlainInterceptorMetadata(InterceptorMetadataReader.java:61)
	at org.jboss.weld.injection.producer.InterceptionModelInitializer.initClassDeclaredEjbInterceptors(InterceptionModelInitializer.java:247)
	at org.jboss.weld.injection.producer.InterceptionModelInitializer.initEjbInterceptors(InterceptionModelInitializer.java:231)
	at org.jboss.weld.injection.producer.InterceptionModelInitializer.init(InterceptionModelInitializer.java:109)
	at org.jboss.weld.injection.producer.BeanInjectionTarget.buildInterceptionModel(BeanInjectionTarget.java:94)
	at org.jboss.weld.injection.producer.BeanInjectionTarget.initializeInterceptionModel(BeanInjectionTarget.java:89)
	at org.jboss.weld.injection.producer.BeanInjectionTarget.initializeAfterBeanDiscovery(BeanInjectionTarget.java:99)
	at org.jboss.weld.injection.producer.InjectionTargetInitializationContext.initialize(InjectionTargetInitializationContext.java:42)
	at org.jboss.weld.injection.producer.InjectionTargetService.initialize(InjectionTargetService.java:63)
	at org.jboss.weld.bootstrap.WeldStartup.deployBeans(WeldStartup.java:463)
	at org.jboss.weld.bootstrap.WeldBootstrap.deployBeans(WeldBootstrap.java:83)
	at org.jboss.as.weld.WeldStartService.start(WeldStartService.java:95)
	at org.jboss.msc.service.ServiceControllerImpl$StartTask.startService(ServiceControllerImpl.java:2032)
	at org.jboss.msc.service.ServiceControllerImpl$StartTask.run(ServiceControllerImpl.java:1955)
	... 3 more
Caused by: org.jboss.weld.exceptions.DefinitionException: WELD-001508: Cannot create an InjectionTarget from [EnhancedAnnotatedTypeImpl] public abstract interface @Inherited @InterceptorBinding @Retention @Target class javax.transaction.Transactional as it is an interface
	at org.jboss.weld.util.InjectionTargets.createNonProducibleInjectionTarget(InjectionTargets.java:65)
	... 27 more
```
thus I merged that with  https://docs.jboss.org/weld/reference/latest/en-US/html/extend.html#_wrapping_an_literal_annotatedtype_literal and I started getting exception that transaction is not available at the method call (the literal is defined with TxType.REQUIRED thus there should be a transaction available I think).

ochaloup:
@mkouba would you have for me some tip what is wrong with my test, please? :)

mkouba:
@ochaloup I can still see `wrapper.addAnnotation(Transactional.class)` in your test: https://github.com/jbosstm/narayana/pull/1211/files#diff-1ead56674a1b9cbc7fc821ad5719c78aR44. Also try to use `@Observes ProcessAnnotatedType<NoAnnotationBean> bean` instead of the `classToProcess` hack.

ochaloup:
@mkouba ok, good point with the `classToProcess`, my wrong, I didn't realize.

To the other part of your note - I think that should not be a problem - or I hope it should not be a problem. The wrapper itself works in way that it never returns the `Transactional` annotation as it is added. It will always return the Literal. The `addAnnotation` is only a `helper` method which gives me way to change the behaviour of the wrapper. What do you consider wrong at that?

mkouba:
> To the other part of your note - I think that should not be a problem ...

I see. Well, the wrapper impl looks unnecessarily complex to me. Why not use `AnnotatedTypeWrapper.addAnnotation(new TransactionalLiteral())` and avoid all the "filtering" stuff in `getAnnotations()` and `getAnnotation()`?

ArjunaJTA/cdi/tests/classes/com/hp/mwtests/ts/jta/cdi/transactional/stereotype/extension/TransactionalChangedByExtensionTest.java
```diff
@@ -0,0 +1,58 @@
...
+public class TransactionalChangedByExtensionTest {
+    @Inject
+    NoAnnotationBean bean;
+
+    @Deployment
+    public static WebArchive createTestArchive() {
+        return ShrinkWrap.create(WebArchive.class, "transactional-test.war")
+            .addPackage(TransactionalChangedByExtensionTest.class.getPackage())
+            .addClass(Transactional.class)
+            .addAsWebInfResource(
```
mkouba: Just a tip - using `ManifestContainer.addAsServiceProvider(Class<?>, Class<?>...)` is the type-safe way to specify a CDI extension.

ochaloup:
@mkouba ok, nice tip with the `addAsServiceProvider` I didn't know about that.

Right, I feel dummy. I was changing the code several time and haven't take then this simple way of adding TransactionalLiteral directly. Ok, changed, thanks.

But still there is the trouble that transaction is not active when `@Transactional` is added by the extension. Yes, I debug various part of the processing and currently I have got a strange finding that the injected bean in the test `@Inject NoAnnotationBean bean` is not a Weld proxy object but only the object of the bean. There is no interceptor binding then. I don't know what are the implementation details - like if there is no around invoke processing then Weld proxy object is not create (?). Then I would need to check the place where this is decided. Would you give me a clue?

ArjunaJTA/cdi/tests/classes/com/hp/mwtests/ts/jta/cdi/transactional/stereotype/extension/AddTransactionalAnnotationExtension.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Extension adding the {@link Transactional} at any processed bean.
+ */
+public class AddTransactionalAnnotationExtension implements Extension {
+
+    private List<Class<?>> classToProcess = Arrays.asList(new Class[] {NoAnnotationBean.class});
+
+    void processAnnotatedType(@Observes ProcessAnnotatedType<NoAnnotationBean> bean) {
+        if(classToProcess.contains(bean.getAnnotatedType().getJavaClass())) {
```
mkouba: This `if` and the list above can be removed completely.

ArjunaJTA/cdi/tests/classes/com/hp/mwtests/ts/jta/cdi/transactional/stereotype/extension/AnnotatedTypeWrapper.java
```diff
@@ -0,0 +1,113 @@
...
+            public Class<? extends Annotation> annotationType() {
+                return classAnnotation;
+            }
+        };
+        addAnnotation(annotation);
+    }
+
+    @Override
+    public <A extends Annotation> A getAnnotation(Class<A> annotationType) {
+        return wrapped.getAnnotation(annotationType);
```
mkouba: You should use the `annotations` field here too, not the `wrapped` reference.

mkouba:
For the record - we've found out that `org.jboss.as.txn.deployment.TransactionDependenciesProcessor` does not add `org.jboss.jts` and `javax.transaction.api` dependencies if there is no `@Transactional` or `@TransactionScoped` found in the deployment. A dummy class with `@Transactional` fixed the test.

ArjunaJTA/cdi/tests/classes/com/hp/mwtests/ts/jta/cdi/transactional/stereotype/extension/AddTransactionalAnnotationExtension.java
ochaloup: thanks, I omitted to remove that part of the code

ArjunaJTA/cdi/tests/classes/com/hp/mwtests/ts/jta/cdi/transactional/stereotype/extension/AnnotatedTypeWrapper.java
ochaloup: true, thanks

ochaloup:
@mkouba thanks a lot. Fixed and pushed here. I created a jira for potential documentation later as we do not have any txn CDI jta documentation (what I'm aware of) where this specific "feature" could be added to: https://issues.jboss.org/browse/JBTM-2933

ochaloup:
@mmusgrov what do you think, is it fine to be merged now?

ochaloup:
@tomjenkinson ok, I'm removing the commit with the `.gitignore` adding `/bin`. I found being an eclipse stuff. But Narayana contains bin directories to its own usage in few folders. I will check this and in case I found that makes me issues I will create another PR.

PR #1219, opened 2017-09-27
JBTM-2934 LRA nested and parent coordinators should optimise calls when in a single VM

mmusgrov:
https://issues.jboss.org/browse/JBTM-2934

When a nested LRA is started and is in the same JVM as the parent then it registers itself as a compensator with the parent using a Java (instead of JAX-RS) call to the coordinator. When the parent LRA is ended and the nested compensator is in the same JVM, then the nested compensator is ended using a direct Java invocation. 

Two points of note:

1, I did this in 2 commits to enhance readability. The first commit shows the necessary changes for the new functionality and is straight forward. The second commit (which is less clear) simply cleans up the code.
2. I believe I have caught all the parent/nested LRA interactions and performed an inVM call accordingly. However it would be great if you could verify that with your openshift demos.


mmusgrov:
@tomjenkinson Can you take my fix instead: https://github.com/jbosstm/narayana/commit/70712eb2eeeace3e3ca1b103391fbbf707d47807

mmusgrov:
I had to make changes to the same files in another pr #1226 and didn't fancy the pain of resolving conflicts so I have cherry picked this commit on that PR and am closing this one accordingly.

PR #1221, opened 2017-10-02
Parent of nested transaction is not pushed back to headers

tomjenkinson:


mmusgrov:
@tomjenkinson  I put my patch in a separate PR: #1222

PR #1222, opened 2017-10-02
JBTM-2936 JAX-RS response headers from an LRA @Nested method doesn't include the parent LRA id

mmusgrov:
#https://issues.jboss.org/browse/JBTM-2936

### DIFF
--- a/rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
+++ b/rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
@@ -259,8 +259,12 @@ public class ServerLRAFilter implements ContainerRequestFilter, ContainerRespons
         // store state with the current thread. TODO for the async version use containerRequestContext.setProperty("lra", Current.peek());
         Current.updateLRAContext(lraId, headers); // make the current LRA available to the called method
 
-        if (newLRA != null)
+        if (newLRA != null) {
+            if (suspendedLRA != null)
+                Current.putState("suspendedLRA", suspendedLRA);
+
             Current.putState("newLRA", newLRA);
+        }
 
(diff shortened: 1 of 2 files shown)


rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -259,8 +259,12 @@ public void filter(ContainerRequestContext containerRequestContext) throws IOExc
         // store state with the current thread. TODO for the async version use containerRequestContext.setProperty("lra", Current.peek());
         Current.updateLRAContext(lraId, headers); // make the current LRA available to the called method
 
-        if (newLRA != null)
+        if (newLRA != null) {
```
tomjenkinson: newLRA is always null in this case because of:
 if (isLongRunning)
            newLRA = null;

mmusgrov: Yes but in the incoming request filter on line 270 there is a call that sets the notion of current lraClient.setCurrentLRA(lraId); 

Then in the response filter there on line 366 the call Current.updateLRAContext(responseContext.getHeaders());
will puts the notion of current into the response headers:

```java
    public static void updateLRAContext(MultivaluedMap<String, Object> headers) {
        URL lraId = Current.peek();

        if (lraId != null)
            headers.putSingle(LRA_HTTP_HEADER, lraId);
        else
            headers.remove(LRA_HTTP_HEADER);
    }
```

mmusgrov:
This is passing the test for the fix so I think it is a worthwhile merging.

If there some extra functionality you want in this area we can raise a separate JIRA and address it accordingly.

@tomjenkinson Can you confirm whether this approach is satisfactory.

PR #1231, opened 2017-10-13
JBTM-2944 LRA spec updates

mmusgrov:

These changes implement the spec updates made in https://github.com/jbosstm/microprofile-sandbox/pull/5 (my plan is to raise the spec PR soon).

https://issues.jboss.org/browse/JBTM-2944

### DIFF
--- a/rts/lra/lra-client/src/main/java/io/narayana/lra/client/LRAClient.java
+++ b/rts/lra/lra-client/src/main/java/io/narayana/lra/client/LRAClient.java
@@ -22,6 +22,7 @@
 package io.narayana.lra.client;
 
 import io.narayana.lra.annotation.Compensate;
+import io.narayana.lra.annotation.CompensatorStatus;
 import io.narayana.lra.annotation.Complete;
 import io.narayana.lra.annotation.Forget;
 import io.narayana.lra.annotation.Leave;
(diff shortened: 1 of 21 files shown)


rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/participant/TerminationException.java
```diff
@@ -0,0 +1,25 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package io.narayana.lra.client.participant;
+
+public class TerminationException extends Throwable {
```
ochaloup: would not be good to add javadoc with explanation what is this exception used for? Plus I don't think it's a good idea to extend directly from `Throwable`.

rts/lra/lra-client/src/main/java/io/narayana/lra/client/LRAClient.java
```diff
@@ -234,7 +236,6 @@ public URL startLRA(String clientID) throws GenericLRAException {
         return startLRA(clientID, 0L);
```
ochaloup: I think `public` methods should have a javadoc describing a purpose of the method.

mmusgrov: This class implements LRAClientAPI and the javadoc is in the interface class - I would rather not duplicate the same javadoc so as to avoid the descriptions diverging.

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/participant/TerminationException.java
mmusgrov: Again the javadoc is in the interface LRAParticipant where it is used.

And, okay I will extend from Exception instead.

ochaloup: @mmusgrov ah, I can see now, I was asking the same question some days ago. ay... Btw. what to use in the class implementation something like

```
/**
 * {@inheritDoc} 
**/
```

mmusgrov: I've not come across that before - I will use it

PR #1232, opened 2017-10-17
JBTM-2947 LRA end calls return the status

mmusgrov:
https://issues.jboss.org/browse/JBTM-2947

The specification used to return participant completion data to the client that closed an LRA. The spec has changed and now it just returns the status of the LRA.

Enables:  jbosstm/quickstart#208


mmusgrov:
The quickstart PR is failing because it is using the master build of narayana, ie "Requires jbosstm/narayana#1232" is having no effect. Therefore I will merge this one and then retest the quickstart. Maybe I am using the wrong "Requires" syntax.

PR #1234, opened 2017-10-19
[JBTM-2948] for failed beforeCompletion failure it could happen

ochaloup:
having empty heuristics list which then fails with NPE

This issue came from the observation on customer log.

I would like present the fix of the issue but currently I do not have a reproducer. I would like to check with you about this issue and if you are okI would prepare a testcase.

https://issues.jboss.org/browse/JBTM-2948

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !PERF

### DIFF
--- a/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/xa/SynchronizationUnitTest.java
+++ b/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/xa/SynchronizationUnitTest.java
@@ -35,11 +35,15 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
 import javax.transaction.Status;
+import javax.transaction.xa.XAException;
+import javax.transaction.xa.Xid;
 
 import org.junit.Test;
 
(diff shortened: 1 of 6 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
```diff
@@ -275,6 +275,8 @@ public final boolean preventCommit ()
             //	    unlockMutex();
         }
 
+        createPreparedLists();
```
tomjenkinson: Does this truly need to be done in arjunaCore or could it be guarded in XTS alone? It is fine if it is an arjunaCore issue and we should code to fix it but I need to see an arjunacore test case that shows this or reasoning it can't be sole change in XTS

ochaloup:
@tomjenkinson @mmusgrov after investigation I humbly agree the NPE seems to be a trouble only for XTS transactions. I created the reprocucer and proposing a fix.

mmusgrov:
Thanks, it is a good fix

PR #1235, opened 2017-10-20
JBTM-2949 - Remove synchronization from ThreadUtil.getThreadId

barreiro:
https://issues.jboss.org/browse/JBTM-2949

This uses the strategy defined in the issue as number 2, to use Thread.getId().
Although the javadoc from Thread.getId() mentiones the following:

> Returns the identifier of this Thread. The thread ID is a positive long number generated when this thread was created. The thread ID is unique and remains unchanged during its lifetime. When a thread is terminated, this thread ID may be reused.

The reality is that IDs are generated from a counter, just like what ThreadUtil does now, and therefore both solutions have roughly the same behavior.

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/utils/ThreadUtil.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/utils/ThreadUtil.java
@@ -20,23 +20,11 @@
  */
 package com.arjuna.ats.arjuna.utils;
 
-import java.util.WeakHashMap;
-import java.util.concurrent.atomic.AtomicLong;
-
 /**
  * Provides utilities to manage thread ids.
  */
 public class ThreadUtil
 {
-    /**
-     * The ID associated with the thread.
-     */
-    private static final WeakHashMap<Thread,String> THREAD_ID = new WeakHashMap<Thread,String>() ;
-    /**
-     * The thread id counter.
-     */
-    private static AtomicLong id = new AtomicLong();
-    
     /**
      * Get the string ID for the current thread.
      * @return The thread id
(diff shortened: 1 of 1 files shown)


barreiro:
I understand the concerns about the uniqueness of Thread.getID(). From my research so far, no JVM reuses the IDs. Actually, most of them rely on the openJDK class library that implements IDs with a counter.

Another question is how fast does an ID being re-used would cause a problem. My assumption is that it would only be problematic if two treads with the same ID coexist within the life-cycle of a transaction.

mmusgrov:
Thanks for the call stack you put on the other PR. I checked the code and you are right we always go through the synchronized method even for top level actions. We could refactor the code to avoid that but that would be higher risk than the solution you propose.

# The JTA Case:

Your assumption is that it would only be potentially problematic if two threads with the same ID coexist within the life-cycle of a transaction. This is because we need a thread id to index into a collection of child actions associated with a top level action. If one of those child threads terminated and was reused in the context of the same transaction (unlikely but possible) then we would loose the reference to the dead thread. But I think this would be fine since we no longer need to interact with a terminated thread (except perhaps to report its status or stack).

# The JTS Case:

JTS uses ThreadUtil to lookup things like the associated child actions, the transaction propagation context, OTS checked actions etc. We also pass around the thread id as part of the IIOP payload on OTS calls for the purpose of suspending and resuming transactions as the leave and enter the JVM.

In the JTS case I find it harder to reason about all of the use cases. They are probably safe but the code is complex but we would need to reason about the impact on a per use case basis which I do not have time for at the moment.

So in conclusion I think the fix is safe for the JTA but I cannot commit to the JTS case. 

@tomjenkinson if the PR was changed to only apply to the JTA case then I would be happy for this fix to be merged but it would mean that we would need to main distinct versions of ThreadUtil.java which has a bigger impact on the codebase. My question is should we go ahead and ask Luis to make the change for JTA only

@barreiro on the other PR you said 
> Running SPEC jEnterprise 2010 with hundreds of threads and it shows that the amount of time threads spend waiting to get an ID is significant (I can quantify if needed). 
would you be able to add a note about that to the JIRA

mmusgrov:
I would be okay with (another) config option.
Would it be one option for both JTA and JTS?
Would the default be what we currently implement?

nmcl:
Are there any public API changes implied by this?

nmcl:
don't you think we have enough configurable options? It's going to get harder and harder to ensure that the cross-product combination of all configurations are QE-ed, let alone pass those tests. Furthermore, if this is really so specific to the implementation of the JVM I think it's as risky as relying on some of the sun.misc.unsafe stuff.

nmcl:
Have you discussed this with the OpenJDK team?

nmcl:
I really dislike the idea of a configurable option, particularly if it's configuring the synchronised keyword from the API.

PR #1240, opened 2017-10-30
[JBTM-2951] providers are registered twice - by annotation and by descriptor

ochaloup:
Proposing a fix for not getting WARN messages on app startup using LRACLient which normally shows duplicate warns.

https://issues.jboss.org/browse/JBTM-2951

Warning means that registration of the REST provider is done several times - only `@Provider` annotation should be enough.

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !AS_TESTS !mysql !postgres !db2 !oracle

### DIFF
--- a/rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ClientLRARequestFilter.java
+++ b/rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ClientLRARequestFilter.java
@@ -25,10 +25,8 @@ import io.narayana.lra.client.Current;
 
 import javax.ws.rs.client.ClientRequestContext;
 import javax.ws.rs.client.ClientRequestFilter;
-import javax.ws.rs.ext.Provider;
 import java.io.IOException;
 
-@Provider
 public class ClientLRARequestFilter implements ClientRequestFilter {
     @Override
     public void filter(ClientRequestContext context) throws IOException {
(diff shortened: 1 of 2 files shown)


ochaloup:
@mmusgrove true, I haven't tested it with quickstart. I will do so. Thank you.

I'm testing it on other apps. Using only `@Provider` vs. `javax.ws.rs.ext.Providers` is pretty ok and works fine. I was thinking to not using `@Provider` for the `ServerLRAFilter` and use only programatic resolution but that way the filter was not activated on incoming lra request. I'm not sure if there could be some trouble that the `@Provider` register the `ServerLRAFilter` during startup as it is, without any filtering on the existence of `@LRA` (so far currently it works the way it's just registered).

mmusgrov:
I can understand why the presence of the FilterRegistration dynamic feature provider might cause the Server filter to be registered twice but if you remove the client registration, ie
rts/lra/lra-filters/src/main/resources/META-INF/services/javax.ws.rs.ext.Providers
then there is nothing to install the client side request and response filters (ie ClientLRAResponseFilter and ClientLRARequestFilter).

ochaloup:
@mmusgrov yes, you are right. I haven't realized that. The `@Provider` annotations will not be used at the client. I changed the fix to prefer the descriptor - meaning not using `@Provider` at the client filters.

The lra quickstart passes with such change.

What do you think to fix in this way?

ochaloup:
@mmusgrov thanks, I will wait to see how the tests pass and then eventually merge this.

ochaloup:
...just changed the formatting for not changing the order of imports

mmusgrov:
... okay

PR #1246, opened 2017-11-05
Add an XAResourceOrphanFilter that can check if there is a subordinate transaction in flight for an Xid

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2685

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/SubordinateAtomicActionRecoveryModule.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/SubordinateAtomicActionRecoveryModule.java
@@ -1,0 +1,135 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2016, Red Hat Middleware LLC, and individual contributors
+ * as indicated by the @author tags.
+ * See the copyright.txt in the distribution for a full listing
+ * of individual contributors.
+ * This copyrighted material is made available to anyone wishing to use,
+ * modify, copy, or redistribute it subject to the terms and conditions
+ * of the GNU Lesser General Public License, v. 2.1.
+ * This program is distributed in the hope that it will be useful, but WITHOUT A
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+ * PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more details.
+ * You should have received a copy of the GNU Lesser General Public License,
+ * v.2.1 along with this distribution; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA  02110-1301, USA.
+ *
+ * (C) 2016,
+ * @author JBoss Inc.
+ */
+package com.arjuna.ats.internal.jta.recovery.arjunacore;
+
+import com.arjuna.ats.arjuna.common.Uid;
+import com.arjuna.ats.arjuna.exceptions.ObjectStoreException;
+import com.arjuna.ats.arjuna.objectstore.RecoveryStore;
+import com.arjuna.ats.arjuna.objectstore.StoreManager;
+import com.arjuna.ats.arjuna.recovery.RecoveryManager;
+import com.arjuna.ats.arjuna.recovery.RecoveryModule;
+import com.arjuna.ats.arjuna.state.InputObjectState;
+import com.arjuna.ats.internal.arjuna.common.UidHelper;
+import com.arjuna.ats.internal.jta.transaction.arjunacore.jca.SubordinationManager;
+import com.arjuna.ats.internal.jta.transaction.arjunacore.subordinate.jca.SubordinateAtomicAction;
+import com.arjuna.ats.jta.logging.jtaLogger;
+
+import javax.transaction.xa.XAException;
+import java.io.IOException;
+import java.util.Vector;
+
+
+/**
+ * This class is purely used by the recovery system to load the transactions into memory so we can be sure
+ * that orphan detection can find the TransactionImple if the EIS has not called XATerminator::recover yet
+ *
+ * At the time of writing this module is utilized by the SubordinationManagerXAResourceOrphanFilter to ensure that
+ * it can check with the SubordinationManager if the transaction is in flight during orphan detection.
+ */
+public class SubordinateAtomicActionRecoveryModule implements RecoveryModule {
+
+    private boolean recoveryScanCompletedWithoutError;
+    private boolean validatePosition;
+
+    @Override
+    public void periodicWorkFirstPass() {
+        /*
+         * Requires going through the objectstore for the states of imported
+         * transactions - this is just to make sure the server control are loaded into memory.
+         *
+         * The EIS will call XATerminator::recover() for actual crash recovery
+         */
+
+        RecoveryStore recoveryStore = StoreManager.getRecoveryStore();
+        InputObjectState states = new InputObjectState();
+
+        // only look in the JCA section of the object store
+        Uid uid = null;
+
+        try {
+            if (recoveryStore.allObjUids(SubordinateAtomicAction.getType(), states)
+                    && (states.notempty())) {
+                while (true) {
+                    uid = UidHelper.unpackFrom(states);
+                    if (uid.notEquals(Uid.nullUid())) {
+                        SubordinationManager
+                                .getTransactionImporter().recoverTransaction(
+                                uid);
+                    } else {
+                        break;
+                    }
+                }
+            }
+            recoveryScanCompletedWithoutError = true;
+        } catch (ObjectStoreException | XAException | IOException e) {
+            jtaLogger.i18NLogger.warn_could_not_recover_subordinate(uid, e);
+            recoveryScanCompletedWithoutError = false;
+        }
+
+        if (!validatePosition()) {
+            recoveryScanCompletedWithoutError = false;
+        }
+    }
+
+    @Override
+    public void periodicWorkSecondPass() {
+        // No-op - recovery is performed by the EIS
+    }
+
+    /**
+     * Used to ensure that the orphan detection has fully loaded the transaction state before asserting
+     * a decision.
+     *
+     * @return Whether the last recovery scan completed without an error
+     */
+    public boolean isRecoveryScanCompletedWithoutError() {
+        return recoveryScanCompletedWithoutError;
+    }
+
+    /**
+     * It is important to verify that the SubordinateAtomicActionRecoveryModule has been loaded before the XARecoveryModule
+     * as otherwise when we check for an imported transaction it may not have been loaded by the  SubordinateAtomicActionRecoveryModule yet.
+     *
+     * @return true if and only if SubordinateAtomicActionRecoveryModule appears before XARecoveryModule in the modules
+     */
+    private boolean validatePosition() {
+        if (!this.validatePosition) {
+            Vector<RecoveryModule> modules = RecoveryManager.manager().getModules();
+            boolean foundSelf = false;
+            for (RecoveryModule module : modules) {
+                if (module instanceof SubordinateAtomicActionRecoveryModule) {
+                    foundSelf = true;
+                } else if (module instanceof XARecoveryModule) {
+                    // If we have found the XARecoveryModule (which in practice must be there because the filter is running
+                    // we need to check to make sure that the class responsible for loading SubordinateAtomicActions is
+                    // configured to run before it
+                    if (!foundSelf) {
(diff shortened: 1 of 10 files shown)


ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/SubordinateAtomicActionRecoveryModule.java
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * This class is purely used by the recovery system to load the transactions into memory so we can be sure
+ * that bottom-up recovery could find a serverControl if the EIS has not called XATerminator::recover yet
+ */
+public class SubordinateAtomicActionRecoveryModule implements RecoveryModule {
+
+    @Override
+    public void periodicWorkFirstPass() {
+                /*
```
ochaloup: a cosmetic remark - the indentation could be adjusted

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/SubordinationManagerXAResourceOrphanFilter.java
```diff
@@ -0,0 +1,93 @@
...
+    @Override
+    public Vote checkXid(Xid xid) {
+        if (xid.getFormatId() != XATxConverter.FORMAT_ID) {
+            // we only care about Xids created by the JTA
+            return Vote.ABSTAIN;
+        }
+
+        if (!TxControl.getXANodeName().equals(XATxConverter.getSubordinateNodeName(new XidImple(xid).getXID()))) {
+            // It either doesn't have a subordinate node name or isn't for this server
+            return Vote.ABSTAIN;
```
ochaloup: could happen that the `XATxConverter.getSubordinateNodeName` to `getXID` will not match but the `NodeNameXAResourceOrphanFilter` will match the `XATxConverter.getNodeName` for the same XID.

This would mean - this filter votes for {{ABSTAIN}} but the orphan filter then votes for rollback.

I can't find some situation to reach this from top of my head. Just I would like to mention such possibility.

tomjenkinson: I see what you are saying: "what if the parent recovery manager also has access to the same resource manager and scans a shared XAResource - won't it try to roll it back"? I will add an extra check into NodeNameXAResourceOrphanFilter to make sure that the original NodeNameXAResourceOrphanFilter does not do that

tomjenkinson: That was not necessary eventually as only the node that is ultimately responsible for NodeNameXAResourceOrphanFilter will match the XATxConverter.getNodeName and that has the further orphan check that looks in the ActionStatusService

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/SubordinateAtomicActionRecoveryModule.java
```diff
@@ -0,0 +1,83 @@
...
+                boolean finished = false;
+
+                do {
+                    Uid uid = UidHelper.unpackFrom(states);
+                    if (uid.notEquals(Uid.nullUid())) {
+                        SubordinationManager
+                                .getTransactionImporter().recoverTransaction(
+                                uid);
+                    } else {
+                        finished = true;
```
mmusgrov: Why not break here instead of using the flag.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/SubordinateAtomicActionRecoveryModule.java
```diff
@@ -0,0 +1,83 @@
...
+                    }
+                }
+                while (!finished);
+            }
+        } catch (ObjectStoreException | XAException | IOException e) {
+            e.printStackTrace();
```
mmusgrov: We need a logging statement here instead.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/SubordinateAtomicActionRecoveryModule.java
```diff
@@ -0,0 +1,83 @@
...
+import com.arjuna.ats.internal.jta.transaction.arjunacore.jca.SubordinationManager;
+import com.arjuna.ats.internal.jta.transaction.arjunacore.subordinate.jca.SubordinateAtomicAction;
+
+import javax.transaction.xa.XAException;
+import java.io.IOException;
+
+
+/**
+ * This class is purely used by the recovery system to load the transactions into memory so we can be sure
+ * that bottom-up recovery could find a serverControl if the EIS has not called XATerminator::recover yet
```
mmusgrov: Can you add more commentary, maybe a javadoc link to the code that expects them to be already loaded and an indication of why that other code could not do the loading.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/SubordinationManagerXAResourceOrphanFilter.java
```diff
@@ -0,0 +1,93 @@
...
+import com.arjuna.ats.jta.xa.XidImple;
+
+import javax.transaction.xa.XAException;
+import javax.transaction.xa.Xid;
+import java.util.Vector;
+
+/**
+ * An XAResourceOrphanFilter which vetos rollback for xids which have an in-flight subordinate transaction.
+ * <p>
+ * The SubordinateAtomicActionRecoveryModule must be loaded in in a position prior to the XARecoveryModule for this to work
```
mmusgrov: Re my earlier comment: can you include some commentary on the mechanics of how the module and filter coordinate their responsibilities.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/SubordinationManagerXAResourceOrphanFilter.java
```diff
@@ -0,0 +1,93 @@
...
+            return Vote.LEAVE_ALONE;
+        }
+        if (importedTransaction != null) {
+            return Vote.LEAVE_ALONE;
+        } else {
+            return Vote.ROLLBACK;
+        }
+    }
+
+    private boolean validatePosition() {
```
mmusgrov: This complicated (and probably non performant) routine is essentially just doing the following:
```java
return modules.contains(SAARM) && modules.contains(XARM))
```
Can we either refactor the code or simplify.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/SubordinateAtomicActionRecoveryModule.java
tomjenkinson: Sure

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/SubordinationManagerXAResourceOrphanFilter.java
tomjenkinson: It's checking the order too (hence the name) - I can add more comments

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/SubordinateAtomicActionRecoveryModule.java
tomjenkinson: Done

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/SubordinateAtomicActionRecoveryModule.java
```diff
@@ -0,0 +1,99 @@
...
+            if (recoveryStore.allObjUids(SubordinateAtomicAction.getType(), states)
+                    && (states.notempty())) {
+                while (true) {
+                    UidHelper.unpackFrom(states);
+                    if (uid.notEquals(Uid.nullUid())) {
+                        SubordinationManager
+                                .getTransactionImporter().recoverTransaction(
+                                uid);
+                    } else {
+                        break;
```
mmusgrov: Here you set recoverySucceeded to true even though nothing was recovered.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/SubordinationManagerXAResourceOrphanFilter.java
```diff
@@ -0,0 +1,112 @@
...
+    /**
+     * It is important to verify that the SubordinateAtomicActionRecoveryModule has been loaded before the XARecoveryModule
+     * as otherwise when we check for an imported transaction it may not have been loaded by the  SubordinateAtomicActionRecoveryModule yet.
+     *
+     * This method also stores a reference to the SubordinateAtomicActionRecoveryModule so that we can verify there
+     * were no failures in recovery transactions for this pass of the recovery cycle.
+     *
+     * @return true if and only if SubordinateAtomicActionRecoveryModule appears before XARecoveryModule in the modules
+     */
+    private boolean validatePosition() {
```
mmusgrov: I still say it is a complicated bit of code. All it is achieving is:
```java
return modules.contains(SAARM) && modules.indexOf(SAARM) < modules.indexOf(XARM))
```

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/SubordinateAtomicActionRecoveryModule.java
tomjenkinson: I have changed the name of this and commented where it is used to try to make that clearer

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/SubordinationManagerXAResourceOrphanFilter.java
tomjenkinson: I don't think I can do that because I don't have a reference for SAARM and XARM (I am having to compare the classnames)

mmusgrov: I wasn't suggesting that my line was an alternative (just that the routine is achieving something quite easy with lots of lines of code and therefore harder to maintain and read. My original comment said:

> Can we either refactor the code or simplify.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/SubordinateAtomicActionRecoveryModule.java
mmusgrov: okay

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/SubordinationManagerXAResourceOrphanFilter.java
tomjenkinson: > can we refactor or simplify
maybe I can move this to the SubordinateAtomicActionRecoveryModule but the code will be the same more or less in that location, plus I will still need to scan the list to obtain the instance of SubordinateAtomicActionRecoveryModule to check for errors.

We do already do similar code elsewhere:
https://github.com/jbosstm/narayana/blob/master/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java#L1254

mmusgrov: > Blaargh! There are better ways to do this!

Someone had similar thought to me here. Anyway, I don't want to labour the issue so I am approving.

PR #1248, opened 2017-11-10
BUILD SCRIPT ONLY: Change paratemers to check for regression

tomjenkinson:


mmusgrov:
So we need to go back to running the job on a dedicated machine.

mmusgrov:
We have a lot of jobs that need checking in order to find an uncontested time slot. Probably best to look at the weekend jobs. Clearly we cannot do it on the PR perf axis job.

PR #1253, opened 2017-11-15
[JBTM-2203] Remove ReentrantLock from StateManager

mmusgrov:
Resurrection of PR 680 originally raised by @jesperpedersen 

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
@@ -33,9 +33,9 @@ package com.arjuna.ats.arjuna;
 
 import java.io.IOException;
 import java.io.PrintWriter;
-import java.util.Enumeration;
-import java.util.Hashtable;
-import java.util.concurrent.locks.ReentrantLock;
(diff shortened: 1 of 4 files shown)


nmcl:
The JIRA even says "rejected" (from 2016) so why is this being worked on? At least keep things in sync!

mmusgrov:
There was a bug in the performance comparison checker

nmcl:
OK but that doesn't explain why the JIRA is still closed (rejected). Either the issue is being worked on or it isn't. This isn't a heisenbug ;) It can't be in two states simultaneously until one of us looks at it and the waveform then collapses ;)

mmusgrov:
Well we wanted to do a quick test to test first ie not actually change any code. If we saw a change then that would be the trigger to reopen the JIRA.

nmcl:
So you're doing work against a closed JIRA because ..? It's not hard to click "re-open" you know :)

jesperpedersen:
@mmusgrov !

mmusgrov:
@jesperpedersen Sorry I was setting up a PR poller job on our upshift installation. I temporarily added a build periodically schedule and it started testing all the PRs. Sorry about that. It is disabled now

PR #1258, opened 2017-11-16
JBTM-1732 sync to memory when a new STM proxy is created

mmusgrov:
https://issues.jboss.org/browse/JBTM-1732

 !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !AS_TESTS !RTS

### DIFF
--- a/STM/src/main/java/org/jboss/stm/internal/reflect/InvocationHandler.java
+++ b/STM/src/main/java/org/jboss/stm/internal/reflect/InvocationHandler.java
@@ -58,6 +58,8 @@ import com.arjuna.ats.txoj.LockManager;
 import com.arjuna.ats.txoj.LockMode;
 import com.arjuna.ats.txoj.LockResult;
 
+import static com.arjuna.ats.txoj.LockResult.GRANTED;
+
 public class InvocationHandler<T> implements java.lang.reflect.InvocationHandler
 {
     /*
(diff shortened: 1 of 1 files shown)


STM/src/main/java/org/jboss/stm/internal/reflect/InvocationHandler.java
```diff
@@ -192,6 +194,18 @@ public InvocationHandler (RecoverableContainer<T> cont, T obj, Uid u)
...
             else
                 _txObject = new LockManagerProxy<T>(obj, cont);  // recoverable or persistent
+
+            AtomicAction action = new AtomicAction();
+
+            action.begin();
+            int result = _txObject.setlock((_optimistic ? new OptimisticLock(LockMode.WRITE) : new Lock(LockMode.WRITE)), 0, 0);
+
+            if (result != GRANTED)
+                if (txojLogger.logger.isInfoEnabled())
```
mmusgrov: I wonder if it might be better to throw an exception here like you do on line 415 at the end of the method interceptor:

```java
throw new LockException(Thread.currentThread()+" could not set "+LockMode.stringForm(cachedLock._lockType)+" lock. Got: "+LockResult.stringForm(result));
```

nmcl:
Looks good to me. This wasn't the issue I was thinking about when I had concerns.

PR #1260, opened 2017-11-22
JBTM-2961 Temporarily disabling the lra cdi checker (fails on mac builds)

mmusgrov:
https://issues.jboss.org/browse/JBTM-2961

NO_TEST


mmusgrov:
The lra cdi checker is failing the mac build so we are temporarily disabling it whist we investigate.

PR #1264, opened 2017-12-11
[JBTM-2973] direct recoverable connection to be recoverable

ochaloup:
when try to commit fails with RETRY. Then recoverable should be still
saved in object store in way that recovery manager is capable to
deserialize it and use it for committing.

The most of the PR are tests the fix is only the line
https://github.com/jbosstm/narayana/pull/1264/files#diff-9613cb84dbca3e6a76d760b2b4626befL1273

https://issues.jboss.org/browse/JBTM-2973

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF  !RTS !AS_TESTS !mysql !postgres !db2 !oracle

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
@@ -1270,7 +1270,6 @@ public class XAResourceRecord extends AbstractRecord implements ExceptionDeferre
 		if (_recoveryObject != null)
 		{
 			_recoveryObject.close();
-			_recoveryObject = null;
 		}
 
 		if (_theTransaction != null)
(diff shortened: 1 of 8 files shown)


ochaloup:
test setup fix (need to setup recovery manager in `@Before` not in each test separately. when defined separately only the first set up is taken)

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
```diff
@@ -1270,7 +1270,6 @@ private final void removeConnection()
 		if (_recoveryObject != null)
 		{
 			_recoveryObject.close();
-			_recoveryObject = null;
```
tomjenkinson: Can we get a comment please?

ochaloup:
Ok, thank you. I'm going to merge.

PR #1271, opened 2018-01-16
[JBTM-2983] Returning a correct error code for one phase resource

ochaloup:
https://issues.jboss.org/browse/JBTM-2983

Based on the question at https://stackoverflow.com/questions/48213996/transaction-commit-error-is-not-captured-in-ejb-code I investigated on the behaviour and I consider it wrong.

The point is that if commit on the non-XA resource fails the caller does not get response that the commit failed. I think the wrong error code is returned here. I think even the comment is wrong as recovery will not be taken for one phase commit resource.

Particular test is expected for WFLY as it's about integration with WFLY settings, txn spi and jca using the LastResource interface https://github.com/ochaloup/wildfly/tree/last-resource-fail-on-commit

!QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS !mysql !postgres !db2 !oracle

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/abstractrecords/LastResourceRecord.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/abstractrecords/LastResourceRecord.java
@@ -160,6 +160,32 @@ public class LastResourceRecord extends AbstractRecord
         }
     }
 
+    public int topLevelOnePhaseCommit ()
+    {
+        if (tsLogger.logger.isTraceEnabled()) {
+            tsLogger.logger.trace("LastResourceRecord::topLevelOnePhase() for " + order());
+        }
+
+        if (_lro == null)
+            return TwoPhaseOutcome.PREPARE_NOTOK;
+
+        switch (_lro.commit())
+        {
+        case TwoPhaseOutcome.FINISH_OK:
+            return TwoPhaseOutcome.FINISH_OK;
+        case TwoPhaseOutcome.ONE_PHASE_ERROR:
+            return TwoPhaseOutcome.ONE_PHASE_ERROR;
+        default:
+            // in case of error
+            // FINISH_ERROR is used for 2PC there was an error and expecting recovery will retry (not available here)
+            //  no exception thrown to the caller
+            // ONE_PHASE_ERROR assuming rollback but we don't know the outcome (was rolled-back or committed in the RM)
+            //  javax.transaction.RollbackException thrown to the caller
+            // HEURISTIC_MIXED used to get javax.transaction.HeuristicMixedException to the caller
+            return TwoPhaseOutcome.HEURISTIC_MIXED;
+        }
+    }
+
     @Override
     public String toString()
     {


ochaloup:
@tomjenkinson ah, you are absolutely right.

But I think the more correct would be {{javax.transaction.HeuristicMixedException}}. The {{SystemException}} first sounds like something wrong is in the HW and second the normal {{XAResource}} ends in one-phase commit with {{HeuristicMixedException}} thus that should be consistent, I think. See https://github.com/jbosstm/narayana/blob/master/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java#L756

I'm just a little in doubts how to implement this in a nice/correct way. My issue is now that the AbstractRecord  ignores the other outcome than `OK` or `ERROR` (https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java#L426). The `XAResourceRecord` overrides the whole `topLevelOnePhaseCommit` method (https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java#L426) so I did for the `LastResourceRecord` in this PR.

ochaloup:
My fix does not follow the CMR work where retry (or rollback) is expected rather then throwing the mixed heuristic. Mixed heuristic is still necessary for  which is needed for the non-XA resource. I changed the fix to impact only the onePhase commit method.

ochaloup:
(I forgot to add the fixed file to the commit 😿 )

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/abstractrecords/LastResourceRecord.java
```diff
@@ -160,6 +160,32 @@ public int topLevelPrepare ()
...
+            return TwoPhaseOutcome.FINISH_OK;
+        case TwoPhaseOutcome.ONE_PHASE_ERROR:
+            return TwoPhaseOutcome.ONE_PHASE_ERROR;
+        default:
+            // in case of error
+            // FINISH_ERROR is used for 2PC there was an error and expecting recovery will retry (not available here)
+            //  no exception thrown to the caller
+            // ONE_PHASE_ERROR assuming rollback but we don't know the outcome (was rolled-back or committed in the RM)
+            //  javax.transaction.RollbackException thrown to the caller
+            // HEURISTIC_MIXED used to get javax.transaction.HeuristicMixedException to the caller
```
tomjenkinson: Remember we are in ArjunaCore code here which is not only JTA. Would this HEURISTIC_MIXED change have undesired impact in raw ArjunaCore?

ochaloup:
@tomjenkinson those considerations were reason why it took me such time. I think it should be ok. But to be absolutely honest I don't have such insight to the core code to say it for 100%.

The `AbstractRecord#topLevelOnePhaseCommit` (overridden in particular implementations as the `LastResourceRecord` does now) is called only from `BasicAction#onePhaseCommit` (https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java#L2386). Then the if expects - `OK`, `FINISH_ERROR` or `ONE_PHASE_ERROR` and whatever else which reports heuristics. From that point the sounds it's ok. 

I can see the JTS can throw heuristic hazard (https://github.com/jbosstm/narayana/blob/master/ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/resources/ResourceRecord.java#L557), the WS code too and the `BasicAction` seems to consider that.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/abstractrecords/LastResourceRecord.java
ochaloup: those considerations are the reason why it took me such time. I think it should be ok. But to be absolutely honest I don't have such insight to the core code to say it for 100%.

The AbstractRecord#topLevelOnePhaseCommit (overridden in particular implementations as the LastResourceRecord does now) is called only from BasicAction#onePhaseCommit (https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java#L2386). Then the if expects - OK, FINISH_ERROR or ONE_PHASE_ERROR and whatever else which reports heuristics. From that point the sounds it's ok.

I can see the JTS can throw heuristic hazard (https://github.com/jbosstm/narayana/blob/master/ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/resources/ResourceRecord.java#L557), the WS code too and the BasicAction seems to consider that.

tomjenkinson: Did you see this? https://developer.jboss.org/thread/277017 It suggests that the user _does not_ get an exception in this (or perhaps some closely related) case.

ochaloup: Ah, I haven't noticed that yet. It's pretty closely related. That's a testcase I was playing with too
https://github.com/ochaloup/wildfly/commit/10a5cce9c0c9489fe0ebef4d455c9028088d076b#diff-3f86971fbf72ccd8cc76481e04ca61ceR86

But by my understanding this is completely fine as for one phase commit optimization Narayana does not write any object store log record as it does for 2PC end of prepare (which is beneficial for performance). Then the commit ends in `HeuristicMixedException` as we can't decide if the transaction was really committed or not.

But I understand that's not much obvious from the first sight.

We can see that the java api
https://docs.oracle.com/javaee/7/api/javax/transaction/UserTransaction.html#commit--
does not count with 1PC (at least it seems to me so) and there is no exeption like "error - unknown state" (which could be misleading in fact too). The `HeuristicMixedException` is defined as `Thrown to indicate that a heuristic decision was made and that some relevant updates have been committed while others have been rolled back.`. That does not fit well for 1PC. Maybe the `SystemException` would be closer - kind of some error happened - we have no idea what was it - HW/system/connection). But `SystemException` does not cover inside of itself that it is not obvious if it was committed or rolled-back. In that way the name `HeuristicMixedException` makes that fact more obvious.

I'm now looking at the XA spec and there is said - `These values may be returned only if
TMONEPHASE is set in flags:` and there is `XA_RBCOMMFAIL] A communication failure occurred within the resource manager.`.  Which sounds that should be returned from RM in case of connection is lost. We handle such exception for throwin `RolllbackException`-> https://github.com/jbosstm/narayana/blob/master/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAOnePhaseResource.java#L154 -> `ONE_PHASE_ERROR` -> `javax.transaction.RollbackException`

ochaloup: for example jdbc api says the `commit`: https://docs.oracle.com/javase/8/docs/api/java/sql/Connection.html#commit-- returns just `SQLException` as a generic error

ochaloup: By XA spec the `RMFAIL`: `[XAER_RMFAIL] An error occurred that makes the resource manager unavailable.` is the same as the `XA_RBCOMMFAIL` but currently ends in ERROR_FINISH: https://github.com/jbosstm/narayana/blob/master/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAOnePhaseResource.java#L164

that leads to either nothing is done or in case of XA in heuristic hazard return code
https://github.com/jbosstm/narayana/blob/master/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java#L757

tomjenkinson: RBCOMMFAIL is slightly different in that it does confirm there is a rollback ("RB_"). If we can't be certain we should not throw something that the transaction manager can say is definitive.

ochaloup: ok, I see. Then `XAER_RMFAIL` sounds the best, right?

tomjenkinson: +1 - behaviour should be as per RMFAIL (i.e. we just don't know).

ochaloup: Would you be fine me merging this change? This PR changes the behaviour of non-XA 1PC in way of the call
https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java#L2386
ends in heuristic mixed exception
https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java#L2448

the `RMFAIL` is error returned in case of default behaviour when error on commit happens
https://github.com/jbosstm/narayana/blob/master/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAOnePhaseResource.java#L162

tomjenkinson: Sure

PR #1274, opened 2018-01-27
JBTM-2987 XAER_NOTA return code handling

primIl:
When doing rollback from Recovery Manager treat XAER_NOTA return code as debug instead of warning

!QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF !RTS !mysql !postgres !db2 !oracle

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
@@ -846,7 +846,16 @@ public class XARecoveryModule implements RecoveryModule
         }
         catch (XAException e1)
         {
-        	jtaLogger.i18NLogger.warn_recovery_xarecovery1(_logName+".xaRecovery", XAHelper.printXAErrorCode(e1), e1);
+            if(e1.errorCode == XAException.XAER_NOTA)
+            {
+                if(jtaLogger.logger.isDebugEnabled()) {
+                    jtaLogger.logger.debug("XAER_NOTA received while rolling back " + XAHelper.xidToString(xid));
+                }
+            }
+            else
+            {
+                jtaLogger.i18NLogger.warn_recovery_xarecovery1(_logName+".xaRecovery", XAHelper.printXAErrorCode(e1), e1);
+            }
 
             switch (e1.errorCode)
             {
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
@@ -849,7 +849,7 @@ public class XARecoveryModule implements RecoveryModule
             if(e1.errorCode == XAException.XAER_NOTA)
             {
                 if(jtaLogger.logger.isDebugEnabled()) {
-                    jtaLogger.logger.debug("XAER_NOTA received while rolling back " + XAHelper.xidToString(xid));
+                    jtaLogger.logger.debug(_logName+".xaRecovery: XAER_NOTA received while rolling back " + XAHelper.xidToString(xid));
                 }
             }
             else


ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
```diff
@@ -846,7 +846,16 @@ else if(vote == XAResourceOrphanFilter.Vote.ROLLBACK)
         }
         catch (XAException e1)
         {
-        	jtaLogger.i18NLogger.warn_recovery_xarecovery1(_logName+".xaRecovery", XAHelper.printXAErrorCode(e1), e1);
+            if(e1.errorCode == XAException.XAER_NOTA)
+            {
+                if(jtaLogger.logger.isDebugEnabled()) {
+                    jtaLogger.logger.debug("XAER_NOTA received while rolling back " + XAHelper.xidToString(xid));
```
ochaloup: I'm not fully sure here... Would not be good to retain the exception stacktrace in the output even it's kind of expected exception?

primIl: I'm seeing now _logName+".xaRecovery" is missing, without that it's not clear where the log is coming from, since jtaLogger is shared for all jta classes. For stack trace, that wouldn't hurt either, but would it be consistent with the rest of the codebase, is there any other place where debug log has stacktrace?

primIl:
We would probably also want to update the comment in `JTAEnvironmentBean.setOrphanSafetyInterval`, currently is says: 

"Set the amount of time to wait before deciding if the Xid is orphaned.
It is important because if this is too short and a transaction completes
between the two recovery scan phases the xids from the RM will be considered
as orphaned. Although this does not cause data integrity issues it can
appear unsettling."

First, this comment was made when JTAActionStatusServiceXAResourceOrphanFilter was not implemented, back then orphanSafetyInterval did actually impact data integrity: if it was too low, recovery manager could run rollback on in-progress xids, if time it takes to finish prepare on all resources is greater than orphanSafetyInterval. This can happen bacause Narayana writes tlog entry only once after prepare finishes on all resources - one or many stuck resources with very slow prepare call could cause this scenario.

Now, with this fix, and with JTAActionStatusServiceXAResourceOrphanFilter enabled, orphanSafetyInterval has this effect:

- setting it too low or zero: sometimes we would unnecessarily hit XAResource with rollback() and get XAER_NOTA, causing unnecessary network activity and load on resource 

- setting it too high: if TM crashes after XAResource prepare is done but before tlog entry is written, these orphan xids would be rolled back by RM only after orphanSafetyInterval, causing longer data unavailability (locks being held for longer time)

Not sure how to best descibe it in javadoc

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
ochaloup: to me returning use of `_logName` sounds as a good approach for not changing the previous behaviour abruptly

my belief is to use the right and enough information in the log to provide a good way to investigate in case of some failures. I personally do not bound this to the reasoning that there is no such necessity in the code around. For sure there are some (but not often) cases of debugging showing stacktrace in the Narayana codebase (https://github.com/jbosstm/narayana/blob/master/ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/recovery/recoverycoordinators/ResourceCompletor.java#L79)

my doubt was what value it will provide for somebody investigation. I tend to think, now, that's for NOTA it's not necessary as Xid will link the information with prior operations.

ochaloup:
@primIl it sounds pretty reasonable to me to adjust the javadoc. Just there should be probably differentiated what is impact for having `JTAActionStatusServiceXAResourceOrphanFilter` configured and the old behaviour when the `JTAActionStatusServiceXAResourceOrphanFilter` is not configured.  I think your two points, as they are, sounds me reasonable to be added to the javadoc. Maybe @tomjenkinson would have some more idea on this?

btw. I updated the description of the PR with CI tags to not running all the CI tests (ignoring those which sounds me not relevant to the change in this PR)

primIl:
@tomjenkinson yes

ochaloup:
@tomjenkinson yes, I consider the change being fine for merge

ochaloup:
@primIl thanks a lot for your contribution. I personally consider this as pretty nice enhancement of the user experience.

PR #1277, opened 2018-02-06
JBTM-2990 performance fix for ThreadUtil lock contention.

jhalliday:
Add ThreadLocal cache to allow locking on WeakHashMap to be avoided for most use cases.


mmusgrov:
@tomjenkinson I would say that @barreiro needs to retest to see if the improvement is sufficient.

mmusgrov:
Rebasing and merging

barreiro:
@mmusgrov I was only able to give it a go last Friday, and although I didn't got any numbers for comparison, I can confirm that the contention we used to see is gone. That was the main concern.

Thank you all for sorting this out!

PR #1281, opened 2018-02-13
[JWS-883] Database allocation via DBAllocator, Docker, H2

Karm:
See https://github.com/Karm/narayana/blob/narayana-tomcat-ts/tomcat/README.md
for description of the new functionality.

TL;DR: The TS was expanded from the former hard coded H2 database to support DBAllocator and Docker provided databases. H2 remains the default though.

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS !AS_TESTS !mysql !postgres !db2 !oracle !MAIN


ochaloup:
@tomjenkinson yes, it's so, working on that

ochaloup:
@Karm @tomjenkinson I fixed the issues with XTS PR troubles, rebase issues of WFLY vs. 5_BRANCH and do a fix in the script to run.

Now the test fails with

```
[INFO] --- maven-failsafe-plugin:2.19.1:integration-test (default) @ tomcat-jta ---

-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running org.jboss.narayana.tomcat.jta.integration.BaseITCase
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 2.067 sec <<< FAILURE! - in org.jboss.narayana.tomcat.jta.integration.BaseITCase
org.jboss.narayana.tomcat.jta.integration.BaseITCase  Time elapsed: 2.067 sec  <<< ERROR!
org.jboss.arquillian.container.spi.client.container.LifecycleException: Could not start container
Caused by: org.jboss.arquillian.container.spi.ConfigurationException: 
Unable to connect to Tomcat manager. The server command (/text/list) failed with responseCode (401) and responseMessage (Unauthorized).

Please make sure that you provided correct credentials to an user which is able to access Tomcat manager application.
These credentials can be specified in the Arquillian container configuration as "user" and "pass" properties.
The user must have aapropriate role specified in tomcat-users.xml file.



Results :

Tests in error: 
  BaseITCase.org.jboss.narayana.tomcat.jta.integration.BaseITCase ? Lifecycle Co...
```
that seems to be already issue at the Tomcat test side. From the message there is probably issue with ` tomcat-users.xml` settings. 

I would expect that the PR does not reflect need of changes in the narayana.sh script. @Karm the all CI builds are run with `scripts/hudson/narayana.sh` script. There is function to run the tomcat tests: https://github.com/jbosstm/narayana/blob/master/scripts/hudson/narayana.sh#L563 and there is the `cat` to `xml` config to fill the credentials for the arquillian. I'm not sure what is the logic in the changes. Do you think you can change this bash function to follow the changes for our CI passes?

Karm:
@ochaloup I'm on it. Gimme a sec. 🐙

ochaloup:
@Karm still the same failure. I would tip (taken from the readme) that the execution is done without passing the `-Duser.name` and `-Duser.passwor` parameters to `mvn build.sh` command at https://github.com/jbosstm/narayana/pull/1281/commits/98d8440b373cdc6853ea598e22b5a3419c45e42d#diff-6779b95748ceb5718e81c25f9ee4a489R582

btw. what about providing the values `arquillian` as default one by specify them as system properties of failsafe plugin(https://github.com/jbosstm/narayana/blob/master/tomcat/tomcat-jta/pom.xml#L221, https://github.com/jbosstm/narayana/blob/master/tomcat/tomcat-jta/src/test/resources/arquillian.xml#L28)

Karm:
@ochaloup Fixed.
BTW: There appears to be no tomcat testing in narayana.bat, at least H2 should be able to run there...right?

ochaloup:
@Karm no, what I know. Most of the CI testing is run only on the Linux. I'm not sure where the question aims to... sure, H2 should be runable on W$.

Karm:
@ochaloup To rephrase: Should I implement the tomcat-jta testing profile in narayana.bat as it exists in narayana.sh?

ochaloup:
@Karm no, I think it would be better to spend time to think how to tear off the DBAllocator from the PR 😄🍭

Karm:
@ochaloup I'll come back to the PR in April and move DBAllocator away 🐙

ochaloup:
@tomjenkinson there is an issue in the narayana poller script which seems are working only with first comment taken from the github api and not with all of them. Investigating to get the 'TESTIT' to get working

Karm:
@ochaloup @tomjenkinson  This is not ready for merging - DBAllocator code still in there. I just want to test the Narayana CI PostgreSQL server and that the big rebase went well.
The "Docker" thing passed :) [https://ci.modcluster.io/job/narayana-tomcat](https://ci.modcluster.io/job/narayana-tomcat/DATABASE_VERSION=postgres%3A10,JDK_VERSION=openjdk-8,TOMCAT_VERSION=9.0.7,label=karm-centos7-x86-64/81/consoleText).

Karm:
@tomjenkinson :+1: Works for me.   thx.

Karm:
@ochaloup @tomjenkinson Ready for review. Highlights:
* Passes [https://ci.modcluster.io/job/narayana-tomcat](https://ci.modcluster.io/job/narayana-tomcat/DATABASE_VERSION=postgres%3A10,JDK_VERSION=openjdk-8,TOMCAT_VERSION=9.0.7,label=karm-centos7-x86-64/83/consoleText)
* DBAllocator and other DB related code moved to "tools" module in the repo, away from the main Tomcat JTA codebase. Does it suffice?
* ```jboss-transaction-api_1.2_spec``` jar is loaded into Tomcat's lib directory by the TS, no need to shuffle things around in bash, see [documentation](https://github.com/Karm/narayana/tree/narayana-tomcat-ts/tomcat#transactions-api-jar-and-jws-976)
* updated documentation with debugging and other handy notes, see [debugging](https://github.com/Karm/narayana/tree/narayana-tomcat-ts/tomcat#debugging)
* refactored DB allocation for your upstream CI Postgres and H2 tests, it should run both via the same DB allocation API as the rest of the databases (DBAllocator, Docker etc.)
* TransactionalDataSourceFactory logging switched to Tomcat Juli (already pulled in with Tomcat DBCP2), [see code](https://github.com/jbosstm/narayana/pull/1281/files#diff-131154e303b7fd01bd18c416261896fdL51), might interest @zhfeng , WDYT?
* @zhfeng , @ochaloup , IMHO there is no point in requiring "DataSource" string name while in fact all we instanciate is "XADataSource" -> [I changed it](https://github.com/jbosstm/narayana/pull/1281/files#diff-131154e303b7fd01bd18c416261896fdL74). It is also updated in all context.xml generator props etc. If it gets accepted, I will go on to refactor the [quickstart too](https://github.com/jbosstm/quickstart/tree/master/dbcp2-and-tomcat).
* Some additional fields in TransactionalDataSourceFactory declared ```final```; I hope you don't mind. No intention to start a "space-or-tab" style flame though :)
* @zhfeng Removed props that do not belong to the cool config and cause FieldNotFound execptions, [see code](https://github.com/jbosstm/narayana/pull/1281/files#diff-131154e303b7fd01bd18c416261896fdL60). I realized just now that you already did that in your   #1302  -- I guess it gets merged first, so I'll rebase and amend my changes after yours...
* Proper default object store location for Tomcat, see [the change](https://github.com/jbosstm/narayana/pull/1281/files#diff-55ad9469d54b576e75eaa89facddc29eR56) to work directory.

zhfeng:
@Karm https://github.com/jbosstm/narayana/pull/1302 has been merged, so feel free to rebase with your branch. It looks good to me to change to use "XADataSource" string name.

@tomjenkinson I had checked the CI output and the tests look good

tools/src/main/java/io/narayana/db/DBAllocator.java
```diff
@@ -0,0 +1,328 @@
...
+        }
+        final URL driverJarSrc;
+        try {
+            driverJarSrc = new URL(dballocatorDriverUrl);
+        } catch (MalformedURLException e) {
+            e.printStackTrace();
+            LOGGER.severe(e.getMessage());
+            return null;
+        }
+        final String[] driverJarFilename = dballocatorDriverUrl.split("/");
```
ochaloup: please, use  https://docs.oracle.com/javase/7/docs/api/java/io/File.html#getName(). In case the handling is needed do not use the `"/"` but https://docs.oracle.com/javase/7/docs/api/java/io/File.html#separator

tools/src/main/java/io/narayana/db/DBAllocator.java
```diff
@@ -0,0 +1,328 @@
...
+        final File driverJarDst = new File(driversDir, driverJarFilename[driverJarFilename.length - 1]);
+        long t = System.currentTimeMillis();
+        // < 1K is a suspicious DB driver jar...
+        while (!fileOK(1024, driverJarDst) && ((System.currentTimeMillis() - t) / 1000 < driverFetchTimeoutSeconds)) {
+            try {
+                LOGGER.finest("Downloading DB driver from " + driverJarSrc.getPath());
+                FileUtils.copyURLToFile(driverJarSrc, driverJarDst);
+            } catch (IOException e) {
+                LOGGER.severe(e.getMessage());
```
ochaloup: the exception should be printed as whole in case of error happens, not only error message. The stack normally facilitate finding the cause.

This pattern is used at more places over the commit. I would personally vote for logging the whole stacktrace at most of the places possible/reasonable. If you have some reason to not doing so, I'm happy to understand and discuss it.

tools/src/main/java/io/narayana/db/CIPostgreAllocator.java
```diff
@@ -0,0 +1,111 @@
...
+        if (StringUtils.isEmpty(servername)) {
+            throw new IllegalArgumentException("pgsql.servername must not be empty");
+        }
+        final String portnumber = System.getProperty("pgsql.portnumber");
+        final int port;
+        if (StringUtils.isEmpty(portnumber)) {
+            throw new IllegalArgumentException("pgsql.portnumber must not be empty");
+        } else {
+            port = Integer.parseInt(portnumber);
+            if (port > 65535 || port < 1025) {
```
ochaloup: I know it's a bad behaviour to run whatever at port lower to 1025 but you never know. I would personally vote not checking here the range and leave the user to handle it himself.

tools/src/main/java/io/narayana/db/DBAllocator.java
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    public boolean cleanDB(final DB db) {
+        if (db == null) {
+            throw new IllegalArgumentException("db must not be null");
+        }
+        if (db.allocationProperties == null) {
+            throw new IllegalArgumentException("db.allocationProperties must not be null");
+        }
+        final String dballocatorHostPort = System.getProperty("dballocator.host.port");
```
ochaloup: would not be benefical to create some utility method which get `property name` and returns value or throws `IAE` if empty? I think it could remove quite a lot of `if` causes from the start of all the methods around.

tools/src/main/java/io/narayana/db/PostgreContainerAllocator.java
```diff
@@ -0,0 +1,400 @@
...
+                throw new IllegalArgumentException("container.database.bind.host.port out of expected range [1025, 65535]");
+            }
+        }
+        final String containerTimeoutWaitingForTcp = System.getProperty("container.timeout.waiting.for.tcp");
+        final long timeout;
+        if (StringUtils.isEmpty(containerTimeoutWaitingForTcp)) {
+            throw new IllegalArgumentException("container.timeout.waiting.for.tcp must not be empty");
+        } else {
+            timeout = Long.parseLong(containerTimeoutWaitingForTcp);
+            if (timeout > 30 * 60 * 1000 || timeout < 500) {
```
ochaloup: what about `TimeUnit.MINUTES.toMillis(30)`?

tomcat/tomcat-jta/src/test/java/org/jboss/narayana/tomcat/jta/integration/AbstractCase.java
```diff
@@ -0,0 +1,191 @@
...
+                        Paths.get(dest),
+                        StandardCopyOption.REPLACE_EXISTING);
+            } catch (IOException e) {
+                fail("Either CATALINA_HOME or version.org.jboss.spec.javax.transaction is invalid, we cannot copy to " + dest);
+                e.printStackTrace();
+            }
+        } catch (IOException | NullPointerException e) {
+            e.printStackTrace();
+            fail(e.getMessage());
+            if (dba != null && db != null) {
```
ochaloup: place under `finally` block, this will be never executed (`fail` would throw `AssertionException` I expect)

tomcat/tomcat-jta/src/test/resources/arquillian.xml
```diff
@@ -25,14 +25,18 @@
...
-            <!-- Need to add the following to the tomcat-users.xml file: <user username="test" password="test" roles="manager-script"/> -->
-            <property name="user">test</property>
-            <property name="pass">test</property>
+            <property name="unpackArchive">true</property>
+            <!-- Add user and pass to your conf/tomcat-users.xml-->
+            <property name="user">${tomcat.user}</property>
+            <property name="pass">${tomcat.pass}</property>
             <!--  -agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=8000 -->
-            <property name="javaVmArguments">${server.jvm.args}</property>
+            <property name="javaVmArguments">-Djava.security.egd=file:/dev/./urandom ${server.jvm.args}</property>
```
ochaloup: this is pretty interesting :grinning:

Karm:
@tomjenkinson 
> Should remove the changes are are about spacing only to make clear what the real changes are - we don't usually take formatting changes.

I understand, yet I took the liberty of treating the Tomcat JTA module as something quite new that one could still reformat and reshape, e.g. not mixing tabs and spaces in pom etc.
If you point out a specific file you want reverted, I'll do it.

> I also wonder why the need to move to the ".app" package. If it's styling I propose you revert it, if it has a practical use lets consider it.

The ```app``` package contains the actual web application code whereas the parent package contains the test cases. It helps me navigate the project. Do I have to put all classes in a one package?

> Before merge it should be squashed to single commit please.

Certainly. As soon as @ochaloup and you ack the requested changes in the latest commit, I'll squash it all.

@zhfeng 
> Rebase ```tomcat/tomcat-jta/src/main/java/org/jboss/narayana/tomcat/jta/TransactionalDataSourceFactory.java```

Done.

@ochaloup 
Ad [attribute](https://github.com/jbosstm/narayana/pull/1281#discussion-diff-183044112R598)
> I'm maybe a bit picky but the attribute ```pgsql.or.h2.or.container.or.dballocator``` seems to me being strangely named. What about something shorter kind of ```datasource.type``` or ```test.db.type```.

Done.

Ad [PostgreSQL](https://github.com/jbosstm/narayana/pull/1281#discussion-diff-183044914R10)
> Here we should not be limited at PostgreSQL as Narayan ci environment runs oracle and mysql as well. I think the section should be rewritten in way where it talks just about db direct connection - aka. url, user, password are provided by some remote DB already running. The name of the properties should be renamed appropriately to not mention psql.

The documentation reflects exactly the current state of the code. There is a generator for PostgreSQL, e.g. having things like ```org.postgresql.xa.PGXADataSourceFactory``` in it etc. at the moment. As soon as I add support for more databases, I naturally update the documentation and refactor as needed, one by one.

Ad [Remove notes](https://github.com/jbosstm/narayana/pull/1281#discussion-diff-183046831R311)
>  this note could be possibly removed

Done.

Ad [Jenkins job in the README.md](https://github.com/jbosstm/narayana/pull/1281#discussion-diff-183048269R136)
> I hesitate if this section should be part of the README.
> ... snip ...
> If this is a preparation for the module being moved somewhere under tomcat repository jurisdiction then by me I think it's fine to be here.

The mod_cluster ci server builds various components, including but not limited to OpenSSL, Wildfly-OpenSSL, httpd... It is used by our JWS/JBCS group, so yes, this note could be understood as related to the future migration to https://github.com/web-servers organization.
Last but not least, IIUC, Narayana CI slaves don't run Docker and my CentOS 7 slave does, hence the note about the Jenkins job.

Ad [HTML markup](https://github.com/jbosstm/narayana/pull/1281#discussion-diff-183072823R40)
>  it's not necessary to add end ```</p>``` tag as it's added automatically. it's probably matter of taste but for me it makes superfluous text noise

Done.

Ad [Allocator constructor](https://github.com/jbosstm/narayana/pull/1281#discussion-diff-183073157R47)
> maybe not needed as it's abstract class and it can't be instantiated

Done.

Ad [Thread interrupt](https://github.com/jbosstm/narayana/pull/1281#discussion-diff-183074319R106)
> if interrupted and not rethrown as a good practice the flag should be set up back to the thread ```Thread.currentThread().interrupt();```

Done.

Ad [year in copyright headers](https://github.com/jbosstm/narayana/pull/1281#discussion-diff-183074465R3)
> maybe the copyright headers should be changed to contain the current year?

Done.

Ad [driverFetchTimeoutSeconds typo](https://github.com/jbosstm/narayana/pull/1281#discussion-diff-183076391R69)
>  ```driverFetchTimeoutSeconds``` is typo I think, it should be ```dballocator.driver.url.timeout.seconds``` in the error message, no?

Done.

Ad [setting this non-existing directory](https://github.com/jbosstm/narayana/pull/1281#discussion-diff-183076632R60)
> maybe add to the exception what was name of the property used for setting this non-existing directory?

Done.

Ad [Long.getLong](https://github.com/jbosstm/narayana/pull/1281#discussion-diff-183077686R67)
> maybe Long.getLong could be used. hm... maybe not....

Why would I do that? I use ```long```, not ```Long```.

Ad [printStackTrace and LOGGER](https://github.com/jbosstm/narayana/pull/1281#discussion-diff-183077908R81)
> if it is possible please do not use stdout but ```LOGGER```. The printStackTrace is used and more places over the code, please, fix them all.

Done.

Ad [dballocatorDriverUrl.split](https://github.com/jbosstm/narayana/pull/1281#discussion-diff-183080026R85)
> please, use https://docs.oracle.com/javase/7/docs/api/java/io/File.html#getName(). In case the handling is needed do not use the "/" but https://docs.oracle.com/javase/7/docs/api/java/io/File.html#separator

Why so? It is not a file system path, the *source* is a URL. See ```dballocator.driver.url```. There will always be ```/```, See [rfc1738](https://tools.ietf.org/html/rfc1738). The *destination* is a ```File``` though, but I'm not tokenizing that...
Either explain more or I leave it as it is.

Ad [printing just messages, not the whole stack](https://github.com/jbosstm/narayana/pull/1281#discussion-diff-183080818R94)
> the exception should be printed as whole in case of error happens, not only error message. The stack normally facilitate finding the cause.

Done.

Ad [port > 65535 || port < 1025](https://github.com/jbosstm/narayana/pull/1281#discussion-diff-183082189R62)
> I know it's a bad behaviour to run whatever at port lower to 1025 but you never know. I would personally vote not checking here the range and leave the user to handle it himself.

Well, do you have a specific case for that such as Narayana CI running databases on privileged ports? I don't understand why it is a problem at this stage. Perhaps you could elaborate. My case is catching obvious typos such as ```543``` instead of ```5432```, ```0``` etc.  If it should block the entire PR I will just delete it though.

Ad [create some utility method which get property name](https://github.com/jbosstm/narayana/pull/1281#discussion-diff-183086560R298)
> would not be benefical to create some utility method which get property name and returns value or throws IAE if empty? I think it could remove quite a lot of if causes from the start of all the methods around.

Done. 

Ad [TimeUnit](https://github.com/jbosstm/narayana/pull/1281#discussion-diff-183087568R117)
> what about TimeUnit.MINUTES.toMillis(30)?

Done.

Ad [code after fail](https://github.com/jbosstm/narayana/pull/1281#discussion-diff-183090800R98)
> place under finally block, this will be never executed (fail would throw AssertionException I expect)

Done.

Ad [java.security.egd=file:/dev/./urandom](https://github.com/jbosstm/narayana/pull/1281#discussion-diff-183091727R39)
> this is pretty interesting :)

If there is next to none entropy on the system, Tomcat doesn't wait for entropy to be gathered and uses insecure pseudorandom bits. We are talking test suite here, so there is no point on requiring secure random numbers, right?


If I have your acks @tomjenkinson, @zhfeng, @ochaloup, I do the squash and it is ready for merging.  Much more stuff is coming though. This PR is suppose to merely establish the first working smoke test.

tomcat/tomcat-jta/src/main/java/org/jboss/narayana/tomcat/jta/TransactionSynchronizationRegistryFactory.java
```diff
@@ -1,6 +1,6 @@
 /*
  * JBoss, Home of Professional Open Source.
- * Copyright 2016, Red Hat, Inc., and individual contributors
+ * Copyright 2018, Red Hat, Inc., and individual contributors
```
tomjenkinson: accidental year change?

PR #1283, opened 2018-02-17
[JBTM-2969] lra annotation checker as maven plugin

ochaloup:
https://issues.jboss.org/browse/JBTM-2969

Changing the `lra-cdi-rest` project from Weld extension checking the LRA annotation which was not good to use and was hard to understand from user point of view to lra annotation checker project as maven plugin. Now if somebody wants to start writing application with LRA it can use this plugin to check LRA annotations prerequisites are satisfied. The prerequisite could be e.g. `@Compensate` annotation expects `@Path` to expose endpoint, the `/forget` endpoint expects `@DELETE` REST call to be used etc.

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !AS_TESTS !mysql !postgres !db2 !oracle


ochaloup:
@mmusgrov thanks. I added the README file. I hope it covers what's demanded.

PR #1290, opened 2018-03-01
[BUILD SCRIPT ONLY] jboss as tests invocation to consider narayana version

ochaloup:
The AS test could fail as there is not changed Narayana to the current one during XTS and RTS tests execusions.

From
```bash
./build.sh -f ./testsuite/integration/xts/pom.xml -B -Pxts.integration.tests.profile "$@" test
```

to
```bash
./build.sh -f ./testsuite/integration/xts/pom.xml -B -Pxts.integration.tests.profile -Dversion.org.jboss.narayana=${NARAYANA_CURRENT_VERSION} "$@" test
```

!MAIN !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS !mysql !postgres !db2 !oracle


ochaloup:
failing because of issue of WFLY where there is wrong versions in pom.xml version checker after WFLY 12.0.0.Final release

see https://github.com/wildfly/wildfly/pull/10963

PR #1300, opened 2018-04-13
JBTM-3012 Close recovery connection to ensure we have a valid connect…

tomjenkinson:
https://issues.jboss.org/browse/JBTM-3012

!RTS !QA_JTS_OPENJDKORB !AS_TESTS !QA_JTA !QA_JTS_JACORB !BLACKTIE !QA_JTS_JDKORB !XTS NO_WIN  !PERF


ochaloup:
@tomjenkinson as I haven't tried I'm just throwing an idea here. I was looking at the `ConnectionListener` and I think it still should be possible to enlist the connection listener even here. I think the underlaying pooled connection will notify (https://docs.oracle.com/javase/7/docs/api/javax/sql/ConnectionEventListener.html). Maybe?

PR #1301, opened 2018-04-17
JBTM-3012 Close recovery connection to ensure we have a valid connect…

tomjenkinson:
https://issues.jboss.org/browse/JBTM-3012

!RTS !QA_JTS_OPENJDKORB !AS_TESTS !QA_JTA !QA_JTS_JACORB !BLACKTIE !QA_JTS_JDKORB !XTS NO_WIN !MAIN !PERF

### DIFF
--- a/tomcat/tomcat-jta/src/main/java/org/jboss/narayana/tomcat/jta/TransactionalDataSourceFactory.java
+++ b/tomcat/tomcat-jta/src/main/java/org/jboss/narayana/tomcat/jta/TransactionalDataSourceFactory.java
@@ -100,26 +100,22 @@ public class TransactionalDataSourceFactory implements ObjectFactory {
             XARecoveryModule xaRecoveryModule = getXARecoveryModule();
             if (xaRecoveryModule != null) {
                 xaRecoveryModule.addXAResourceRecoveryHelper( new XAResourceRecoveryHelper() {
+                    private Object lock = new Object();
                     private XAConnection connection;
                     @Override
                     public boolean initialise(String p) throws Exception {
-                        initialiseConnection();
                         return true;
                     }
 
                     @Override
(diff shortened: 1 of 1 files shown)


tomcat/tomcat-jta/src/main/java/org/jboss/narayana/tomcat/jta/TransactionalDataSourceFactory.java
```diff
@@ -108,26 +109,29 @@ public boolean initialise(String p) throws Exception {
...
-
-                            if (user != null && password != null) {
-                                return new XAResource[]{xaDataSource.getXAConnection(user, password).getXAResource()};
-                            } else {
-                                return new XAResource[]{xaDataSource.getXAConnection().getXAResource()};
+                    public synchronized XAResource[] getXAResources() throws Exception {
+                        synchronized (lock) {
+                            initialiseConnection();
+                            try {
+                                String user = properties.getProperty(PROP_USERNAME);
```
ochaloup: this is a bit strange to gather value for `user`/`password` and not using it further in the `try-block` and then the block for `user != null` is the same of functionality as if `user == null`

ochaloup: this is a bit strange to gather value for `user`/`password` and not using it further in the `try-block` and then the block for `user != null` is the same of functionality as if `user == null`

tomjenkinson: Agreed!

ochaloup:
@tomjenkinson yes, that's fine with me

zhfeng:
@tomjenkinson it looks good to me

PR #1303, opened 2018-04-23
[SCRIPTS ONLY] changing comment NO_TEST for github label NoCI

ochaloup:
Removing the `NO_TEST` github comment handling from narayana.sh in substitution with `NoCI` label. When the `NoCI` label is used then the `-poller` narayana ci script does not start the test build.

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS !AS_TESTS !TOMCAT !mysql !postgres !db2 !oracle !MAIN

/cc @zhfeng @mmusgrov

### DIFF
--- a/scripts/hudson/narayana.sh
+++ b/scripts/hudson/narayana.sh
@@ -75,10 +75,6 @@ function init_test_options {
         export AS_BUILD=0 NARAYANA_BUILD=0 NARAYANA_TESTS=0 BLACKTIE=0 XTS_AS_TESTS=0 XTS_TESTS=0 TXF_TESTS=0 txbridge=0
         export RTS_AS_TESTS=0 RTS_TESTS=0 JTA_CDI_TESTS=0 QA_TESTS=0 SUN_ORB=0 OPENJDK_ORB=0 JAC_ORB=0 JTA_AS_TESTS=0
         export PERF_TESTS=0 OSGI_TESTS=0 TOMCAT_TESTS=0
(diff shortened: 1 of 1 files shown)


ochaloup:
http://narayanaci1.eng.hst.ams2.redhat.com/view/Pulls/job/btny-pulls-narayana-poller/build?delay=0sec

ochaloup:
@tomjenkinson sure, I'm just in the phase of testing now :) What I want is to get running the poller job again not the narayana build itself. The poller job covers more jobs than just the narayana.

ochaloup:
@tomjenkinson ok, I consider this as ready-to-go. Would you be so kind and check the code changes again, plus take a look at the message which would be reported by the poller job and if you're ok that the poller job is the on being invoked. The only 'issue' is that the link won't directly build but you need to click through confirmation button.

btw. github offers webhook in case of label is created or deleted at the pull request but there is the same issue that github can't access to VPN thus the idea with link is the simplest and best suitable I think

ochaloup:
@tomjenkinson the poller runs checks of the SCM (github repository) each 5 min. If there are some new changes then the script defined in the poller job is executed. I thought that when you run poller manually you skip this check for the new commits. But I can see I was wrong partly. The poller execution runs for the last commit, for the last changed PR only.

Thus I changed the text to what you suggested (see the pull comment above).

PR #1305, opened 2018-04-24
[JWS-883] not throwing NPE when wrong db test type is provided

ochaloup:
When tomcat is run and a non-existent test type is specified (e.g. `psql` instead of `pgsql`, aka. `-Dtest.db.type=psql`) then  NPE is thrown at the end of the test instead of being reported only(!) the test fails because of non-existent test type.

!MAIN !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS !AS_TESTS !mysql !postgres !db2 !oracle

### DIFF
--- a/tomcat/tomcat-jta/src/test/java/org/jboss/narayana/tomcat/jta/integration/AbstractCase.java
+++ b/tomcat/tomcat-jta/src/test/java/org/jboss/narayana/tomcat/jta/integration/AbstractCase.java
@@ -111,7 +111,7 @@ public abstract class AbstractCase {
             dba.deallocateDB(db);
         }
         // @see https://issues.jboss.org/browse/JWS-976
-        if (libFiles[0] != null) {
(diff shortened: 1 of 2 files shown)


ochaloup:
testing purposes of closing and opening

ochaloup:
@tomjenkinson yes, absolutely true. Sorry Tom for misguiding you. I just wanted to run the JBTM-3015 reproducer on my machine but I haven's spent enough time to fix it. While I was doing so I found just a small thing where NPE is thrown when you define wrong test type e.g. `psql` instead of `pgsql`, aka. `-Dtest.db.type=psql`. I was fighting with the label ci stuff yesterday so I used this PR as a small fix testing case. This fix is really just small improvement to the original JWS-883 - aka. small fix for test execution. There is no idea in this PR to fix the JBTM-3015. If you consider this too small or something I can close it.

PR #1311, opened 2018-05-09
JBTM-3016 Update to use the BasicManagedDataSource

zhfeng:
https://issues.jboss.org/browse/JBTM-3016

!RTS !QA_JTS_OPENJDKORB !AS_TESTS !QA_JTA !QA_JTS_JACORB !BLACKTIE !QA_JTS_JDKORB !XTS NO_WIN !MAIN !PERF

### DIFF
--- a/tomcat/tomcat-jta/src/main/java/org/jboss/narayana/tomcat/jta/TransactionalDataSourceFactory.java
+++ b/tomcat/tomcat-jta/src/main/java/org/jboss/narayana/tomcat/jta/TransactionalDataSourceFactory.java
@@ -80,6 +80,11 @@ public class TransactionalDataSourceFactory implements ObjectFactory {
         final XADataSource xaDataSource = (XADataSource) getReferenceObject(ref, context, PROP_XA_DATASOURCE);
 
         if (transactionManager != null && xaDataSource != null) {
+            /*
+             * There is a trick to fix DBCP-215 so we have to remove the "initialSize" that
+             * the BaiscDataSourceFactory.createDataSource(properties) will not create the connections in the pool.
+             * And it will create the connections with the BaiscManagedDataSource later if the initialSize > 0.
+             */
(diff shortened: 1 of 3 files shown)


tomcat/tomcat-jta/src/main/java/org/jboss/narayana/tomcat/jta/TransactionalDataSourceFactory.java
```diff
@@ -119,50 +67,41 @@ public Object getObjectInstance(Object obj, Name name, Context context, Hashtabl
...
         final XADataSource xaDataSource = (XADataSource) getReferenceObject(ref, context, PROP_XA_DATASOURCE);
 
         if (transactionManager != null && xaDataSource != null) {
-            final DataSourceXAConnectionFactory xaConnectionFactory =
-                    new DataSourceXAConnectionFactory(transactionManager, xaDataSource);
-            final PoolableConnectionFactory poolableConnectionFactory =
-                    new PoolableConnectionFactory(xaConnectionFactory, null);
-            final GenericObjectPoolConfig config = new GenericObjectPoolConfig();
-            final AbandonedConfig abandonedConfig = new AbandonedConfig();
+            String initialSize = properties.getProperty("initialSize");
```
tomjenkinson: I was wondering why initialSize is different to other properties? Perhaps you can add a comment into the code to explain?

zhfeng: There is a trick to fix [DBCP-215](https://issues.apache.org/jira/browse/DBCP-215) so we have to remove the "initialSize" that the BaiscDataSourceFactory.createDataSource(properties) will not create the connections in the pool. And it will create with the BaiscManagedDataSource later.

I will add these comments in the code.

PR #1319, opened 2018-05-20
JBTM-2977 Handle network failures when ending participants

mmusgrov:
https://issues.jboss.org/browse/JBTM-2977

### DIFF
--- a/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
+++ b/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
@@ -52,9 +52,11 @@ import java.time.LocalTime;
 import java.util.Arrays;
 import java.util.SortedMap;
 import java.util.TreeMap;
+import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantLock;
 
 import static org.eclipse.microprofile.lra.client.LRAClient.LRA_HTTP_HEADER;
 import static org.eclipse.microprofile.lra.client.LRAClient.LRA_HTTP_RECOVERY_HEADER;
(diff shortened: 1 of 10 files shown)


rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -213,7 +215,21 @@ public int topLevelCommit() {
     }
 
     private int doEnd(boolean compensate) {
+        assert lraService != null;
+
+        ReentrantLock lock = lraService.lockTransaction(lraId);
+
+        try {
+            return tryDoEnd(compensate);
```
ochaloup: could not be here get a deadlock kind of behaviour? What if the http call at `tryDoEnd` takes long time (issues on the network or similar). Then the second call could be locked because of the first call can't proceed. What's expectation of such case?

mmusgrov: The application thread will be blocked until the participant replies - I think this is reasonable behaviour.
The recovery thread does a tryLock and if it cannot obtain the lock then it returns and will try again on the next recovery pass - hence there is no opportunity for deadlock.

mmusgrov: @ochaloup  But I agree that blocked application threads can cause a problem with shutting down the container. I am addressing that problem in https://issues.jboss.org/browse/JBTM-2982 (LRA participant completion calls during recovery do not timeout)

ochaloup: ok, I see. 

Still I can imagine there could be long lasting call trying to end the LRA. The client's first request is hanging thus it tries to call second one and it hangs too. Client has no idea what's happening and one connection depends on other one. It's ok in general just I think in discrete network environment this could be considered.

When thinking about this to be honest I have another issue. I think it would be nice if there is a way to solve the issue without the JVM locks if possible. As we were considering possibility of scaling LRA coordinator then JVM locks added here will need to be refactored later. Which is not possibly an issue as refactoring later is still possible, just if we would think in some general design way of avoiding it that would be nice.
I'm not deep into this just regarding the JBTM-2982, I'm not sure if applicable, just thinking about some timeouts and catching signal of the container shutdown.

mmusgrov:
The lock does not introduce scalability issues because each LRA gets its own lock.

With regards to the client getting confused about why his request is hanging, he has the option of asking the coordinator for its status.

ochaloup:
I got the point of the LRA locking. Especially I haven't recognized there is use of the `synchronized` blocks. As we discussed this could be addressed by other issue or by some refactoring (especially regarding asynchronicity).

mmusgrov:
@ochaloup The last commit adds the test

PR #1322, opened 2018-05-22
Removing release profile from the jboss as build

ochaloup:
It's not needed to build the WFLY as release as Narayana uses it only for tests.

This is test to get not having CI failures with error message

```
[ERROR] Failed to execute goal org.jboss.galleon:galleon-maven-plugins:1.0.0.CR3:provision (server-provisioning) on project wildfly-build: Failed to provision the state: Failed to locate /tmp/ecb7b9e6-6be5-4af7-aba6-b8fba83b78b5/staged/docs/licenses/servlet-feature-pack-licenses.xml -> [Help 1]
```

!MAIN !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS  !TOMCAT !mysql !postgres !db2 !oracle

### DIFF
--- a/scripts/hudson/narayana.sh
+++ b/scripts/hudson/narayana.sh
@@ -384,13 +384,13 @@ function build_as {
   export MAVEN_OPTS="-XX:MaxPermSize=512m -XX:+UseConcMarkSweepGC $MAVEN_OPTS"
   if [ $AS_TESTS = 1 ]; then
     JAVA_OPTS="-Xms1303m -Xmx1303m -XX:MaxPermSize=512m $JAVA_OPTS" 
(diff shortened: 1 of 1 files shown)


ochaloup:
@tomjenkinson I think when the `-Prelease` is removed then the licenses are not downloaded and the issue with OOM could not appear. It seems that this change helped to get the as build working.

In other words - what I understand is that the `release` does some post build actions to release could be processed. One of that is the licensing. While we want just take the distro and test with it as with a SNAPSHOT our CI does not need to run as `release`.

ochaloup:
My last change was only about changing title of the commit to contain `SCRIPTS ONLY` wording. The shell script content was not changed. Removing parameters of release and downloading licenses seem to get the as build working.

PR #1330, opened 2018-05-25
JBTM-3028 Temporarily disable SpecIT#acceptTest - can fail on a slow …

mmusgrov:
…network

https://issues.jboss.org/browse/JBTM-3028

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !AS_TESTS !TOMCAT !mysql !postgres !db2 !oracle !RTS

### DIFF
--- a/rts/lra/lra-test/src/test/java/io/narayana/lra/participant/SpecIT.java
+++ b/rts/lra/lra-test/src/test/java/io/narayana/lra/participant/SpecIT.java
@@ -67,6 +67,7 @@ import org.junit.After;
 import org.junit.AfterClass;
 import org.junit.Before;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


mmusgrov:
@ochaloup Our CI has a long queue and I reckon this temporary fix is needed urgently. I think that this can be safely merged without a CI run, WDYT?

rts/lra/lra-test/src/test/java/io/narayana/lra/participant/SpecIT.java
```diff
@@ -619,7 +619,7 @@ public void testUserData() {
         assertTrue(activity.contains("endData='" + testData));
     }
 
-    @Test
+    //@Test
```
ochaloup: would not be better to use `@Ignored` for the maven plugin list the test as existing but ignored?

PR #1332, opened 2018-06-07
JBTM-2982 Drive LRA participants from a separate thread

mmusgrov:
https://issues.jboss.org/browse/JBTM-2982

### DIFF
--- a/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
+++ b/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
@@ -239,8 +239,6 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
 
     private int tryDoEnd(boolean compensate) {
         URL endPath;
-        boolean isRecovering = status != null &&
-                (status.equals(CompensatorStatus.Compensating) || status.equals(CompensatorStatus.Completing));
 
         // cancel any timer associated with this participant
         if (scheduledAbort != null) {
@@ -271,8 +269,6 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
 
         // NB trying to compensate when already completed is allowed (for nested LRAs)
 
-        Current.push(lraId); // make sure the lra id is set so that it can be included in the headers
-
         int httpStatus = -1;
 
         if (accepted) {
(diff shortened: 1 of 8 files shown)


rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -379,15 +371,19 @@ private int tryDoEnd(boolean compensate) {
...
+        updateStatus(compensate);
 
         // if the the request is still in progress (ie accepted is true) let recovery finish it
         return accepted ? TwoPhaseOutcome.HEURISTIC_HAZARD : TwoPhaseOutcome.FINISH_OK;
     }
 
+    private void updateStatus(boolean compensate) {
+           if (compensate)
+            status = accepted ? CompensatorStatus.Compensating : CompensatorStatus.Compensated;
+        else
```
ochaloup: maybe fixing indentation?

rts/lra/lra-test/pom.xml
```diff
@@ -35,14 +35,14 @@
              -Xrunjdwp:transport=dt_socket,address=${coordinator.debug.port},server=y,suspend=y
              -->
         <coordinator.debug.port>8787</coordinator.debug.port>
-        <coordinator.debug.jdwp>-Xrunjdwp:transport=dt_socket,address=${coordinator.debug.port},server=y,suspend=y</coordinator.debug.jdwp>
+        <coordinator.debug.jdwp>-Xrunjdwp:transport=dt_socket,address=${coordinator.debug.port},server=y,suspend=n</coordinator.debug.jdwp>
         <coordinator.debug.params></coordinator.debug.params>
 
         <!-- config for the test JAX-RS resource, for exampe to enable debugging use
```
ochaloup: do we need here this comment when the `swarm.debug.jdwp` attribute covers this?

rts/lra/lra-test/pom.xml
```diff
@@ -35,14 +35,14 @@
              -Xrunjdwp:transport=dt_socket,address=${coordinator.debug.port},server=y,suspend=y
```
ochaloup: do we need here this comment when the swarm.debug.jdwp attribute covers this?

rts/lra/lra-test/src/main/java/io/narayana/lra/participant/api/ActivityController.java
```diff
@@ -560,8 +560,25 @@ private void endCheck(Activity activity) {
...
+                } else {
+                    int ms = 0;
+
+                    try {
+                        ms = Integer.getInteger(arg, 0);
+                    } catch (Exception ignore) {
+                    }
+
+                    try {
+                        Thread.sleep(ms <= 0 ? Long.MAX_VALUE : ms); // delay the end call
```
ochaloup: I would like to understand the rationalization of giving to user a way of delaying end call?
It's seems as some workaround to something. I can be a wrong in understanding but letting the client to drive timing from outside sounds to me supportive of race conditions.
I know we discussed some timeouts to be involved but I can't connect this with the discussion.

rts/lra/lra-test/pom.xml
mmusgrov: What do you mean, do you want me to add an explanation of how the debugging config works?

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
mmusgrov: done

rts/lra/lra-test/pom.xml
ochaloup: sure. I'm sorry for not being clear. I just meant that if there is declared `coordinator.debug.jdwp` with debug parameters the comment could be removed completely. The information is doubled.

rts/lra/lra-test/pom.xml
ochaloup: I just meant if information is not doubled and if you consider so the comment could be removed competely.

ochaloup:
@mmusgrov hm... I have probably different version, I can't find the `isRecovering` at more places in the `LRARecord`.  If I take what is here in the PR: https://github.com/mmusgrov/narayana/blob/43e5c9c9a93cf58db1b25b8a1a86b76e27622bd2/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java#L242 then the variable is declared at the `l242` and then I can't find any usage of it. What is your version of the code?

rts/lra/lra-test/src/main/java/io/narayana/lra/participant/api/ActivityController.java
ochaloup: ouch, I see. Even I was checking the package I missed that. I'm sorry,  this is fine.

rts/lra/lra-test/pom.xml
mmusgrov: okay

mmusgrov:
@ochaloup re *isRecovering*, it was me that was looking at a wrong version, sorry.

So I have pushed a commit with that, the extra tab and the jdwp debug comment removed. None of those are code changes so if you are satisfied then I don't think we need to wait for CI?

rts/lra/lra-test/src/main/java/io/narayana/lra/participant/api/ActivityController.java
mmusgrov: Notice that I also added a comment to indicate what the recovery scan is testing here.

PR #1335, opened 2018-06-20
JBTM-3035 Ensure LRA context is propagated across non-LRA aware services

mmusgrov:
https://issues.jboss.org/browse/JBTM-3035

### DIFF
--- a/rts/lra/lra-test/src/main/java/io/narayana/lra/participant/api/StandardController.java
+++ b/rts/lra/lra-test/src/main/java/io/narayana/lra/participant/api/StandardController.java
@@ -1,0 +1,91 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+package io.narayana.lra.participant.api;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.ws.rs.HeaderParam;
+import javax.ws.rs.NotFoundException;
+import javax.ws.rs.PUT;
+import javax.ws.rs.Path;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.client.ClientBuilder;
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.UriInfo;
+
+import static io.narayana.lra.participant.api.ActivityController.ACTIVITIES_PATH;
+import static org.eclipse.microprofile.lra.client.LRAClient.LRA_HTTP_HEADER;
+
+@ApplicationScoped
+@Path(StandardController.ACTIVITIES_PATH3)
+public class StandardController {
+    public static final String ACTIVITIES_PATH3 = "activities3";
+    public static final String NON_TRANSACTIONAL_WORK = "/work";
+
+    @Context
+    private UriInfo context;
+
+    @PUT
+    @Path(NON_TRANSACTIONAL_WORK)
+    public Response work2(@HeaderParam(LRA_HTTP_HEADER) String lraId) throws NotFoundException {
+
+        if (lraId != null) {
+            return Response.status(Response.Status.PRECONDITION_FAILED).entity(Entity.text("Unexpected LRA context")).build();
+        }
+
+        WebTarget resourcePath = ClientBuilder.newClient().target(context.getBaseUri())
+                .path(ACTIVITIES_PATH)
+                .path(ActivityController.MANDATORY_LRA_RESOURCE_PATH);
+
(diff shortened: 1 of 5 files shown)


rts/lra/lra-client/src/main/java/io/narayana/lra/client/Current.java
```diff
@@ -156,7 +156,10 @@ public static void updateLRAContext(ClientRequestContext context) {
         if (lraId != null) {
             context.getHeaders().putSingle(LRA_HTTP_HEADER, lraId);
         } else {
-            Object lraContext = context.getProperty(LRA_HTTP_HEADER);
+            Object lraContext = getState("suspendedLRA");
```
tomjenkinson: Is it not worth using a "final static String" here?

rts/lra/lra-test/src/main/java/io/narayana/lra/participant/api/StandardController.java
```diff
@@ -0,0 +1,85 @@
...
+import javax.ws.rs.client.ClientBuilder;
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.UriInfo;
+import java.util.logging.Logger;
+
+import static io.narayana.lra.participant.api.ActivityController.ACTIVITIES_PATH;
+import static io.narayana.lra.participant.api.StandardController.ACTIVITIES_PATH3;
```
tomjenkinson: Is this a cyclic import?

rts/lra/lra-test/src/main/java/io/narayana/lra/participant/api/StandardController.java
```diff
@@ -0,0 +1,85 @@
...
+    @PUT
+    @Path(NON_TRANSACTIONAL_WORK)
+    public Response work2(@HeaderParam(LRA_HTTP_HEADER) String lraId)
+            throws NotFoundException {
+
+        if (lraId != null) {
+            return Response.status(Response.Status.PRECONDITION_FAILED).entity(Entity.text("Unexpected LRA context")).build();
+        }
+
+        WebTarget resourcePath = ClientBuilder.newClient().target(context.getBaseUri()).path(ACTIVITIES_PATH).path("nestedActivity");
```
tomjenkinson: I am assuming nestedActivity is already defined elsewhere in the project

rts/lra/lra-test/src/main/java/io/narayana/lra/participant/api/StandardController.java
```diff
@@ -0,0 +1,85 @@
...
+
+    private static final Logger LOGGER = Logger.getLogger(ActivityController.class.getName());
+
+    @Context
+    private UriInfo context;
+
+    @PUT
+    @Path(NON_TRANSACTIONAL_WORK)
+    public Response work2(@HeaderParam(LRA_HTTP_HEADER) String lraId)
+            throws NotFoundException {
```
tomjenkinson: Query here: Can I check, my assumption here would be that there is an LRA context I can get hold of and manually register a participant with if I wanted to. Is that possible using LRAClient at this point?

rts/lra/lra-test/src/main/java/io/narayana/lra/participant/api/StandardController.java
tomjenkinson: This is not a "request for changes" element of the review - just a query

rts/lra/lra-client/src/main/java/io/narayana/lra/client/Current.java
mmusgrov: +1

rts/lra/lra-test/src/main/java/io/narayana/lra/participant/api/StandardController.java
mmusgrov: No. I could do

> @Path(StandardController.ACTIVITIES_PATH2)

instead but I think the static import is neater

rts/lra/lra-test/src/main/java/io/narayana/lra/participant/api/StandardController.java
mmusgrov: It is defined in the other resource/controller (ActivityController) that we are using for tests. But, it's a good point, I will add a static final for it (and for the other paths too).

rts/lra/lra-test/src/main/java/io/narayana/lra/participant/api/StandardController.java
mmusgrov: Yes this change only affects the annotations. The non LRA aware resource would need to 

> @Inject
> private LRAClient lraClient;

and in the resource method he can then obtain the context:

> URL current = lraClient.getCurrent();

and then use lraClient to do the usual LRA stuff

rts/lra/lra-test/src/main/java/io/narayana/lra/participant/api/StandardController.java
tomjenkinson: OK, I disagree (slightly) but not holding the PR up for this one.

rts/lra/lra-client/src/main/java/io/narayana/lra/client/Current.java
mmusgrov: I am going to do the same for the resource paths as well which are currently defined as literal strings but are referenced from different places (in the resource @PATH annotations and in the SpecIT tests).

rts/lra/lra-test/src/main/java/io/narayana/lra/participant/api/StandardController.java
mmusgrov: Okay but since I am making a number of other changes I will go with StandardController.ACTIVITIES_PATH2 (since the recommendation is not overuse static imports)

rts/lra/lra-test/src/main/java/io/narayana/lra/participant/api/StandardController.java
mmusgrov: Agreed, I am going to turn all these @Path references to static final String

PR #1336, opened 2018-06-23
JBTM-3037 Ensure the LRA context is set correctly in the JAX-RS respo…

mmusgrov:
https://issues.jboss.org/browse/JBTM-3037

Notes for the reviewer:

LRA transaction propagation is implemented using a combination of thread locals and JAX-RS filters. The following text describes how the implementation keeps track of the currently active LRA:

Client Side:

When a service makes a JAX-RS invocation the ClientLRARequestFilter is triggered. This filter
checks to see if there is an LRA on the calling thread and, if so, it includes it in the outgoing headers.

When the invocation returns the ClientLRAResponseFilter is triggered. If the response header
contains an LRA then it is put on the thread as a thread local. Otherwise the filter checks whether
or not there was an LRA on the outgoing invocation and, if so, it puts that back on the thread.

**The current code does not perform this last check and this PR provides the fix for that.**

Server Side:

When a request enters a JAX-RS server the ServerLRAFilter is triggered. This filter checks for
LRA annotations on the target resouce method and decides whether a new LRA needs to be started
or whether an existing one needs to be suspended or ended (this is explained in the javadoc for @LRA).
The thread local context is then modified accordingly.

### DIFF
--- a/rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ClientLRAResponseFilter.java
+++ b/rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ClientLRAResponseFilter.java
@@ -46,6 +46,10 @@ public class ClientLRAResponseFilter implements ClientResponseFilter {
          */
         if (incomingLRA != null) {
             Current.push(new URL(incomingLRA.toString()));
+        } else {
+            // any previous context must have been ended by the invoked service otherwise incomingLRA
+            // would have been present
+            Current.pop();
         }
     }
 }
--- a/rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
+++ b/rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
@@ -182,7 +182,7 @@ public class ServerLRAFilter implements ContainerRequestFilter, ContainerRespons
 
             if (incommingLRA != null) {
                 Current.push(incommingLRA);
-                Current.putState(SUSPENDED_LRA_PROP, incommingLRA);
+                containerRequestContext.setProperty(SUSPENDED_LRA_PROP, incommingLRA);
             }
 
             return; // not transactional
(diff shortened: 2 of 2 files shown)


rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ClientLRAResponseFilter.java
```diff
@@ -36,16 +36,19 @@
     public void filter(ClientRequestContext requestContext, ClientResponseContext responseContext) throws IOException {
         Object incomingLRA = responseContext.getHeaders().getFirst(LRA_HTTP_HEADER);
 
-        if (incomingLRA == null) {
-            incomingLRA = requestContext.getProperty(LRA_HTTP_HEADER);
-        }
-
         /*
```
zhfeng: Why this is removed ?

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ClientLRAResponseFilter.java
```diff
@@ -36,16 +36,19 @@
...
+         * put it back on the thread.
          */
         if (incomingLRA != null) {
             Current.push(new URL(incomingLRA.toString()));
+        } else {
+            Object outgoingLRA = requestContext.getHeaders().getFirst(LRA_HTTP_HEADER);
+
+            if (outgoingLRA != null) {
+                Current.push((URL) outgoingLRA); // put it back on the request
+            }
```
zhfeng: The origin comment says
```
note we never popped the context in the request filter so we don't need to push outgoingLRA
```
So it does pop the context in the request filter now ? I can not find it in the ClientLRARequestFilter.

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ClientLRAResponseFilter.java
mmusgrov: It was a cut and paste error. BTW the test suite still passed so something is not right, I will put this PR on HOLD while I investigate further. Thanks

zhfeng:
Thanks @mmusgrov and it looks good to me.

mmusgrov:
Thanks for the review Amos

PR #1337, opened 2018-06-25
JBTM-3038 Use thorntail checkstyle ruleset for LRA

mmusgrov:
https://issues.jboss.org/browse/JBTM-3038

### DIFF
--- a/rts/lra/pom.xml
+++ b/rts/lra/pom.xml
@@ -21,6 +21,8 @@
 
     <properties>
         <version.wildfly-swarm>2017.8.1</version.wildfly-swarm>
+        <version.wildfly.swarm.checkstyle>3</version.wildfly.swarm.checkstyle>
+
         <version.json.api>1.1</version.json.api>
         <version.jaxrs.api>1.0.0.Final</version.jaxrs.api>
         <version.jackson>2.8.9</version.jackson>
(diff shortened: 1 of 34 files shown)


mmusgrov:
@ochaloup Like I said in the PR description wfly and thorntail use the same ruleset. So if I change the dependency will that address your concern

rts/lra/pom.xml
```diff
@@ -131,9 +134,36 @@
...
+                  <artifactId>checkstyle</artifactId>
+                  <version>${version.checkstyle}</version>
+                </dependency>
+                <dependency>
+                  <groupId>org.wildfly.swarm</groupId>
+                  <artifactId>checkstyle-config</artifactId>
+                  <version>${version.wildfly.swarm.checkstyle}</version>
+                </dependency>
+              </dependencies>
+              <executions>
```
tomjenkinson: Why do we need this when in the main pom we use configLocation: https://github.com/jbosstm/narayana/blob/431f3f759e0630d4a3105c52c678deb83aa5e02e/pom.xml#L381

mmusgrov:
@ochaloup you're right puppycrawal isn't needed (it is only needed for Eclpse integration).
@ochaloup as discussed this morning I would like LRA to conform with thorntail rules since that may well end up being the ultimate home for LRA

PR #1339, opened 2018-06-26
JBTM-3017 RecoveryMonitor should call System.exit(n) with a non zero …

mmusgrov:
https://issues.jboss.org/browse/JBTM-3017

I have reopened JBTM-3017 so that the program returns a standard code according to the conventions defined in System.exit(n).

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/RecoveryMonitor.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/RecoveryMonitor.java
@@ -47,6 +47,11 @@ public class RecoveryMonitor
 	private static String systemOutput = "";
 
 	public static void main (String[] args)
+	{
+		System.exit(main2(args));
+	}
+
+	public static int main2 (String[] args)
     {
 	String host = null;
 	int port = 0;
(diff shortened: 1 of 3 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/RecoveryMonitor.java
```diff
@@ -47,20 +47,26 @@
 	private static String systemOutput = "";
 
 	public static void main (String[] args)
+	{
+		System.exit(main2(args));
+	}
+
+	public static int main2 (String[] args)
```
ochaloup: I would consider to use more descriptive name for the method. The method name should be descriptive to express what's the functionality about. I think `main2` does not show the purpose. And it would be good to add some javadoc comment which says what the return `int` means (like simply - follow linux exit code standards where `0` is no-errors, `>0` is error)

mmusgrov: What a nit picker you are. Please suggest an alternative.

ochaloup: :)
I would consider something like `runRecovery` or `processRecovery`. As you split the `main` method to two and considering the point Nicola wrote to the email.

...what about to change the behaviour in way that the `main` processes the arguments verification and writing data to the systemout and calling System.exit and use the `processRecovery` as method which would make the remote recovery call while returning the outcome of it as method return value?

mmusgrov:
No, the solution in the PR does what he needs already and your suggestion is merely adding extra work.

Besides I plan to give him what he needs via the recovery management bean.

PR #1340, opened 2018-06-28
JBTM-3017 Reset recovery failure flag on first recovery pass

mmusgrov:
https://issues.jboss.org/browse/JBTM-3017


mmusgrov:
@tomjenkinson Would it be possible to get an early review on this so as to speed things along.
I added extra asserts for the programmatic access that Nicola asked for.
I also added an extra test to validate that the recoveryProblems flag was being reset by calling the test twice (see RecoveryScanTest): once for failure and again for the successful route).

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoveryScanTest.java
```diff
@@ -0,0 +1,167 @@
...
+        ut.commit();
+
+        manager.scan();
+
+        assertFalse(jtaLogger.isRecoveryProblems());
+    }
+
+    @Test
+    public void testBoth() throws Exception{
+        testRecoveryMonitorWithFailure();
```
tomjenkinson: Good test, this is the one I think we need to see where a resource transitions from failing to passing

PR #1344, opened 2018-07-11
Made getTransactionManager() protected instead of private

ljnelson:
This pull request makes the `getTransactionManager()` method in `TransactionContext` `protected` instead of `private`.  In my use case, I have no need of JNDI.  It would be nice to be able to override this method so that I can simply use, for example, `com.arjuna.ats.jta.common.TransactionManager.transactionManager()`.

### DIFF
--- a/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/TransactionContext.java
+++ b/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/TransactionContext.java
@@ -134,7 +134,7 @@ public class TransactionContext implements Context {
         }
     }
 
-    private TransactionManager getTransactionManager() {
+    protected TransactionManager getTransactionManager() {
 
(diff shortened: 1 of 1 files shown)


mmusgrov:
You could use UserTransaction:

UserTransaction ut = com.arjuna.ats.jta.UserTransaction.userTransaction();

ljnelson:
OK; will start a discussion.  For posterity I'm simply trying to get the transactional interceptors to work in CDI SE without JNDI.  All I need for this to work is for the transaction manager to be supplied in a DI-friendly way rather than looked up in JNDI.

jitendrakmwt:
Getting so many emails from today , already unsubscribed and still getting emails.

mmusgrov:
@jitendrakmwt Sorry I was setting up a PR poller job on our upshift installation. I temporarily added a build periodically schedule and it started testing all the PRs. Sorry about that. It is disabled now

PR #1346, opened 2018-07-18
Make TransactionContext and TransactionalInterceptorBase DI-friendly

ljnelson:
This pull request makes the `TransactionContext` and `TransactionalInterceptorBase` classes more dependency-injection friendly so that, for example, JNDI can be an optional rather than a required dependency.  My intention is that these changes are backwards compatible.

A full discussion of the issues involved can be found here: https://developer.jboss.org/thread/278367

The related issue can be found here: https://issues.jboss.org/browse/JBTM-3044

!QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF !RTS  !TOMCAT !mysql !postgres !db2 !oracle

### DIFF
--- a/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/AbstractBean.java
+++ b/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/AbstractBean.java
@@ -1,0 +1,110 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2018, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package com.arjuna.ats.jta.cdi;
+
+import java.lang.annotation.Annotation;
+import java.util.Collections;
+import java.util.Objects;
+import java.util.Set;
+
+import javax.enterprise.context.Dependent;
+import javax.enterprise.context.spi.CreationalContext;
+import javax.enterprise.inject.Default;
+import javax.enterprise.inject.spi.Bean;
+import javax.enterprise.inject.spi.InjectionPoint;
+import javax.enterprise.util.AnnotationLiteral;
+
+/**
+ * An opinionated and skeletal {@link Bean} implementation.  It is, by
+ * default, {@linkplain #getScope() in <code>Dependent</code> scope},
+ * has {@linkplain #getQualifiers() only the <code>Default</code>
+ * qualifier}, has {@linkplain #getStereotypes() no stereotypes},
+ * {@linkplain #isAlternative() is not an alternative}, has
+ * {@linkplain #getInjectionPoints() no injection points} and
+ * {@linkplain #destroy(Object, CreationalContext) does not need
+ * cleanup}.
+ *
+ * <p>Subclasses may override any of the methods in this class to
+ * customize its behavior.</p>
+ *
+ * @author <a href="https://about.me/lairdnelson"
+ * target="_parent">Laird Nelson</a>
+ */
+abstract class AbstractBean<T> implements Bean<T> {
+
+    protected AbstractBean() {
+        super();
+    }
+    
+    @Override
+    public Class<?> getBeanClass() {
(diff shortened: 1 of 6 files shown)


ljnelson:
Completed (1) and (2) above.  My git clumsiness is showing: I'll read up on what (3) means.

ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/AbstractBean.java
```diff
@@ -0,0 +1,100 @@
...
+
+import javax.enterprise.context.Dependent;
+import javax.enterprise.context.spi.CreationalContext;
+import javax.enterprise.inject.Default;
+import javax.enterprise.inject.spi.Bean;
+import javax.enterprise.inject.spi.InjectionPoint;
+import javax.enterprise.util.AnnotationLiteral;
+
+/**
+ * An opinionated and skeletal {@link Bean} implementation.
```
ochaloup: maybe the javadoc comment could be updated to contain the purpose of this class rather than saying that's "an implementation"

ljnelson: Noted; will fix.

ljnelson:
Hi, @tomjenkinson I'd like to try squashing my commits myself so I learn something new (and useful).  It may take me a while since I want to be deliberate and careful about it.

If I'm reading things right I want to use `git rebase -i` on my branch in my forked project to rewrite its commit history, even though strictly speaking this rewrites the (public) history of the analogous branch in Github's copy of my forked project.  I guess the thinking is: if someone is dumb enough to build off of my PR branch and it changes out from under them, they get what they deserve. 😄 

I'll plan to follow [this guide](https://www.devroom.io/2011/07/05/git-squash-your-latests-commits-into-one/) and hopefully won't muck things up too badly.

ljnelson:
OK.  I'll address the javadoc comment issue and then squash and we should be good to go.  Stand by.

ljnelson:
OK; I followed your recipe; let me know if things look good to you.

ljnelson:
Derp; OK; applied your recipe; kindly check again (sorry!).

mmusgrov:
@ljnelson Hi, these changes cause the jta standalone quickstart to fail. It fails with a 'javax.enterprise.inject.CreationException' exception at the line  https://github.com/jbosstm/quickstart/blob/master/jta-1_2-standalone/src/test/java/org/jboss/narayana/quickstarts/jta/TestCase.java#L83 The exception message is

> Not a compound name: TransactionManager

Any thoughts about how we can fix this, is it a test issue or do you think there is a problem with the changes to TransactionExtension#afterBeanDiscovery

ljnelson:
I'll look at this straightaway.  Offhand I don't see anything obvious.  I wonder if there's a JNDI context initialization sequence that I inadvertently caused to get bypassed in this case?

ljnelson:
I can confirm that the `InitialContext` used by `TestCase` indirectly (as part of the `bindJTAImplementation()` call) contains a binding for `java:/TransactionManager`, but the `InitialContext` used internally by `TransactionExtension` et al. does not have such a binding (!).  Given what I know about JNDI this seems very odd: `InitialContext` is effectively a wrapper around a shared store, so at first glance this shouldn't be possible—the call to `bindJTAImplementation()` should be sufficient, as it was in the past when two `InitialContext` instances were being used successfully.

More to the point, I'm not immediately sure how to address this, nor am I sure how simply using an `InitialContext` instance from within CDI somehow causes that binding call not to "stick".  Investigating further.

ljnelson:
Aha; I know what the issue is.

The JNDI tutorial indicates that all `String` names are `CompositeName`s.  But I don't think JBoss' JNDI implementation honors this.

So the string `java:/TransactionManager` is, strictly speaking, a `CompositeName` using [JNDI composite name syntax](https://docs.oracle.com/javase/8/docs/api/javax/naming/CompositeName.html).  It has two parts: one component (`java:`) and another one (`TransactionManager`).

I create just such a `CompositeName` here: https://github.com/jbosstm/narayana/blob/bedaefdb1009dc6e258e4735ab917809d2d715ab/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/TransactionExtension.java#L77

But I think JBoss' JNDI implementation assumes that `Name` implementations passed to `lookup(Name)` will be non-`CompositeName` instances (specifically `CompoundName` instances).  This is strictly speaking a violation of the JNDI specification, but obviously the JBoss JNDI implementation has been around for a long time.

I think the solution will be for me to supply a `String` instead of a `Name` so that JBoss' JNDI implementation can continue creating whatever `Name` implementation it wants, and so that other JNDI implementations can create the ones they want.

I'll submit a new PR shortly.  CC: @tomjenkinson

ljnelson:
Yep; confirming this.  Here's where the JBoss `Context` `lookup(String)` method incorrectly turns a `String` into a `CompoundName` instead of a `CompositeName`: https://github.com/bobmcwhirter/jboss-naming/blob/387ff284a396eaf2fad13941503df7f641ee729a/jnpserver/src/main/java/org/jnp/interfaces/NamingContext.java#L629

That ends up calling this: https://github.com/bobmcwhirter/jboss-naming/blob/387ff284a396eaf2fad13941503df7f641ee729a/jnpserver/src/main/java/org/jnp/interfaces/NamingParser.java#L62

That's incorrect; see for example https://docs.oracle.com/javase/jndi/tutorial/beyond/names/string.html.

ljnelson:
Pull request filed: https://github.com/jbosstm/narayana/pull/1351

emmartins:
I’m not responsible for that naming component, I believe it’s David, we actually need to switch that on JIRA.

@Kabir ^^

PR #1347, opened 2018-07-20
Update README.md

jfclere:
the upstream is now in https://github.com/web-servers/narayana-tomcat


ochaloup:
@tomjenkinson  should not be deleted whole the `src` folder and `pom.xml` and left only the `README` file in the `tomcat` directory which refers to the new location of the code?

PR #1348, opened 2018-07-23
JBTM-3030 Moved Tomcat to https://github.com/web-servers/narayana-tomcat

tomjenkinson:
https://issues.jboss.org/browse/JBTM-3030

!MAIN !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS !mysql !postgres !db2 !oracle !AS_TESTS


zhfeng:
@tomjenkinson https://github.com/jbosstm/narayana/tree/master/tools/src/main/java/io/narayana/db these files have also been moved to the web-server repo. So I think it could be safe to remove.

tomcat/README.md
```diff
@@ -1,3 +0,0 @@
-# Tomcat Narayana testing
-
-The code has been moved to https://github.com/web-servers/narayana-tomcat
```
zhfeng: It could be better to keep this information in the README.md

scripts/hudson/narayana.sh
```diff
@@ -603,11 +603,13 @@ javax.sql.level = ALL
 org.apache.tomcat.tomcat-dbcp.level = ALL
 com.arjuna.level = ALL
 EOT
+    rm -rf narayana-tomcat
+    git clone https://github.com/jbosstm/narayana-tomcat.git
```
zhfeng: I wonder that why we can not use https://github.com/web-servers/narayana-tomcat.git ?

tomjenkinson: Just in case they are ever slot to merge one of our PR. I did have it set to that originally but then I found that test-utils/tomcat-jta order bug which they did merge quickly but I wanted to test this and so it failed. It's a bit like how we test against our 5_BRANCH of wildfly - does that make sense?

zhfeng: OK, I understand.

PR #1349, opened 2018-07-24
[JBTM-2847] refactoring based on reports from static code analysis

ochaloup:
Few minor code changes found by the static code analysis. There are about

- null pointer exception checks
- unused code
- typos

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !TOMCAT !mysql !postgres !db2 !oracle !AS_TEST

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/abstractrecord/RecordTypeManager.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/abstractrecord/RecordTypeManager.java
@@ -67,7 +67,8 @@ public class RecordTypeManager
     {
         for (int i = 0; i < _map.size(); i++)
         {
-            if (_map.get(i).getRecordClass().equals(c))
+            Class<? extends AbstractRecord> recordClass = _map.get(i).getRecordClass();
+            if (recordClass != null && recordClass.equals(c))
                 return _map.get(i).getType();
         }
         
--- a/ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/ConnectionImple.java
+++ b/ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/ConnectionImple.java
@@ -963,7 +963,7 @@ public class ConnectionImple implements Connection
 					getModifier();
 
 					if (_theModifier == null) {
-						jdbcLogger.i18NLogger.info_closingconnectionnull(_theConnection.toString());
+						jdbcLogger.i18NLogger.info_closingconnectionnull(theConnection.toString());
 
 						// no indication about connections, so assume close immediately
 
@@ -982,7 +982,7 @@ public class ConnectionImple implements Connection
 	                         * terminated, so register a Synchronization here.
 	                         */
 
-							jdbcLogger.i18NLogger.debug_closingconnection(_theConnection.toString());
+							jdbcLogger.i18NLogger.debug_closingconnection(theConnection.toString());
 
 							jtaPropertyManager.getJTAEnvironmentBean().getTransactionSynchronizationRegistry().registerInterposedSynchronization(new ConnectionSynchronization(this));
 						}
--- a/ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/recovery/BasicXARecovery.java
+++ b/ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/recovery/BasicXARecovery.java
@@ -41,6 +41,7 @@ import javax.transaction.xa.XAResource;
 import com.arjuna.ats.jdbc.TransactionalDriver;
 import com.arjuna.ats.jdbc.logging.jdbcLogger;
 import com.arjuna.ats.jta.recovery.XAResourceRecovery;
+import com.arjuna.ats.jta.xa.RecoverableXAConnection;
 import com.arjuna.common.util.propertyservice.PropertiesFactory;
 
 /**
@@ -157,7 +158,10 @@ public class BasicXARecovery implements XAResourceRecovery
 			connections = new JDBC2RecoveryConnection[numberOfConnections];
 		}
 		try {
-			return getConnection().recoveryConnection().getResource();
+			RecoverableXAConnection recoveryConnection = getConnection().recoveryConnection();
+			if(recoveryConnection == null)
+				throw new IllegalStateException("Recovery connection of connection " + getConnection() + " is null");
+			return recoveryConnection.getResource();
 		} finally {
 			connectionIndex++;
 		}
--- a/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
+++ b/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
@@ -289,9 +289,9 @@ public class LRAService {
                                     String compensatorData) {
         if (lra ==  null) {
             lraTrace(null, "Error missing LRA header in join request");
+        } else {
+            lraTrace(lra, "join LRA");
         }
-
-        lraTrace(lra, "join LRA");
 
         Transaction transaction = getTransaction(lra);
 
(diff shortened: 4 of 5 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/abstractrecord/RecordTypeManager.java
```diff
@@ -67,7 +67,8 @@ public synchronized int getType (Class<? extends AbstractRecord> c)
     {
         for (int i = 0; i < _map.size(); i++)
         {
-            if (_map.get(i).getRecordClass().equals(c))
+            RecordTypeMap typeMap = _map.get(i);
```
mmusgrov: you could avoid the second call to getRecordClass by doing
```
AbstractRecord rec = _map.get(i).getRecordClass()
```
unless you think _map.get(i) could return null in which case you need to be checking for that too.

tomjenkinson: I am not altogether sure of what is going on here, is it possible for typeMap to be null in practice? If not perhaps we should add an exclusion for this bit. If it is possible to be null then I think we would report that elsewhere too if possible

ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/recovery/BasicXARecovery.java
```diff
@@ -157,7 +158,10 @@ public synchronized XAResource getXAResource () throws SQLException {
 			connections = new JDBC2RecoveryConnection[numberOfConnections];
 		}
 		try {
-			return getConnection().recoveryConnection().getResource();
+		    RecoverableXAConnection recoveryConnection = getConnection().recoveryConnection();
+		    if(recoveryConnection == null)
+		        throw new IllegalStateException("Recovery connection of connection " + getConnection() + " is null");
```
tomjenkinson: I assume the XARM will be able to deal with this OK?

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/abstractrecord/RecordTypeManager.java
ochaloup: @mmusgrov agree, thanks!
@tomjenkinson  the issue here is that the `getRecordClass()` can return null, e.g. for https://github.com/jbosstm/narayana/blob/master/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java#L58

ochaloup: @mmusgrov agree, thanks!
@tomjenkinson the issue here is that the getRecordClass() can return null, e.g. for https://github.com/jbosstm/narayana/blob/master/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java#L58

ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/ConnectionImple.java
```diff
@@ -963,7 +963,7 @@ protected final synchronized void registerDatabase() throws SQLException
 					getModifier();
 
 					if (_theModifier == null) {
-						jdbcLogger.i18NLogger.info_closingconnectionnull(_theConnection.toString());
```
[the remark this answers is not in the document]
ochaloup: the point here is that the `_theConnection` could be `null`: https://github.com/jbosstm/narayana/blob/master/ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/ConnectionImple.java#L896 but `theConnection` is won't be null (the `getConnection()` is called and `if` statement ensures it). That way I thought using `theConnection` will be safer. There could be used global `_theConnection` but I think it should be not called `toString()` on it but just provided as a nullable object to the logger?

ArjunaJTA/jdbc/classes/com/arjuna/ats/internal/jdbc/recovery/BasicXARecovery.java
ochaloup: @tomjenkinson yes, I assume so. The XARM catches any exception and logs it (https://github.com/jbosstm/narayana/blob/master/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java#L551) (e.g. `#getXAResource` throws the SQLE). I think let the NPE being thrown non-intentionally should be avoided thus I changed it to explicitly stating the state is illegal/unexpected. But maybe the NPE could be left here.

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
```diff
@@ -289,10 +289,10 @@ public synchronized int joinLRA(StringBuilder recoveryUrl, URL lra, long timeLim
                                     String compensatorData) {
         if (lra ==  null) {
             lraTrace(null, "Error missing LRA header in join request");
+        } else {
+            lraTrace(lra, "join LRA");
```
ochaloup: hi @mmusgrov, if you don't mind I have one more doubt on here on the `LRAService`. From the check it seems the `lra` variable could be `null` but then the lra is used in way it should not be null (see https://github.com/jbosstm/narayana/blob/master/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java#L309). I refer to the call `participant = transaction.enlistParticipant(lra, ...)` where `enlistParticipant` provides the `lra` to the next method which calls `coordinatorUrl.toExternalForm()` and could end with NPE possibly.

ochaloup:
@tomjenkinson thanks Tom, I have a final doubt about the LRAService and the NPE situation: https://github.com/jbosstm/narayana/pull/1349#pullrequestreview-139984983. I asked for help @mmusgrov via email. I expect this to merge soon.

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
mmusgrov: I added the trace method during the early prototype and can now be safely deleted

ochaloup: @mmusgrov ok, I did it so. Thank you for the explanation.

mmusgrov: @ochaloup I meant that the lra_trace method can now be removed from the source file. Altrnatively just add the null check at line 340:

> if (lraId != null && lras.containsKey(lraId)) {

ochaloup: @mmusgrov I see, thanks for the hint. I think having the trace log helps during issue investigation. I think your alternative approach is better.

On top of this my worry was about `lra` being null at call `enlistParticipant` (https://github.com/jbosstm/narayana/blob/master/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java#L309). If `lra` is null it leaks to the method and will `NPE` will be thrown somewhere during execution. What to throw NPE directly here? https://github.com/jbosstm/narayana/blob/master/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java#L291

ochaloup:
I discussed my doubts with Mike, I did a change to adjust the code based on Mike's advice and now I just wait for the CI to run and then I expect to merge this.

PR #1351, opened 2018-07-24
JBTM-3045: Fixes Standalone JTA 1.2 Quickstart failure

ljnelson:

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF !RTS !TOMCAT !mysql !postgres !db2 !oracle

### DIFF
--- a/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/JNDIBean.java
+++ b/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/JNDIBean.java
@@ -48,14 +48,18 @@ import javax.naming.NamingException;
  */
 final class JNDIBean<T> extends AbstractBean<T> {
 
-    private final Name name;
+    private final String name;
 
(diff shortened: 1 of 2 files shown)


ljnelson:
This pull request should fix [JBTM-3045](https://issues.jboss.org/browse/JBTM-3045).

The discussion can be found [in the conversation section of a prior pull request](https://github.com/jbosstm/narayana/pull/1346#issuecomment-407431157).

ochaloup:
I would like just mentioned that the `AS_TEST` failures is because of cluster test `HotRodPersistenceWebFailoverTestCase`. I tested that on my machine and the test passes. This seems to be a testsuite instability. But strangely the WildFly CI passes without any issue on the same test.

mmusgrov:
@tomjenkinson the hanging test was an old performance test. I will check that we have something equivalent in the perf repo and then delete/move the test.

PR #1353, opened 2018-08-03
[JBTM-3047] adjusting stopping listener for avoid deadlock on client scan call

ochaloup:
https://issues.jboss.org/browse/JBTM-3047

!QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !PERF !mysql !postgres !db2 !oracle

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/Listener.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/Listener.java
@@ -190,21 +190,33 @@ public class Listener extends Thread
         notifyAll();
     }
 
+    /**
+     * <p>
+     * Ensuring the socket does not accept more requests by closing it
+     * and setting flag to inform the listener is stopped.
+     * <p>
+     * Method does not close opened connections, run the {@link #stopListener()} method for that.
+     */
+    public synchronized void closeListenerSockets() {
+        _stop_listener = true;
+
+        try
+        {
+            _listener_socket.close();  // in case we're still in accept
+        }
+        catch (final Exception ex)
+        {
+        }
+    }
+
    /**
     * Halts running of the listener thread.
     */
 
    public synchronized void stopListener()
    {
-      _stop_listener = true;
+       closeListenerSockets();
 
-       try
-       {
-           _listener_socket.close();  // in case we're still in accept
-       }
-       catch (final Exception ex)
-       {
-       }
       // there is no need for this as the close will interrupt any i/o that is in progress
       // this.interrupt();
 
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/PeriodicRecovery.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/PeriodicRecovery.java
@@ -164,9 +164,9 @@ public class PeriodicRecovery extends Thread
      */
    public void shutdown (boolean async)
    {
-       // stop the lsitener from adding threads which can exercise the worker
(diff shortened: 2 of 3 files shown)


ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoveryMonitorTestDeadlock.java
```diff
@@ -0,0 +1,133 @@
...
+        recoveryModule.addXAResourceRecoveryHelper(xaResourceRecoveryHelper);
+        manager.addModule(recoveryModule); // we only need to test the XARecoveryModule
+        manager.startRecoveryManagerThread(); // start periodic recovery
+
+        String host = recoveryEnvironmentBean.getRecoveryAddress(); // the recovery listener host
+        String rcPort = String.valueOf(recoveryEnvironmentBean.getRecoveryPort()); // the recovery listener port
+
+        // running the recovery scan which will be delayed by sleepingXAResource
+        manager.scan(null);
+        countdown.await(TEST_TIMEOUT_S, TimeUnit.SECONDS);
```
mmusgrov: So you trigger a background scan and then wait for 1 second for the latch to hit zero
(and PeriodicRecovery will take at least one second because the XAR#recovery call sleeps for 1 second).
I would expect that the countdown.await call would timeout rather than hit zero in this case. Can you add an assert here to indicate what your expectation is.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoveryMonitorTestDeadlock.java
```diff
@@ -0,0 +1,133 @@
...
+
+        // sending client SCAN request while there is still running the recovery process
+        // the timeout on socket is defined so small that the call immediately returns here
+        // but SCAN invocation is already happening inside of the Naryaana
+        RecoveryMonitor.main(new String[] {"-verbose", "-port", rcPort, "-host", host, "-timeout", "10"});
```
mmusgrov: You have triggered a scan via a socket with an SO timeout of 10 msecs which will leave a WorkerService running inside periodic recovery. Next you trigger a RecoveryManager suspend and shutdown in a background thread (and wait 5 seconds for that to complete) and finally you validate that the shutdown request isn't deadlocked.

This all makes sense but I am not clear about how the countdown hits zero without the await call timing out?

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoveryMonitorTestDeadlock.java
ochaloup: yes, pretty good point, thanks. I will change this.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoveryMonitorTestDeadlock.java
ochaloup: the current countdown (based on your previous comment I will add there one more to not depend on timeout in recovery) hits zero as there is first call `manager.scan(null)` which starts recovery. Then the `@Test` waits on countdown to ensure the `XAResource.recover()` is in progress and that's where the countdown is set to zero. After that the `RecoveryMonitor.main` is called.

ochaloup:
@mmusgrov I updated the test for not being dependent on timeout inside of the `XAResource.recover()` method and add the check you recommended.

PR #1363, opened 2018-08-22
[JBTM-3053] printing to debug the node identifier which is used

ochaloup:
for xa orphan detection filters

https://issues.jboss.org/browse/JBTM-3053

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS !TOMCAT !mysql !postgres !db2 !oracle

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
@@ -33,6 +33,7 @@ import org.jboss.tm.usertx.UserTransactionOperationsProvider;
 import org.jboss.tm.usertx.client.ServerVMClientUserTransactionOperationsProvider;
 
 import com.arjuna.ats.internal.jta.resources.arjunacore.XAResourceRecordWrappingPlugin;
+import com.arjuna.ats.jta.logging.jtaLogger;
 import com.arjuna.ats.jta.recovery.XAResourceOrphanFilter;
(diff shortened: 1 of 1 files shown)


ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
```diff
@@ -410,6 +411,11 @@ else if(instance != oldInstance)
      */
     public void setXaRecoveryNodes(List<String> xaRecoveryNodes)
     {
+        if(jtaLogger.logger.isDebugEnabled()) {
+            jtaLogger.logger.debugf("Setting up node identifiers '%s' for which recovery"
+                + " will be performed", xaRecoveryNodes);
```
mmusgrov: Does putting a List<String> here actually print out each element of the list?

ochaloup: @mmusgrov it depends on the `toString()` method of the `List` implementation and the `ArrayList` is used here which really prints all the records. This is not true e.g. for arrays in java but from my experience it's true for all the collection java classes in the java util package.

mmusgrov: System.out.println("%s", arrayList) prints out each element as you suggest so I agree and would expect any logger should do the same.

ochaloup:
Thanks Mike.

AS tests fails on LDAP security tests.

PR #1368, opened 2018-10-01
Don't run tests that require WildFly for the MAIN build axis

tomjenkinson:
!QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF !RTS !TOMCAT !mysql !postgres !db2 !oracle !JACOCO !QA_JTA


ochaloup:
I think there is several integration tests in the Narayana which depend on the WFLY to be built (CDI is one of them). Should not we consider flags `AS_BUILD` and `AS_TEST`? For the `AS_BUILD` there will be run only the build with `-DskipTests` to get `JBOSS_HOME`.

PR #1376, opened 2018-11-01
JBTM-2685 update to make sure the openjdk qa tests running with the jdk9

zhfeng:
!QA_JTS_JACORB !BLACKTIE !XTS !PERF !RTS !TOMCAT !mysql !postgres !db2 !oracle !JACOCO !QA_JTA NO_WIN !AS_TESTS

I just need to check QA_JTS_OPENJDK and QA_JTS_JDKORB. It should be OK to just put the openjdk-orb in the classpath and no need to use the --patch-module. I will close the [WFLY-9569](https://issues.jboss.org/browse/WFLY-9569) if this PR is merged.

### DIFF
--- a/scripts/hudson/narayana.sh
+++ b/scripts/hudson/narayana.sh
@@ -729,11 +729,12 @@ function qa_tests_once {
   sed -e "s#^COMMAND_LINE_0=.*#COMMAND_LINE_0=${JAVA_HOME}/bin/java#" TaskImpl.properties > "TaskImpl.properties.tmp" && mv "TaskImpl.properties.tmp" "TaskImpl.properties"
   [ $? = 0 ] || fatal "sed TaskImpl.properties failed"
 
+  # store the origin orbtype
+  origin_orbtype=orbtype
+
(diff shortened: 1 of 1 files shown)


zhfeng:
@mmusgrov 

> Why the change in name from orbtype to orb.type.
Because the ant does not modify the property value as it is immutable. And I have to change the orbtype to "idlj" if the value is "openjdk" in the build.xml and build-jts.xml
We just do this change in the script/hudson/narayana.sh before.

mmusgrov:
@zhfeng I was referring to "the change in the name of the property from orbtype to orb.type", I am okay with the change if you think it isn't reference from anywhere else but it seemed an unnecessary change that does carry risk.

mmusgrov:
@zhfeng The property value can be changed by passing in the value externally (it only becomes immutable once the ant script sets the value).

Also what is the relationship between two properties orbtype and orb.type (it needs to be documented in the scripts otherwise this will be very difficult code to maintain).

zhfeng:
Thanks @mmusgrov I will change it in the narayana.sh

PR #1387, opened 2018-11-29
[JBTM-3079] inboundbridge recovery abort

ochaloup:
Applying Jonathan's fix and adding a test for it (+a little bit more error messages for `XAException`)
Please, let me know objections about the test or `XAException` messages.

_NOTE:_ I would squash the two commits when it's approved to be fine

/cc @jhalliday 

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !PERF !RTS  !TOMCAT !mysql !postgres !db2 !oracle


txbridge/src/main/java/org/jboss/jbossts/txbridge/inbound/InboundBridgeManager.java
```diff
@@ -135,13 +146,16 @@ private static synchronized void createMapping(String externalTxId)
...
         String participantId = org.jboss.jbossts.txbridge.inbound.BridgeDurableParticipant.TYPE_IDENTIFIER+new Uid().toString();
         transactionManager.enlistForDurableTwoPhase(bridgeDurableParticipant, participantId);
 
         BridgeVolatileParticipant bridgeVolatileParticipant = new BridgeVolatileParticipant(externalTxId, xid);
         transactionManager.enlistForVolatileTwoPhase(bridgeVolatileParticipant, new Uid().toString());
 
-        inboundBridgeMappings.put(externalTxId, new InboundBridge(xid));
+        InboundBridge inboundBridge = new InboundBridge(xid);
+        inboundBridgeMappings.put(externalTxId, inboundBridge);
+        liveXids.add(inboundBridge.getXid());
```
tomjenkinson: My experience of XATerminator is that a recovery scan might make (certain - not necessarily these) subordinate transactions live again - I don't know if it would be a problem for you if it was recovered into memory by the recovery manager but wasn't added into the liveXids map?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/XATerminatorImple.java
```diff
@@ -94,23 +94,32 @@ public void commit (Xid xid, boolean onePhase) throws XAException
             SubordinateTransaction tx = SubordinationManager
                     .getTransactionImporter().getImportedTransaction(xid);
 
-            if (tx == null)
-                throw new XAException(XAException.XAER_INVAL);
+            if (tx == null) {
+                XAException xaException = new XAException("No subordinate transaction to  drive for commit with xid: " + xid);
```
tomjenkinson: I think we should localise the exception messages

ochaloup: ah, thanks! I haven't realized that.

txbridge/src/main/java/org/jboss/jbossts/txbridge/inbound/InboundBridgeManager.java
ochaloup: I do not know all the details in what circumstances the `XATerminator` make the subordinate transaction live again.
My understanding is that if transaction is already handled by recovery manager it will never be added under some thread of business logic. It stays to be handled by recovery manager at the end. The purpose for the `liveXids` is the set of xids which are active for in-memory running transactions.
But honestly I could be missing something.

tomjenkinson: I think the question is rather what is responsible for recovering the transactions. I do not think XATerminator will be in this case, it is just quite a similar pattern. If it is always top-down, what method recovers this bridges on this server? That should be checked to make sure it hits this liveXids.add() call...

ochaloup: @tomjenkinson I think you are right and I'm learning it in hard way.

But first. I still think the purpose of the `liveXids` is to store what's happening in-memory. When server crashes there should not be any clash between transaction manager and recovery manager. But... well, maybe

...I created a [new test case](https://github.com/ochaloup/narayana/commit/091beaaca26ee5eae3e7c42bd13e45302751e78e#diff-4b0f88e8bd98f6704571555b0dbe4712R152) which was not part of the testsuite and it does not pass neither with the updated behaviour or original one. That's strange and I struggle to find out if it's an issue of the code or test.

The test should be like having a client deployed on the WFLY. Testcase calls the client which starts JTA transaction and then invokes WS to the WFLY. This webservice did some work by enlisting a mock TestXAResource in the business method. Then 2PC starts, the prepare phase passes (there is  `BridgeDurableParticipant`, `BridgeVolatileParticipant` and `TestXAResource`, and JVM crashes.
Then during recovery the `ATParticipantRecoveryModule` runs `ParticipantEngine`[1] to resend prepare and all goes wrong (https://github.com/jbosstm/narayana/blob/master/XTS/WS-T/dev/src/com/arjuna/wst11/messaging/engines/ParticipantEngine.java#L315).
The calls ends at the [`CoordinatorProcessorImpl`](https://github.com/jbosstm/narayana/blob/master/XTS/WS-T/dev/src/com/arjuna/wst11/messaging/CoordinatorProcessorImpl.java#L150), it says there is no active coordinator (which is correct I expect as it's recovery) and goes with rollback. That's where I ended so far. I need to investigate more deeper to understand. If you would have some pointer for me, that would be great! Thanks.

[1]
```
Thread [Periodic Recovery] (Suspended)
	ParticipantEngine.recovery() line: 315
	ATParticipantRecoveryRecord.activate() line: 68
	XTSATRecoveryManagerImple.recoverParticipants() line: 219
	ATParticipantRecoveryModule.processParticipantsStatus() line: 265
	ATParticipantRecoveryModule.periodicWorkSecondPass() line: 137
	PeriodicRecovery.doWorkInternal() line: 816
	PeriodicRecovery.run() line: 382
```

tomjenkinson: It's not an area I am totally familiar with but what I could say is that for XATerminator there has to be a method to reload the subordinate transactions from disk into memory in this case it would be when the prepare was resent if there was nothing in memory we would (if the analogy holds) determine that there is nothing in memory and then try to reload it from disk rather than go to rollback.

ochaloup: @tomjenkinson I see. On my way of investigating on this I would say that it's desirable that the `liveXids` are not updated during recovery as it's purpose is a defence against the recovery manager to touch the `Xid` which is active in context of WS inbound message (https://github.com/jbosstm/narayana/blob/master/txbridge/src/main/java/org/jboss/jbossts/txbridge/inbound/InboundBridgeManager.java#L59)

From that I vote for accepting this PR.
@tomjenkinson do you think I should provide something more here to get this PR accepted?

I will keep the other part of my doubt seems to be a separate problem and I would leave it as a new issue (or rather a forum topic) when this PR is resolved.

tomjenkinson: Thanks for checking that Ondra, it must be different to the way that XATerminator/EJBremoting interact then

PR #1391, opened 2018-12-04
[JBTM-3138] XTS txbridge does not run commit during recovery when 1PC is used

ochaloup:
https://issues.jboss.org/browse/JBTM-3138
https://issues.jboss.org/browse/JBTM-3147

A follow-up to https://issues.jboss.org/browse/JBTM-3079 where active xids were considered to not be processed by recovery manager.

Here adding a testcase which crashes after prepare phase is finished and I would expect the commit should be processed. Discussion on this topic is at the forum https://developer.jboss.org/thread/279243

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE NO_WIN !RTS !AS_TESTS !TOMCAT !JACOCO !LRA !PERF
!MAIN XTS

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
@@ -3677,11 +3677,14 @@ public class BasicAction extends StateManager
     {
         if (TxControl.onePhase)
         {
-            return (((pendingList == null) || (pendingList.size() == 1)) ? true
-                    : false);
+            if(pendingList == null) {
+                return true;
+            }
+            if (pendingList.size() == 1) {
+                return pendingList.peekFront().isPermittedTopLevelOnePhaseCommit();
+            }
         }
-        else
-            return false;
+        return false;
     }
 
     /*
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/XATerminatorImple.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/XATerminatorImple.java
@@ -196,8 +196,11 @@ public class XATerminatorImple implements javax.resource.spi.XATerminator, XATer
             SubordinateTransaction tx = SubordinationManager
                     .getTransactionImporter().getImportedTransaction(xid);
 
-            if (tx == null)
-                throw new XAException(XAException.XAER_INVAL);
+            if (tx == null) {
+                XAException xaException = new XAException(jtaLogger.i18NLogger.get_no_subordinate_txn_for("forget", xid));
+                xaException.errorCode = XAException.XAER_INVAL;
+                throw xaException;
+            }
 
             tx.doForget();
         }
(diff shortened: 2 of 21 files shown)


ochaloup:
@tomjenkinson my observation is that the 1PC for the XTS is not implemented correctly. Based on the existence of the https://issues.jboss.org/browse/JBTM-396 the implementation was planned but not implemented. The code of the https://github.com/jbosstm/narayana/blob/5.9.5.Final/XTS/WSTX/classes/com/arjuna/mwlabs/wst/at/participants/DurableTwoPhaseCommitParticipant.java#L253 shows that method confirmOnePhase simulates the 2PC (`prepare`, `confirm`) but in the wrong way. The prepare phase does not store anything to the object store. Thus the recovery can't  behave correctly after JVM crash.

My suggestion is to disable the 1PC for XTS records. I took similar approach as recommended by you at https://developer.jboss.org/message/989130#989130.

Can you please check what is your opinion on this?

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
```diff
@@ -3677,11 +3677,14 @@ private final synchronized boolean doOnePhase ()
     {
         if (TxControl.onePhase)
         {
-            return (((pendingList == null) || (pendingList.size() == 1)) ? true
-                    : false);
+            if(pendingList == null) {
+                return true;
+            }
+            if (pendingList.size() == 1) {
+                return pendingList.peekFront().isPermittedTopLevelOnePhaseCommit();
```
tomjenkinson: I expect we could mix records in one transaction, should not the whole list be checked?

tomjenkinson: Sorry, pending list == 1, I see it now

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/XATerminatorImple.java
```diff
@@ -95,7 +95,7 @@ public void commit (Xid xid, boolean onePhase) throws XAException
                     .getTransactionImporter().getImportedTransaction(xid);
 
             if (tx == null) {
-                XAException xaException = new XAException(jtaLogger.i18NLogger.get_no_subordinate_txn_for_commit(xid));
+                XAException xaException = new XAException(jtaLogger.i18NLogger.get_no_subordinate_txn_for("commit", xid));
```
mmusgrov: Shouldn't we deprecate the log message first before removing it?

tomjenkinson: Agreed - let's go back to the original message for this one as people may be using the error code

ochaloup: @mmusgrov @tomjenkinson yeap, agree, I returned back the log message.

ochaloup:
@mmusgrov @tomjenkinson I managed to fix the test failures and added several enhancements for logging. I fixed the objection about the log deprecation.
Would you be so kind to re-review this PR?

PR #1403, opened 2019-02-11
[JBTM-3106][WFLY-11858] Fire appropriate scope events for TransactionScoped scope. Remove bean for InitialContext.

ljnelson:

This pull request:

1. Corrects a few generics issues in `TransactionContext`
2. Introduces the `DelegatingTransactionManager` abstract class
3. Introduces the `NarayanaTransactionManager` concrete class that extends from `DelegatingTransactionManager`
   1. This class fires `@Initialized(TransactionScoped.class)` from its `begin()` method
   2. This class fires `@Destroyed(TransactionScoped.class)` from its `commit()` and `rollback()` methods
4. Moves JNDI detection logic from `TransactionExtension` into `NarayanaTransactionManager`

The original forum discussion behind all this is here: https://developer.jboss.org/thread/279595

https://issues.jboss.org/browse/JBTM-3106
https://issues.jboss.org/browse/WFLY-11858

Requires: https://github.com/jbosstm/jboss-as/pull/80


ljnelson:
@ochaloup I see a few tests failed; I am not a Red Hat/IBM employee so I cannot see why that would be.  This PR builds the `cdi` artifact fine.  Let me know if there are changes I need to make.

mmusgrov:
@ljnelson The failures are know issues (jacoco not pulling in LRA snapshot builds and a WildFly integration test issue) and are unrelated to your PR so can be safely ignored.

ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/NarayanaTransactionManager.java
```diff
@@ -0,0 +1,294 @@
...
+        // bean types.
+        //
+        // Action: create a new InitialContext by hand with no
+        // arguments.
+        
+        InitialContext temp = null;
+        try {
+          temp = new InitialContext();
+        } catch (final NamingException namingException) {
+          // TODO: log
```
mmusgrov: If this exception is thrown then the code on line 156 will blow up because ic is set to the value null in the finally block.

mmusgrov:
@ljnelson the patch looks good. I just had one query about InitialContext

ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/NarayanaTransactionManager.java
ljnelson: Nice catch!

mmusgrov:
@ljnelson We occasionally get false positives on the performance test axis and the failure is unrelated to your code changes so you can ignore that failure.
The failure on the windows test run is also unrelated and I have raised https://issues.jboss.org/browse/JBTM-3105 to look into it.

ochaloup:
@ljnelson perfect. Thanks a lot. If you won't be angry with me I created a [JIRA](https://issues.jboss.org/browse/JBTM-3106) for us being able to track the change in the release. If you don't mind I assigned that to you.
Up to that I changed the description of this PR to not running all our CI tests.
For merging, please, change the commit message by prefixing it with `[JBTM-3106]`.

ljnelson:
Done!

ljnelson:
Hi; just noticed your comment.  I'll digest and figure out what to do.

ljnelson:
Hi, @ochaloup; regarding your second issue, would the replacement of this line:

https://github.com/jbosstm/narayana/blob/98a59c38b2f8c7c5164a7256c60e03951ee9f137/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/NarayanaTransactionManager.java#L177

…with the logic found here:

https://github.com/jbosstm/narayana/blob/98a59c38b2f8c7c5164a7256c60e03951ee9f137/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/NarayanaTransactionManager.java#L124-L166

…do what you want?  Obviously if I make this change I'd refactor as well.

ljnelson:
Hi, @ochaloup; see what you think about my latest round of changes.

ochaloup:
@ljnelson yeap, this seems good now. I did some quick testing and works fine for me. Thank you.
Do you think you could add some test for this PR, please?

ljnelson:
@ochaloup Will do.  I am nervous about all the test failures reported by jbosstm-bot; everything is actually OK?

ochaloup:
@ljnelson no, unfortunately everything is not ok, our CI went mad and is a bit unstable :-)

I'm sorry for not pointing that out. Everything is fine with your PR. I was checking the CI test runs, particularly those important to this PR - aka. Narayana main testsuite and WFLY testsuite, and all runs without failures.

I made adjustment to fix our mad JDK11 runs and I hope your next push to this PR will run CI cleanly.

ochaloup:
@ljnelson as I reviewed the failed WFLY tests. They seem to be caused by misconfigured dependencies in the tests (https://github.com/jbosstm/jboss-as/pull/80). I've started the CI to see.

ochaloup:
@ljnelson the tests pass. The only failure in `AS_TESTS` is intermittent (`ArchiveDefaultScriptNamesTestCase#testDeployUndeployArchive`) not relevant to this PR.

ljnelson:
I note that unit tests are skipped by default in the `cdi` subproject's `pom.xml`, and tests are clearly meant to be run by Arquillian.  I've tried doing `mvn test -Parq` but all unit tests fail if I do that.  I am no kind of Arquillian expert.  How can I run my tests locally, or should I just let CI do it for me?

ochaloup:
@ljnelson you are right, it's not clear. Sorry about that. I created an issue https://issues.jboss.org/browse/JBTM-3119 to document at least a bit about this.

Nevertheless, you are right the CDI tests are integration one and using arquillian profile `-Parq` to be started. They are deactivated by default. You need two things to get them running

1. activate with `-Parq`
2. [http://wildfly.org/downloads/](download WildFly), unpack it to some directory and set up `JBOSS_HOME` env variable to point to that directory, `export JBOSS_HOME=...`
3. `cd ArjunaJTA/cdi/; mvn test -Parq`

Let me know if you have any issue with Arquillian I will try to respond about it (btw. for Arquillian I like this tutorial https://rpestano.wordpress.com/2014/06/08/arquillian/).

Sorry for inconveniences and thank you!

ljnelson:
Thanks; one last dumb question.

In my test that I'm writing, I would very much like to do:

    @Inject private TransactionManager tm;

…since that's part of what this PR allows.  But it would seem in this case that injection is being performed by Wildfly, using whatever transaction engine is installed in Wildfly (presumably an older, released version of Narayana).  I want this injection to be satisfied by the Narayana classes that are under test.  Other tests in this directory simply acquire a `TransactionManager` using the `jtaPropertyManager`.

What must I do to test the CDI features of the `cdi` subproject?

ljnelson:
For reference, here is the test that I have at the moment: 

https://gist.github.com/ljnelson/7581a3f55dbb0f027e6463bd60033583

My interpretation of the fact that this does not succeed is that the transaction machinery in play in Wildfly as exercised by Arquillian is not the transaction machinery in this PR.  Adding the transaction machinery in this PR to the ShrinkWrap archive obviously doesn't work either, as Wildfly (I'm guessing) doesn't look at those classes first in its classloader hierarchy.

If I were writing tests myself, I'd of course just start a CDI container by hand and ignore Arquillian entirely; testing in this case would be simple, straightforward and properly isolated from other side effects.  Is that a reasonable thing to do in this case?

ljnelson:
Hi, @tomjenkinson; I sort of see.  😄 

What…so what should I do here, exactly?  If I'm reading that script right, then CI/CD will run the test properly with a hefty amount of shell scripting mojo (once I check it in) and although it doesn't run via a simple `mvn test -Parq` invocation, that is OK because really the CI/CD pipeline is the authoritative tester anyhow.  That suggests that I should ensure my test compiles, and then should check it in and let the CI/CD pipeline tell me what mistakes I've made.  Is that about right?

ochaloup:
@ljnelson I think the easiest way is to compile the WFLY instance as whole (but it takes a while)

```
# make changes in Narayana code and build the whole project
./build.sh clean install -Pcommunity -DskipTests -Didlj-enabled=true
# take the sources of WFLY and build that with the Narayana that was just built
git clone https://github.com/wildfly/wildfly  
cd wildfly  
./build.sh install -B -DskipTests  -Dversion.org.jboss.narayana=5.9.4.Final-SNAPSHOT
# use the built WFLY as the JBOSS_HOME for the integration tests
export JBOSS_HOME="$PWD/dist/target/wildfly-17.0.0.Beta1-SNAPSHOT"
# run the tests
cd $NARAYNA_HOME/ArjunaJTA/cdi/
mvn test -Parq
```

If you need to do more changes and test how that works check the document here https://developer.jboss.org/wiki/BuildingNarayanaAndUpdatingWildFlyModules with few tips. And let us know if there is some issue with that.

ljnelson:
OK.  I am concerned about checking in a test that will break unless it is run in Wildfly.  What I mean is, if someone were to specify `-DskipTests=false` and do `mvn test` my Arquillian-based test would fail, right?  That seems wrong.  If that's how it's done, I will conform, but perhaps in this case a simple standalone CDI container test would be better, since all that's being tested are CDI scope events and bean discovery?

ljnelson:
Right, I understand.  My point is, profile or no, I can do `-DskipTests=false` and I'd expect many tests to break.  I'd really like it if the tests I add do not fall victim to this problem.  If they have to, they have to, but I think I can do better.

I'll look to see if I can write my test using the standalone features of Weld instead and then we can discuss it here in this ever-more-impressive PR thread.  😄

ljnelson:
With appropriate imports and using the version of Weld governed by the `${version.org.jboss.weld}` property, this test works: https://gist.github.com/ljnelson/15f0dfe964941538020d9f8fde4fd805

ljnelson:
Hello, @tomjenkinson, @ochaloup; I've added a test to my PR that uses Weld standalone.

ochaloup:
@ljnelson our CI fails with the error. Can you please check the error below?

```
[ERROR] org.jboss.as.test.integration.microprofile.opentracing.EarOpenTracingWithWeldProbeTestCase  Time elapsed: 0.213 s  <<< ERROR!
org.jboss.arquillian.container.spi.client.container.DeploymentException: 
Cannot deploy EarOpenTracingTestCase.ear: {"WFLYCTL0062: Composite operation failed and was rolled back. Steps that failed:" => {"Operation step-1" => {"WFLYCTL0080: Failed services" => {"jboss.deployment.unit.\"EarOpenTracingTestCase.ear\".WeldStartService" => "Failed to start service
    Caused by: org.jboss.weld.exceptions.DeploymentException: WELD-001436: Type com.arjuna.ats.jta.cdi.NarayanaTransactionManager is not proxyable because it has a private constructor [EnhancedAnnotatedConstructorImpl] @Inject private com.arjuna.ats.jta.cdi.NarayanaTransactionManager(BeanManager, @Initialized Event<Transaction>, @Destroyed Event<Object>) - class com.arjuna.ats.jta.cdi.NarayanaTransactionManager."}}}}
```

> OK. I am concerned about checking in a test that will break unless it is run in Wildfly. What I mean is, if someone were to specify -DskipTests=false and do mvn test my Arquillian-based test would fail, right? That seems wrong. If that's how it's done, I will conform, but perhaps in this case a simple standalone CDI container test would be better, since all that's being tested are CDI scope events and bean discovery?

> With appropriate imports and using the version of Weld governed by the ${version.org.jboss.weld} property, this test works: https://gist.github.com/ljnelson/15f0dfe964941538020d9f8fde4fd805

> It is also (understandably) much, much faster than the Arquillian-based analog.

For the first part that's not right, at least I think :) If you specify `-DskipTests=false` then arquillian tests are still not run. You need to activate them with `-Parq` profile for them to start running.

I understand your point about the CDI container. Feel free to create a JBTM jira on enhancement of the behaviour.
But... the WFLY is one of the main targets for Narayana integartion and we need a profile that would run these tests with WFLY too.
Arquillian provides the embedded Weld provider (http://weld.cdi-spec.org/news/2017/02/01/tip4-testing/) so it could be a chance to check that. If that works it could be the best as the tests could be left as they are only there could be added the embedded WELD testing as an option to the WFLY one.

I agree that sole Weld could be much faster but with Arquillian you can possibility to run with different containers which is handy for us. The embedded weld (if works) still could be sufficiently faster in comparison with WFLY.

ljnelson:
@ochaloup Sure; I'll make these changes.

I find the error interesting, particularly given http://docs.jboss.org/cdi/spec/1.2/cdi-spec.html#unproxyable.  As you can see from my PR, the `NarayanaTransactionManager` _does_ have a non-`private` zero-argument constructor.  It's obviously also proxyable at least _some_ of the time, or _none_ of the tests would work.  I'll make its other constructor non-`private` as well and see if that does the trick.

ljnelson:
@ochaloup Sigh; turns out that the Arquillian Weld-embedded container is not actually a CDI container; it just fakes injection.  So among other things the method parameter injection points that use `BeanManager` don't work.  That would be a step backwards.

I think that since I've proven the test works, I'll check it in with `@RunWith(Arquillian.class)` but I'll just leave the existing Arquillian container alone and see if that works.  Here's hoping!

ochaloup:
@ljnelson great, it seems all works  fine now.

> I find the error interesting, particularly given http://docs.jboss.org/cdi/spec/1.2/cdi-spec.html#unproxyable. As you can see from my PR, the NarayanaTransactionManager does have a non-private zero-argument constructor.

The test which were failing was test of the EAR archive in the WildFly testsuite. I had some problematic experience with the EAR deployments as there was some issue of `BeanManager` and classloader. It could be some WFLY classloading "feature" or kind of clash of EE spec and CDI. I'm not sure. But as it works with protected constructor I'm personally fine with it as you did.

Regarding the weld container. I think  WELD  embedded `<artifactId>arquillian-weld-embedded</artifactId>` should work but some dependency management would be needed. I just quickly tried that and I got to the spiral of `Unsatisfied dependencies` failures. I think the arquillian get the weld core and then there needs to be defined correct dependencies for the core have all what is needed to construct the test application. Not sure though.

Either way the current state of the PR seems good to me. Do you think you can squash your PRs for one commit which would have the jira number as prefix of its name?

ljnelson:
Hi, @ochaloup; I'm no git expert so I'm going to describe what I'm going to do and you can tell me if it looks right.

I'm on my local branch on my machine.  I've added a remote called `upstream` which references the official Narayana repo.

I'm going to find my merge base by doing: `git merge-base scope-events upstream/master`.  That should give me a hash.

Next, I'm going to do an interactive rebase, which as I understand it is the official name for a squash operation: `git rebase -i ${HASH_FROM_PREVIOUS_STEP}`.

I'm going to `pick` the first commit and `squash` the others.  I'm then going to edit the commit message so it says something like `[JBTM-3106] Fire transaction scope events`.

I'm then going to do whatever is necessary to commit this One True Commit™.   Then I'm going to `push` as normal.

Does that all sound right?

ochaloup:
@ljnelson yes, that should be so. When you push the squashed commit you will need to force it (`git push origin -f`)

Just adding that I normally try to rebase against the most up to date master to cover possible merge confilicts.

```
git checkout scope-events
git fetch upstream master # expecting the 'git remote -v' shows upstream to be the narayana repo
git rebase upstream/master
git rebase -i upstream/master
# now picking the first commit 'Initial PR revision' and for other ones choosing 'squash' (or just 's')
# saving the editor, replacing the message as you said and force pushing
```

ochaloup:
I approved the PR with the remark that commits should be squashed and then tested by CI before merging this.

ljnelson:
OK; took a backup and am going to follow this guide: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request.

ljnelson:
Seems to have worked; let me know if anything is amiss:

```
LANELSON-mac:narayana LANELSON$ git checkout scope-events
Switched to branch 'scope-events'
Your branch is up to date with 'origin/scope-events'.
LANELSON-mac:narayana LANELSON$ git merge-base scope-events upstream/master
d8093b27c4baaae7cdcb36d42a4ea3dc2d65909a
LANELSON-mac:narayana LANELSON$ git rebase -i d8093b27c4baaae7cdcb36d42a4ea3dc2d65909a
[detached HEAD f9b2b693a] [JBTM-3106] Fire appropriate scope events for TransactionScoped scope.
 Date: Sun Feb 10 22:26:14 2019 -0800
 7 files changed, 801 insertions(+), 54 deletions(-)
 create mode 100644 ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/DelegatingTransactionManager.java
 create mode 100644 ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/NarayanaTransactionManager.java
 create mode 100644 ArjunaJTA/cdi/etc/META-INF/beans.xml
 create mode 100644 ArjunaJTA/cdi/tests/classes/com/hp/mwtests/ts/jta/cdi/transactionScoped/TransactionScopeLifecycleEventsTest.java
Successfully rebased and updated refs/heads/scope-events.
LANELSON-mac:narayana LANELSON$ git status
On branch scope-events
Your branch and 'origin/scope-events' have diverged,
and have 1 and 9 different commits each, respectively.
  (use "git pull" to merge the remote branch into yours)

nothing to commit, working tree clean
LANELSON-mac:narayana LANELSON$ git rebase upstream/master
First, rewinding head to replay your work on top of it...
Applying: [JBTM-3106] Fire appropriate scope events for TransactionScoped scope.
LANELSON-mac:narayana LANELSON$ git push -f
Counting objects: 28, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (16/16), done.
Writing objects: 100% (28/28), 8.89 KiB | 2.22 MiB/s, done.
Total 28 (delta 7), reused 0 (delta 0)
remote: Resolving deltas: 100% (7/7), completed with 7 local objects.
To github.com:ljnelson/narayana.git
 + 40dcdccd9...43597edc4 scope-events -> scope-events (forced update)
LANELSON-mac:narayana LANELSON$ 
```

ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/TransactionExtension.java
```diff
@@ -65,27 +63,39 @@
...
+    public void afterBeanDiscovery(@Observes AfterBeanDiscovery event, BeanManager manager) {        
 
-        if (maybeAddInitialContextBean) {
+        Set<Bean<?>> beans = manager.getBeans(TransactionSynchronizationRegistry.class);
+        if (beans == null || beans.isEmpty()) {
+            event.addBean(new JNDIBean<>(jtaPropertyManager.getJTAEnvironmentBean().getTransactionSynchronizationRegistryJNDIContext(), TransactionSynchronizationRegistry.class));
             beans = manager.getBeans(InitialContext.class);
-            if (beans.isEmpty()) {
+            if (beans == null || beans.isEmpty()) {
                 event.addBean(new AbstractBean<InitialContext>() {
```
ochaloup: @ljnelson sorry for asking a little bit unrelated thing to this PR but when I'm going through the code I'm thinking that this part where there is created `new AbstractBean<InitialContext>` is a bit hard to read. Do you think it's reasonable to pull the definition off and creating a new class that would be just referenced here? Similar to how `JNDIBean` is defined?

ochaloup:
@ljnelson a side question about the code. We got an issue https://issues.jboss.org/browse/WFLY-11858. Would you have some comment for this kind of change in the `JNDIBean`? : https://github.com/ochaloup/narayana/commit/964222d60c6f64849430057cd744d5ed6d1bba54

ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/TransactionExtension.java
ljnelson: I can do this, sure.  Given the discussion going on on [WFLY-11858](https://issues.jboss.org/browse/WFLY-11858), would you like me to make its scope `Dependent` as well?

ljnelson:
Another thing I noticed was someone added a `getId()` method to `AbstractBean` (in order to implement `PassivationCapable`) that references `JNDIBean` (one of its subclasses).  This is bad practice (to have a superclass reference one of its possible subclasses).  It also has a `protected abstract String getTypeName()` method, whose semantics are already covered by the CDI `Bean` class' `getTypes()` method.  I will correct these issues here as well.

ochaloup:
@ljnelson I see. The reason for the pasivation capability is based on the issue https://issues.jboss.org/browse/JBTM-3110. Feel free to adjust the code, just bear in mind this issue. Thanks!

ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/TransactionExtension.java
ochaloup: @ljnelson as I commented on the WFLY-11858 I think the `Singleton` scope sound fine to me.

ljnelson:
Hello, @ochaloup; I've included the fixes I mentioned above, and also put in what I think will be a fix for [WFLY-11858](https://issues.jboss.org/browse/WFLY-11858).  Let me know what you think.

ljnelson:
Hi, @ochaloup; I've updated this PR with feedback from [WFLY-11858](https://issues.jboss.org/browse/WFLY-11858).

ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/TransactionExtension.java
ljnelson: I've resolved this in my latest pushes by incorporating feedback from [WFLY-11858](https://issues.jboss.org/browse/WFLY-11858)—I've removed all the acquisition of `InitialContext` instances by any means other than direct construction.  This issue is therefore resolved.

ljnelson:
Hi, @ochaloup; hoping you can tell me what the failures are.

ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/JNDIBean.java
```diff
@@ -46,39 +47,108 @@
...
-        } catch (final NamingException namingException) {
-            throw new CreationException(namingException.getMessage(), namingException);
+            initialContext = this.getNewInitialContext();
+            return this.type.cast(initialContext.lookup(this.name));
+        } catch (final NamingException namingException) {            
+            e = new CreationException(namingException.getMessage(), namingException);
+            throw e;
+        } finally {
+            try {
+                initialContext.close();
```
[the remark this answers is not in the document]
ljnelson: Ah, nice catch; will fix.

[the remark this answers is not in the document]
ljnelson: Resolved.

ochaloup:
@ljnelson ah, one more think. When you will be squashing the commits could you please prefix the commit message with both jiras this touches. That means with JBTM-3106 and WFLY-11858 with description?
```
[JBTM-3106][WFLY-11858] ... some text ...
```

ljnelson:
Hi, @ochaloup; I believe (again!) that I'm all set here.

ochaloup:
@ljnelson yes, it seems all good. Just let's wait for the final check from the CI.

mmusgrov:
TESTIT

ochaloup:
@mmusgrov I'm not sure if the `TESTIT` message was intentional but as the PR was already merged the CI won't run the tests for this PR.

mmusgrov:
@ochaloup I wasn't sure. I have just noticed that the PR disabled perf testing and since the PR wraps the TM and also generates events on a per transaction basis so I needed to check that there was no performance hit.

ochaloup:
hi @ljnelson , I wanted to update our quickstarts with this new capability and I hit an issue with the transaction context deactivation.
When using `@Transactional(Transactional.TxType.MANDATORY)` for the whole method there is thrown exception `ARJUNA016110: Transaction is required for invocation`. I think the `TransactionalInterceptorMandatory` should probably being able to find out that the `Observer` is called and not being forcing the transaction to be active.
Would you have some observations about this.

The code is here:
https://github.com/ochaloup/quickstart-jbosstm/blob/JBTM-3106-testing/jta-1_2-standalone/src/main/java/org/jboss/narayana/quickstarts/jta/MandatoryCounterManager.java#L37

```
javax.transaction.TransactionalException: ARJUNA016110: Transaction is required for invocation
        at com.arjuna.ats.jta.cdi.transactional.TransactionalInterceptorMandatory.doIntercept(TransactionalInterceptorMandatory.java:57)
        at com.arjuna.ats.jta.cdi.transactional.TransactionalInterceptorBase.intercept(TransactionalInterceptorBase.java:88)
        at com.arjuna.ats.jta.cdi.transactional.TransactionalInterceptorMandatory.intercept(TransactionalInterceptorMandatory.java:51)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.jboss.weld.interceptor.reader.SimpleInterceptorInvocation$SimpleMethodInvocation.invoke(SimpleInterceptorInvocation.java:73)
        at org.jboss.weld.interceptor.proxy.InterceptorMethodHandler.executeAroundInvoke(InterceptorMethodHandler.java:84)
        at org.jboss.weld.interceptor.proxy.InterceptorMethodHandler.executeInterception(InterceptorMethodHandler.java:72)
        at org.jboss.weld.interceptor.proxy.InterceptorMethodHandler.invoke(InterceptorMethodHandler.java:56)
        at org.jboss.weld.bean.proxy.CombinedInterceptorAndDecoratorStackMethodHandler.invoke(CombinedInterceptorAndDecoratorStackMethodHandler.java:79)
        at org.jboss.weld.bean.proxy.CombinedInterceptorAndDecoratorStackMethodHandler.invoke(CombinedInterceptorAndDecoratorStackMethodHandler.java:68)
        at org.jboss.narayana.quickstarts.jta.MandatoryCounterManager$Proxy$_$$_WeldSubclass.transactionScopeDectivated(Unknown Source)
```

ljnelson:
This is a tricky one.  I think the situation is—unfortunately—exactly as CDI wants it to be.

You have a method, `transactionScopeDeactivated`, which is effectively annotated with `@Transactional(Transactional.TxType.MANDATORY)`.  It is an observer method, which means that CDI will apply relevant interceptors, decorators, etc.  So CDI begins to do this.  One of the interceptors Narayana's transactional interceptor, and boom.

If we could use CDI 2.0, then you could observe the `@BeforeDestroyed(TransactionScoped.class)` qualifier, and this method could be notified right before the transaction is deactivated.

In summary, I think that a `MANDATORY` method that observes the (post-) destruction of the transaction scope is nonsensical, so this exception doesn't surprise me.

ochaloup:
@ljnelson I see, thanks for input. I was thinking if the UX would be better of making the interceptor being aware of an observer methods. But as you say it's the expected behaviour and ok for how the CDI works.

Good point about the `@BeforeDestroyed`. Thanks for the mentioning it.

ljnelson:
Oh, I see what you're asking, I think: if you're asking, "Should we change the interceptors to be aware of these observer methods in some way?" I would say no, don't do that.

Also you'll note that the `NarayanaTransactionManager` fires a `String` as the event qualified with `@Destroyed(TransactionScoped.class)` on purpose instead of a `Transaction` object for similar reasons.

ochaloup:
@ljnelson yeap, ok, thanks. 

I already wondered about the fact that the `String` is fired instead of the `Transaction`. But then I found the reason about the empty (null) transaction context.

ljnelson:
Right; the point being that really observers should just be observing `Object`, since it doesn't really matter what the payload is.

ochaloup:
@ljnelson thanks for the reply!

I have another doubt ojn my mind. I wanted to run the CDI code without need to define `InitialContext()` binding. My understanding was that's the idea of your fixes at the end. And I liked that idea.

I want to get running standalone Java SE example. I did `new Weld().initialize()` (https://github.com/ochaloup/quickstart-jbosstm/blob/JBTM-3106-testing-lookup-failure/jta-1_2-standalone/src/test/java/org/jboss/narayana/quickstarts/jta/CDIBindingTestCase.java#L49). But that's where the error[1] happens.

The reason is that the `JNDIBean` (https://github.com/jbosstm/narayana/blob/5.9.5.Final/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/JNDIBean.java#L122) and in the same manner the `NarayanaTransactionManager` (https://github.com/jbosstm/narayana/blob/5.9.5.Final/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/NarayanaTransactionManager.java#L183) catches the `NamingException` and throws that up. Even the NarayanaTransactionManager defines fallback to `com.arjuna.ats.jta.TransactionManager.transactionManager()` (https://github.com/jbosstm/narayana/blob/5.9.5.Final/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/NarayanaTransactionManager.java#L204) which I would be happy to be used in my example it's never used. The `InitialContext.lookup()` ends with the `NamingException` either if there is `connection timeout` (aka. JNDI server is defined but is not started) or if no JNDI context exists[2]. In both cases the `NamingException` causes the code just ends with error and do not use the defined fallback.

Please @ljnelson , do you have some code using this approach and can you point me to reason this does not work? Or can I consider it rather as an issue of the current implementation which should be adjusted? (which I consider more probable).

Thanks a lot
o.

[1]
```
testTransactionScoped(org.jboss.narayana.quickstarts.jta.CDIBindingTestCase)  Time elapsed: 5.879 sec  <<< ERROR!
javax.enterprise.inject.CreationException: Need to specify class name in environment or system property, or as an applet parameter, or in an application resource file:  java.naming.factory.initial
 at com.arjuna.ats.jta.cdi.JNDIBean.create(JNDIBean.java:123)
 at org.jboss.weld.contexts.AbstractContext.get(AbstractContext.java:96)
 at org.jboss.weld.bean.ContextualInstanceStrategy$DefaultContextualInstanceStrategy.get(ContextualInstanceStrategy.java:100)
 at org.jboss.weld.bean.ContextualInstance.get(ContextualInstance.java:50)
 at org.jboss.weld.manager.BeanManagerImpl.getReference(BeanManagerImpl.java:694)
 at org.jboss.weld.manager.BeanManagerImpl.getReference(BeanManagerImpl.java:717)
 at org.jboss.weld.util.ForwardingBeanManager.getReference(ForwardingBeanManager.java:64)
 at org.jboss.weld.bean.builtin.BeanManagerProxy.getReference(BeanManagerProxy.java:87)
 at com.arjuna.ats.jta.cdi.TransactionExtension.lambda$afterBeanDiscovery$1(TransactionExtension.java:107)
 at com.arjuna.ats.jta.cdi.TransactionContext.get(TransactionContext.java:107)
 at com.arjuna.ats.jta.cdi.TransactionContext.get(TransactionContext.java:139)
 at org.jboss.weld.contexts.PassivatingContextWrapper$AbstractPassivatingContextWrapper.get(PassivatingContextWrapper.java:78)
 at org.jboss.weld.bean.ContextualInstanceStrategy$DefaultContextualInstanceStrategy.getIfExists(ContextualInstanceStrategy.java:89)
 at org.jboss.weld.bean.ContextualInstance.getIfExists(ContextualInstance.java:63)
 at org.jboss.weld.bean.proxy.ContextBeanInstance.getInstance(ContextBeanInstance.java:87)
 at org.jboss.weld.bean.proxy.ProxyMethodHandler.getInstance(ProxyMethodHandler.java:131)
 at org.jboss.narayana.quickstarts.jta.Counter$Proxy$_$$_WeldClientProxy.get(Unknown Source)
 at org.jboss.narayana.quickstarts.jta.RequiredCounterManager.getCounter(RequiredCounterManager.java:75)
 at org.jboss.narayana.quickstarts.jta.RequiredCounterManager$Proxy$_$$_WeldSubclass.getCounter(Unknown Source)
 at org.jboss.narayana.quickstarts.jta.CDIBindingTestCase.testTransactionScoped(CDIBindingTestCase.java:78)
 at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
 ...
Caused by: javax.naming.NoInitialContextException: Need to specify class name in environment or system property, or as an applet parameter, or in an application resource file:  java.naming.factory.initial
 at javax.naming.spi.NamingManager.getInitialContext(NamingManager.java:662)
 at javax.naming.InitialContext.getDefaultInitCtx(InitialContext.java:313)
 at javax.naming.InitialContext.getURLOrDefaultInitCtx(InitialContext.java:350)
 at javax.naming.InitialContext.lookup(InitialContext.java:417)
 at com.arjuna.ats.jta.cdi.JNDIBean.create(JNDIBean.java:121)
 ... 50 more
```

[2] InitialContext.lookup failure
```
Exception in thread "main" java.lang.RuntimeException: javax.naming.NoInitialContextException: Need to specify class name in environment or system property, or as an applet parameter, or in an application resource file:  java.naming.factory.initial
 at org.jboss.narayana.quickstarts.jta.MyMain.main(MyMain.java:11)
Caused by: javax.naming.NoInitialContextException: Need to specify class name in environment or system property, or as an applet parameter, or in an application resource file:  java.naming.factory.initial
 at javax.naming.spi.NamingManager.getInitialContext(NamingManager.java:662)
 at javax.naming.InitialContext.getDefaultInitCtx(InitialContext.java:313)
 at javax.naming.InitialContext.getURLOrDefaultInitCtx(InitialContext.java:350)
 at javax.naming.InitialContext.lookup(InitialContext.java:417)
 at org.jboss.narayana.quickstarts.jta.MyMain.main(MyMain.java:9)
```

ljnelson:
I'll look later today.  I am pretty sure that it is a simple oversight: https://github.com/jbosstm/narayana/blob/5.9.5.Final/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/NarayanaTransactionManager.java#L183 catches `NoInitialContextException`, whereas https://github.com/jbosstm/narayana/blob/5.9.5.Final/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/JNDIBean.java#L122 does not, and probably should.

ljnelson:
Hi, @ochaloup; I filed https://issues.jboss.org/browse/JBTM-3131 to track this; I'll have a PR shortly.

ochaloup:
@ljnelson thanks a lot! I will check it out with my code.

mmusgrov:

PR #1407, opened 2019-02-13
[SCRIPTS ONLY] info on what is the JDK the pull request is running/failing on

ochaloup:
the pull request info should be enhance with information what is the JDK we run at

!MAIN !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !PERF WIN !RTS !AS_TESTS !TOMCAT !JACOCO !mysql !postgres !db2 !oracle JDK11 JDK9

### DIFF
--- a/scripts/hudson/narayana.sh
+++ b/scripts/hudson/narayana.sh
@@ -216,10 +216,12 @@ function comment_on_pull
 {
     if [ "$COMMENT_ON_PULL" = "" ]; then echo $1; return; fi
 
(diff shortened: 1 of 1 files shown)


ochaloup:
@tomjenkinson  all right, correct. I didn't realized that in fact. It was not visible enough for me. But you are right. That's a duplicity information.

mmusgrov:

PR #1409, opened 2019-02-15
[JBTM-3111][WFLY-11706] ordering XARecoveryModule locks

ochaloup:
to avoid potential deadlock scenario

https://issues.jboss.org/browse/JBTM-3111
https://issues.jboss.org/browse/WFLY-11706

 !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF !RTS !TOMCAT !JACOCO !mysql !postgres !db2 !oracle

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
@@ -49,7 +49,6 @@ import javax.transaction.xa.Xid;
 
 import com.arjuna.ats.arjuna.common.Uid;
 import com.arjuna.ats.arjuna.exceptions.ObjectStoreException;
-import com.arjuna.ats.arjuna.logging.tsLogger;
 import com.arjuna.ats.arjuna.objectstore.RecoveryStore;
 import com.arjuna.ats.arjuna.objectstore.StateStatus;
 import com.arjuna.ats.arjuna.objectstore.StoreManager;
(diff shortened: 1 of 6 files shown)


ochaloup:
hi @tomjenkinson,

I intentionally set up the `NoCI` for now as I would like to ask you for review first. I didn't think that first time but it got me more effort than I  thought. I was testing and trying to understand the code to come with some nicer solution than this ordering one. But I didn't found any other. If you have an idea how to do this better I will be happy to learn it.
_After your review I will run the tests_

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
```diff
@@ -112,24 +115,33 @@ public void removeXAResourceRecoveryHelper(XAResourceRecoveryHelper xaResourceRe
...
-							System.out.println("Was null");
-						}
-					}
-				}
+                } else {
+                    // synchronized(this) is needed as removing on the _resources; to avoid deadlock
+                    //   we need to preserve the order of acquiring locks - first 'this' then second 'scanStateLock'
+                    scanStateLock.unlock();
+                    synchronized (this) {
+                        scanStateLock.lock();
```
tomjenkinson: It looks a little bit strange that you release the lock and then reacquire it. From what I can see though we do only need to check for the state BETWEEN_PASSES to do the removes but I wonder if we should check again for something like 

```
isHelperInUse(xaResourceRecoveryHelper)) {
                    waitForScanState(ScanStates.IDLE);
}
```

Under the second scanStateLock.lock() code?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
```diff
@@ -112,24 +115,33 @@ public void removeXAResourceRecoveryHelper(XAResourceRecoveryHelper xaResourceRe
...
+                            if (xaResources != null) {
+                                for (XAResource xar : xaResources) {
+                                    xaRecoverySecondPass(xar);
+                                    _resources.remove(xar);
+                                }
+                            } else {
+                                jtaLogger.logger.debugf("nothing to remove - xa resources of recovery helper '%s' were null",
+                                        xaResourceRecoveryHelper);
+                            }
+                        }
```
tomjenkinson: i.e. to add it here:

```
else (if isHelperInUse(xaResourceRecoveryHelper)) {
  waitForScanState(ScanStates.IDLE);
}
```

ochaloup:
@tomjenkinson yes, you are absolutely right. This is wrong. But the check you offered could not be enough. After the `scanStateLock.unlock()` is called the state could be changed to whatever one. That means there could be started the first pass again (`ScanStates.FIRST_PASS`) which means the need for the waiting `waitForNotScanState(ScanStates.FIRST_PASS)` could be needed.

I will recap my thinking here

* my proposal was releasing the lock and then acquiring it as the deadlock was caused by running in parallel the `removeXAResourceRecoveryHelper` which acquires the `scanState` lock and `xaRecoverySecondPass` which acquires whole `this` lock. Then both methods wanted to acquire the locks of each other later on. Releasing the lock let me change the order of lock acquisition to be ordered in one way. Both methods then needed to acquire first `this` and then `scanState`. As the lock acquisition order was guaranteed then the deadlock was mitigated.

* lock on `scanState` is acquired by methods `setScanState` and the `removeXAResourceRecoveryHelper`
* method `setScanState` is used only by methods `periodicWorkFirstPass` and `periodicWorkSecondPass`
* both methods `periodicWorkFirstPass` and `periodicWorkSecondPass` acquire the lock `this`
* the methods `removeXAResourceRecoveryHelper` is written in way to wait for state would be changed. Because of that is not possible(?) to synchronized the whole method. If it's done so then no state change could happen as the state change will be blocked by the `removeXAResourceRecoveryHelper` holding the `this` lock
* the `periodicWorkFirstPass` acquires `this` and changes state first to `ScanStates.FIRST_PASS` and then to some state, the most possibly to `ScanStates.BETWEEN_PASSES`
* the `periodicWorkSecondPass` acquires `this` and changes state first to `ScanStates.SECOND_PASS` and then to `ScanStates.IDLE`

Even I wanted to avoid global lock on `this` I can't find out the way how to do without. Please @tomjenkinson, take a look at the new changes here: https://github.com/jbosstm/narayana/pull/1409/files

ochaloup:
@tomjenkinson I was wrong, sorry. I don't mean the `xaRecoverySecondPass` but I mean the method `periodicWorkSecondPass` (which is started by periodic recovery thread). I already updated my previous comment to not saying the misguiding fact.

ochaloup:
@tomjenkinson ah, ok. Interesting for me is that the change [JBTM-2242](https://issues.jboss.org/browse/JBTM-2242) came with the removing any synchronization from method call `addXAResourceRecoveryHelper`. There is no synchronization anymore when you want to add new recovery helper. I expect it should be fine as there is used `CopyOnWriteArraySet`. As the `addXAResourceRecoveryHelper` does not define any synchronization currently this PR does not touch that functionality.

My understanding was that if we are `BETWEEN_PASSES` (see https://github.com/jbosstm/narayana/blob/5.9.2.Final/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java#L117) we want to speed up shutdown/reload of the server for not waiting whole back off period + second pass. The code during between passes pass just speed up with first finishing the second phase and run the removal.
That way the recovery could be stopped as soon as possible and helpers could be removed (https://issues.jboss.org/browse/WFLY-10841).

ochaloup:
hi @tomjenkinson, I updated the code. Thank you for hints. Please, take a look on the current state and review it. I hope I addressed the issue and there should not be the stackoveflow now. My thinking is there is the recursion call but the `scanState` is already changed to a different value. The longest chain of call what I can see is

. `removeXAResourceRecoveryHelper` the `scanState` is `FIRST_PASS`, waiting for the state different to `FIRST_PASS`, changes to `SECOND_PASS`
. recurrent of  `removeXAResourceRecoveryHelper` the `scanState` is `SECOND_PASS`. As synchronized there is no change of the state. Waiting for `IDLE`.
. recurrent of  `removeXAResourceRecoveryHelper` the `scanState` is `IDLE` and we can proceed with removal

In fact this chain should never happen as if we are in `FIRST_PASS` then the change state could be either for the `BETWEEN_PASS` or to `IDLE. Both of the state does not run recursion call to the method itself.

ochaloup:
* The `AS_TESTS` contains our known intermittent failures.
* For the `MAIN` test I found an issue I fixed, plus I needed to change the `XARecoveryModuleHelpersUnitTest#testTimelyXAResourceRecoveryHelperRemoval3` as with the current change the `XARecoveryModule#removeXAResourceRecoveryHelper` has to wait for finishing the `SECOND_PASS`. The test still checks that the `XARecoveryModule#addXAResourceRecoveryHelper` is fine to add helper even the recovery is in progress.

ochaloup:
or maybe adding a new lock which protects the access to the `_resources`. But I think it's complicated. Such a try could be seen here: https://github.com/ochaloup/narayana/commit/238b7a03dafac9b9dff7c7df9eb0b68244d34de0

I still think the synchronized on the `this` is not such an issue as it waits more only in case of running second phase of recovery and only if there are no recovery helpers active. If I take the WFLY example then during shutdown it needs to remove all the helpers for being stopped. I needs to wait for the second phase to finish as

* either there is some helper and some processing is needed to be finished so such helper delays the shutdown (xa remove)
* or there is no helper, then here is no processing and the second phase is just quick enough to not delay the shutdown

ochaloup:
Ok, thanks. I hope I do not bring some unseen issue here. Anyway I'm merging it now.

mmusgrov:

PR #1412, opened 2019-02-19
[JBTM-3112] Remove dependency on LRA client from LRA coordinator

xstefank:
https://issues.jboss.org/browse/JBTM-3112

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN  !AS_TESTS !TOMCAT !JACOCO !mysql !postgres !db2 !oracle


ochaloup:
TESTIT

rts/lra/lra-service-base/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.jboss.logging</groupId>
+      <artifactId>jboss-logging</artifactId>
+      <version>${version.org.jboss.logging.jboss-logging}</version>
+    </dependency>
+  </dependencies>
+  
+  <build>
+    <finalName>LRA Service base</finalName>
```
mmusgrov: Isn't this element used as the final name the build artefact, in which case we need a better name (in particular no spaces or capitals). Easiest is to just follow the name of the maven module that builds the artefact.

Also it isn't clear what this module is for so can you add a description element for the module.

rts/lra/lra-service-base/src/main/java/io/narayana/lra/LRAUtils.java
```diff
@@ -0,0 +1,77 @@
...
+import org.eclipse.microprofile.lra.client.GenericLRAException;
+
+import java.io.UnsupportedEncodingException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLEncoder;
+
+import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
+
+public abstract class LRAUtils {
```
mmusgrov: I'm not keen on Utils type classes, they become a dumping ground for miscellaneous bits of code people think may or may not be generally useful.

rts/lra/lra-service-base/pom.xml
xstefank: you are right, I will update this

rts/lra/lra-service-base/src/main/java/io/narayana/lra/LRAUtils.java
xstefank: @mmusgrov generally true. But these methods are needed both in coordinator and in the client. They are all stateless static methods. All they do is transformations of the String and URL variants of lraId. Maybe we can change name to something different? Maybe `LraIdTransformator`?

ochaloup:
failures in the MAIN CI tests are caused by LRA TCK api changes. The fix is here: https://github.com/jbosstm/narayana/pull/1415

mmusgrov:
Let's wait until @ochaloup has merged #1415 before merging this PR

mmusgrov:
TESTIT

since the dependant PR has just been merged

mmusgrov:
@xstefank and @ochaloup I tried fixing up the conflicts between both of your branches since you have both been making similar changes but I realised that I am going to have to back out some of the changes in order to resolve the conflicts. Can you both decide which changes to take. Since Martin's PR was stalled waiting for Ondra's PR (1415) to be merged it probably makes sense take Ondra's changes where conflicts arise.

rts/lra/pom.xml
```diff
@@ -171,6 +171,7 @@
         <module>lra-filters</module>
         <module>lra-annotation-checker</module>
         <module>lra-test</module>
+        <module>lra-service-base</module>
```
ochaloup: I think the `lra-service-base` module should be moved to the top of the list. I think maven runs compilation/installation in sequence by order defined in pom.xml. If the `lra-service-base` is used by other modules it should be compiled before other ones.

xstefank: that's not totally true. Maven first checks dependencies of the modules so if there is dependency in lra-client to lra-service-base for instance, lra-service-base must be build before lra-client so it is buildable in clean environment. But I can definitely make this change, however I find it unnecessary and CI is already running for this and force push would retrigger it:) @ochaloup you want to update this?

ochaloup: ok, then it's my ignorance. sorry for that. I have no issue with this in such case.

rts/lra/lra-service-base/src/main/java/io/narayana/lra/LRAUtils.java
mmusgrov: @xstefank I see that you force pushed a change that removed Utils making me think that you had accepted my proposal. But now I see that all you did was move the code into a different class (LRAIdConverter.java) and hide the change amongst a load of other changes. That was very sneaky, especially as I had already studied this large PR so assumed that since your force push removed Utils that the new commit was in response to this comment so I did not re-review the changes. Very poor form.

xstefank: @mmusgrov that was my mistake. I forgot to include that as part of the force push and I didn't expect this PR to be merged so soon prioritizing other work. Sorry for confusion, I will make another PR tomorrow amending these changes

mmusgrov:

PR #1414, opened 2019-02-22
[SCRIPT ONLY] after fixing JBTM-3067 there is no need to rebase jdk11

ochaloup:
… against a special branch. Especially when there is no such branch as `jdk-11` :)

MAIN !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS !AS_TESTS !TOMCAT !JACOCO JDK11


mmusgrov:

PR #1415, opened 2019-02-23
[JBTM-3109][lra#94] renaming LRA status complete/compensate to close/cancel

ochaloup:
https://issues.jboss.org/browse/JBTM-3109

Based on the changes in LRA TCK the LRA status methods were renamed to
close and cancel.

https://github.com/eclipse/microprofile-lra/issues/94
https://github.com/eclipse/microprofile-lra/pull/98

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS !AS_TESTS !TOMCAT !JACOCO

### DIFF
--- a/rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
+++ b/rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
@@ -105,7 +105,7 @@ public class NarayanaLRAClient implements LRAClient, Closeable {
     private static final String startLRAUrl = "/start";///?ClientId=abc&timeout=300000";
     private static final String recoveryQueryUrl = "/recovery";
     private static final String getAllLRAsUrl = "/";
-    private static final String getRecoveringLRAsUrl = "?status=Compensating";
+    private static final String getRecoveringLRAsUrl = "?status=Cancelling";
     private static final String getActiveLRAsUrl = "?status=";
 
     private static final String confirmFormat = "/%s/close";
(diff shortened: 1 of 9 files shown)


ochaloup:
@mmusgrov I'm on PTO next week, if there are needed some changes of this PR, please feel free to close this one and open new one with adjustments for you being able to merge it.

ochaloup:
@xstefank as we talked there is some strange way how the shrinkwrap evaluates dependencies and way how the maven brings the dependencies in. There happens to be the dependency version clash. As the `org.eclipse.microprofile.lra:microprofile-lra-tck` defines dependency to `org.eclipse.microprofile.lra:microprofile-lra-api:${project.version}` (https://github.com/eclipse/microprofile-lra/blob/master/tck/pom.xml#L59) and the project version is `1.0-SNAPSHOT` then the latest LRA snapshot is taken and causes the clash. It seems the shrinkwrap and for example `maven source plugin` takes this latest version based on the definition from the tck. What I did was to change the scope from `test` to `provided` and in that way the dependency was resolved correctly. Unfortunately I don't understand in details why it is so.

The second failure was the 

```
java.lang.RuntimeException: java.lang.ClassNotFoundException: org.glassfish.jersey.client.JerseyClientBuilder
        at org.eclipse.microprofile.lra.tck.TckCancelOnTests.beforeClass(TckCancelOnTests.java:76)
Caused by: java.lang.ClassNotFoundException: org.glassfish.jersey.client.JerseyClientBuilder
        at org.eclipse.microprofile.lra.tck.TckCancelOnTests.beforeClass(TckCancelOnTests.java:76)
```

I thought I already fixed this issue by using `META-INF/services/javax.ws.rs.client.ClientBuilder` descriptor in the auxiliary file. But for some reason this does not fix the issue. For that reason I followed the advices https://developer.jboss.org/thread/241325 and defined dependency on the resteasy client to the pom.xml. That way it works for me.

ochaloup:
I'm going to merge this while it fixes the issue with the changes in interface lra#94 and should fix the trouble with the dependency failures. If there is some more troubles with dependencies I will address it in the next PR.

mmusgrov:

PR #1379, opened 2018-11-21
Revert "disable the javadoc with the JDK9"

tomjenkinson:
https://issues.jboss.org/projects/JBTM/issues/JBTM-3057?filter=allissues
!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !PERF NO_WIN !RTS !XTS !AS_TESTS !JACOCO !TOMCAT

JDK11

PR #1416, opened 2019-02-27
JBTM-3057 javadoc failures on JDK9+

zhfeng:
!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !PERF NO_WIN !RTS !XTS !AS_TESTS !JACOCO !TOMCAT

JDK9 and JDK11

https://issues.jboss.org/projects/JBTM/issues/JBTM-3057

This PR replaces the https://github.com/jbosstm/narayana/pull/1379

### DIFF
--- a/narayana-full/pom.xml
+++ b/narayana-full/pom.xml
@@ -391,15 +391,5 @@
           </dependency>
       </dependencies>
     </profile>
-    <profile>
-      <id>j9</id>
-      <activation>
-        <jdk>[1.9,)</jdk>
-      </activation>
-      <properties>
-        <maven.javadoc.skip>true</maven.javadoc.skip>
-	<maven.antrun.skip>true</maven.antrun.skip>
-      </properties>
-    </profile>
   </profiles>
 </project>
--- a/narayana-full/pom.xml
+++ b/narayana-full/pom.xml
@@ -186,7 +186,35 @@
       <version>${version.org.jboss.integration}</version>
       <scope>provided</scope>
     </dependency>
-
+    <dependency>
+      <groupId>org.jboss.resteasy</groupId>
+      <artifactId>resteasy-jaxrs</artifactId>
+      <version>${version.org.jboss.resteasy}</version>
+      <scope>provided</scope>
+      <!-- filter out unwanted jars -->
+      <exclusions>
+        <exclusion>
+          <groupId>commons-httpclient</groupId>
+          <artifactId>commons-httpclient</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>javax.servlet</groupId>
+          <artifactId>servlet-api</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>commons-httpclient</groupId>
+      <artifactId>commons-httpclient</artifactId>
+      <version>${version.commons-httpclient}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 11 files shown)


narayana-full/pom.xml
```diff
@@ -186,7 +186,35 @@
       <version>${version.org.jboss.integration}</version>
       <scope>provided</scope>
     </dependency>
-
+    <dependency>
+      <groupId>org.jboss.resteasy</groupId>
+      <artifactId>resteasy-jaxrs</artifactId>
```
tomjenkinson: Can you explain why we needed to add this dependency here please?

pom.xml
```diff
@@ -452,7 +452,7 @@
     <version.jacorb>2.3.1jboss.patch01-brew</version.jacorb>
     <version.org.jboss.openjdk-orb>8.0.8.Final</version.org.jboss.openjdk-orb>
     <version.javax.annotation>1.2</version.javax.annotation>
-    <version.javax.enterprise>1.1</version.javax.enterprise>
+    <version.javax.enterprise>1.2</version.javax.enterprise>
```
tomjenkinson: Please can you explain why we needed to update this dependency?

narayana-full/pom.xml
```diff
@@ -391,5 +425,64 @@
...
+             <jdk>[9,)</jdk>
+          </activation>
+          <properties>
+             <jvm.args.modular>--add-modules=java.se</jvm.args.modular>
+	     <version.org.jboss.spec.javax.xml.bind.jboss-jaxb-api_2.3_spec>1.0.1.Final</version.org.jboss.spec.javax.xml.bind.jboss-jaxb-api_2.3_spec>
+	     <version.org.jboss.spec.javax.xml.ws.jboss-jaxws-api_2.3_spec>1.0.0.Final</version.org.jboss.spec.javax.xml.ws.jboss-jaxws-api_2.3_spec>
+	     <version.jaxws-ri>2.3.0</version.jaxws-ri>
+	  </properties>
+	  <dependencies>
+	     <dependency>
```
tomjenkinson: Is this because these are no longer in the JDK? I am not sure why these are only needed in narayana-full in order to generate the Javadoc?

narayana-full/pom.xml
zhfeng: It needs to generate the restat javadoc. There are warnings with the JDK 8 about can not found the class but the errors with the JDK9+.

pom.xml
zhfeng: The javax.enterprise.cdi-api 1.1depends on the org.jboss.spec.javax.interceptor.jboss-interceptors-api_1.1_spec 1.0.0.Beta1 which is out of date and causes the ClassNotFound exception when building the javadoc of ArjunaJTA/cdi

narayana-full/pom.xml
zhfeng: yeah, it has been removed in the JDK-11. In JDK-9 it needs to add the "--add-modules=java.xml.ws,java.xml.bind" with the javadoc options but it through the error with can not find "javax.annotation.Priority" even if the the javax.annotation-api jar is in the classpath . I think this could be a split package issue. 

The java.xml.ws depends on the java.xml.ws.annotations which contains the package "javax.annotation" but does not have the "javax.annotation.Priority". I tried to add the --patch-module  but the javadoc can not find the source files with this options.

mmusgrov:

PR #1417, opened 2019-02-27
JBTM-3115 Do not run LRA TCK by default

mmusgrov:
https://issues.jboss.org/browse/JBTM-3115
By default skip LRA TCK testing
Note that this change does not disable running a coordinator which is a more invasive change and I would prefer that @ochaloup does that.


mmusgrov:
@tomjenkinson This is a small change but quite urgent since the jenkins CI is clogged up with a backlog of jobs that are going to fail. Can you approve before CI runs the job so that we can get the change in early?

mmusgrov:

PR #1420, opened 2019-03-08
[JBTM-3104] adding LRA tests profile for LRA not being executed as part of the MAIN

ochaloup:
https://issues.jboss.org/browse/JBTM-3104

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !AS_TESTS !TOMCAT !JACOCO !MAIN !RTS

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -30,6 +30,7 @@ if [ "$?" -ne "1" ]; then
   fi
 fi
 
+ORIG_WORKING_DIR=`pwd`
 PROGNAME=`basename $0`
 DIRNAME=`dirname $0`
 GREP="grep"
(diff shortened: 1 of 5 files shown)


ochaloup:
@mmusgrov @tomjenkinson 
The LRA fails as there is the `TckCancelOnTests.cancelFromRemoteCall` test failing.
I needed to do a change in `build.sh` script because of unfortunate issue of swarm issue https://issues.jboss.org/browse/THORN-2049 which causes the `-f .../pom.xml` is not possible to be used.

ochaloup:
I'm merging this. If there is some follow-up CI issues regarding this change I will try to solve them as soon as possible.

mmusgrov:

PR #1425, opened 2019-03-14
[WFLY-11849] txbridge import transaction to use default timeout not null

ochaloup:
The integration with WFLY with change of the WFTC-54 causes
that timeout set to 0 means the timeout elapsed and participant
is not permitted to be enlisted to transaction.
Let's use default timeout instead.

https://issues.jboss.org/browse/WFLY-11849

requires: https://github.com/jbosstm/jboss-as/pull/81
(adding the integration tests for txbridge to wfly)

Documentation is over here: https://github.com/jbosstm/documentation/pull/49

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !PERF !TOMCAT !JACOCO !LRA

### DIFF
--- a/rts/at/bridge/src/main/java/org/jboss/narayana/rest/bridge/inbound/InboundBridge.java
+++ b/rts/at/bridge/src/main/java/org/jboss/narayana/rest/bridge/inbound/InboundBridge.java
@@ -35,6 +35,7 @@ import javax.transaction.xa.Xid;
 import com.arjuna.ats.jta.recovery.SerializableXAResourceDeserializer;
 import org.jboss.logging.Logger;
 
+import com.arjuna.ats.arjuna.coordinator.TxControl;
 import com.arjuna.ats.internal.arjuna.FormatConstants;
 import com.arjuna.ats.internal.jta.transaction.arjunacore.jca.SubordinationManager;
 import com.arjuna.ats.jta.TransactionManager;
@@ -217,7 +218,8 @@ public final class InboundBridge implements XAResource, SerializableXAResourceDe
         final Transaction transaction;
 
         try {
-            transaction = SubordinationManager.getTransactionImporter().importTransaction(xid);
+            transaction = SubordinationManager.getTransactionImporter()
+                    .importTransaction(xid, TxControl.getDefaultTimeout());
         } catch (XAException e) {
             LOG.warn(e.getMessage(), e);
             throw new InboundBridgeException("Failed to import transaction.", e);
--- a/XTS/WSTX/classes/com/arjuna/mw/wst11/UserTransaction.java
+++ b/XTS/WSTX/classes/com/arjuna/mw/wst11/UserTransaction.java
@@ -129,4 +129,10 @@ public abstract class UserTransaction
         throws UnknownTransactionException, SecurityException, SystemException, WrongStateException;
 
     public abstract String transactionIdentifier ();
+
+    /**
+     * Returning timeout of the transaction
+     * as it was defined when it begun.
+     */
+    public abstract int getTimeout();
 }
--- a/XTS/WSTX/classes/com/arjuna/mwlabs/wst11/at/remote/UserSubordinateTransactionImple.java
+++ b/XTS/WSTX/classes/com/arjuna/mwlabs/wst11/at/remote/UserSubordinateTransactionImple.java
@@ -59,4 +59,8 @@ public class UserSubordinateTransactionImple extends UserTransaction
     public String transactionIdentifier() {
         return UserTransactionImple.getUserTransaction().transactionIdentifier();
     }
+
+    public int getTimeout() {
+        return UserTransactionImple.getUserTransaction().getTimeout();
+    }
 }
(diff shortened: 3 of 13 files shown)


ochaloup:
Unfortunately, there is a trouble in RTS tests. I need to investigate what is the reason behind.

@tomjenkinson I wanted to check the behaviour described at the forum https://developer.jboss.org/message/988408 and found I'm not able to run the quickstarts at all.
It's strange to me that we haven't hit it during the Narayana testing on CI. My suspision is that the job http://narayanaci1.eng.hst.ams2.redhat.com/job/narayana-quickstarts/ was not run for 14 days. But I don't understand why it was not run.

ochaloup:
for correct working (hopefully it  will be confirmed by smooth CI run) it needs the changes in WFLY testsuite (see https://github.com/jbosstm/jboss-as/pull/81). I ported WFLY txbridge test to WFLY integration testsuite plus change the RTS test which was written in a bit non-resilient way. It started to fail because of the `setTimeout` was called after my fix. I was checking the behaviour in the debugger and it seems it works fine.

ochaloup:
failures of AS_TESTS seems intermittent not connected to this PR (`org.jboss.as.test.integration.messaging.jms.context.notClosingInjectedContext.NotClosingInjectedContextTestCase`)
failures of RTS is my error as I forgot to fix one testcase. fixed now.

ochaloup:
@tomjenkinson I'm not sure where this should be documented. We haven't documented the prior behaviour either - that the bridge JTA is configured with no timeout. For me it's quite internal behaviour.

I can create some documentation to the upstream narayana doc, would that be fine?

mmusgrov:
Note that REST-AT do use timeouts on transactions and importing a transaction should respect the timeout.

rts/at/bridge/src/main/java/org/jboss/narayana/rest/bridge/inbound/InboundBridge.java
```diff
@@ -217,7 +218,8 @@ private Transaction getTransaction() {
         final Transaction transaction;
 
         try {
-            transaction = SubordinationManager.getTransactionImporter().importTransaction(xid);
+            transaction = SubordinationManager.getTransactionImporter()
+                    .importTransaction(xid, TxControl.getDefaultTimeout());
```
mmusgrov: The timeout on the bridged JTA transaction should be set to never time out since its lifecycle is under the control of the REST-AT transaction

ochaloup:
@mmusgrov I see what you mean but I'm not sure if I agree on this.

If the parent transaction is not capable to finish the subordinate (bridged) one then the subordinate can be blocking resources for infinite time. I think it's more reasonable to define the subordinate with timeout. The subordinate transaction can then be rolled-back independently on the fact that parent transaction/connection/host is dead. Isn't the no timeout behaviour more dangerous than beneficial?
What is the real issue in taking timeout provided by protocol or set up the default timeout of the environment?

The other thing is then an implementation question here. The integration layer in WildFly does not permit timeout to be set to `0`. `0` is taken as elapsed timeout. Importing transaction with timeout being set to `0` is refused. For "infinite" timeout we could use some big number (e.g. `Integer.MAX_VALUE`) but there is a trouble that some of the resource adapters do not accept such big number. So some arbitrary big number would need to be chosen. Which is a fragile approach.

mmusgrov:
Good points, you are correct the subordinate needs its own timeout which should be set to the remaining timeout. But we do not propagate that information on REST-AT calls so I am not sure what to suggest other than documenting the behaviour (which should be okay since the REST-AT bridge is community only).

ochaloup:
@mmusgrov thanks. I was looking into this just quickly so I can't say it for sure but I found there is some timeout field for REST-AT in Narayana implementation. I plan to investigate more deeply and I will let you know back here to discuss my findings.

ochaloup:
@mmusgrov really, I haven't found the REST-AT would pass the timeout value (https://github.com/jbosstm/narayana/blob/master/rts/at/util/src/main/java/org/jboss/jbossts/star/util/TxLinkNames.java). So my fix is about providing the default timeout.

@tomjenkinson I changed the XTS for using timeout provided by protocol and if there is no timeout or not defined then using the default timeout.

I changed the tests as well to be hitting this issue. Please take a look and tell me what you think about this now.

mmusgrov:
> @mmusgrov really, I didn't find the REST-AT would pass the timeout value (https://github.com/jbosstm/narayana/blob/master/rts/at/util/src/main/java/org/jboss/jbossts/star/util/TxLinkNames.java). So my fix is about providing the default timeout as it was.

I don't know what your response is about but my input on this issue was:

> ... But we do not propagate that information [the timeout] on REST-AT calls so I am not sure what to suggest other than documenting the behaviour (which should be okay since the REST-AT bridge is community only).

Perhaps you missed my message and was responding to an earlier one? What I am calling for is *for the behaviour to be documented* since the implementation does not propagate the remaining timeout.

ochaloup:
@mmusgrov I reacted on your comment and my point which followed your one. I was investigating if the implementation does not provide a way to pass the timeout in the context. I didn't find such way.

Agree, the behaviour should be documented. I created PR for documentation update here:
https://github.com/jbosstm/documentation/pull/49

Is there something more to do for you giving the approval for this PR?

PR #1429, opened 2019-03-20
SCRIPTS ONLY: disable LRA for JACOCO

ochaloup:


ochaloup:
@tomjenkinson hm... I can create a jira if you wish.

From the LRA TCK perspective - TCK could be changed asynchronously to the Narayana implementation. Even some  TCK failure is fixed now there could come new ones in future. The TCK failure that we were talking about is `cancelFromRemoteCall` which should be already fixed in the LRA TCK (https://github.com/eclipse/microprofile-lra/pull/118). For that fix will be "applied" we would need to switch to the newer SNAPSHOT version of LRA spec artifact. There I don't know if occurs some new incompatibilities meanwhile.

mmusgrov:

PR #1432, opened 2019-03-29
JBTM-3126 Do not start nested LRAs unles @NestedLRA is present

mmusgrov:
https://issues.jboss.org/browse/JBTM-3126

This PR is a bug fix and an upgrade of the MP-LRA dependency

### DIFF
--- a/rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
+++ b/rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
@@ -270,7 +270,8 @@ public class ServerLRAFilter implements ContainerRequestFilter, ContainerRespons
 //                    previous = AtomicAction.suspend();
                 suspendedLRA = incommingLRA;
                 lraTrace(containerRequestContext, suspendedLRA, "ServerLRAFilter before: REQUIRES_NEW start new LRA");
-                newLRA = lraId = startLRA(incommingLRA, method, getTimeOut(method));
+                URI parent = nested ? incommingLRA : null;
+                newLRA = lraId = startLRA(parent, method, getTimeOut(method));
 
                 break;
             case SUPPORTS:
@@ -394,26 +395,38 @@ public class ServerLRAFilter implements ContainerRequestFilter, ContainerRespons
 
         try {
             if (current != null && isJaxRsCancel(requestContext, responseContext)) {
-                lraClient.cancelLRA(current);
-                if (current.toASCIIString().equals(
-                        Current.getLast(requestContext.getHeaders().get(LRA_HTTP_HEADER)))) {
-                    // the callers context was ended so invalidate it
-                    requestContext.getHeaders().remove(LRA_HTTP_HEADER);
-                }
+                try {
+                    lraClient.cancelLRA(current);
+                } catch (NotFoundException ignore) {
+                    // must already be cancelled (if the intercepted method caused it to cancel)
+                    // or completed (if the intercepted method caused it to complete)
+                } finally {
+                    if (current.toASCIIString().equals(
+                            Current.getLast(requestContext.getHeaders().get(LRA_HTTP_HEADER)))) {
+                        // the callers context was ended so invalidate it
+                        requestContext.getHeaders().remove(LRA_HTTP_HEADER);
+                    }
 
-                if (toClose != null && toClose.toASCIIString().equals(current.toASCIIString())) {
-                    toClose = null; // don't attempt to finish the LRA twice
+                    if (toClose != null && toClose.toASCIIString().equals(current.toASCIIString())) {
+                        toClose = null; // don't attempt to finish the LRA twice
+                    }
                 }
             }
 
             if (toClose != null) {
-                lraClient.closeLRA(toClose);
-                requestContext.getHeaders().remove(LRA_HTTP_HEADER);
-
-                if (toClose.toASCIIString().equals(
-                        Current.getLast(requestContext.getHeaders().get(LRA_HTTP_HEADER)))) {
-                    // the callers context was ended so invalidate it
+                try {
+                    lraClient.closeLRA(toClose);
+                } catch (NotFoundException ignore) {
+                    // must already be cancelled (if the intercepted method caused it to cancel)
+                    // or completed (if the intercepted method caused it to complete
+                } finally {
                     requestContext.getHeaders().remove(LRA_HTTP_HEADER);
+
+                    if (toClose.toASCIIString().equals(
+                            Current.getLast(requestContext.getHeaders().get(LRA_HTTP_HEADER)))) {
+                        // the callers context was ended so invalidate it
+                        requestContext.getHeaders().remove(LRA_HTTP_HEADER);
+                    }
                 }
             }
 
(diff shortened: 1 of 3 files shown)


mmusgrov:
The single test failure (mixedMultiLevelNestedActivity) is a known issue (https://issues.jboss.org/browse/JBTM-3127) and fails without my PR.

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -270,7 +270,7 @@ public void filter(ContainerRequestContext containerRequestContext) throws IOExc
 //                    previous = AtomicAction.suspend();
                 suspendedLRA = incommingLRA;
                 lraTrace(containerRequestContext, suspendedLRA, "ServerLRAFilter before: REQUIRES_NEW start new LRA");
-                newLRA = lraId = startLRA(incommingLRA, method, getTimeOut(method));
+                newLRA = lraId = startLRA(nested ? incommingLRA : null, method, getTimeOut(method));
```
ochaloup: just thinking if it is better readable to use
```
if(nested) {
...
}
```
as it's used in other `case` branches in the code around. It just came to my mind when I'm looking at the PR. It could be subjective.

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -394,26 +394,36 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
...
-                lraClient.cancelLRA(current);
-                if (current.toASCIIString().equals(
-                        Current.getLast(requestContext.getHeaders().get(LRA_HTTP_HEADER)))) {
-                    // the callers context was ended so invalidate it
-                    requestContext.getHeaders().remove(LRA_HTTP_HEADER);
-                }
+                try {
+                    lraClient.cancelLRA(current);
+                } catch (NotFoundException ignore) {
+                    // must already be cancelled
```
ochaloup: I think it is useful to write a trace log in situation like this.

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -394,26 +394,36 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
...
-                lraClient.closeLRA(toClose);
-                requestContext.getHeaders().remove(LRA_HTTP_HEADER);
-
-                if (toClose.toASCIIString().equals(
-                        Current.getLast(requestContext.getHeaders().get(LRA_HTTP_HEADER)))) {
-                    // the callers context was ended so invalidate it
+                try {
+                    lraClient.closeLRA(toClose);
+                } catch (NotFoundException ignore) {
+                    // must already be closed
```
ochaloup: similarly as above, it would be good to have a trace log here (with information what LRA id has to be already closed)

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
tomjenkinson: I think that this would be easier to read:
```
String value = nested ? incommingLRA : null;
newLRA = lraId = startLRA(value, method, getTimeOut(method));
```
As there are different stylistic ways to do this and they have different merits I don't think we can force a particular approach here.

mmusgrov: Is this second coding style satisfactory?

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: If the LRA has already cancelled (either explicitly by the application or because it timed out then the exception path is normal behaviour and therefore does not need to be logged.

On the other hand if the application closed it then I would argue it is an application decision and therefore, again does not need to be logged.

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: This is the same comment and I have the same response.

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
tomjenkinson: I think all three styles are valid so we should go with the authors choice

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
tomjenkinson: I agree, we should ensure that the spec is clear that cancelOn only applies when the transaction is still in a state it can be cancelled. I think the doc here should be updated to say:
// must already be completed

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
tomjenkinson: Same comment as above^^

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
ochaloup: no problem, I agree it's the authors choice

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
ochaloup: I think there are cases when some error condition occurs and logging such situation to `trace` or `debug` is beneficial. Even it's assumed such situation it's expected or it should never happen.

tomjenkinson: I do agree with the point that more verbose logging is useful typically I think the reason where this might not be appropriate to add in this case is for example we are not logging 
```
lraClient.cancelLRA(current);
traceln("Was cancelled);
```
In the same way Mike has analysed this and would say that the NotFoundException is expected in some cases.

ochaloup: I would argue that exception is exception and it should be considered exceptional in any case. Still I'm fine the PR would be left as it is.

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: If we think that certain java language constructs should not be used then we should update the style checker.

mmusgrov:
I raised https://github.com/eclipse/microprofile-lra/issues/129 to cover the final change request on this PR

PR #1433, opened 2019-04-01
RELEASE SCRIPTS ONLY: adding few more information to the error messages

ochaloup:
A small addition to the fixed release script.

NO_TEST

### DIFF
--- a/narayana-release-process.sh
+++ b/narayana-release-process.sh
@@ -126,7 +126,7 @@ cd ~/tmp/narayana/$CURRENT/sources/documentation/
 git checkout $CURRENT
 if [[ $? != 0 ]]
 then
-  echo Tag did not exist
+  echo 1>&2 documentation: Tag '$CURRENT' did not exist
   exit
 fi
 mvn clean install -Prelease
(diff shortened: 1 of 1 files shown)


narayana-release-process.sh
```diff
@@ -150,32 +150,32 @@ fi
 mvn clean install -DskipTests -gs tools/maven/conf/settings.xml -Dorson.jar.location=$ORSON_PATH -Pcommunity
 if [[ $? != 0 ]]
 then
-  echo Could not install narayana
+  echo 1>&2 Could not install narayana
```
tomjenkinson: Hi Ondra, I will marked as approved as the main changes look good. I am not sure of the need for the 1>&2 on the echo, can you explain please?

ochaloup: It's used in the script (e.g. https://github.com/jbosstm/narayana/blob/master/narayana-release-process.sh#L55) so I think that was some decision to write the errors to the `STD_ERR` as defaul way. If that's not the good idea for these places I can remove it and let it being written to `STD_OUT`.

tomjenkinson: I think it is fine - thanks for the explanation!

PR #1434, opened 2019-04-01
SCRIPTS ONLY: unary operator expected error fix

ochaloup:
fix for error message

```
./scripts/hudson/narayana.sh: line 811: [: =: unary operator expected
```
!MAIN !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS !AS_TESTS !TOMCAT !JACOCO !LRA DB_TESTS mysql postgres db2 oracle

### DIFF
--- a/scripts/hudson/narayana.sh
+++ b/scripts/hudson/narayana.sh
@@ -808,7 +808,7 @@ function qa_tests_once {
     [ x$QA_TARGET = x ] || target=$QA_TARGET # the caller can force the build to run a specific target
 
     # run the ant target (QA_TESTMETHODS is a list of method names in QA_TESTGROUP to be executed)
-    if [ $QA_TRACE = 1 ]; then
+    if [ "x$QA_TRACE" = "x1" ]; then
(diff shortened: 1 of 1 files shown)


ochaloup:
@tomjenkinson I think there was this issue all the time. The functionality is not broken event without this fix. It just writes error to the console. I'm helping Ivan with investigation of some jdbc errors in testsuite so I was looking into CI and I just spotted this error thus I fixed it.

Shell script output for this issue
```
$ unset TEST
$ if [ $TEST = 1 ]; then echo "TEST variable is 'one'"; else echo "TEST variable is *NOT* 'one'"; fi
bash: [: =: unary operator expected
TEST variable is *NOT* 'one'
$ TEST=1
$ if [ $TEST = 1 ]; then echo "TEST variable is 'one'"; else echo "TEST variable is *NOT* 'one'"; fi
TEST variable is 'one'
```

PR #1436, opened 2019-04-02
JBTM-3129 disable the site plugin for the codeCoverage profile on JDK 11

mmusgrov:
https://issues.jboss.org/browse/JBTM-3129

!MAIN !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF !RTS !AS_TESTS !TOMCAT JACOCO !LRA !DB_TESTS !mysql !postgres !db2 !oracle !QA_JTA JDK11 !JDK8

### DIFF
--- a/vertx/module/pom.xml
+++ b/vertx/module/pom.xml
@@ -69,10 +69,6 @@
     <maven.compiler.plugin.version>3.0</maven.compiler.plugin.version>
     <maven.resources.plugin.version>2.6</maven.resources.plugin.version>
     <maven.clean.plugin.version>2.5</maven.clean.plugin.version>
-    <maven.surefire.plugin.version>2.14</maven.surefire.plugin.version>
-    <maven.failsafe.plugin.version>2.14</maven.failsafe.plugin.version>
-    <maven.surefire.report.plugin.version>2.14</maven.surefire.report.plugin.version>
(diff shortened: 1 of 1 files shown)


ochaloup:
The jira says the `codeCoverage` fails on Windows JDK11. Should not be the activation properties combined with the both criteria?

mmusgrov:
Thanks for the spot on the JIRA. I wasn't ready for the review yet since I have not been able to reproduce the issue on my lenovo laptop and was wondering if the same issue would show on the CI PR run.

mmusgrov:
The vertx mod-stm pom is the only one that uses the maven-surefire-report-plugin (which is the one causing the failure). Maybe just removing that is sufficient to fix the failure with the justification of making it consistent with the rest of the artefacts created by the project.

mmusgrov:
I updated the JIRA with a suggested "resolution":

> The failure occurs when running the maven-surefire-report-plugin. The vertx mod-stm pom is the only artefact in the narayana project using this plugin. Would there be any problem if we just removed it from the build. If we decide later that the plugin is useful then we should apply it to all of the modules.

mmusgrov:
I have removed the maven surefire report plugin as I suggest in the associated JIRA.

mmusgrov:
@tomjenkinson as long the incremental changes eventually converge to the correct solution ... ;)

PR #1440, opened 2019-04-10
JBTM-3135 Temporarily disable the BlackTie JNI tests on Windows

tomjenkinson:
https://issues.jboss.org/browse/JBTM-3135

This is to temporarily remove the JNI tests on Windows JDK11 as instructions are currently for a 32-bit JDK and a 32-bit JVM and JDK11 is only available as 32-bit

!MAIN !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB BLACKTIE !XTS !PERF !RTS !AS_TESTS !TOMCAT !JACOCO !LRA !DB_TESTS JDK11

### DIFF
--- a/blacktie/integration-tests/pom.xml
+++ b/blacktie/integration-tests/pom.xml
@@ -283,6 +283,27 @@
 	</properties>
 
         <profiles>
+          <!-- Temporarily disable the tests on Windows https://issues.jboss.org/browse/JBTM-3135 -->
+          <profile>
+            <id>jdk11</id>
+            <activation>
+              <os>
+                <family>Windows</family>
+              </os>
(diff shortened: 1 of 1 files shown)


blacktie/integration-tests/pom.xml
```diff
@@ -283,6 +283,27 @@
 	</properties>
 
         <profiles>
+          <!-- Temporarily disable the tests on Windows https://issues.jboss.org/browse/JBTM-3135 -->
+          <profile>
+            <id>jdk11</id>
```
ochaloup: maybe the profile could be named differently when it disable tests?

PR #1442, opened 2019-04-12
JBTM-3105 STM with nested transactions and fix for taxonomy test failure

mmusgrov:
https://issues.jboss.org/browse/JBTM-3105

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB BLACKTIE !XTS !PERF !RTS !AS_TESTS !TOMCAT !JACOCO !LRA !DB_TESTS !JDK11 !WIN !BLACKTIE

This update fixes the taxonomy test for verifying that STM implements the closed nested transaction model. It also improves/hardens those tests.

### DIFF
--- a/STM/src/test/java/org/jboss/stm/TaxonomyNestedTest.java
+++ b/STM/src/test/java/org/jboss/stm/TaxonomyNestedTest.java
@@ -1,0 +1,432 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2018, JBoss Inc., and individual contributors as indicated
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.jboss.stm;
+
+import com.arjuna.ats.arjuna.AtomicAction;
+import com.arjuna.ats.arjuna.coordinator.ActionStatus;
+import junit.framework.TestCase;
+import org.jboss.stm.annotations.Nested;
+import org.jboss.stm.annotations.ReadLock;
+import org.jboss.stm.annotations.Transactional;
+import org.jboss.stm.annotations.WriteLock;
+import org.jboss.stm.internal.RecoverableContainer;
+import org.junit.BeforeClass;
+
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * <p>
+ * There are no standards for STM and there are different models for how nested transactions behave.
+ * The following tests investigate the behaviour of the Narayana Software Transactional Memory
+ * implementation with respect to nested transactions.
+ * </p>
+ *
+ * <p>
+ * The tests are meant to supplement the STM discussion introduced
+ * <a href="https://jbossts.blogspot.com/2018/09/tips-on-how-to-evaluate-stm.html">in an online article</a>
+ * and include:
+ * <ul>
+ *   <li>checking that committing changes made by a closed nested transaction should be visible to the parent</li>
+ *   <li>checking that aborting changes made by a closed nested transaction should not be visible to the parent</li>
+ * </ul>
+ * The tests are ran in two configurations:
+ *   <li>run the nested action in the same thread as the parent action</li>
+ *   <li>run the nested action in a different thread from the one used by the parent action</li>
+ * </p>
+ */
+public class TaxonomyNestedTest extends TestCase {
+    @Transactional // define some transactional memory
+    @Nested // with nesting semantics
+
+    public interface AtomicInt {
+        int get() throws Exception;
+        void set(int value) throws Exception;
+    }
+
+    public class AtomicIntImpl implements AtomicInt {
+        private int state;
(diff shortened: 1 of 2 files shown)


STM/src/test/java/org/jboss/stm/TaxonomyNestedTest.java
```diff
@@ -0,0 +1,407 @@
...
+import org.jboss.stm.annotations.Transactional;
+import org.jboss.stm.annotations.WriteLock;
+import org.jboss.stm.internal.RecoverableContainer;
+import org.junit.BeforeClass;
+
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.atomic.AtomicInteger;
+
+// these tests are discussed in the article https://jbossts.blogspot.com/2018/09/tips-on-how-to-evaluate-stm.html
```
ochaloup: * It coudl be a good to add a summary of what the tests are about and then refers to details with the link to the article.
* As this is the comment covering the whole test class it would be good to use the `/** */ format of the comment to that being the generated javadoc.
* It could be better to use the link with `<a href="...`">...</a>` as the javadoc is then generated with the clicable link. 

Maybe I would consider to use the javadoc format of comments (`/** */` instead of `//`) in other places in the file as well. I'm not going to list them in this review as that could be intentional as this is just a test case and not an API expected to be used.

STM/src/test/java/org/jboss/stm/TaxonomyNestedTest.java
```diff
@@ -0,0 +1,407 @@
...
+
+        Message(MSG_TYPE what, Integer value) {
+            this(what, value, -1, null);
+        }
+    }
+
+    enum NESTED_TYPE {TOP_LEVEL, COMMIT, ABORT};
+
+    // define a task for performing nested reads and writes to transactional memory
+    private class ThreadedReaderWriter extends Thread {
```
ochaloup: I assume this inner class is quite complex. I could be good to consider it for being a "standalone" class.

STM/src/test/java/org/jboss/stm/TaxonomyTest.java
```diff
@@ -52,157 +52,6 @@ public void set(int value) throws Exception {
...
-        }
-
-        // since the child aborted and nesting follows the closed mode the value should still be 2
-        assertEquals(2, ai.get());
-    }
-
-    /**
-     * committing changes made by a closed nested transaction should be visible to the parent
-     */
-    public void testIsClosedNestedCommit() throws Exception {
```
ochaloup: As I don't understand all the inner details of the tescases I just wonder if this `testIsClosedNestedCommit` matches in functionality with `testNestedCommit` or if the `testNestedCommit` is a different testcase and the `testIsClosedNestedCommit` was removed for some reason.

STM/src/test/java/org/jboss/stm/TaxonomyNestedTest.java
mmusgrov: okay

STM/src/test/java/org/jboss/stm/TaxonomyNestedTest.java
mmusgrov: It is not particularly complex and rightly belongs to the class it resides in. If I ever need to reuse its functionality (in non nested scenarios) in later blogs in this series about STM then it would make sense to refactor it into a separate class.

STM/src/test/java/org/jboss/stm/TaxonomyTest.java
mmusgrov: They are testing different things: the first test runs the nested action in a separate thread whereas the second test runs the nested action in the same thread as the parent. I do have some brief javadoc and java comments above each method that explain what the tests are doing but I will add more text to make that clearer.

Secondly, the first test still exists but I moved it to TaxonomyNestedTest.java and the reason I did that is because the tests for nested actions require the transactional memory to be marked with @Nested (whereas the other tests should not carry this annotation). This change is part of the bug fix for JBTM-3105 that this PR is fixing:

> There are two issues with the nested tests:

> the transactional object is not marked with @Nested
> the second thread that performs a nested update catches Exceptions with fail() in the catch  block. This does not seem to cause the test to fail. The fix is twofold, catch Throwable and assert > failure in the main thread (see PR for the fix)

Thirdly, I also improve coverage so you will see some extra tests.

And finally, I also improved the testability by adding some communication between the parent and child threads so that the parent can check that the child thread is behaving correctly. This change simplifies how the test operates and makes it more resilient since it uses a different approach to inter thread communication (basically it uses message passing between the threads rather than synchronisation primitives which is easier to reason about).

STM/src/test/java/org/jboss/stm/TaxonomyNestedTest.java
mmusgrov: I added javadoc for the class and for each test.

The tests are meant to supplement the article and should be read in an IDE - ie they are not meant for API usage but rather for the twofold purpose of:
- illumination
- validating our STM behaviour

STM/src/test/java/org/jboss/stm/TaxonomyNestedTest.java
mmusgrov: For example, the next article in this series will cover concurrency (thread interaction, pessimistic, optimistic, when does the STM implementation detects failures, ... etc) and I would imagine this inner class could be extended to support that at which point it would make sense to refactor it into its own class file.

STM/src/test/java/org/jboss/stm/TaxonomyTest.java
ochaloup: +1 thank you for elaboration

STM/src/test/java/org/jboss/stm/TaxonomyNestedTest.java
```diff
@@ -0,0 +1,430 @@
...
+ * There are no standards for STM and there are different models for how nested transactions behave.
+ * The following tests investigate the behaviour of the Narayana Software Transactional Memory
+ * implementation with respect to nested transactions.
+ *
+ * The tests are meant to supplement the STM discussion introduced
+ * <a href="https://jbossts.blogspot.com/2018/09/tips-on-how-to-evaluate-stm.html">in an online article</a>.
+ *
+ * The tests include:
+ *
+ * - checking that committing changes made by a closed nested transaction should be visible to the parent
```
ochaloup: I would only mention that the javadoc is generated from html. It seems the markdown is used. The `-` will not be generated as list. The `<ul><li>` could be used for that being the list.

mmusgrov: Fixed

PR #1446, opened 2019-04-29
JBTM-3087 NESTED LRA type element value

mmusgrov:

 - NESTED LRA type element value
 - parent context header on nested LRAs

### DIFF
--- a/rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
+++ b/rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
@@ -29,7 +29,6 @@ import org.eclipse.microprofile.lra.annotation.Complete;
 import org.eclipse.microprofile.lra.annotation.Forget;
 import org.eclipse.microprofile.lra.annotation.ws.rs.LRA;
 import org.eclipse.microprofile.lra.annotation.ws.rs.Leave;
-import org.eclipse.microprofile.lra.annotation.ws.rs.NestedLRA;
 import org.eclipse.microprofile.lra.annotation.Status;
 
 import javax.inject.Inject;
@@ -44,7 +43,6 @@ import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.Provider;
-import java.io.IOException;
 import java.lang.reflect.Method;
 import java.net.URI;
 import java.net.URISyntaxException;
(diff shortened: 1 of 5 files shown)


mmusgrov:
I think the failures were caused by missing MP-LRA updates - when I updated the pom to use the latest MP-LRA snapshot builds it is passing

rts/lra/lra-test/tck/pom.xml
```diff
@@ -168,6 +168,7 @@
                                         <argument>-Djava.net.preferIPv4Stack=true</argument>
                                         <argument>${lra.coordinator.debug.params}</argument>
                                         <argument>${lra.coordinator.trace.params}</argument>
+                                        <argument>-DRecoveryEnvironmentBean.periodicRecoveryPeriod=10</argument>
```
ochaloup: should not be considered to change the back off time as well? I mean when periodic recovery is changed from 2min to 10 s then could be the backoff to 3 seconds could be enough? Or do I understand the purpose here wrongly?

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -215,7 +215,14 @@ public void filter(ContainerRequestContext containerRequestContext) throws IOExc
...
+        if (nested) {
+            type = LRA.Type.REQUIRED;
+
+            if (incommingLRA != null) {
+                headers.putSingle(LRA_HTTP_PARENT_CONTEXT_HEADER, incommingLRA.toASCIIString());
+            }
+        }
 
         switch (type) {
             case MANDATORY: // a txn must be present
```
xstefank: if type is `LRA.Type.NESTED` so `nested == true` none of these switch cases will ever be entered so I don't see any reason for keeping `if (nested)` blocks inside these cases (e.g. line 224). I would also prefer to move `if (nested)` block from line 218 to a new case branch in this switch so the approach is unified.

mmusgrov: We may need to revisit the ability to set attributes on nested LRAs so I don't see any harm in leaving the tests. When/if we need to look at performance we can look into it then.

rts/lra/lra-test/tck/pom.xml
mmusgrov: A backoff of zero is fine so I will update the PR to set that value

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
xstefank: sorry, I don't understand. We've changed the spec so now nested LRA is defined by type and there will never be a case that the type will be MANDATORY and NESTED at the same time so this nested ifs are never to be executed.

ochaloup: @mmusgrov: I agree with @xstefank . I haven't recognized this relation to the spec. Re-factoring the switch would make the code nicer and enhance the readability. I really consider such clean-up as a  benefit. (In case there is a some change of the spec in future there is still the history in the git.)

ochaloup: My concern is readability and understandability of the code. Layering the clean-up as a next commit sounds good to me.

mmusgrov:
The latest commit should address your concerns, it also cleans up some IDE warnings.

mmusgrov:
@ochaloup and @xstefank I trust that my push addressed your comments sufficiently. I cherry picked this commit onto PR 1448 which was merged so I am closing this PR.

PR #1448, opened 2019-05-01
Jbtm 3087 mp lra issue149

mmusgrov:

1. This PR updates LRA to use JAX-RS 2.1 (which adds reactive features to JAX-RS and is a version supported by Thorntail)
2. The PR also includes the commit to accommodate MP-LRA issue 117 which replaced the @NestedLRA annotation with an LRA.Type.Nested element (and exposes the parent context in a JAX-RS header when a nested LRA is active)

### DIFF
--- a/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
+++ b/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
@@ -61,6 +61,7 @@ import java.util.concurrent.locks.ReentrantLock;
 import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
 import static javax.ws.rs.core.Response.Status.PRECONDITION_FAILED;
 import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_CONTEXT_HEADER;
+import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_PARENT_CONTEXT_HEADER;
 import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_RECOVERY_HEADER;
 
 public class LRARecord extends AbstractRecord implements Comparable<AbstractRecord> {
@@ -71,6 +72,7 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
 
     private Transaction lra;
     private URI lraId;
+    private URI parentId;
     private URI recoveryURI;
     private String participantPath;
 
@@ -120,6 +122,8 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
             }
 
             this.lraId = new URI(lraId);
+            this.parentId = lraService.getTransaction(this.lraId).getParentId();
+
             this.lraService = lraService;
             this.participantPath = linkURI;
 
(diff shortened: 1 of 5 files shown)


rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -305,8 +309,10 @@ private int tryDoEnd(boolean compensate) {
                 // ask the participant to complete or compensate
                 Future<Response> asyncResponse = target.request()
                         .header(LRA_HTTP_CONTEXT_HEADER, lraId.toASCIIString())
+                        .header(LRA_HTTP_PARENT_CONTEXT_HEADER, parentId) // make the context available to participants
                         .header(LRA_HTTP_RECOVERY_HEADER, recoveryURI.toASCIIString())
-                        .property(LRA_HTTP_CONTEXT_HEADER, lraId) // make the context available to the jaxrs filters
+                        .property(LRA_HTTP_CONTEXT_HEADER, lraId) // make the context available to the jaxrs filters)
```
tomjenkinson: It might be something I don't see in the diff but seems to have gained a spurious ")" at the end of the line

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -214,51 +185,32 @@ public void filter(ContainerRequestContext containerRequestContext) throws IOExc
...
+                lraId = null; // must not run with any context
 
                 break;
+            case NESTED:
+                // FALLTHRU
             case REQUIRED:
                 if (incommingLRA != null) {
-                    if (nested) {
+                    if (type == NESTED) {
+                        headers.putSingle(LRA_HTTP_PARENT_CONTEXT_HEADER, incommingLRA.toASCIIString());
```
tomjenkinson: Does anything ever use LRA_HTTP_PARENT_CONTEXT_HEADER - I just see this put and some passing around earlier in the diff?

rts/lra/lra-test/tck/pom.xml
```diff
@@ -168,6 +167,8 @@
                                         <argument>-Djava.net.preferIPv4Stack=true</argument>
                                         <argument>${lra.coordinator.debug.params}</argument>
                                         <argument>${lra.coordinator.trace.params}</argument>
+                                        <argument>-DRecoveryEnvironmentBean.periodicRecoveryPeriod=10</argument>
```
tomjenkinson: this might be too short in CI, I guess we will find out

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: The MP-LRA test suite validates that it is present on participant callbacks and on the business method where the nested LRA is creates

rts/lra/lra-test/tck/pom.xml
mmusgrov: It seems to work fine (on CI and the laptop) - the goal is to minimise the time on the TCK tests (since we no longer have a "trigger recovery mechanism")

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
tomjenkinson: OK, thanks

mmusgrov: There are a few uses but an example of where we validate that it is available on participant callbacks is:  https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ContextTckResource.java#L277

rts/lra/lra-test/tck/pom.xml
mmusgrov: Also it is only for testing against the LRA axis where we use arquillian to launch a thorntail MP container so will not affect any other jobs.

PR #1450, opened 2019-05-20
JBTM-3087 Ensure the LRA implementation is up to date

xstefank:

This PR updates LRA to:
```xml
<version.microprofile.lra.api>1.0-20190519.061148-384</version.microprofile.lra.api>
<version.microprofile.lra.tck>1.0-20190519.061150-384</version.microprofile.lra.tck>
```

It provides mainly [spec issue 35](https://github.com/eclipse/microprofile-lra/issues/35) implementation and the update for the removal of LRAManagement from the spec.

Please note that possible CompletionStage optimization is left for another PR.

### DIFF
--- a/rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRACDIExtension.java
+++ b/rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRACDIExtension.java
@@ -1,0 +1,118 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2019, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package io.narayana.lra.client.internal.proxy.nonjaxrs;
+
+import io.narayana.lra.client.internal.proxy.nonjaxrs.jandex.DotNames;
+import io.narayana.lra.logging.LRALogger;
+import org.jboss.jandex.AnnotationInstance;
+import org.jboss.jandex.AnnotationTarget;
+import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.Index;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.event.Observes;
+import javax.enterprise.inject.Any;
+import javax.enterprise.inject.spi.AfterBeanDiscovery;
+import javax.enterprise.inject.spi.Bean;
+import javax.enterprise.inject.spi.BeanManager;
+import javax.enterprise.inject.spi.Extension;
+import javax.enterprise.util.AnnotationLiteral;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * This CDI extension collects all LRA participants that contain
+ * one or more non-JAX-RS participant methods. The collected classes are stored
+ * in {@link LRAParticipantRegistry}.
+ */
+public class LRACDIExtension implements Extension {
+
+    private ClassPathIndexer classPathIndexer = new ClassPathIndexer();
+    private final Map<String, LRAParticipant> participants = new HashMap<>();
+
+    public void observe(@Observes AfterBeanDiscovery event, BeanManager beanManager) throws IOException, ClassNotFoundException {
+        Index index = classPathIndexer.createIndex();
+
+        List<AnnotationInstance> annotations = index.getAnnotations(DotName.createSimple("javax.ws.rs.Path"));
+
+        for (AnnotationInstance annotation : annotations) {
+            ClassInfo classInfo;
+            AnnotationTarget target = annotation.target();
+
+            if (target.kind().equals(AnnotationTarget.Kind.CLASS)) {
+                classInfo = target.asClass();
+            } else if (target.kind().equals(AnnotationTarget.Kind.METHOD)) {
+                classInfo = target.asMethod().declaringClass();
+            } else {
+                continue;
+            }
+
+            LRAParticipant participant = getAsParticipant(classInfo);
+            if (participant != null) {
+                participants.put(participant.getJavaClass().getName(), participant);
+                Set<Bean<?>> participantBeans = beanManager.getBeans(participant.getJavaClass(), new AnnotationLiteral<Any>() {});
+                if (participantBeans.isEmpty()) {
+                    // resource is not registered as managed bean so register a custom managed instance
+                    try {
+                        participant.setInstance(participant.getJavaClass().newInstance());
+                    } catch (InstantiationException | IllegalAccessException e) {
+                        LRALogger.i18NLogger.error_cannotProcessParticipant(e);
+                    }
+                }
+            }
+        }
+
+        event.addBean()
+            .read(beanManager.createAnnotatedType(LRAParticipantRegistry.class))
+            .beanClass(LRAParticipantRegistry.class)
+            .scope(ApplicationScoped.class)
+            .createWith(context -> new LRAParticipantRegistry(participants));
+    }
+
+    /**
+     * Collects all non-JAX-RS participant methods in the defined Java class
+     *
+     * @param classInfo a Jandex class info of the class to be scanned
+     * @return Collected methods wrapped in {@link LRAParticipant} class or null if no non-JAX-RS methods have been found
+     */
+    private LRAParticipant getAsParticipant(ClassInfo classInfo) throws ClassNotFoundException {
+        if (isNotLRAParticipant(classInfo)) {
+            return null;
+        }
+
+        Class<?> javaClass = getClass().getClassLoader().loadClass(classInfo.name().toString());
+
+        LRAParticipant participant = new LRAParticipant(javaClass);
+        return participant.hasNonJaxRsMethods() ? participant : null;
+    }
+
+    private boolean isNotLRAParticipant(ClassInfo classInfo) {
+        Map<DotName, List<AnnotationInstance>> annotations = classInfo.annotations();
+        return !annotations.containsKey(DotNames.LRA) ||
+            (!annotations.containsKey(DotNames.COMPENSATE) &&
+                !annotations.containsKey(DotNames.AFTER_LRA));
+    }
+}
--- a/rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRAParticipant.java
+++ b/rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRAParticipant.java
@@ -1,0 +1,407 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2019, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package io.narayana.lra.client.internal.proxy.nonjaxrs;
+
+import io.narayana.lra.logging.LRALogger;
+import org.eclipse.microprofile.lra.annotation.AfterLRA;
+import org.eclipse.microprofile.lra.annotation.Compensate;
+import org.eclipse.microprofile.lra.annotation.Complete;
+import org.eclipse.microprofile.lra.annotation.Forget;
+import org.eclipse.microprofile.lra.annotation.LRAStatus;
+import org.eclipse.microprofile.lra.annotation.ParticipantStatus;
+import org.eclipse.microprofile.lra.annotation.Status;
+
+import javax.enterprise.inject.spi.CDI;
+import javax.ws.rs.DELETE;
+import javax.ws.rs.GET;
+import javax.ws.rs.HEAD;
+import javax.ws.rs.OPTIONS;
+import javax.ws.rs.POST;
+import javax.ws.rs.PUT;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.UriBuilder;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.net.URI;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.CompletionStage;
+
+import static io.narayana.lra.LRAConstants.AFTER;
+import static io.narayana.lra.LRAConstants.COMPENSATE;
+import static io.narayana.lra.LRAConstants.COMPLETE;
+import static io.narayana.lra.LRAConstants.FORGET;
+import static io.narayana.lra.LRAConstants.STATUS;
+
+/**
+ * Keeps references to individual non-JAX-RS paraticipant methods in
+ * single LRA participant class.
+ */
+public class LRAParticipant {
+
+    private Class<?> javaClass;
+    private Method compensateMethod;
+    private Method completeMethod;
+    private Method statusMethod;
+    private Method forgetMethod;
+    private Method afterLRAMethod;
+    private Object instance;
+
+    private Map<URI, Optional<?>> participantStatusMap = new HashMap<>();
+
+    LRAParticipant(Class<?> javaClass) {
+        this.javaClass = javaClass;
+
+        Arrays.stream(javaClass.getDeclaredMethods()).forEach(this::processParticipantMethod);
+    }
+
+    Class<?> getJavaClass() {
+        return javaClass;
+    }
+
+    synchronized Response compensate(URI lraId, URI parentId) {
+        if (participantStatusMap.containsKey(lraId)) {
+            processCompletionStageResult(compensateMethod, lraId, parentId, COMPENSATE);
+        }
+
+        return invokeParticipantMethod(compensateMethod, lraId, parentId, COMPENSATE);
+    }
+
+    synchronized Response complete(URI lraId, URI parentId) {
+        if (participantStatusMap.containsKey(lraId)) {
+            processCompletionStageResult(completeMethod, lraId, parentId, COMPLETE);
+        }
+
+        return invokeParticipantMethod(completeMethod, lraId, parentId, COMPLETE);
+    }
+
+    synchronized Response status(URI lraId, URI parentId) {
+        if (participantStatusMap.containsKey(lraId)) {
+            return processCompletionStageResult(statusMethod, lraId, parentId, STATUS);
+        }
+
+        return invokeParticipantMethod(statusMethod, lraId, parentId, STATUS);
+    }
+
+    synchronized Response forget(URI lraId, URI parentId) {
+        return invokeParticipantMethod(forgetMethod, lraId, parentId, FORGET);
+    }
+
+    synchronized void afterLRA(URI lraId, LRAStatus lraStatus) {
+        invokeMethod(AFTER, afterLRAMethod, getInstance(), lraId, lraStatus);
+    }
+
+    /**
+     * Process participant method for non JAX-RS related processing
+     * defined by the specification and verify its signature
+     *
+     * @param method method to be processed
+     */
+    private void processParticipantMethod(Method method) {
+
+        if (isJaxRsMethod(method)) {
+            return;
+        }
+
+        if (setAfterLRAAnnotation(method) || !setParticipantAnnotation(method)) {
+            return;
+        }
+
+        verifyReturnType(method);
+
+        Class<?>[] parameterTypes = method.getParameterTypes();
+
+        if (parameterTypes.length > 2) {
+            throw new IllegalStateException(String.format("%s: %s",
+                method.toGenericString(), "Participant method cannot have more than 2 arguments"));
+        }
+
+        if (parameterTypes.length > 0 && !parameterTypes[0].equals(URI.class)) {
+            throw new IllegalStateException(String.format("%s: %s",
+                method.toGenericString(), "Invalid argument type in LRA participant method: " + parameterTypes[0]));
+        }
+
+        if (parameterTypes.length > 1 && !parameterTypes[1].equals(URI.class)) {
+            throw new IllegalStateException(String.format("%s: %s",
+                method.toGenericString(), "Invalid argument type in LRA participant method: " + parameterTypes[1]));
+        }
+    }
+
+    private boolean setAfterLRAAnnotation(Method method) {
+        if (!method.isAnnotationPresent(AfterLRA.class)) {
+            return false;
+        }
+
+        // spec defined signature is "void afterLRA(URI, LRAStatus)
+        if (!method.getReturnType().equals(Void.TYPE)) {
+            throw new IllegalStateException(String.format("%s: %s",
+                method.toGenericString(), "Invalid return type for @AfterLRA method: " + method.getReturnType()));
+        }
+
+        Class<?>[] parameterTypes = method.getParameterTypes();
+        if (parameterTypes.length > 2) {
+            throw new IllegalStateException(String.format("%s: %s",
+                method.toGenericString(), "@AfterLRA method cannot have more than 2 arguments"));
+        }
+
+        if (parameterTypes.length > 0 && !parameterTypes[0].equals(URI.class)) {
+            throw new IllegalStateException(String.format("%s: %s",
+                method.toGenericString(), "Invalid first argument of @AfterLRA method: " + parameterTypes[0]));
+        }
+
+        if (parameterTypes.length > 1 && !parameterTypes[1].equals(LRAStatus.class)) {
+            throw new IllegalStateException(String.format("%s: %s",
+                method.toGenericString(), "Invalid second argument of @AfterLRA method: " + parameterTypes[1]));
+        }
+
+        afterLRAMethod = method;
+        return true;
+    }
+
+    private boolean isJaxRsMethod(Method method) {
+        return method.isAnnotationPresent(GET.class) ||
+            method.isAnnotationPresent(POST.class) ||
+            method.isAnnotationPresent(PUT.class) ||
+            method.isAnnotationPresent(DELETE.class) ||
+            method.isAnnotationPresent(HEAD.class) ||
+            method.isAnnotationPresent(OPTIONS.class);
+    }
+
+    private boolean setParticipantAnnotation(Method method) {
+        if (method.isAnnotationPresent(Compensate.class)) {
+            compensateMethod = method;
+            return true;
+        } else if (method.isAnnotationPresent(Complete.class)) {
+            completeMethod = method;
+            return true;
+        } else if (method.isAnnotationPresent(Status.class)) {
+            statusMethod = method;
+            return true;
+        } else if (method.isAnnotationPresent(Forget.class)) {
+            forgetMethod = method;
+            return true;
+        }
+
+        return false;
+    }
+
+    private void verifyReturnType(Method method) {
+        Class<?> returnType = method.getReturnType();
+
+        if (returnType.equals(CompletionStage.class)) {
+            Type genericReturnType = method.getGenericReturnType();
+            ParameterizedType parameterizedType = (ParameterizedType) genericReturnType;
+            verifyReturnType((Class<?>) parameterizedType.getActualTypeArguments()[0], method.toGenericString(), true);
+            return;
+        }
+
+        verifyReturnType(returnType, method.toGenericString(), false);
+    }
+
+    private void verifyReturnType(Class<?> returnType, String methodGenericString, boolean inCompletionStage) {
+        if (!returnType.equals(Void.TYPE) &&
+            !returnType.equals(Void.class) &&
+            !returnType.equals(ParticipantStatus.class) &&
+            !returnType.equals(Response.class)) {
+            throw new IllegalStateException(String.format("%s: %s",
+                methodGenericString, "Invalid return type for participant method "
+                    + (inCompletionStage ? "CompletionStage<" + returnType + ">" : returnType)));
+        }
+    }
+
+    private Response processCompletionStageResult(Method method, URI lraId, URI parentId, String type) {
+        Optional<?> optional = participantStatusMap.get(lraId);
+        if (optional.isPresent()) {
+            participantStatusMap.remove(lraId);
+
+            // only Optionals with ParticipantResult instances are put into the participantStatusMap so this cast is safe
+            Object result = optional.get();
+
+            if (shouldInvokeParticipantMethod(result)) {
+                LRALogger.i18NLogger.warn_participantReturnsImmediateStateFromCompletionStage(
+                    getJavaClass().getName(), lraId.toASCIIString());
+                return invokeParticipantMethod(method, lraId, parentId, type);
+            }
+
+            return processResult(result, lraId, method, type);
+        } else {
+            // participant is still compensating / compeleting
+            return Response.accepted().build();
+        }
+    }
+
+    private boolean shouldInvokeParticipantMethod(Object result) {
+        if (result instanceof ParticipantStatus) {
+            ParticipantStatus participantStatus = (ParticipantStatus) result;
+            if (participantStatus.equals(ParticipantStatus.Compensating) ||
+                participantStatus.equals(ParticipantStatus.Completing)) {
+                return true;
+            }
+        } else if (result instanceof Response && ((Response) result).getStatus() == Response.Status.ACCEPTED.getStatusCode()) {
+            return true;
+        }
+
+        return false;
+    }
+
+    private Response invokeParticipantMethod(Method method, URI lraId,
+                                             URI parentId, String type) {
+        Object participant = getInstance();
+        Object result;
+
+        switch (method.getParameterCount()) {
+            case 0:
+                result = invokeMethod(type, method, participant);
+                break;
+            case 1:
+                result = invokeMethod(type, method, participant, lraId);
+                break;
+            case 2:
+                result = invokeMethod(type, method, participant, lraId, parentId);
+                break;
+            default:
+                throw new IllegalStateException(
+                    method.toGenericString() + ": invalid number of arguments: " + method.getParameterCount());
+        }
+
+        return processResult(result, lraId, method, type);
+    }
+
+    private Object getInstance() {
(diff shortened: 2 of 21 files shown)


mmusgrov:
I am back onto this again. How did you test your changes. Here's I what I tried:

> git fetch upstream
> git co pr/1450
> ./build.sh clean install -DskipTests
> cd rts/lra
> mvn verify -Parq

which produces some strang error about foreign imports:

> [ERROR] Number of foreign imports: 1
> [ERROR] import: Entry[import  from realm ClassRealm[maven.api, parent: null]]

xstefank:
@mmusgrov the same error occurs on current master. After some digging I found our that it's a maven issue - https://issues.jboss.org/browse/JBTM-3149

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRACDIExtension.java
```diff
@@ -0,0 +1,150 @@
...
+            throw newInvalidLRAParticipantDefinitionException("Participant method cannot have more than 2 arguments");
+        }
+
+        Arrays.stream(parameterTypes).forEach(this::verifyParamType);
+
+        if (method.getAnnotation(Compensate.class) != null) {
+            participant.setCompensateMethod(method);
+        } else if (method.getAnnotation(Complete.class) != null) {
+            participant.setCompleteMethod(method);
+        } else if (method.getAnnotation(Status.class) != null) {
```
mmusgrov: Remark: This reminds me that I need to add a spec update to say that if multiple methods are marked with Status or Forget then an arbitrary one is chosen (in your case the last one wins).

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRACDIExtension.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.Arrays;
+import java.util.concurrent.CompletionStage;
+
+public class LRACDIExtension implements Extension {
+
+    private String currentMethodGenericString;
+
+    private LRAParticipantRegistry participantRegistry = LRAParticipantRegistry.getInstance();
+
+    public void observe(@Observes @WithAnnotations(Path.class) ProcessAnnotatedType<?> type) {
```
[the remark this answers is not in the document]
mmusgrov: Ideally we would use standard Java annotations, then the implementation will have more general applicability.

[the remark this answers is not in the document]
xstefank: JAX-RS specification states that each resource (@Path) needs to be a CDI bean (https://download.oracle.com/otn-pub/jcp/jaxrs-2_1-final-eval-spec/jaxrs-2_1-final-spec.pdf?AuthParam=1559135898_a2cb061a2106994cbc4f71db60edcf47 section 11.2.3)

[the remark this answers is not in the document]
mmusgrov: Is that new in 2.1?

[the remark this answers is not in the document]
mmusgrov: Are you referring to the sentence:

> In a product that supports CDI, implementations MUST support the use of CDI-style Beans as root resource classes, providers and Application subclasses

If so, that would mean our implementation can only be used in CDI + JAX-RS containers whereas currently it should be possible to use it in just a JAX-RS container. Since standard java annotation processing is pretty straight forward, introducing the restriction seems unnecessary.

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRAParticipantResource.java
```diff
@@ -0,0 +1,99 @@
...
+    @Produces(MediaType.TEXT_PLAIN)
+    @Forget
+    public Response forget(@PathParam("participantId") String participantId,
+                           @HeaderParam(LRA_HTTP_CONTEXT_HEADER) String lraId,
+                           @HeaderParam(LRA_HTTP_PARENT_CONTEXT_HEADER) String parentId) {
+        return lraParticipantRegistry.getParticipant(participantId).forget(createURI(lraId), createURI(parentId));
+    }
+
+    private URI createURI(String value) {
+        return value != null ? URI.create(value) : null;
```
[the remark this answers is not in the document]
xstefank: these values will be passed to the actual participant invocations so they should be null if they are not received by the invocation (they are taken from the header values).

xstefank:
@mmusgrov just rebase on current master

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRAParticipant.java
```diff
@@ -0,0 +1,209 @@
...
+            // wait for result the same way as JAX-RS filters do
+            try {
+                result = ((CompletionStage) result).toCompletableFuture().get();
+            } catch (InterruptedException | ExecutionException e) {
+                throw new RuntimeException(e);
+            }
+        }
+
+        if (result == null) {
+            builder.status(Response.Status.NOT_FOUND);
```
mmusgrov: result will be null if the return type of the method is void so returning NOT_FOUND is incorrect.

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRAParticipantResource.java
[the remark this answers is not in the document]
mmusgrov: The participant will not be invoked because
```java
URI.create(value)
```
with throw a null pointer which will be caught by resteasy and converted to a generic 500 code

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRAParticipant.java
```diff
@@ -0,0 +1,209 @@
...
+
+    void setStatusMethod(Method statusMethod) {
+        this.statusMethod = statusMethod;
+    }
+
+    void setForgetMethod(Method forgetMethod) {
+        this.forgetMethod = forgetMethod;
+    }
+
+    Method getCompensateMethod() {
```
mmusgrov: This is the only getter method that you need since if the compensate method is absent then by definition it is not a participant (see my earlier comment about how to identify a participant resource), ie the other getters can be removed.

mmusgrov:
@xstefank Unfortunately I added my comments as another review instead of directly responding to your responses. Do you want me to redo my responses or can you figure out which of your responses I am responding to?

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRACDIExtension.java
```diff
@@ -0,0 +1,150 @@
...
+    /**
+     * Process participant method for non JAX-RS related processing
+     * defined by the specification and verify its signature
+     * @param method method to be processed
+     *
+     * @throws InvalidLRAParticipantDefinitionException if error is detected
+     */
+    private void processParticipantMethod(Method method, LRAParticipant participant) {
+
+        if (isJaxRsMethod(method) || isNotParticipantMethod(method)) {
```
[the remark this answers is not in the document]
mmusgrov: Related to this is one of my comments on your update to this line:

> You have already (effectively) performed this set of tests when you called getParticipantAnnotation on line 93. You should set the method as soon as you discover the annotation type.

So I will leave this conversation as unresolved for now since it is related.

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRAParticipantRegistry.java
```diff
@@ -0,0 +1,54 @@
...
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package io.narayana.lra.client.internal.proxy.nonjaxrs;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class LRAParticipantRegistry {
+
+    private static Map<Class<?>, LRAParticipant> lraParticipants = new HashMap<>();
```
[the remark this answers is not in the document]
mmusgrov: See my other comment about adding an application scoped bean in a CDI extension.

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRAParticipantResource.java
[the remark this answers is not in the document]
mmusgrov: My last comment was invalid, thanks for the answer. I will resolve this converstation.

xstefank:
@mmusgrov addressed all your comments (if I didn't miss anything). Can you please take a look?

rts/lra/pom.xml
```diff
@@ -22,14 +22,14 @@
...
+        <version.cdi-api>2.0</version.cdi-api>
         <version.junit>4.12</version.junit>
 
         <version.thorntail>2.4.0.Final</version.thorntail>
         <version.arquillian>1.2.1.Final</version.arquillian> <!-- cannot use the up-to-date Arquillian https://issues.jboss.org/browse/THORN-2090 -->
 
-        <version.microprofile.lra.api>1.0-20190515.061205-378</version.microprofile.lra.api>
-        <version.microprofile.lra.tck>1.0-20190515.061206-378</version.microprofile.lra.tck>
+        <version.microprofile.lra.api>1.0-20190606.061200-408</version.microprofile.lra.api>
+        <version.microprofile.lra.tck>1.0-20190606.061201-408</version.microprofile.lra.tck>
```
mmusgrov: This is quite old (Thu Jun 06).

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -320,7 +320,19 @@ private int tryDoEnd(boolean compensate) {
 
                 httpStatus = response.getStatus();
 
-                accepted = httpStatus == Response.Status.ACCEPTED.getStatusCode();
+                ParticipantStatus participantStatus = null;
```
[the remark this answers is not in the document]
xstefank: @mmusgrov yes, this one other bug I discovered when running newly added TCK. I will raise a separate JIRA for it then.

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRAParticipant.java
```diff
@@ -0,0 +1,249 @@
...
+
+        return invokeParticipantMethod(statusMethod, lraId, parentId, STATUS);
+    }
+
+    private boolean shouldInvokeStatusMethod(Object result) {
+        if (result instanceof ParticipantStatus) {
+            ParticipantStatus participantStatus = (ParticipantStatus) result;
+            if (participantStatus.equals(ParticipantStatus.Compensating) ||
+                participantStatus.equals(ParticipantStatus.Completing)) {
+                return true;
```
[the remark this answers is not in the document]
xstefank: ok

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRAParticipant.java
```diff
@@ -0,0 +1,249 @@
...
+
+    Response status(URI lraId, URI parentId) {
+        if (participantStatusMap.containsKey(lraId)) {
+            Optional<?> optional = participantStatusMap.get(lraId);
+            if (optional.isPresent()) {
+                participantStatusMap.remove(lraId);
+                
+                Object result = optional.get();
+                if (shouldInvokeStatusMethod(result)) {
+                    return invokeParticipantMethod(statusMethod, lraId, parentId, STATUS);
```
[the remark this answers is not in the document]
xstefank: I am relying here on coordinator because it is calling status on the participant right now it must have found out that this participant has a status endpoint thus method. However, this a good catch because in case there is no status compensate or complete will get reinvoked, so we should apply same logic to both of this methods. I will update the PR later today. Thanks!

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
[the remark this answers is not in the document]
ochaloup: @mmusgrov as I'm looking at these code changes in `LRARecord` they fix the issue reported as the JBTM-3156. I think they should be fine to be accepted for the PR.
Please, what is the functional trouble with these code changes from your perspective?

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRAParticipant.java
```diff
@@ -0,0 +1,249 @@
...
+    private Response processResult(Object result, URI lraId, Method method, String type) {
+        Response.ResponseBuilder builder = Response.status(Response.Status.OK);
+
+        if (result instanceof CompletionStage) {
+            // store the CompletionStage result and respond compensating / completing
+            participantStatusMap.put(lraId, Optional.empty());
+            ((CompletionStage<?>) result)
+                .thenAccept(res -> participantStatusMap.replace(lraId, Optional.ofNullable(res)))
+                .exceptionally(throwable -> {
+                    participantStatusMap.replace(lraId, Optional.ofNullable(throwable));
```
mmusgrov: So you correctly remember that the participant threw an exception but you never act upon it. In the `if then else` block below you need:

```java
        } else if (result instanceof Throwable) {
            if type is compensate
                return ParticipantStatus.FailedToCompensate.name();
            else if type is complete
                return ParticipantStatus.FailedToComplete.name();
```

I still think that all of these optional signatures is overly complex for both the application developer and for the spec implementer: doing specs by committee is well known to result in suboptimal results and that is what we have here. I think the mistakes you have made in the last review and in this latest review only serve to highlight these facts. My original solution did not suffer from either of these problems, mainly because the original feature was simply stated. I think that any future potential spec implementors will take one look at this stuff and decide, no, it is too complicated.

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRAParticipant.java
```diff
@@ -0,0 +1,249 @@
...
+        if (method.getReturnType().equals(Void.TYPE)) {
+            // void return type and no exception was thrown
+            builder.entity(type.equals(COMPLETE) ? ParticipantStatus.Completed.name() : ParticipantStatus.Compensated.name());
+        } else if (result == null) {
+            builder.status(Response.Status.NOT_FOUND);
+        } else if (result instanceof ParticipantStatus) {
+            builder.entity(((ParticipantStatus) result).name());
+        } else if (result instanceof Response) {
+            return (Response) result;
+        } else {
```
mmusgrov: This is where you would end up if the participant business logic throws an exception.

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRAParticipant.java
```diff
@@ -0,0 +1,249 @@
...
+            Optional<?> optional = participantStatusMap.get(lraId);
+            if (optional.isPresent()) {
+                participantStatusMap.remove(lraId);
+                
+                Object result = optional.get();
+                if (shouldInvokeStatusMethod(result)) {
+                    return invokeParticipantMethod(statusMethod, lraId, parentId, STATUS);
+                }
+                
+                return processResult(optional.get(), lraId, statusMethod, STATUS);
```
mmusgrov: You already assigned the value of optional.get() to result on line 80

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRAParticipant.java
mmusgrov: A *participant must define a compensate method* and all of the other methods are optional. Therefore, if this method is present you need to process the java class and if it is not present you should ignore the class (even if it contains any of the optional annotations since mixing and matching with JAX-RS annotated methods will be a recipe for disaster).

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRAParticipantRegistry.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Singleton tracking individual collected LRA participants that
+ * contain one or more non-JAX-RS participant methods in their definitions.
+ */
+public class LRAParticipantRegistry {
+
+    private final Map<String, LRAParticipant> lraParticipants;
+    
+    // required for Weld to be able to create proxy
```
mmusgrov: But does it have to be public?

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRAParticipantRegistry.java
```diff
@@ -0,0 +1,47 @@
...
+    public LRAParticipantRegistry() {
+        lraParticipants = new HashMap<>();
+    }
+
+    public LRAParticipantRegistry(Map<String, LRAParticipant> lraParticipants) {
```
mmusgrov: this can be package private

rts/lra/pom.xml
```diff
@@ -147,7 +147,7 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-checkstyle-plugin</artifactId>
                 <configuration>
-                    <skip>false</skip>
+                    <skip>true</skip>
```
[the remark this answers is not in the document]
xstefank: I forgot to change this back

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRAParticipantRegistry.java
xstefank: addressed in last force push

xstefank:
@mmusgrov @ochaloup this is ready for review again. Can you take a look, please?

rts/lra/lra-test/tck/pom.xml
```diff
@@ -86,11 +81,28 @@
...
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jboss.arquillian.junit</groupId>
+            <artifactId>arquillian-junit-container</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.weld.se</groupId>
+            <artifactId>weld-se-shaded</artifactId>
+            <version>3.0.3.Final</version>
```
ochaloup: this could be good to be extracted to a property as then it could be redefined during the test run with `-Dversion....=...`

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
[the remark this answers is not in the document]
ochaloup: @mmusgrov @xstefank I can see there is a new issue on this in the LRA spec https://github.com/eclipse/microprofile-lra/issues/173 and the https://issues.jboss.org/browse/JBTM-3156 was rejected. I'm not sure I do understand what is the intended behaviour. Is it really in way that the `Compensating` and `Completing` state is really wrong (not `accepted`) here? I probably don't understand the spec at this place and I would need more elaboration on this.

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRACDIExtension.java
```diff
@@ -0,0 +1,235 @@
...
+
+            Class<?> clazz = getClass().getClassLoader().loadClass(annotation.target().asClass().name().toString());
+
+            LRAParticipant participant = processJavaClass(clazz);
+            if (participant != null) {
+                participants.put(participant.getId().getName(), participant);
+                Set<Bean<?>> participantBeans = beanManager.getBeans(clazz, new AnnotationLiteral<Any>() {});
+                if (participantBeans.isEmpty()) {
+                    // resource is not registered as managed bean
+                    event.addBean()
```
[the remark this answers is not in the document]
mmusgrov: But JAX-RS resources do not need to be CDI beans. Are you turning everything into an ApplicaitionScoped bean. Strictly speaking, if you want to go down this route then we would need to introduce a new CDI scope for LRAs (similar to TransactionScoped), so I am still unconvinced about turning every JAX-RS resource into a bean. An alternative would be to maintain an application wide map to hold the participants.

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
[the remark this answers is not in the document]
mmusgrov: @ochaloup Yes, the new section on "Eventual Consistency" introduced the ambiguity.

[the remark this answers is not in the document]
xstefank: @mmusgrov @ochaloup however, can we please merge this PR in its current form and adjust eclipse/microprofile-lra#173 afterwards. It will definitely require TCK changes so additional PR on top of this one will be clear distinction IMO.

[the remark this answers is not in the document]
ochaloup: yes, I'm fine with that approach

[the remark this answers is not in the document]
mmusgrov: This conversation hasn't been resolved in you latest commit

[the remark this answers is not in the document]
xstefank: @mmusgrov this is waiting for your input, can we go with these changes and adjust if necessary after eclipse/microprofile-lra#173 is resolved?

xstefank:
@mmusgrov Updated to manual tracking of non-CDI participants instances, can you please take a look again?

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRACDIExtension.java
```diff
@@ -0,0 +1,249 @@
...
+                    try {
+                        participant.setInstance(clazz.newInstance());
+                    } catch (InstantiationException | IllegalAccessException e) {
+                        LRALogger.i18NLogger.error_cannotProcessParticipant(e);
+                    }
+                }
+            }
+        }
+
+        event.addBean()
```
mmusgrov: We discussed this in the Narayana chat room and you said you were going to do more investigation.

The new non JAX-RS functionality does not require CDI and it is easy to implement the functionality without requiring CDI. Your PR is now restricting where we can use our implementation and there is no benefit to mandating CDI.

xstefank: We will need to bring the dependency on CDI because this class is a CDI extension. Is there any other way how I can catch the initialization of the underlying application in Narayana, please?

mmusgrov: There are at least a couple of JAX-RS specific approaches that I am aware of:

1) implement a javax.servlet.ServletContextListener
2) implement a JAX-RS dynamic feature which will be processed at deployment time.

BTW you never said that you were using CDI as a workaround for hooking into the application deployment lifecycle. So if that is the reason then okay we can go with CDI (and can rework it if there is ever a need to use our LRA implementation in a JAX-RS only environment).

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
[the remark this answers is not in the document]
mmusgrov: There are no MP-LRA tests that would exercise this change of behaviour you have added so I don't accept it should be in this PR. 202 Accepted is the way that participants report in progress compensations (it has been that way since before the sandbox). If the outcome of PR 173 is not to use 202 then yes we would need to modify the implementation but let's not make changes before they are required.

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRACDIExtension.java
```diff
@@ -0,0 +1,249 @@
...
+
+        boolean shouldProcess = participant.getCompensateMethod() != null ||
+            participant.getCompleteMethod() != null ||
+            participant.getStatusMethod() != null ||
+            participant.getForgetMethod() != null;
+
+        return shouldProcess ? participant : null;
+    }
+
+    private boolean isNotLRAParticipant(Class<?> javaClass) {
```
mmusgrov: I found this method difficult to parse. Why not just:

```java
    private boolean isNotLRAParticipant(Class<?> javaClass) {
        for (Method method : javaClass.getMethods())
            if (method.isAnnotationPresent(LRA.class) || method.isAnnotationPresent(Compensate.class))
                return true;
        }
       return false;
   }
```

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRACDIExtension.java
```diff
@@ -0,0 +1,249 @@
...
+     * @return Collected methods wrapped in {@link LRAParticipant} class or null if no non-JAX-RS methods have been found
+     */
+    private LRAParticipant processJavaClass(Class<?> javaClass) {
+        if (isNotLRAParticipant(javaClass)) {
+            return null;
+        }
+
+        LRAParticipant participant = new LRAParticipant(javaClass);
+
+        Arrays.stream(javaClass.getDeclaredMethods()).forEach(m -> processParticipantMethod(m, participant));
```
mmusgrov: Didn't you just stream all the methods in the call to isNotLRAParticipant - maybe you could do both jobs in one pass.

mmusgrov: Actually this code is ran once at deployment time so performance is less off an issue. Feel free to ignore this comment if you don't want to refactor the code to improve its performance.

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRACDIExtension.java
xstefank: @mmusgrov thank you, I would like to keep CDI approach because it's easy to respect user decisions if users define CDI annotations themselves.

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRACDIExtension.java
xstefank: but we need both `@LRA` and `@Compensate` method in class for it to be LRA participant if I'm not mistaken. I'm a little confused by your suggestion. For instance:

```java
@Path("not-lra")
public class NotLRAParticipant {
  @GET
  @Path("status")
  @Status
  public void status() {
    // ...
  }
}
```

The `isNotLRAParticipant(NotLRAParticipant.class)` returns false which indicates that NotLRAParticipant is a participant which not true.

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRACDIExtension.java
xstefank: Thanks, I agree. It will be ran only once during deployment and the benefit is that we won't process (verify signatures and save references) for classes that are not participants (the first pass through declared methods).

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRACDIExtension.java
mmusgrov: Ah you are right, both annotations need to be present (not either). 

Why not use the jandex index and the ClassInfo objects you generated in the AfterBeanDiscovery observer since that contains all the data you require and has query methods. I hope you didn't mind but I took a look at get the same info from the index:

```java
public class CDIExtension {
    private Index index;
    ...
    void processDeployment() {
        for (ClassInfo ci : index.getKnownClasses()) {
            LRAParticipant p = getAsParticipant(ci);

            if (p != null && p.shouldProcess()) {
                // update participants and add as a CDI bean
                System.out.println(p);
            }
        }
    }

    private LRAParticipant getAsParticipant(ClassInfo ci) {
        // isNotLRAParticipant
        if (!hasAnnotations(ci, Compensate.class.getName(), LRA.class.getName()))
            return null;

        // isJaxRsMethod
        if (hasAnnotations(ci,
                GET.class.getName(),
                POST.class.getName(),
                PUT.class.getName(),
                DELETE.class.getName(),
                HEAD.class.getName(),
                OPTIONS.class.getName())) {
            return null;
        }

        LRAParticipant participant = new LRAParticipant(ci.getClass());

        // setParticipantAnnotation
        participant.setCompensateMethod(ci.annotations().get(
                  DotName.createSimple(Compensate.class.getName())));
        participant.setCompleteMethod(ci.annotations().get(
                  DotName.createSimple(Complete.class.getName())));
        participant.setStatusMethod(ci.annotations().get(
                  DotName.createSimple(Status.class.getName())));
        participant.setForgetMethod(ci.annotations().get(
                  DotName.createSimple(Forget.class.getName())));

        return participant;
    }

    private boolean hasAnnotations(ClassInfo ci, String... annotations) {
        for (String annotation : annotations) {
            if (!hasAnnotation(ci, DotName.createSimple(annotation)))
                return false;
        }

        return true;
    }

    private boolean hasAnnotation(ClassInfo ci, DotName name) {
        return ci.annotations().containsKey(name);
    }
}
```

mmusgrov: And instead of having participant setter methods just do

>         LRAParticipant participant = new LRAParticipant(ci);

and set the methods internally in the constructor.

mmusgrov: My suggested getAsParticipant() method is checking for JAX-RS annotations on the whole class instead of on each method. To correct that and to check each method do something like:

```java
    private LRAParticipant getAsParticipant(ClassInfo ci) throws ClassNotFoundException {
        // isNotLRAParticipant
        if (!hasAnnotations(ci, Compensate.class.getName(), LRA.class.getName()))
            return null;

        LRAParticipant participant = new LRAParticipant(ci);
        for (MethodInfo mi : ci.methods()) {
            // isJaxRsMethod
            if (!hasAnyAnnotation(mi,
                    GET.class.getName(),
                    POST.class.getName(),
                    PUT.class.getName(),
                    DELETE.class.getName(),
                    HEAD.class.getName(),
                    OPTIONS.class.getName())) {
                // process the class if has any participant annotation:
                participant.process(mi);
            }
        }

        return participant;
    }
```
You can still use your existing method validation code by doing:

```java
        id = Class.forName(ci.name().toString());
```
to get the java class instead of the jandex ClassInfo. Alternatively the jandex type MethodInfo should give you sufficient information to do the validation (but I did not try it).

xstefank: I will try to rewrite it to use only Jandex, but as you mentioned, we will still need to cycle through declared methods I am afraid. Also, we will need to keep the Method reference, which I guess must by somehow obtainable from MethodInfo. Well, I will try and let you know if it's possible to rewrite to Jandex. But the current solution can be reused even in case there is no Jandex which may come in handy when reintroduce programmatic registration again. Otherwise, we will have to create the Jandex index dynamically, which I am not sure may or may not be a problem.

xstefank: @mmusgrov I finally got to try to rewrite annotation scanning to Jandex and I run into a fundamental problem that there is no way how to obtain reflection information from Jandex or the other way around. I actually spoke even to Jandex creator and he confirmed that Jandex is designed to run before the reflection kicks in. I was actually able to get it running but its rather complex and I must somehow "guess" the methods in participant class because there is no way how to obtain Method object from MethodInfo (switch on the number of arguments as now all parameters must be Strings so method is compensate(), compensate(String) or compensate(String, String) -- which is not really maintainable in a long term). Another reason is that we would need to dynamically create Jandex Type objects for every parameter and return type in every method instead of simple equals() on class objects. However, it does help the check whether a class is a participant (handy annotations() method you've suggested) but I don't think it's worth creating index just for this. I can update the PR, that if the index present we can optimize this part if you want. Or I can finish up the rewrite to Jandex if you would like to compare.

After all, I think the solution like it stands now is the safest bet.

mmusgrov: > After all, I think the solution like it stands now is the safest bet.

The jandex code that I proposed does most of what you need. The final step of getting the method info to do the invocations can be done using your existing code as I suggested in my comment https://github.com/jbosstm/narayana/pull/1450#discussion_r301944907

mmusgrov: My motivation for this conversation was the initial review comment

> I found this method difficult to parse ...

and when I noticed that you had already created a Jandex index I gave an example of how to simplify the method isNotLRAParticipant() and the other methods in this class using the contents of the index. This had the benefit of:

- making the code more readable
- more efficient since you already had the index

In the comment #1450 of this conversation I said "You can still use your existing method validation code (that's in LRAParticipant) by doing:  id = Class.forName(ci.name().toString()); to get the java class instead of the jandex ClassInfo.

To summarise, since you have already gone to the trouble of creating an index I would use that for all annotation processing. And for refection (since as you state is not available from the index) I would use Class.forName(ci.name().toString()) and the code you wrote for LRAParticipant.

mmusgrov: Note that the code I included in my previous comment is using the Jandex index for the annotation processing so we know it is feasible.

xstefank:
@mmusgrov I moved the verification to the LRAParticipant and the implementation is now using Jandex ClassInfo to determine whether the class is a participant. From your comments, I am missing only the isJaxRsMethod where because of the problem with the mapping between reflection Method and Jandex MethodInfo I cannot see the right way now with a nice solution. If you think that some more refactoring can be appropriate, can this PR be merged (after it passes TCK, the spec PR is open) and then you can refactor in a separate PR later?

xstefank:
@mmusgrov rebased and I removed changes in `LRARecord`. The client now always sends 202 for Compensating/Completing states.

mmusgrov:
> @mmusgrov I moved the verification to the LRAParticipant and the implementation is now using Jandex ClassInfo to determine whether the class is a participant. From your comments, I am missing only the isJaxRsMethod where because of the problem with the mapping between reflection Method and Jandex MethodInfo I cannot see the right way now with a nice solution. If you think that some more refactoring can be appropriate, can this PR be merged (after it passes TCK, the spec PR is open) and then you can refactor in a separate PR later?

When I suggested that you make use of the Jandex index it was because the `isNotLRAParticipant()` method was overly complex (I notice that this method has still not changed). I presented some code that simplified the parsing in this class as a whole by using the index that you had already created  but your latest commit ignores my suggestion (which is okay because you are not compelled to follow the reviewers' advice). 

I think this review could be ongoing so I am going to abstain from approving it. But since Ondra has already approved the PR, and we only need one approval, I will merge with only his approval.

xstefank:
@mmusgrov I may be missing something, but I did change this method according to your suggestion - https://github.com/jbosstm/narayana/pull/1450/files#diff-85202fddd1a60d6168e3f1e82a462622R112

PR #1452, opened 2019-05-28
[JBTM-3148] returning correct error code on one phase commit failure

ochaloup:
https://issues.jboss.org/browse/JBTM-3148

1PC for subordinate XAResource to return heuristics if failure on commit happens.

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !RTS  !TOMCAT !JACOCO !LRA !PERF !XTS !AS_TESTS
MAIN

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/subordinate/SubordinateAtomicAction.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/subordinate/SubordinateAtomicAction.java
@@ -242,6 +242,22 @@ public class SubordinateAtomicAction extends
 	    if (status == ActionStatus.ABORT_ONLY || doBeforeCompletion())
 	    {
 	        status = super.End(true);
+
+	        switch (super.getHeuristicDecision())
+	        {
+	        // commit was processed correctly but some of the resources failed to finish
+	        // for 2PC this means no trouble as in case of a crash at this time
+	        // there is recovery which works with prepared transaction
+	        // for 1PC where there is nothing prepared and nothing stored at object store
+	        // we need to inform parent about error by setting up heuristic state
+	        case TwoPhaseOutcome.PREPARE_OK:
+	        case TwoPhaseOutcome.FINISH_OK:
+	            if (super.failedList != null && super.failedList.size() > 0) {
+	                status = ActionStatus.H_COMMIT;
+	            }
+	        default:
+	            break;
+	        }
 	    }
 	    else
 	    {
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/jtaI18NLogger.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/jtaI18NLogger.java
@@ -580,12 +580,16 @@ public interface jtaI18NLogger {
     @Message(id = 16142, value = "Not actived transaction ''{0}'' for xid: {1}", format = MESSAGE_FORMAT)
     String get_not_activated_transaction(Transaction txn, Xid xid);
 
-    @Message(id = 16143, value = "Problem during waiting for lock '{0}' whilst in state {1}", format = MESSAGE_FORMAT)
+    @Message(id = 16143, value = "Problem during waiting for lock ''{0}'' whilst in state {1}", format = MESSAGE_FORMAT)
     @LogMessage(level = WARN)
     public void warn_intteruptedExceptionOnWaitingXARecoveryModuleLock(XARecoveryModule recModule, String state, @Cause() InterruptedException arg0);
 
     @Message(id = 16144, value = "No subordinate transaction to drive {0}, xid: {1}", format = MESSAGE_FORMAT)
     String get_no_subordinate_txn_for(String actionToDrive, Xid xid);
+
+    @Message(id = 16145, value = "One phase commit for transaction ''{0}'' does not store data in the object store. "
+            + "Recovery is won''t able to decide about outcome. Transaction is marked as heuristic to be decided by administrator.", format = MESSAGE_FORMAT)
+    String get_onephase_heuristic_commit_failure(StateManager action);
 
     /*
         Allocate new messages directly above this notice.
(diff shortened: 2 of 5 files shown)


ochaloup:
@mmusgrov  @tomjenkinson this is my proposal for JBTM-3148. The issue which came from discussion at forum about 1PC XTS (https://developer.jboss.org/thread/279243). The idea of the solution is to report heuristics back to the parent transaction (parent JBoss EAP) as it was implemented by https://issues.jboss.org/browse/JBTM-2443

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/subordinate/SubordinateAtomicAction.java
```diff
@@ -242,6 +242,22 @@ public int doOnePhaseCommit ()
...
+	        {
+	        // commit was processed correctly but some of the resources failed to finish
+	        // for 2PC this means no trouble as in case of a crash at this time
+	        // there is recovery which works with prepared transaction
+	        // for 1PC where there is nothing prepared and nothing stored at object store
+	        // we need to inform parent about error by setting up heuristic state
+	        case TwoPhaseOutcome.PREPARE_OK:
+	        case TwoPhaseOutcome.FINISH_OK:
+	            if (super.failedList != null && super.failedList.size() > 0) {
+	                return ActionStatus.H_COMMIT;
```
mmusgrov: Is it okay to skip the afterCompletion(H_COMMIT) call that was previously performed at line 252

ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/jtaI18NLogger.java
```diff
@@ -580,10 +580,14 @@
     @Message(id = 16142, value = "Not actived transaction ''{0}'' for xid: {1}", format = MESSAGE_FORMAT)
     String get_not_activated_transaction(Transaction txn, Xid xid);
 
-    @Message(id = 16143, value = "Problem during waiting for lock '{0}' whilst in state {1}", format = MESSAGE_FORMAT)
+    @Message(id = 16143, value = "Problem during waiting for lock ''{0}'' whilst in state {1}", format = MESSAGE_FORMAT)
```
mmusgrov: Won't this change affect any existing users that are potentially parsing this output? If so you should deprecate first.

ochaloup: This fixes a formatting bug.
The use of `'{0}'` causes that data are not pasted there. Aka the call to method `public void warn_intteruptedExceptionOnWaitingXARecoveryModuleLock(XARecoveryModule recModule, String state, ...` expects to inform user about what recovery module is in use. Now the recovery module is **not** printed. The use of `'{0}'` causes the data is not placed to the output. The change to `''{0}''` makes the information to be printed.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/subordinate/SubordinateAtomicAction.java
ochaloup: I took the same approach as it's in the `doCommit`: https://github.com/jbosstm/narayana/blob/5.9.5.Final/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/subordinate/SubordinateAtomicAction.java#L170

Plus my understanding is - when the transaction ends with the heuristic outcome we can't say it was completed. Aka [the doc says](https://docs.oracle.com/javaee/7/api/javax/transaction/Synchronization.html#afterCompletion-int-) that the callback is called after transaction is rolled-back or committed. Here we are in *unspecified* state.

ochaloup:
@mmusgrov Thank you for pointing this out. I was wondering about the use of the `TransactionImple.removeTransaction(this);` and I understand it as the transaction is removed from the active transaction to be tracked. It does not mean the participant record is deleted from object store. 

The call is for me: this transaction is not active anymore but reporting back to the caller an error with throwing an exception.
What I was testing the participant record is left in the object store with this code (when the heuristic excepetion is thrown)
Let me know how this could be done better or what are the bad expectations from this perspective. Thanks a lot!

ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/jtaI18NLogger.java
mmusgrov: OK

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/subordinate/SubordinateAtomicAction.java
mmusgrov: The approach taken in `doCommit` for the heuristic outcome is:
https://github.com/jbosstm/narayana/blob/5.9.5.Final/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/subordinate/SubordinateAtomicAction.java#L179
which does call afterCompletion.

mmusgrov:
> @mmusgrov Thank you for pointing this out. I was wondering about the use of the `TransactionImple.removeTransaction(this);` and I understand it as the transaction is removed from the active transaction to be tracked. It does not mean the participant record is deleted from object store.
> 
> The call is for me: this transaction is not active anymore but reporting back to the caller an error with throwing an exception.
> What I was testing the participant record is left in the object store with this code (when the heuristic excepetion is thrown)
> Let me know how this could be done better or what are the bad expectations from this perspective. Thanks a lot!

Okay

ochaloup:
@mmusgrov I found I should add add a specific test for this to testsuite so I'm in middle of work now.

ochaloup:
I worked for tuning end2end test in the QA crashrec testsuite and I added a testcase to the narayana testsuite.

ochaloup:
Pretty strange. There is a failure in the test I added for testing the code changes in this PR. It passes smoothly at my laptop. Investigating...

PR #1453, opened 2019-05-31
JBTM-3150 Upgrade maven version to 3.6.1

mmusgrov:
https://issues.jboss.org/browse/JBTM-3150

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB BLACKTIE !XTS !PERF !AS_TESTS !TOMCAT !JACOCO !LRA !DB_TESTS !JDK11 !WIN !BLACKTIE
RTS LRA


mmusgrov:
The build failures only occur with JDK 11 and are caused by https://issues.jboss.org/browse/JBTM-3151

rts/lra/pom.xml
```diff
@@ -142,21 +136,6 @@
                     <skip>true</skip>
                 </configuration>
             </plugin>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
```
ochaloup: the checkstyle should not be removed from the project. It could be used the one configured as default for the Narayana project (JBTM-989). The changes which are needed are to to define plugin here in `pom.xml` to be

```xml
 <plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-checkstyle-plugin</artifactId>
  <configuration>
    <skip>false</skip>
  </configuration>
</plugin>
```
and change the version of the wfly checkstyle plugin to version `1.0.8.Final` at the root `pom.xml`
https://github.com/jbosstm/narayana/blob/5.9.5.Final/pom.xml#L491

rts/pom.xml
```diff
@@ -31,9 +31,9 @@ Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
   <url>http://www.jboss.org/jbosstm</url>
  
   <properties>
-    <version.wildfly-swarm>2017.2.0</version.wildfly-swarm>
+    <version.wildfly-swarm>2.3.0.Final</version.wildfly-swarm>
```
ochaloup: possibly renamed for `thorntail`?

mmusgrov: They're just names and besides the name change could brake any existing scripts/notes that use the property. Why do you feel it's important?

rts/lra/pom.xml
mmusgrov: That what I wanted to happen so good spot, thanks

rts/lra/lra-proxy/test/pom.xml
```diff
@@ -26,8 +23,8 @@
         <lra.http.port>8080</lra.http.port>
         <service.http.port>${swarm.http.port}</service.http.port>
 
-        <swarm.debug.params></swarm.debug.params>
-        <swarm.logging.params></swarm.logging.params>
+        <swarm.debug.params/>
```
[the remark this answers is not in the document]
mmusgrov: They're just names and besides the name change could brake any existing scripts/notes that use the property. Why do you feel it's important?

rts/pom.xml
tomjenkinson: good point, we shouldn;t just drop these externally facing property values as they could be in other scripts (e.g. user ones). I think it would be a good follow up task though

rts/lra/lra-proxy/test/pom.xml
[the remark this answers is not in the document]
ochaloup: Because we should aim to provide clear view for any external reader or contributor (and for us in the future) what project do we used. It's a good practice to name with descriptive names. Using descriptive name makes the code/script/configuration file being easier to maintain and less error-prone.
In summary it's not a good practice to refer to an old, deprecated project which is not used in the code base.
It's the reason for doing refactoring. Makes things clearer. We can break some script/notes but as the point of the whole of this jira is movement to thorntail it's logic the swarm is not what is about to be used.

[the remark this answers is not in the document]
tomjenkinson: I agree with both opinions. I do think that we should look at these properties somewhat like an API though as people may be overriding some of these properties in their own scripts. That said I am now of the opinion that the version number property can safely be changed from version.wildfly-swarm to version.thorntail because the rest of the GAV has changed. If the swarm.debug.params are equivalent between the two then it seems more like we want to treat with deprecation approach, if they have incompatible APIs for the input then I agree that during the change of the dependency it should be updated (either now or as a follow up action)

rts/pom.xml
ochaloup: I think we should rename it as we are doing refactoring for the swarm to thorntail and we should not contribute to the technical debt for the project and for the bad understandability.

tomjenkinson: I agree, for this one it seems definitely viable to change to thorntail as the whole GAV has changed so it is very unlikely that a -Dversion.wildfly-swarm=2017.2.0 or similar would work with the new artifact we are using

rts/lra/lra-proxy/test/pom.xml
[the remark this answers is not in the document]
mmusgrov: My latest push replaces all occurrences of swarm with thorntail with the justification that the source is implementing a spec that is still at version 1.0-SNAPSHOT and therefore liable to change in arbitrary ways at any time.

mmusgrov:
> * two points more. there was removed the `arquillian-swarm.xml`. it could not be needed (but e.g. it defines a way to run the test with more verbose logging) but if stated to not be needed it should not be mentioned as `arquillian-thorntail.xml` in `pom.xml`

Not sure what this means but all I did was replace  `arquillian-swarm.xml` with  `arquillian-thorntail.xml`

mmusgrov:
@ochaloup So I think I have addressed your concerns may I go ahead and merge?

mmusgrov:
@ochaloup There is no file called rts/lra/lra-test/tck/src/test/resources/arquillian-swarm.xml in this PR

PR #1456, opened 2019-06-05
JBTM-3087 Upgrade the version of the MP-LRA dependency

mmusgrov:

This is an interim upgrade of our LRA dependencies while we wait for pr 1450 (which is in progress). We need it as soon as possible since all of our builds are currently failing.

Note that we need an immediate fix because MP-LRA builds are expunged after 30 days

### DIFF
--- a/rts/lra/pom.xml
+++ b/rts/lra/pom.xml
@@ -28,8 +28,8 @@
         <version.thorntail>2.4.0.Final</version.thorntail>
         <version.arquillian>1.2.1.Final</version.arquillian> <!-- cannot use the up-to-date Arquillian https://issues.jboss.org/browse/THORN-2090 -->
 
-        <version.microprofile.lra.api>1.0-20190430.132220-362</version.microprofile.lra.api>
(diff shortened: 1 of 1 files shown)


xstefank:
I understand that current LRA snapshot is already deleted so Narayana build won't pass, but this change introduces significant conflicts with #1450. Would it be possible to update to versions
```
<version.microprofile.lra.api>1.0-20190515.061205-378</version.microprofile.lra.<version.microprofile.lra.tck>1.0-20190515.061206-378</version.microprofile.lra.tck>
```
This would still give us 9 days to merge #1450 but I acknowledge that it is my delay on 1450 that causes the problem so if you prefer go ahead and merge this and I will resolve conflicts in 1450.

mmusgrov:
The changeset is the minimal set of changes that allow the module to compile and pass the TCK. The extra changes are because of the removal of the LRAManagement which happened 21 days ago so "only updating the artifact api" version is insufficient.

mmusgrov:
@xstefank I updated the PR to the version you suggested

xstefank:
@mmusgrov thank you! Much appreciated.

PR #1461, opened 2019-06-20
JBTM-3145 (MP-LRA issue146) allow LRA compensators to be called with …

mmusgrov:
…any HTTP method

https://issues.jboss.org/browse/JBTM-3145

MP-LRA issue 146 allows compensator callback endpoints to be invoked with any HTTP method (DELETE, PUT, POST or GET)

### DIFF
--- a/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
+++ b/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
@@ -29,12 +29,17 @@ import com.arjuna.ats.arjuna.state.InputObjectState;
 import com.arjuna.ats.arjuna.state.OutputObjectState;
 
 import io.narayana.lra.Current;
+import io.narayana.lra.LRAConstants;
 import io.narayana.lra.coordinator.domain.service.LRAService;
 import io.narayana.lra.logging.LRALogger;
 import org.eclipse.microprofile.lra.annotation.LRAStatus;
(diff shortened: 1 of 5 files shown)


rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -555,6 +566,50 @@ private int retryGetEndStatus(URI endPath, boolean compensate) {
...
+                if (pair.contains("=")) {
+                    String[] qp = pair.split("=");
+
+                    if (qp[0].equals(LRAConstants.HTTP_METHOD_NAME)) {
+                        switch (qp[1]) {
+                            case "javax.ws.rs.GET":
+                                return asyncInvoker.get();
+                            case "javax.ws.rs.PUT":
+                                return asyncInvoker.put(Entity.text(null));
+                              //  return asyncInvoker.put(Entity.entity(compensatorData, mediaType));
```
xstefank: why we cannot already use compensatorData here? It's being used in default case (switch (method) {...}) later in this method

mmusgrov: Because it isn't supported in MP-LR milestone 1. We will re-enable it in the next one

xstefank: I though you are on PTO :) then go ahead and merge this please, it's passing TCK!! Great job!

PR #1462, opened 2019-06-28
JBTM-3157 LRA participant does not respect JAX-RS path definitions

xstefank:
https://issues.jboss.org/browse/JBTM-3157


ochaloup:
I just checked the faulures and the failures in `PERF` tests are not related to this PR. I think those are intermittent failures.
The failed test is `com.hp.mwtests.ts.arjuna.atomicaction.CheckedActionTest` which works with `AtomicAction` which are run in multiple threads. This PR does not touch this functionality.
Up to that there is improvement for 15% in test `com.hp.mwtests.ts.arjuna.performance.Performance1.onePhase` which is suspicious in the same way, indicating the tests are unreliable.

PR #1464, opened 2019-07-15
JBTM-3087 Add support for LRA issues 164 and 173

mmusgrov:

This PR adds support for the following MP-LRA issues:
- issue164 Support notifications that an LRA has finished (@AfterLRA)
- issue173 Clarify how to report in progress compensations (in particular use 500 to report FailedToCompensate/FailedToComplete)

### DIFF
--- a/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
+++ b/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
@@ -372,27 +372,14 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
             }
         }
 
-        if (responseData != null) {
-            String failureReason = null;
-
-            if (httpStatus == Response.Status.INTERNAL_SERVER_ERROR.getStatusCode()) {
-                // the body should contain a valid ParticipantStatus
-                try {
-                    failureReason = ParticipantStatus.valueOf(responseData).name();
-                } catch (IllegalArgumentException ignore) {
-                    // ignore the body and let recovery discover the status of the participant
-                }
-            } else if (httpStatus == Response.Status.OK.getStatusCode()) {
-                // see if any content contains a failed status
-                if (compensate && LRAStatus.FailedToCancel.name().equals(responseData)) {
-                    failureReason = responseData;
-                } else if (!compensate && LRAStatus.FailedToClose.name().equals(responseData)) {
-                    failureReason = responseData;
-                } // else recovery will discover the status
-            }
-
-            if (failureReason != null) {
-                return reportFailure(compensate, endPath.toASCIIString(), failureReason);
+        if (responseData != null &&
+                httpStatus == Response.Status.INTERNAL_SERVER_ERROR.getStatusCode()) {
+            // the body should contain a valid ParticipantStatus
+            try {
+                return reportFailure(compensate, endPath.toASCIIString(),
+                        ParticipantStatus.valueOf(responseData).name());
+            } catch (IllegalArgumentException ignore) {
+                // ignore the body and let recovery discover the status of the participant
             }
         }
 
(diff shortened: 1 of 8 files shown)


rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -364,12 +372,27 @@ private int tryDoEnd(boolean compensate) {
...
+            if (httpStatus == Response.Status.INTERNAL_SERVER_ERROR.getStatusCode()) {
+                // the body should contain a valid ParticipantStatus
+                try {
+                    failureReason = ParticipantStatus.valueOf(responseData).name();
+                } catch (IllegalArgumentException ignore) {
+                    // ignore the body and let recovery discover the status of the participant
+                }
+            } else if (httpStatus == Response.Status.OK.getStatusCode()) {
+                // see if any content contains a failed status
+                if (compensate && LRAStatus.FailedToCancel.name().equals(responseData)) {
```
xstefank: shouldn't this be ParticipantStatus.FailedToCompensate? Similarly for FailedToComplete.

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
```diff
@@ -149,7 +149,9 @@ public void finished(Transaction transaction, boolean fromHierarchy) {
             // the LRA is top level or it's a nested LRA that was closed by a
             // parent LRA (ie when fromHierarchy is true) then it's okay to forget about the LRA
 
-            remove(ActionStatus.stringForm(transaction.status()), transaction.getId());
+            if (transaction.afterLRANotification()) {
+                remove(ActionStatus.stringForm(transaction.status()), transaction.getId());
+            }
```
xstefank: sorry, but will this be called repeatedly? I don't see it from a quick look in the code. In other case, the specification says we should log a warning when we stop retrying - https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/AfterLRA.java#L62

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
tomjenkinson: Ah, I think that might well be right actually. @mmusgrov, please can you comment? I will move it to request changes to block the merge until we get an answer

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
tomjenkinson: I think as a SHOULD we don't actually have to do it, but I agree it would be useful to add

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
mmusgrov: > LGTM - but I think we need a test beyond the TCK [one](https://github.com/eclipse/microprofile-lra/pull/174/files#diff-3eeea840a3958c8e4d50c676076e87a5R259). For example crash recovery one? It would be OK to add the Jira to do this work after IMO

I created a TCK issue for it https://github.com/eclipse/microprofile-lra/issues/190

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
mmusgrov: I created issue https://github.com/eclipse/microprofile-lra/issues/190

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
mmusgrov: The code that used the wrong status is now dead code since we no longer use the body in 200 responses. I have just removed it.

PR #1468, opened 2019-07-28
[JBTM-3163] AfterLRA notifications are delivered twice

xstefank:
https://issues.jboss.org/browse/JBTM-3163

No need to test. There is a TCK test in non-JAX-RS cases which is asserting that the afterLRA is called only once.

### DIFF
--- a/rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
+++ b/rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
@@ -33,6 +33,7 @@ import static io.narayana.lra.LRAConstants.STATUS;
 import static io.narayana.lra.LRAConstants.TIMELIMIT_PARAM_NAME;
 import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
 import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
+import static javax.ws.rs.core.Response.Status.NOT_FOUND;
 import static javax.ws.rs.core.Response.Status.SERVICE_UNAVAILABLE;
 import static javax.ws.rs.core.Response.Status.NOT_ACCEPTABLE;
 
@@ -677,7 +678,7 @@ public class NarayanaLRAClient implements Closeable {
 
             if (response.getStatus() == Response.Status.PRECONDITION_FAILED.getStatusCode()) {
                 LRALogger.i18NLogger.error_tooLateToJoin(lraId, response);
-                throw new WebApplicationException(lraId + ": Too late to join with this LRA ", BAD_REQUEST);
+                throw new WebApplicationException(lraId + ": Too late to join with this LRA ", NOT_FOUND);
             } else if (response.getStatus() == Response.Status.NOT_FOUND.getStatusCode()) {
                 LRALogger.logger.infof("Failed enlisting to LRA '%s', coordinator '%s' responded with status '%s'",
                         lraId, base, Response.Status.NOT_FOUND.getStatusCode());
--- a/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
+++ b/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
@@ -247,8 +247,6 @@ public class LRAService {
             }
         }
 
-        finished(transaction, fromHierarchy);
-
         if (transaction.isTopLevel()) {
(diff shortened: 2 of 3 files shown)


xstefank:
@mmusgrov TckTests fails on different errors. Should I enable them then? Or do I need to implement changes required for them pass even if they are not required for this PR?

mmusgrov:
They passed on our last CI run (Thursday): http://narayanaci1.eng.hst.ams2.redhat.com/job/narayana/PROFILE=LRA,jdk=jdk8.latest,label=linux/

I will push an update today that builds against the latest MP-LRA artifacts (if that is the issue).

mmusgrov:
Also JBTM-3163 does not provide enough information.
Also there is nothing in the spec that says the notification cannot be sent more than once.
The PR or JIRA does not give any hint as to how the duplicates arise or how the PR stops duplicates being sent.

xstefank:
I will update the JIRA description if it's not enough

xstefank:
@mmusgrov I've updated the JIRA with my findings

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
```diff
@@ -677,7 +678,7 @@ private URI enlistCompensator(URI uri, long timelimit, String linkHeader, String
 
             if (response.getStatus() == Response.Status.PRECONDITION_FAILED.getStatusCode()) {
                 LRALogger.i18NLogger.error_tooLateToJoin(lraId, response);
-                throw new WebApplicationException(lraId + ": Too late to join with this LRA ", BAD_REQUEST);
+                throw new WebApplicationException(lraId + ": Too late to join with this LRA ", NOT_FOUND);
```
mmusgrov: What is the motivation of this change?

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
```diff
@@ -247,8 +247,6 @@ public LRAStatusHolder endLRA(URI lraId, boolean compensate, boolean fromHierarc
             }
         }
 
-        finished(transaction, fromHierarchy);
```
mmusgrov: Will this still work when the periodic recovery thread is running recovery

rts/lra/lra-test/tck/pom.xml
```diff
@@ -43,6 +43,8 @@
                         <exclude>**/TckInvalidParticipantSignaturesTests.java</exclude>
                         <exclude>**/TckParticipantTests.java</exclude>
                         <exclude>**/TckInvalidSignaturesTests.java</exclude>
+                        <!-- https://github.com/eclipse/microprofile-lra/pull/192 -->
+                        <exclude>**/TckTests.java</exclude>
```
mmusgrov: We need to run the main test suite to validate that our implementation is working (we need to fix the TCK if it is wrong, not disable it). Is the issue that we need to update which TCK build artefact we are pulling in?

mmusgrov:
> They passed on our last CI run (Thursday): http://narayanaci1.eng.hst.ams2.redhat.com/job/narayana/PROFILE=LRA,jdk=jdk8.latest,label=linux/
> 
> I will push an update today that builds against the latest MP-LRA artifacts (if that is the issue).

The tests still pass so this is not the issue. I will still take a look at updating the mp-lra snapshot dependency though, hopefully this week.

xstefank:
I am already working on it to prove this is an issue.

mmusgrov:
But do you agree we need to [change the test to assert that the notification](https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java#L259) was called at least once.

xstefank:
Personally, I don't see any issue with delivering AfterLRA notification several times (but note that this is a different issue). Should I update that in the spec?

mmusgrov:
The spec does not need to say how many times it should deliver the notification. We already say in the spec that messages can be lost.

But we do need a PR to change the test to assert that the notification is called at least once.

xstefank:
> But we do need a PR to change the test to assert that the notification is delivered at least once.

yes, this is what I meant. We can change this in the specification. I will create a PR later today.

xstefank:
Closing this as out of date. I will open a new PR when ready.

mmusgrov:
@xstefank  I can submit a small fix if you haven't started yet

xstefank:
@mmusgrov no I haven't. Go ahead please.

PR #1472, opened 2019-07-31
JBTM-3166 TransactionListener's should not be enabled by default

stuartwdouglas:
They are no longer used in WildFly and have a performance cost

### DIFF
--- a/ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/BaseTransactionManagerDelegate.java
+++ b/ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/BaseTransactionManagerDelegate.java
@@ -40,6 +40,7 @@ import org.jboss.tm.listener.TransactionListener;
 import org.jboss.tm.listener.TransactionListenerRegistry;
 import org.jboss.tm.listener.EventType;
 
+import java.util.Collections;
 import java.util.EnumSet;
 import java.util.Map;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/BaseTransactionManagerDelegate.java
```diff
@@ -198,6 +205,12 @@ public Transaction suspend()
     // return all the event listeners associated with this thread
     private Collection<TransactionListener> getListeners(Transaction transaction, boolean create)
     {
+        if(!LISTENERS_ENABLED) {
+            if(create) {
+                throw new IllegalStateException("Transaction listeners are disabled and should not be used. If you need them they can be enabled via -D" + LISTENERS_ENABLED_PROP + "=true");
```
ochaloup: @mmusgrov I assume the exception message should be internationalized. Will you create a PR for that or should do I?

ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/BaseTransactionManagerDelegate.java
```diff
@@ -55,6 +56,12 @@
 {
     private static final String LISTENER_MAP_KEY = "__TX_LISTENERS";
 
+    protected static final String LISTENERS_ENABLED_PROP = "com.arjuna.ats.jbossatx.BaseTransactionManagerDelegate.LISTENERS_ENABLED";
```
ochaloup: @mmusgrov should not be the configuration props gathered under the `Bean`s? Should not be transferred under like `JTSEnvironmentBean` or similar?

PR #1474, opened 2019-08-01
[JBTM-3167][JBTM-3166] fixes for test failures brought by JBTM-3166

ochaloup:
https://issues.jboss.org/browse/JBTM-3167

https://issues.jboss.org/browse/JBTM-3166
original PR for JBTM-3166: https://github.com/jbosstm/narayana/pull/1472

@mmusgrov @tomjenkinson 
I'm fixing the test failures and I fixed the issues that I consider the original PR has. Please review and comment with remarks.

MAIN AS_TESTS
!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS PERF NO_WIN !RTS !TOMCAT !JACOCO !LRA

### DIFF
--- a/ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/BaseTransactionManagerDelegate.java
+++ b/ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/BaseTransactionManagerDelegate.java
@@ -31,6 +31,11 @@ import javax.transaction.TransactionManager;
 import javax.transaction.Status;
 
 import org.jboss.tm.listener.TransactionTypeNotSupported;
+
+import com.arjuna.ats.jbossatx.logging.jbossatxLogger;
+import com.arjuna.ats.jta.common.JTAEnvironmentBean;
+import com.arjuna.ats.jta.common.jtaPropertyManager;
+
 import org.jboss.tm.TransactionLocal;
 import org.jboss.tm.TransactionLocalDelegate;
 import org.jboss.tm.TransactionTimeoutConfiguration;
@@ -40,7 +45,6 @@ import org.jboss.tm.listener.TransactionListener;
 import org.jboss.tm.listener.TransactionListenerRegistry;
 import org.jboss.tm.listener.EventType;
 
-import java.util.Collections;
 import java.util.EnumSet;
 import java.util.Map;
 import java.util.HashMap;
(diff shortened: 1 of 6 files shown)


ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/BaseTransactionManagerDelegate.java
```diff
@@ -205,9 +203,10 @@ public Transaction suspend()
     // return all the event listeners associated with this thread
     private Collection<TransactionListener> getListeners(Transaction transaction, boolean create)
     {
-        if(!LISTENERS_ENABLED) {
+        if(!jtaPropertyManager.getJTAEnvironmentBean().isTransactionToThreadListenersEnabled()) {
             if(create) {
-                throw new IllegalStateException("Transaction listeners are disabled and should not be used. If you need them they can be enabled via -D" + LISTENERS_ENABLED_PROP + "=true");
+                throw new IllegalStateException(jbossatxLogger.i18NLogger.get_transaction_listeners_disabled(
```
tomjenkinson: Can this be parameterized rather than use string concatenation?

ochaloup: I understand your point but from my perspective is clearer to parametrize the message with name of property than to create API which requires some class and then a name of a property which will be later "somehow" prefixed with the class name. What if the configuration will be done differently than with the `*EnvironmentBean`s? If you really prefer doing something like 
```
@Message(id=32037,...)
public String get_transaction_listeners_disabled(Class<?> beanClass, String transactionListenersPropertyName)
```
I can change its, for sure.

ochaloup:
@tomjenkinson I assume that the changes like this have a good performance benefit but I have no deep insight to the `PERF` tests.

ArjunaJTS/integration/src/main/java/com/arjuna/ats/jbossatx/BaseTransactionManagerDelegate.java
tomjenkinson: I guess as this is an exceptional circumstance it should be OK. If it was logging only we would have to avoid concatenation. My concern is purely from a performance point of view but given this is an exception it would not be normally expected

ochaloup:
@tomjenkinson I would merge this, just found one doubt here. @stuartwdouglas used the the code for using property named `com.arjuna.ats.jbossatx.BaseTransactionManagerDelegate.LISTENERS_ENABLED` this PR changes it for property named `com.arjuna.ats.jta.common.JTAEnvironmentBean.transactionToThreadListenersEnabled`. Is it something that should be announced somehow?

ochaloup:
Ok, I see. And the default behaviour is to deactivate the functionality either way. (This is not a backport, it's just adding a new configuration property which has never been available before.)

PR #1478, opened 2019-08-07
BUILD SCRIPT ONLY: narayana-codeCoverage job does not exists

ochaloup:
anymore (is disabled), removing it from the release checks

/cc @mmusgrov


ochaloup:
I'm merging this as I'm going to do a release. This is the change only in the `narayana-release-process.sh` which does not impact anything else.

PR #1479, opened 2019-08-07
Jbtm 3169

mmusgrov:

https://issues.jboss.org/browse/JBTM-3169

MP-LRA issue 195 (see https://github.com/eclipse/microprofile-lra/issues) introduced some changes to response codes that relate to how unknown LRA id's are handled by implementations. The associated tests for the issue also uncovered a bug in how recovery handles in progress LRA's. The new TCK tests are in pull request 205.

The PR also adds a none zero value for the MP-LRA TCK config property `lra.tck.consistency.delay` since the new tests require two passes of the recovery thread.

### DIFF
--- a/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
+++ b/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
@@ -72,7 +72,7 @@ import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_RECOVER
 
 public class LRARecord extends AbstractRecord implements Comparable<AbstractRecord> {
     private static String TYPE_NAME = "/StateManager/AbstractRecord/LRARecord";
-    private static long PARTICIPANT_TIMEOUT = 1; // number of seconds to wait for requests
+    static long PARTICIPANT_TIMEOUT = 1; // number of seconds to wait for requests
     private static final String COMPENSATE_REL = "compensate";
     private static final String COMPLETE_REL = "complete";
 
@@ -363,10 +363,8 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
                     responseData = response.readEntity(String.class);
                 }
             } catch (Exception e) {
-                if (LRALogger.logger.isInfoEnabled()) {
-                    LRALogger.logger.infof("LRARecord.doEnd put %s failed: %s",
+                 LRALogger.logger.infof("LRARecord.doEnd put %s failed: %s",
                             target.getUri(), e.getMessage());
-                }
             } finally {
                 client.close();
             }
@@ -488,9 +486,10 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
                 // if the attempt times out the catch block below will return a heuristic
                 Response response = asyncResponse.get(PARTICIPANT_TIMEOUT, TimeUnit.SECONDS);
 
-                // 200 is the only valid response code for reporting the participant status
-                // NB 412 used to be used before the ParticipantStatus#Active state was added to the state model
-                if (response.getStatus() == Response.Status.OK.getStatusCode() &&
+                // 200 and 410 are the only valid response code for reporting the participant status
+                if (response.getStatus() == Response.Status.GONE.getStatusCode()) {
+                    return TwoPhaseOutcome.FINISH_OK;
+                } else if (response.getStatus() == Response.Status.OK.getStatusCode() &&
                         response.hasEntity()) {
                     // the participant is available again and has reported its status
                     String s = response.readEntity(String.class);
--- a/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
+++ b/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
@@ -61,11 +61,16 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.locks.ReentrantLock;
+
+import static io.narayana.lra.coordinator.domain.model.LRARecord.PARTICIPANT_TIMEOUT;
 
 public class Transaction extends AtomicAction {
     private static final String LRA_TYPE = "/StateManager/BasicAction/TwoPhaseCoordinator/LRA";
@@ -141,6 +146,15 @@ public class Transaction extends AtomicAction {
             } else {
                 os.packBoolean(true);
                 os.packString(status.name());
+            }
+
+            if (afterLRAListeners == null) {
+                os.packInt(0);
+            } else {
+                os.packInt(afterLRAListeners.size());
+                for (URI uri : afterLRAListeners) {
+                    os.packString(uri.toASCIIString());
+                }
             }
         } catch (IOException e) {
             return false;
@@ -237,6 +251,17 @@ public class Transaction extends AtomicAction {
             finishTime = finishMillis == 0 ? null :
                     LocalDateTime.ofInstant(Instant.ofEpochMilli(finishMillis), ZoneOffset.UTC);
             status = os.unpackBoolean() ? LRAStatus.valueOf(os.unpackString()) : null;
+
+            int cnt = os.unpackInt();
+
+            if (cnt == 0) {
+                afterLRAListeners = null;
+            } else {
+                afterLRAListeners = new ArrayList<>();
+                for (int i = 0; i < cnt; i++) {
+                    afterLRAListeners.add(new URI(os.unpackString()));
+                }
+            }
 
             return true;
         } catch (IOException | URISyntaxException e) {
@@ -347,6 +372,31 @@ public class Transaction extends AtomicAction {
         }
     }
 
+    private void updateAfterLRAListeners(RecordList list) {
+        if (list == null || list.size() == 0) {
+            return;
+        }
+
+        if (afterLRAListeners == null) {
+            afterLRAListeners = new ArrayList<>();
+        } else {
+            afterLRAListeners.clear();
+        }
+
+        RecordListIterator i = new RecordListIterator(list);
+        AbstractRecord r;
+
+        while ((r = i.iterate()) != null) {
+            if (r instanceof LRARecord) {
+                URI endNotification = ((LRARecord) r).getEndNotificationUri();
+
+                if (endNotification != null) {
+                    afterLRAListeners.add(endNotification);
+                }
+            }
+        }
+    }
+
     // in this version close need to run as blocking code {@link Vertx().executeBlocking}
     private int doEnd(boolean compensate) {
         inFlight = false;
@@ -377,6 +427,7 @@ public class Transaction extends AtomicAction {
 
                 pending.forEach(r -> pendingList.putRear(r));
 
+                updateAfterLRAListeners(pendingList);
                 updateState(LRAStatus.Cancelling);
 
                 super.phase2Abort(true);
@@ -395,6 +446,7 @@ public class Transaction extends AtomicAction {
                     pendingList.putRear(pendingList.getFront());
                 }
 
+                updateAfterLRAListeners(pendingList);
                 // tell each participant that the LRA canceled
                 updateState(LRAStatus.Cancelling);
 
@@ -406,6 +458,7 @@ public class Transaction extends AtomicAction {
                 super.phase2Abort(true); // this route to abort forces a log write on failures and heuristics
                 res = super.status();
             } else {
+                updateAfterLRAListeners(pendingList);
                 // tell each participant that the LRA completed ok
                 updateState(LRAStatus.Closing);
                 res = super.End(true);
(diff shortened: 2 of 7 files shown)


rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -790,15 +830,32 @@ public boolean afterLRANotification() {
...
+                    Response response = null;
+
+                    try {
+                        Future<Response> asyncResponse = client.target(uri)
+                                .request()
+                                .header(LRA.LRA_HTTP_ENDED_CONTEXT_HEADER, id)
+                                .async()
+                                .put(Entity.text(status.name()));
+
+                        response = asyncResponse.get(1000, TimeUnit.SECONDS);
```
xstefank: Can we also use PARTICIPANT_TIMEOUT here? I am not sure 1s is enough in normal network conditions.

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -790,15 +830,32 @@ public boolean afterLRANotification() {
...
+                            listeners.remove();
+                        } else {
+                            notifiedAll = false;
+                        }
+
+                        response = null;
+                    } catch (WebApplicationException | InterruptedException | ExecutionException | TimeoutException e) {
+                        if (LRALogger.logger.isInfoEnabled()) {
+                            LRALogger.logger.infof("Could not notify AfterLRA listener at %s (%s)", uri, e.getMessage());
+                        }
```
xstefank: notifiedAll should be set to false also in case of the exception

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
mmusgrov: Modern networks are fast and a second is a long time.
LRA provides guarantees so if 1 s isn't enough it will be retried on the next recovery pass - ie a low value is transactionally safe.

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -790,15 +832,34 @@ public boolean afterLRANotification() {
...
+                        } else {
+                            notifiedAll = false;
+                        }
+
+                        response = null;
+                    } catch (WebApplicationException | InterruptedException | ExecutionException | TimeoutException e) {
                         notifiedAll = false;
+
+                        if (LRALogger.logger.isInfoEnabled()) {
+                            LRALogger.logger.infof("Could not notify AfterLRA listener at %s (%s)", uri, e.getMessage());
```
ochaloup: I wonder in general here, why the caught exception is printed as the info message. Should not be a warning when it's something exceptional?

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -122,6 +122,12 @@ private void doRecoverTransaction(Uid recoverUid) {
 
                 if (!inFlight && lra.isRecovering()) {
                     lra.replayPhase2();
+
+                    if (!lra.isRecovering()) {
```
ochaloup: would not be better to check for `isFinished` or something similar? This is a bit hard to understand, plus it makes possibility to leak a race condition when `lra.replayPhase2()` does some strange thing like changing the state to `Active` or to some currently unknown state

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -122,6 +122,12 @@ private void doRecoverTransaction(Uid recoverUid) {
...
+                        if (lraService != null) {
```
ochaloup: would not be nice to concat the two conditions under one `if` statement?

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -223,7 +223,7 @@ public void filter(ContainerRequestContext containerRequestContext) {
                         // if there is an LRA present nest a new LRA under it
                         suspendedLRA = incommingLRA;
                         lraTrace(containerRequestContext, suspendedLRA, "ServerLRAFilter before: REQUIRED start new LRA");
-                        newLRA = lraId = startLRA(incommingLRA, method, getTimeOut(method));
+                        newLRA = lraId = startLRA(incommingLRA, method, getTimeOut(transactional));
```
ochaloup: could happen that the `transactional` is null? asking as the statement above checks `transactional` of not being `null`

rts/lra/lra-test/tck/pom.xml
```diff
@@ -21,7 +21,7 @@
         <lra.tck.suite.thorntail.trace.params></lra.tck.suite.thorntail.trace.params>
         <lra.tck.suite.debug.params></lra.tck.suite.debug.params> <!-- has content when -Ddebug.tck is specified -->
         <lra.tck.suite.debug.port>8788</lra.tck.suite.debug.port>
-        <lra.tck.consistency.delay>0</lra.tck.consistency.delay>
+        <lra.tck.consistency.delay>20000</lra.tck.consistency.delay>
```
ochaloup: What is the reason of adding here this delay? It's little big high number, isn't it?

mmusgrov: The delay has to be big enough to allow the next periodic recovery pass to happen. I can reduce the periodic recovery period and this delay but the full suite seems to takes 10-20 minutes which I think is fine (but not ideal) since it use to only take a few minutes prior to the new tests introduced by Rudy.

Note that we are already setting the recovery period quite low and setting it too low is not advisable.

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: On line 180 the method has already returned if transactional is null hence it can never be null at this point in the code.

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mmusgrov: +1

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mmusgrov: This code is following a well known programming idiom:

```
check condition
run some code that can alter the condition
recheck the condition
```

A race is not possible.

The isRecovering() method tests that the LRA has entered the termination phase and has not yet finished and since this is called from periodic recovery the method name is intuitive - it says "is the LRA still in need of recovery?".

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
mmusgrov: The participant that is being notified may not be running at the moment and it will be retried later. This is not an exceptional condition is a distributed system. I can log it at debug level if you prefer but I thought info would be more helpful.

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -790,15 +832,34 @@ public boolean afterLRANotification() {
...
+                    Response response = null;
+
+                    try {
+                        Future<Response> asyncResponse = client.target(uri)
+                                .request()
+                                .header(LRA.LRA_HTTP_ENDED_CONTEXT_HEADER, id)
+                                .async()
+                                .put(Entity.text(status.name()));
+
+                        response = asyncResponse.get(PARTICIPANT_TIMEOUT, TimeUnit.SECONDS);
```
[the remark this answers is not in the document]
mmusgrov: Because the expectation is that the network call should complete within milliseconds. Setting the value to 1000 is just protection against an unavailable server locking up the calling thread indefinitely. The network call will be reattempted on the next period recovery pass so it is transactionally safe to do this.

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -365,7 +365,7 @@ private int tryDoEnd(boolean compensate) {
             } catch (Exception e) {
                 if (LRALogger.logger.isInfoEnabled()) {
                     LRALogger.logger.infof("LRARecord.doEnd put %s failed: %s",
-                            target.getUri(), e.getMessage());
+                            target.getUri(), e);
```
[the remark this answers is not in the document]
mmusgrov: I changed this line when I was debugging something. I have reverted this change. We do want to log exception stacks when they are "normal" conditions. The attempt to contact the participant will be replayed on the next recovery pass.

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -365,7 +365,7 @@ private int tryDoEnd(boolean compensate) {
             } catch (Exception e) {
                 if (LRALogger.logger.isInfoEnabled()) {
```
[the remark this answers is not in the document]
mmusgrov: +1

rts/lra/lra-test/tck/pom.xml
mmusgrov: Also note that we can set the delay on just Rudy's tests after https://github.com/eclipse/microprofile-lra/pull/219 is merged.

ochaloup: ok, but the testsuite running 15 minutes instead of 2 mins sounds me rather as critical. I personally don't think this being a minor. Especially when some timeout needs to be setup to "a big number". This sounds rather as hiding some race condition functionality issue, doesn't it?

I confess that I do not follow the work on LRA spec in details so I could be missing something. But I really don't like fixing things with prolonging a timeout.

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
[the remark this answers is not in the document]
ochaloup: ok, I see

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
ochaloup: I see but I think the unavailable resource is exceptional. It's the same when JTA recovery tries to run 'rollback'. It's not expected behaviour and thus is for WARN.
It could be about the log level interpretation but a caught exception is an exception even in the distributed system.

rts/lra/lra-filters/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
ochaloup: ok, I see now (but it's a bit difficult to understand this code)

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
ochaloup: The code could follow the well know idiom. My question was rather if it's appropriate for this place. How well understandable is such idiom for the reader.

The `replayPhase2` strives to finish the LRA. From that it could be better fitting to check if the LRA is really finished.

The race is not possible now but what happen when the state model is enhanced and the LRA starts to use new action status. Could be then the `toLRAStatus` falls through to be Active?

mmusgrov:
> I added few comments. My main concern is about prolonging the timeout.

I didn't see any extra comments. You should [contribute to the discussion we are having in issue 218](https://github.com/eclipse/microprofile-lra/issues/218). The non zero timeout is because Rudy does not want to call `replayEndPhase` in his new TCK tests.

ochaloup:
@mmusgrov ok, I see. I wrote the comments at time of posting my review. Eg. https://github.com/jbosstm/narayana/pull/1479#discussion_r312199200

I do not follow the TCK and LRA in details but forcing the Narayana testsuite to be run 10times slower is sounds wrong to me and forcing any timeout to be so long sound as hiding some race condition.
When @xstefank or @rdebusscher are in agreement with such changes and impact to the implementatis then it's probably fine. But it does not seem fine to me.

mmusgrov:
@ochaloup I agree wholeheartedly but if Rudy does not want to trigger another attempt at contacting the participant what can we do?

The TCK delay is to give "eventual consistency" the opportunity to play out. Implementations are supposed to complete within this delay period. The narayana implementation works just like other Narayana transaction models in the sense that we have a recovery system reading logs from the object store which replays the end phase. The system operates by periodically running and we cannot make that period too short (and 20 seconds is very short for the recovery thread).

But you know all this so you should be discussing it with the LRA community on the issue I referred to above. This PR is to pass the TCK and is the only way to pass it with the way the TCK is currently implemented.

rts/lra/lra-test/tck/pom.xml
mmusgrov: There is no race condition, we are setting the delay long enough for "eventual consistency" to kick in. In theory the delay could be set to an arbitrarily large value since that would conform to the definition of eventual consistency.

You do not need to know the details of the spec, you just need to understand that the only guarantee provided by LRA is that the system will eventually become consistent. The question then is, in practice, how long is it reasonable to wait for consistency before failing the TCK test.

ochaloup: Ok, I see and I do understand. No objection then :-)

PR #1482, opened 2019-08-12
JBTM-3171 Set recovery header on AfterLRA notifications.

mmusgrov:
https://issues.jboss.org/browse/JBTM-3171

MAIN !TOMCAT !WIN !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF !RTS !AS_TESTS !JACOCO

LRA includes the capability to receive JAX-RS notifications when an LRA is finished but the implementation does not set the "recovery header". This requirement was added in MP-LRA issue eclipse/microprofile-lra#209 with new tests to verify that the recovery header is set correctly.

The PR also updates the implementation to use the new name for MP-LRA config properties (that replace lra.tck.consistency.delay)

### DIFF
--- a/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
+++ b/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
@@ -71,6 +71,7 @@ import java.util.concurrent.TimeoutException;
 import java.util.concurrent.locks.ReentrantLock;
 
 import static io.narayana.lra.coordinator.domain.model.LRARecord.PARTICIPANT_TIMEOUT;
+import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_RECOVERY_HEADER;
 
 public class Transaction extends AtomicAction {
     private static final String LRA_TYPE = "/StateManager/BasicAction/TwoPhaseCoordinator/LRA";
@@ -79,7 +80,7 @@ public class Transaction extends AtomicAction {
     private URI parentId;
     private String clientId;
     private List<LRARecord> pending;
-    private List<URI> afterLRAListeners;
+    private List<AfterLRAListener> afterLRAListeners;
     private LRAStatus status;
     private String responseData;
     private LocalDateTime startTime;
@@ -152,8 +153,9 @@ public class Transaction extends AtomicAction {
                 os.packInt(0);
             } else {
                 os.packInt(afterLRAListeners.size());
-                for (URI uri : afterLRAListeners) {
-                    os.packString(uri.toASCIIString());
+                for (AfterLRAListener participant : afterLRAListeners) {
+                    os.packString(participant.notificationURI.toASCIIString());
+                    os.packString(participant.recoveryId.toASCIIString());
                 }
             }
         } catch (IOException e) {
@@ -259,7 +261,7 @@ public class Transaction extends AtomicAction {
             } else {
                 afterLRAListeners = new ArrayList<>();
                 for (int i = 0; i < cnt; i++) {
-                    afterLRAListeners.add(new URI(os.unpackString()));
+                    afterLRAListeners.add(new AfterLRAListener(new URI(os.unpackString()), new URI(os.unpackString())));
                 }
             }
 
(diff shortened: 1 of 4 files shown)


rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -869,4 +873,14 @@ public boolean afterLRANotification() {
 
         return notifiedAll;
     }
+
+    static class Pair {
```
ochaloup: Would not be good to name it in way to explain what is the purpose of the `Pair`? As it's a inline class it's expected to be used in the closed context of the class and the name can express what is that for. It could be like `LRANotificationPair` or `LRANotificationListenerData` or something similar.

xstefank: I agree, it's hard to read `Pair participant = ...`. I think this is `AfterLRAParticipant` which consists of required afterLRA URI and optional recoveryURI if it is also LRA participant in the normal sense.

rts/lra/pom.xml
```diff
@@ -138,13 +138,13 @@
     </build>
 
     <modules>
+        <module>lra-service-base</module>
```
tomjenkinson: Does this relate to https://developer.jboss.org/message/990612?et=watches.email.thread#990612

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -389,9 +391,10 @@ private void updateAfterLRAListeners(RecordList list) {
         while ((r = i.iterate()) != null) {
             if (r instanceof LRARecord) {
                 URI endNotification = ((LRARecord) r).getEndNotificationUri();
+                URI recoveryCoordinatorURI = ((LRARecord) r).getRecoveryCoordinatorURI();
 
-                if (endNotification != null) {
-                    afterLRAListeners.add(endNotification);
+                if (endNotification != null && recoveryCoordinatorURI != null) {
```
tomjenkinson: Is it ever possible for recoveryCoordinatorURI to be == null but endNotification be != null. Is that covered by spec as an error?

mmusgrov: No, neither of them can be null. I did the check just to be ultra cautious. The background to the two fields is:

A class annotated with both `@AfterLRA` and `@LRA` will receive LRA end notifications if the `@LRA` method is invoked. When the method is invoked we enlist the entity with our coordinator by creating an instance of LRARecord (https://github.com/jbosstm/narayana/blob/master/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java#L552). The LRARecord constructor initialises the endNotification URI. Immediately after construction [we set the recoveryCoordinatorURI](https://github.com/jbosstm/narayana/blob/master/rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java#L557) (I did it that way because it is easier to get the arjuna uid of the record after creation and I didn't see a huge benefit in generating the recovery URI inside the constructor). But it does mean that initialisation the two fields happens in different methods but if there was a crash in between then nothing would make it to disk. Therefore the null check is redundant but I regard it as a harmless check (if I didn't have the check I guess I'd have to document in the code why they cannot both be null).

rts/lra/pom.xml
mmusgrov: I did see the forum question about lra-service-base but wasn't sure. I did add this change to make sure that the lra common module is the first thing to be built (which it should be) but I am not sure if it is related (just that the forum post made me check the code).

rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
mmusgrov: @tomjenkinson actually I remember why I don't initialise the recovery URI in the LRARecord constructor: it is because I need to know the JAX-RS base URL of the server that generates the recovery URI and the class that constructs the LRARecord knows what that is.

PR #1486, opened 2019-08-20
JBTM-3178 Add warning for AbstractRecord create failure

mmusgrov:
https://issues.jboss.org/browse/JBTM-3178

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !AS_TESTS !TOMCAT !JACOCO !mysql !postgres !db2 !oracle !RTS !LRA

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
@@ -473,21 +473,14 @@ public abstract class AbstractRecord extends StateManager
         }
 	    catch (final Throwable ex)
 	    {
-			try {
-				Class rt = RecordType.typeToClass(type);
-				String className = (rt == null) ? "null" : rt.getName();
-				int mb = 1024 * 1024;
-				Runtime runtime = Runtime.getRuntime();
-				long used = (runtime.totalMemory() - runtime.freeMemory()) / mb;
-				long free = runtime.freeMemory() / mb;
+            Class rt = RecordType.typeToClass(type);
+            String className = (rt == null) ? "null" : rt.getName();
 
-				tsLogger.i18NLogger.warn_coordinator_AbstractRecord_create_failure(
-						type, className, used, free, ex);
-			} catch (final Throwable t) {
-				ex.printStackTrace();
-			}
+            tsLogger.i18NLogger.warn_coordinator_AbstractRecord_create_failure(type, className);
 
-			return null;
+            ex.printStackTrace();
+
+            return null;
 	    }
 	}
 
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
@@ -473,9 +473,21 @@ public abstract class AbstractRecord extends StateManager
         }
 	    catch (final Throwable ex)
 	    {
-	        ex.printStackTrace();
-	        
-	        return null;
+			try {
+				Class rt = RecordType.typeToClass(type);
+				String className = (rt == null) ? "null" : rt.getName();
+				int mb = 1024 * 1024;
+				Runtime runtime = Runtime.getRuntime();
+				long used = (runtime.totalMemory() - runtime.freeMemory()) / mb;
+				long free = runtime.freeMemory() / mb;
+
+				tsLogger.i18NLogger.warn_coordinator_AbstractRecord_create_failure(
+						type, className, used, free, ex);
+			} catch (final Throwable t) {
+				ex.printStackTrace();
+			}
+
+			return null;
 	    }
 	}
 
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/logging/arjunaI18NLogger.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/logging/arjunaI18NLogger.java
@@ -1566,6 +1566,12 @@ public interface arjunaI18NLogger {
     @Message(id = 12391, value = "Could not initialize object store '{0}' of type '{1}'", format = MESSAGE_FORMAT)
     public String init_StoreManager_instantiate_class_failure(String name, String type);
 
+    @Message(id = 12392, value = "AbstractRecord.create {0} failed to create record class {1} (memory: used {2}MB, {3}MB free).",
+           format = MESSAGE_FORMAT)
+    @LogMessage(level = WARN)
+    public void warn_coordinator_AbstractRecord_create_failure(
+        Integer arg1, String arg2, Long arg3, Long arg4, @Cause() Throwable arg0);
+
     /*
         Allocate new messages directly above this notice.
           - id: use the next id number in numeric sequence. Don't reuse ids.
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/logging/arjunaI18NLogger.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/logging/arjunaI18NLogger.java
@@ -1566,11 +1566,10 @@ public interface arjunaI18NLogger {
     @Message(id = 12391, value = "Could not initialize object store '{0}' of type '{1}'", format = MESSAGE_FORMAT)
     public String init_StoreManager_instantiate_class_failure(String name, String type);
 
-    @Message(id = 12392, value = "AbstractRecord.create {0} failed to create record class {1} (memory: used {2}MB, {3}MB free).",
+    @Message(id = 12392, value = "AbstractRecord.create {0} failed to create record class {1}.",
            format = MESSAGE_FORMAT)
     @LogMessage(level = WARN)
-    public void warn_coordinator_AbstractRecord_create_failure(
-        Integer arg1, String arg2, Long arg3, Long arg4, @Cause() Throwable arg0);
+    public void warn_coordinator_AbstractRecord_create_failure(Integer arg1, String arg2);
 
     /*
         Allocate new messages directly above this notice.


ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
```diff
@@ -473,9 +473,21 @@ public static AbstractRecord create (int type)
         }
 	    catch (final Throwable ex)
 	    {
-	        ex.printStackTrace();
-	        
-	        return null;
+			try {
+				Class rt = RecordType.typeToClass(type);
+				String className = (rt == null) ? "null" : rt.getName();
+				int mb = 1024 * 1024;
```
ochaloup: My general impression is that the memory consumpsion data is a specific entry which is not meant for a general usage. It's more about to discover an issue. But issue discovery should be processed with a dev branch not on master.
It would be nice to change the message to log WARN instead of exception being printed on stdout. And information about what is the record type is really useful. But I don't think the same about the memory calculation.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
```diff
@@ -473,9 +473,21 @@ public static AbstractRecord create (int type)
...
+				Runtime runtime = Runtime.getRuntime();
+				long used = (runtime.totalMemory() - runtime.freeMemory()) / mb;
+				long free = runtime.freeMemory() / mb;
+
+				tsLogger.i18NLogger.warn_coordinator_AbstractRecord_create_failure(
+						type, className, used, free, ex);
+			} catch (final Throwable t) {
+				ex.printStackTrace();
```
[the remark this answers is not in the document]
mmusgrov: This code was already present in the original code and I have not changed it.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
mmusgrov: > My general impression is that the memory consumpsion data is a specific entry which is not meant for a general usage. It's more about to discover an issue. But issue discovery should be processed with a dev branch not on master.

Disagree, the message is providing more than what was already there (which was just an exception trace). This code path is almost never seen so we need something that will appear under general testing conditions.

> It would be nice to change the message to log WARN instead of exception being printed on stdout. 

Disagree, the message is already printed at WARN level (have you read the rest of the code immediately following this line).

> And information about what is the record type is really useful.

Disagree, the message is already printing the record type.

> But I don't think the same about the memory calculation.

Disagree, This code path is almost never seen so we need something that will appear under general testing conditions.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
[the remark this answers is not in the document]
ochaloup: Yes, and with enhancing our code practice it will be nice to improve it.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
ochaloup: I probably haven't expressed my ideas with the right words. I will try that better, sorry for that.
I meant it in way: The state of code before was printing stacktrace to stdout. Then using logger and WARN with printing the record type is nice and it's needed.
Still I think printing memory consumption into the WARN message is not right.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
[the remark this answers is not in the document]
tomjenkinson: I agree with Ondra, I think this (new catch final Throwable t) handling should use the logger but I am about to add another comment you might disagree with Ondra

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
```diff
@@ -473,9 +473,21 @@ public static AbstractRecord create (int type)
...
+			try {
+				Class rt = RecordType.typeToClass(type);
+				String className = (rt == null) ? "null" : rt.getName();
+				int mb = 1024 * 1024;
+				Runtime runtime = Runtime.getRuntime();
+				long used = (runtime.totalMemory() - runtime.freeMemory()) / mb;
+				long free = runtime.freeMemory() / mb;
+
+				tsLogger.i18NLogger.warn_coordinator_AbstractRecord_create_failure(
+						type, className, used, free, ex);
```
tomjenkinson: I think this should log ex.printStackTrace();after the warn as before in case people are watching system.err for the stack trace...

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
tomjenkinson: I agree with Ondra's summary, it seems usual to add in the memory to the tracing. Perhaps there is a reason that this is needed for this case and in that situation it may well be warranted. Per default I was expecting the new statement to just be something like:
```
Class rt = RecordType.typeToClass(type);
String className = (rt == null) ? "null" : rt.getName();
tsLogger.i18NLogger.warn_coordinator_AbstractRecord_create_failure(type, className, ex);
```
It's not clear how the memory information helps here from my point of view but as I say if there is some reason to add it we can do so, though by the nature of the volatilty of the JVM the numbers for the memory that we log could well be different at the time we catch the exception to the time when the event that the exception was raised were.

mmusgrov: I added it because we have no idea about why we get the InstantiationException and printing some data about memory usage may be useful.

tomjenkinson: IMO I think that is too speculative. Perhaps you could separately trace the message but even that seems speculative. It could be a problem with diskspace for example?

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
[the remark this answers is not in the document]
mmusgrov: Why not raise a separate issue for it since we have lots of code that prints stacktraces so picking on this PR is an ad hoc extra requirement that is orthogonal to the purpose of the PR.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
mmusgrov: Try that last commit instead

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
mmusgrov: None of the standard reasons for instantiation exceptions explain what we see. The mem usage data may have given a hint as to whether or not the JVM is stressed. But I have removed the extra data in my last commit.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
ochaloup: My opinion is that we should try to avoid use of the `Exception.printStackTrace()` calls when logging is available. It's usual good practice to java to use loggers
https://pmd.github.io/latest/pmd_rules_java_bestpractices.html#avoidprintstacktrace
Logger offers better control on when and what will be shown, logger is usually used to gather and redirect logs over network etc.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/logging/arjunaI18NLogger.java
```diff
@@ -1566,6 +1566,11 @@ public void warn_objectstore_JDBCImple_over_max_image_size(int imageSize,
     @Message(id = 12391, value = "Could not initialize object store '{0}' of type '{1}'", format = MESSAGE_FORMAT)
     public String init_StoreManager_instantiate_class_failure(String name, String type);
 
+    @Message(id = 12392, value = "AbstractRecord.create {0} failed to create record class {1}.",
+           format = MESSAGE_FORMAT)
+    @LogMessage(level = WARN)
+    public void warn_coordinator_AbstractRecord_create_failure(Integer arg1, String arg2);
```
ochaloup: Maybe it would be better to name the arguments with the `type` and the `className` rather then using arg1 and arg2?

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
tomjenkinson: On the one hand I do agree that we should definitely make sure any new log statements follow best practice, once the code is in the wild we have to consider whether making the change outweighs the benefit of users who may be relying on the old behaviour. This is not a precise decision always as the pro/con have to be evaluated each time.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/logging/arjunaI18NLogger.java
mmusgrov: This is not a valid objection. The parameter names follows the general style of the other methods in the file. We have agreed in the past to following the style that was already present when making changes.

tomjenkinson: Ah, yes I see. Looking at the diff it shows the one above does have it but that style seems to be in the minority. I suspect as this file is not really re-used too much due to the nature of its purpose. Thanks for looking @mmusgrov

PR #1492, opened 2019-08-29
JBTM-3184 LRA TckParticipantTests do not finish started LRAs

xstefank:
https://issues.jboss.org/browse/JBTM-3184

### DIFF
--- a/rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRAParticipant.java
+++ b/rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRAParticipant.java
@@ -48,7 +48,6 @@ import java.net.URI;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Optional;
 import java.util.concurrent.CompletionStage;
 
 import static io.narayana.lra.LRAConstants.AFTER;
@@ -71,7 +70,7 @@ public class LRAParticipant {
     private Method afterLRAMethod;
     private Object instance;
 
-    private Map<URI, Optional<?>> participantStatusMap = new HashMap<>();
+    private Map<URI, ParticipantResult> participantStatusMap = new HashMap<>();
 
     LRAParticipant(Class<?> javaClass) {
         this.javaClass = javaClass;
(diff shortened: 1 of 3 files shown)


rts/lra/lra-coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -847,7 +847,7 @@ public boolean afterLRANotification() {
 
                         response = asyncResponse.get(PARTICIPANT_TIMEOUT, TimeUnit.SECONDS);
 
-                        if (response.getStatus() == 200) {
+                        if (response.getStatusInfo().getFamily() == Status.Family.SUCCESSFUL) {
```
mmusgrov: The spec requires a 200 OK response

xstefank: ok, let me revisit this

xstefank: you were right, we were returning 204 from non-JAX-RS AfterLRA methods, so I updated rather that.

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRAParticipant.java
```diff
@@ -111,8 +110,10 @@ synchronized Response forget(URI lraId, URI parentId) {
         return invokeParticipantMethod(forgetMethod, lraId, parentId, FORGET);
     }
 
-    synchronized void afterLRA(URI lraId, LRAStatus lraStatus) {
+    synchronized Response afterLRA(URI lraId, LRAStatus lraStatus) {
         invokeMethod(AFTER, afterLRAMethod, getInstance(), lraId, lraStatus);
+        // above invocation throws RuntimeException in case an unexpected exception is thrown from client method
+        return Response.ok().build();
```
mmusgrov: If the invokeMethod call fails then the implementation needs to know so that it can keep retrying the afterLRA notification.

xstefank: RuntimeException translates to 500 internal server error, so the coordinator will retry the call.

Please note that this will interpret WebApplicationException (as in other participant methods invocations) so if the user includes 200 status code in this exception it will be interpreted as a successful call. Is that a problem?

mmusgrov: You have some code in invokeMethod which will returns processThrowable(e.getTargetException(), type) which returns an HTTP status code if the target method throws an exception. But your afterLRA method ignores that return value.

xstefank:
@mmusgrov thanks, PR updated.

PR #1494, opened 2019-09-02
JBTM-3180 Do not use null Entities on JAX-RS PUT requests

mmusgrov:
https://issues.jboss.org/browse/JBTM-3180


mmusgrov:
No. This is an HTTP compliance issue for JAX-RS clients and the one we use (resteasy) does not complain about null entities. To test it we'd need to use a client that does issue such a warning such as Jersey (https://jersey.github.io/apidocs/latest/jersey/org/glassfish/jersey/client/ClientProperties.html#SUPPRESS_HTTP_COMPLIANCE_VALIDATION).. 

So this would be a test issue against a particular JAX-RS client.

mmusgrov:
And it looks like the behaviour is still under discussion: https://github.com/eclipse-ee4j/jaxrs-api/issues/632

So this is not something we need to test. But it does no harm to fix our usage of PUT just in case resteasy every changes its' current behaviour.

mmusgrov:
@paulparkinson thanks for the report

PR #1496, opened 2019-09-03
JBTM-3187 Add MessageBodyWriter for ParticipantStatus and application…

xstefank:
…/octet-stream media type

https://issues.jboss.org/browse/JBTM-3187

### DIFF
--- a/rts/lra/narayana-lra/src/main/java/io/narayana/lra/provider/ParticipantStatusOctetStreamProvider.java
+++ b/rts/lra/narayana-lra/src/main/java/io/narayana/lra/provider/ParticipantStatusOctetStreamProvider.java
@@ -1,0 +1,33 @@
+package io.narayana.lra.provider;
+
+import org.eclipse.microprofile.lra.annotation.ParticipantStatus;
+
+import javax.ws.rs.Produces;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.ext.MessageBodyWriter;
+import javax.ws.rs.ext.Provider;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+
+@Provider
+@Produces(MediaType.APPLICATION_OCTET_STREAM)
+public class ParticipantStatusOctetStreamProvider implements MessageBodyWriter<ParticipantStatus> {
+    @Override
+    public boolean isWriteable(Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
+        return type.isAssignableFrom(ParticipantStatus.class);
+    }
+
+    @Override
+    public long getSize(ParticipantStatus participantStatus, Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
+        return -1;
+    }
+
+    @Override
+    public void writeTo(ParticipantStatus participantStatus, Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType, MultivaluedMap<String, Object> httpHeaders, OutputStream entityStream) throws IOException, WebApplicationException {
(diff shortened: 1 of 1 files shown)


rts/lra/narayana-lra/src/main/java/io/narayana/lra/provider/ParticipantStatusOctetStreamProvider.java
```diff
@@ -0,0 +1,33 @@
...
+@Produces(MediaType.APPLICATION_OCTET_STREAM)
+public class ParticipantStatusOctetStreamProvider implements MessageBodyWriter<ParticipantStatus> {
+    @Override
+    public boolean isWriteable(Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
+        return type.isAssignableFrom(ParticipantStatus.class);
+    }
+
+    @Override
+    public long getSize(ParticipantStatus participantStatus, Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
+        return 0;
```
mmusgrov: The spec says:
```
As of JAX-RS 2.0, the method has been deprecated and the value returned by the method is ignored by a JAX-RS runtime. All MessageBodyWriter implementations are advised to return -1 from the method. 
```

rts/lra/narayana-lra/src/main/java/io/narayana/lra/provider/ParticipantStatusOctetStreamProvider.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    @Override
+    public void writeTo(ParticipantStatus participantStatus, Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType, MultivaluedMap<String, Object> httpHeaders, OutputStream entityStream) throws IOException, WebApplicationException {
+        entityStream.write(participantStatus.toString().getBytes());
```
mmusgrov: It is better to use

``` java
participantStatus.name()
```

since it cannot be overridden whereas toString() can.

rts/lra/narayana-lra/src/main/java/io/narayana/lra/provider/ParticipantStatusOctetStreamProvider.java
```diff
@@ -0,0 +1,33 @@
...
+import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.ext.MessageBodyWriter;
+import javax.ws.rs.ext.Provider;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+
+@Provider
+@Produces(MediaType.APPLICATION_OCTET_STREAM)
```
mmusgrov: Since the actual body that the writer is writing is a java String (enum#name()) why did you choose an octet stream in preference to MediaType.TEXT_PLAIN?

rts/lra/narayana-lra/src/main/java/io/narayana/lra/provider/ParticipantStatusOctetStreamProvider.java
xstefank: ok, I will update this

rts/lra/narayana-lra/src/main/java/io/narayana/lra/provider/ParticipantStatusOctetStreamProvider.java
xstefank: ok, I will update this

rts/lra/narayana-lra/src/main/java/io/narayana/lra/provider/ParticipantStatusOctetStreamProvider.java
xstefank: This is exactly the point of this issue. application/octet-stream is a default media type in RESTEasy, so this provider allows to declare JAX-RS Complete, Compensate, Status, etc. without `@Produces()` annotation.

PR #1499, opened 2019-09-12
JBTM-3191 Align LRA client expectations for LRA coordinator default port

xstefank:
https://issues.jboss.org/browse/JBTM-3191


xstefank:
@tomjenkinson yes, that's the same error I am seeing locally. I will look deeper into quickstarts and make necessary updates.

xstefank:
@tomjenkinson swarm.http.port is the same as thorntail.http.port thus changing the port trip-client application is running on. Let me go through the quickstarts, I will also try to fix the compilation errors.

PR #1504, opened 2019-09-21
JBTM-3190 Transform LRA coordinator runtime to Quarkus

xstefank:
https://issues.jboss.org/browse/JBTM-3190

Please note that swagger-ui is included also in production mode (by default it's only dev and test)
```properties
# allow Swagger in prod mode
quarkus.swagger-ui.always-include=true
```

because originally Thorntail also exposed it by default. If this is not desirable I can remove the property.


mmusgrov:
@xstefank We can decide whether to expose the openapi interface when/if we productise the coordinator.

xstefank:
@mmusgrov Thanks!

PR #1510, opened 2019-10-02
JBTM-3204 Upgrade LRA TCK recovery handling to accommodate specification changes

xstefank:
https://issues.jboss.org/browse/JBTM-3204

### DIFF
--- a/rts/lra/lra-test/lra-test-arquillian-extension/pom.xml
+++ b/rts/lra/lra-test/lra-test-arquillian-extension/pom.xml
@@ -21,6 +21,11 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
+      <groupId>org.eclipse.microprofile.lra</groupId>
+      <artifactId>microprofile-lra-tck</artifactId>
+      <version>${version.microprofile.lra.tck}</version>
+    </dependency>
+    <dependency>
       <groupId>org.jboss.narayana.rts</groupId>
       <artifactId>narayana-lra</artifactId>
       <version>${project.version}</version>
@@ -35,11 +40,6 @@
       <artifactId>shrinkwrap-api</artifactId>
       <version>${version.org.jboss.shrinkwrap}</version>
       <scope>compile</scope>
-    </dependency>
-    <dependency>
-      <groupId>io.thorntail</groupId>
-      <artifactId>arquillian</artifactId>
-      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.jboss.arquillian.junit</groupId>
--- a/rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
+++ b/rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
@@ -1,0 +1,64 @@
+package io.narayana.lra.arquillian.spi;
+
+import io.narayana.lra.LRAConstants;
+import io.narayana.lra.client.NarayanaLRAClient;
+import org.eclipse.microprofile.lra.tck.service.spi.LraRecoveryService;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.ClientBuilder;
+import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.Response;
+import java.net.URI;
+
+@ApplicationScoped
+public class NarayanaLRARecovery implements LraRecoveryService {
+
+    @Override
+    public void triggerRecovery() {
+        doTriggerRecovery(null, System.getProperty(NarayanaLRAClient.LRA_COORDINATOR_HOST_KEY),
+            Integer.parseInt(System.getProperty(NarayanaLRAClient.LRA_COORDINATOR_PORT_KEY)));
+    }
+
+    @Override
+    public void triggerRecovery(URI lraId) {
+        doTriggerRecovery(lraId, lraId.getHost(), lraId.getPort());
+    }
+
+    private void doTriggerRecovery(URI lraId, String host, int port) {
+        if (invokeCoordinatorRecovery(lraId, host, port)) {
+            // first recovery scan probably collided with periodic recovevery which started
+            // before the test execution so try once more
+            invokeCoordinatorRecovery(null, host, port);
+        }
+    }
+
+    /**
+     * Invokes LRA coordinator recovery REST endpoint and returns whether the recovery of intended LRA happended
+     *
+     * @param lraId LRA id of the LRA that is expected to recover
+     * @param host the LRA coordinator host address
+     * @param port the LRA coordinator port
+     * @return true if the expected LRA recovered and no repeated recovery is necessary, false otherwise
+     */
+    private boolean invokeCoordinatorRecovery(URI lraId, String host, int port) {
+        // trigger a recovery scan
+        Client recoveryCoordinatorClient = ClientBuilder.newClient();
+
+        try {
+            String recoveryCoordinatorUrl = String.format("http://%s:%d/%s/recovery",
+                host, port, LRAConstants.RECOVERY_COORDINATOR_PATH_NAME);
+            WebTarget recoveryTarget = recoveryCoordinatorClient.target(URI.create(recoveryCoordinatorUrl));
+
+            // send the request to the recovery coordinator
+            Response response = recoveryTarget.request().get();
+            String json = response.readEntity(String.class);
+            response.close();
+
+            return lraId != null && !json.contains(lraId.toASCIIString());
+        } finally {
+            recoveryCoordinatorClient.close();
+        }
+
+    }
+}
--- a/rts/lra/narayana-lra/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
+++ b/rts/lra/narayana-lra/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
@@ -22,7 +22,6 @@
 package io.narayana.lra.filter;
 
 import io.narayana.lra.Current;
-import io.narayana.lra.LRAConstants;
 import io.narayana.lra.client.NarayanaLRAClient;
 import io.narayana.lra.client.internal.proxy.nonjaxrs.LRAParticipant;
 import io.narayana.lra.client.internal.proxy.nonjaxrs.LRAParticipantRegistry;
@@ -37,9 +36,6 @@ import org.eclipse.microprofile.lra.annotation.ws.rs.Leave;
 import javax.inject.Inject;
 import javax.ws.rs.NotFoundException;
 import javax.ws.rs.WebApplicationException;
-import javax.ws.rs.client.Client;
-import javax.ws.rs.client.ClientBuilder;
-import javax.ws.rs.client.WebTarget;
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.container.ContainerRequestFilter;
 import javax.ws.rs.container.ContainerResponseContext;
@@ -413,8 +409,6 @@ public class ServerLRAFilter implements ContainerRequestFilter, ContainerRespons
                 }
             }
 
-            checkForRecovery(requestContext);
-
             if (responseContext.getStatus() == Response.Status.OK.getStatusCode() &&
                     NarayanaLRAClient.isAsyncCompletion(resourceInfo.getResourceMethod())) {
                 LRALogger.i18NLogger.warn_lraParticipantqForAsync(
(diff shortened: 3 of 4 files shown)


mmusgrov:
@xstefank Is the forget test failure related to what we discussed on hangouts?

xstefank:
@mmusgrov sorry, I don't know what you mean. But probably it isn't. This is an intermittent failure (even in some other tests) caused probably by collisions with periodic recovery in which case the call to /recovery just wait for periodic recovery to end a thus it doesn't know about the LRAs started after the periodic recovery was triggered by the timer. (thanks for the pointers @mmusgrov)

xstefank:
@mmusgrov scratch that, the answer to your question is yes. I just thought hangouts is the video conferencing not the chat :) It should be fixed with the latest commit but I sometimes get forget called two times, however, I can't reproduce it anymore locally. Let's see how CI does.

xstefank:
Ci reproduced my local error, forget is now being called twice which means that the first recovery pass doesn't close the LRA properly.

xstefank:
@tomjenkinson yes, it is. I am working on it now.

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLraRecovery.java
```diff
@@ -0,0 +1,57 @@
...
+    public void triggerRecovery(URI lraId) {
+        doTriggerRecovery(lraId.getHost(), lraId.getPort());
+    }
+
+    private void doTriggerRecovery(String host, int port) {
+        if (!invokeCoordinatorRecovery(host, port)) {
+            // first recovery scan probably collided with periodic recovevery which started
+            // before the test execution so try once more
+            invokeCoordinatorRecovery(host, port);
+        }
```
tomjenkinson: The two recovery attempts could be due to this? /cc @xstefank @mmusgrov 

Perhaps there is something in the json payload below even in the case we expect and that might cause forget twice (PS I am not sure if forget should be called twice even in two recovery attempts - @mmusgrov might be able to comment on that - perhaps after forget is called once Narayana should remember that and in the absence of a crash it could save that info)

xstefank: yes, it is because of this. But recovery should forget about LRA once forget call succeeds for the first time. I have debug logs locally on the coordinator, it seems that the same LRA is registered for recovery twice with a different id. Need to look a little more.

tomjenkinson: Very strange. 

I do note that the console output shows the have the same LRA ID (I think) but perhaps this is not the ID you are referring to?:
> [INFO] [lra-coordinator-it-test] 2019-10-02 22:53:00,405 INFO  [io.nar.lra] (executor-thread-1) RecoveringLRA.replayPhase2 for http://localhost:8080/lra-coordinator/0_ffffac11000c_81fb_5d952910_6d ended with status: Closing
> [INFO] [lra-coordinator-it-test] 2019-10-02 22:53:05,440 INFO  [io.nar.lra] (executor-thread-1) RecoveringLRA.replayPhase2 for http://localhost:8080/lra-coordinator/0_ffffac11000c_81fb_5d952910_6d ended with status: Closing

xstefank: no, this is correct. triggerRecovery() is called two times in the test. My other comments about duplicate id were a dead end (different issue - quarkus is now by default creating objectstore on the filesystem so it's not just enough to kill coordinator process now to clear out LRAs). 

The issue is indeed in how to determine that we collided with periodic recovery (and thus need to replay recovery call). The issue with testForget is:

* test completes an LRA with participant returning 503 status
* triggerRecovery is called
** first /recovery call calls participant /status which again return 503 (thus LRA is still active)
** now, because the /recovery returned non-empty JSON we will replay the /recovery call (this error) -- HERE we need to know that we do not need to replay this call
* then the test changes participant return statuses to 200
* triggerRecovery is called again
* participant /status is called which now returns 200
* participant /forget is called which returns 200

@mmusgrov @tomjenkinson is there a way we can tell if recovery scan is running?

Note that there is nothing wrong with this behavior (status can be called multiple times) but the test is not adjusted to it.

tomjenkinson: @xstefank I didn't quite follow the quarkus issue. Isn't the objectstore always on disk? If there is some problem there @mmusgrov please take a look. To the case in hand, regarding knowing if the recovery scan is in progress I think we should be OK to use something like: https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/RecoveryManager.java#L256 and https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/RecoveryManager.java#L282 - or perhaps we can start the recovery manager in non-threaded mode if we  have the callbacks we need in order to trigger recovery: https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/RecoveryManagerImple.java#L63

mmusgrov: I am not really following the issue here but:
- calling forget multiple times is within spec (for example suppose the participant gets the notification but the response goes missing;
- calling scan will block if there is a current scan in progress and if the record wasn't recovered by that scan then calling scan a second time will recover it completely (if the participants can be contacted).

Will you clearly state what the issue is again?

xstefank: @mmusgrov First issue:

if the /recovery call collides with the periodic recovery then LRAs ended after periodic recovery started but before /recovery call is received are not recovered as /recovery call is just waiting for the periodic recovery scan to finish
* solved by calling /recovery twice

Second issue:

in the test `testForget` the **`/status`** is called twice (not /forget!) because of the test scenario which is described in detail in https://github.com/jbosstm/narayana/pull/1510#discussion_r330947731

mmusgrov: So I still do not see any issue, that is all expected behaviour. The PR looks fine as is.

mmusgrov: With one caveat: why doesn't the test trigger the second recovery invocation. The typical behaviour would be for the TCK:
- to test a condition and trigger recovery if required
- retest the condition and rerun recovery if required

If you don't to burden the code with this responsibility then an possible alternative would be for the TCK test to pass in a lamda to validate the condition.

With the way you have coded it any test that fails with pending LRAs will now fail.

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLraRecovery.java
```diff
@@ -0,0 +1,57 @@
...
+                host, port, LRAConstants.RECOVERY_COORDINATOR_PATH_NAME);
+            WebTarget recoveryTarget = recoveryCoordinatorClient.target(URI.create(recoveryCoordinatorUrl));
+
+            // send the request to the recovery coordinator
+            Response response = recoveryTarget.request().get();
+            String json = response.readEntity(String.class);
+            response.close();
+
+            // empty JSON array is returned if recovery coordinator doesn't know about any recovering LRA
+            return json.equals("[]");
```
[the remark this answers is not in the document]
xstefank: This might be just my point of view, but this code is narayana specific so if any previous test fails with active LRAs that is an error on its own and we should add a test to the TCK to verify that is not the case or prune the coordinator object store before every test.

And this line only makes the /recovery endpoint to be called again. Would that be a problem? It will only replay end phase 2 on the pending LRAs from the previous tests -- unnecessary but it shouldn't fail the test.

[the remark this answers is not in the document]
xstefank: @mmusgrov also can you please respond on google hangouts to @tomjenkinson proposal? It would eliminate the need to call /recovery endpoint from the participant side twice.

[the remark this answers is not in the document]
mmusgrov: You are over complicating the fix. Just check whether or not the return JSON contains the LRA id you are looking for, ie just do this:

```java
return json.contains(lraid);
```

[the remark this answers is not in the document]
tomjenkinson: That looks like it could help to me.

If I followed correctly though, the following test would also need to be made more robust to ensure that it can tolerate forget to be called multiple times? https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckContextTests.java#L173

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLraRecovery.java
```diff
@@ -0,0 +1,57 @@
...
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.ClientBuilder;
+import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.Response;
+import java.net.URI;
+
+@ApplicationScoped
+public class NarayanaLraRecovery implements LraRecoveryService {
```
mmusgrov: Since changes have been requested can I also ask you to change the class name to NarayanaLRARecovery since we normally use camel case. LRA is an acronym and changing it to Lra implies that Lra is a genuine word (as opposed to an acronym).

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLraRecovery.java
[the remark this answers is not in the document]
xstefank: no, forget is not an issue here but it wouldn't hurt of course.

xstefank:
@mmusgrov updated with your comments and TCK passed. Ready for review.

mmusgrov:
LRA RC1 provides baseline functionality that users can rely upon. This PR moves to a snapshot release and these builds only survive for 30 days.

Since the version change does not bring in any new functionality (it provides an alternative mechanism for triggering recovery) I would advise that we either schedule merging this PR when LRA is closer to RC2 (we still need to make some progress on MP-LRA issues 237, 214, 201, 6 and 208) or if those are not progressing well then we can do a Narayana build.

xstefank:
@mmusgrov fair enough. Do you mind creating a new branch here where we could cherry-pick PRs that are updating to the next version, please? This will help us moderate work and avoid conflicts with the other fixes. Also, I would suggest adding a label that will mark PRs that are good to be merged when the time is right (it will be easier for the merger to go through the line of PRs).

xstefank:
@tomjenkinson yes, that's up to us.

xstefank:
Just one comment, Narayana released I think 5.9.8.Final with MP LRA 1.0-RC1. Can't we just continue upstream development in master and integrate 5.9.8.Final (sorry if I got the version wrong) into Quarkus etc.?

xstefank:
@tomjenkinson My point is that I don't understand why we need to change the model we had prior to the RC1 and we need a way to combine changes because we are still adding and changing specification. So, for instance, each new LRA change must be based on this PR as it otherwise won't compile and another one like this (for already merged spec change is coming shortly). 

However, that being said, I agree that depending and releasing with snapshots is a bad idea as even now older versions of Narayana just won't compile as the snapshots are not available anymore. So I believe LRA feature branch would be the best solution (option 2) where we can depend on snapshots (master stays at RC1) and once we release spec RC2 we can fast-forward feature branch to master and update directly to RC2.

mmusgrov:
> @mmusgrov do you have some preference. I think that leaving this in a branch open ended it could get stale.
> 
> I would suggest either to:
> 
> 1. merge it now, one of us can track to get the an upgrade for the dependency in 29 days recurring or we request an RC2 from the community and wait to get that in.
> 2. Create another branch but then we need to have CI running on it periodically.
>    2.i. Perhaps we can do something with our pull request job to periodically retest them (it already rebases) in which case it could just stay in this "branch" - though that will not help with conflicting PR's
> 3. Request an RC2 and hold until we have that
> 
> I think adding 2.i will be a nice addition anyway.

The reason I don't want an RC2 is that we haven't added any functionality since RC1. The new PR only changes how we trigger recover under the covers. I said early that I want to see some movement on the issues before we consider a second candidate release.

And I have a similar reason for not wanting to move back to snapshots. When we have new features then that would be the time to move back onto the snapshot cycle.

But before moving back to snapshots we will cut a release of narayana that depends on RC1.

mmusgrov:
> IIUC then @mmusgrov you are going to 1) update to LRA RC1 on master 2) do a 5.10.0.Final release from master 3) merge this which moves back to snapshots on master?
> 
Master is already on RC1 and we have just released 5.10.0.Final.

> If 3) is not possible, it does seem like we will need a feature branch to continue work with LRA

that's reasonable (although we had that initially and it caused issues though I cannot recall the details).

> 
> I would be OK with creating a feature branch for LRA. We could then resync master with this branch when we are working with the LRA RC2 / GA then.

xstefank:
I won't really have time to get back to this for the next week or two but I don't want to get this conversation into a stale state :) Did we reach a resolution, please?

xstefank:
Sorry for the late response. I am very busy with conferences lately. I will open a new PR to LRA branch. Thanks @tomjenkinson.

PR #1514, opened 2019-10-12
JBTM-3210 recovery listener to be stuck

ochaloup:
https://issues.jboss.org/browse/JBTM-3210

Not being stuck with `IllegalArgument` exception during recovery and closing recovery listener connection when there is no plan of processing it.

MAIN XTS AS_TESTS
!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !PERF NO_WIN !RTS !TOMCAT !JACOCO !LRA

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/Listener.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/Listener.java
@@ -142,9 +142,11 @@ public class Listener extends Thread
             new_conn.start();
             }
          }
-         catch ( final InterruptedIOException ex )
+         catch ( final InterruptedIOException iioex )
          {
-            // timeout on the listener socket expired.
+             if (tsLogger.logger.isDebugEnabled()) {
+                 tsLogger.logger.debug("Timeout on the listener socket expired. Exception: " + iioex);
+             }
          }
          catch (final SocketException ex)
          {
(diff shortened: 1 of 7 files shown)


ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/jca/TransactionImporterImple.java
```diff
@@ -95,8 +95,9 @@ public SubordinateTransaction importTransaction(Xid xid, int timeout)
 	public TransactionImportResult importRemoteTransaction(Xid xid, int timeout)
 			throws XAException
 	{
-		if (xid == null)
-			throw new IllegalArgumentException();
+		if (xid == null) {
+			throw new IllegalArgumentException(jtaLogger.i18NLogger.get_error_imported_transaction_uid_is_null());
```
tomjenkinson: Perhaps this should be get_error_xid_is_null?

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/Listener.java
```diff
@@ -175,19 +175,28 @@ public synchronized boolean addConnection(Socket conn)
...
             return false;
         }
     }
 
     public synchronized void removeConnection(Socket conn)
     {
-        connections.remove(conn);
-        notifyAll();
+        try {
+            connections.remove(conn);
```
tomjenkinson: I think it would be clearer if the connections.remove was outside of this block as I don't think it will be throwing IOException, unless you think the notifyAll always done even with a problem on the linkedlist.remove() but if that is the case perhaps the conn.close() should be in it's own try block?

ochaloup: agree. I will move the `connections.remove` outside of the block as it better corresponds with the prior functionality.

ochaloup:
@tomjenkinson I updated the PR, can you please re-review when having time? Thanks!

PR #1515, opened 2019-10-14
JBTM-3189 Cancel LRAs as soon as their timeLimit is reached

mmusgrov:
https://issues.jboss.org/browse/JBTM-3189
https://issues.jboss.org/browse/JBTM-3211

The ultimate fix could be to wait for Java 11 HttpClient but this PR uses something similar (it replaces RestEasy client with Apache HttpClient). NB the coordinator uses an http client to contact participants and the server JAX-RS filter uses it to contact a coordinator.

### DIFF
--- a/rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
+++ b/rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
@@ -63,9 +63,6 @@ import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.client.Client;
-import javax.ws.rs.client.ClientBuilder;
-import javax.ws.rs.client.Entity;
-import javax.ws.rs.client.WebTarget;
 import javax.ws.rs.container.Suspended;
 import javax.ws.rs.core.Link;
 import javax.ws.rs.core.MediaType;
@@ -73,6 +70,8 @@ import javax.ws.rs.core.Response;
 
 import io.narayana.lra.Current;
 import io.narayana.lra.LRAConstants;
+import io.narayana.lra.RequestBuilder;
+import io.narayana.lra.ResponseHolder;
 import org.eclipse.microprofile.lra.annotation.Compensate;
 
 import io.narayana.lra.logging.LRALogger;
@@ -84,6 +83,7 @@ import org.eclipse.microprofile.lra.annotation.ws.rs.Leave;
 import org.eclipse.microprofile.lra.annotation.Status;
 
 import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_CONTEXT_HEADER;
+import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_RECOVERY_HEADER;
 
 /**
  * A utility class for controlling the lifecycle of Long Running Actions (LRAs) but the prefered mechanism is to use
@@ -111,11 +111,11 @@ public class NarayanaLRAClient implements Closeable {
 
     public static final long DEFAULT_TIMEOUT_MILLIS = 0L;
 
-    private static final String startLRAUrl = "/start";
+    private static final String startLRAUrl = "start";
 
-    private static final String confirmFormat = "/%s/close";
-    private static final String compensateFormat = "/%s/cancel";
-    private static final String leaveFormat = "/%s/remove";
+    private static final String confirmFormat = "%s/close";
+    private static final String compensateFormat = "%s/cancel";
+    private static final String leaveFormat = "%s/remove";
 
     private static final String LINK_TEXT = "Link";
 
@@ -206,15 +206,8 @@ public class NarayanaLRAClient implements Closeable {
         return lraId == null ? null : lraId.replaceFirst(".*/([^/?]+).*", "$1");
     }
 
-    private WebTarget getTarget() {
-        // return target; // TODO can't share the target if a service makes multiple JAX-RS requests
-        if (client != null) {
-            client.close(); // hacking
-        }
-
-        client = ClientBuilder.newClient();
-
-        return client.target(base);
+    private RequestBuilder getTarget() {
+        return new RequestBuilder(base);
     }
 
     public void setCurrentLRA(URI coordinatorUri) {
@@ -258,7 +251,7 @@ public class NarayanaLRAClient implements Closeable {
     }
 
     public URI startLRA(URI parentLRA, String clientID, Long timeout, ChronoUnit unit) throws WebApplicationException {
-        Response response = null;
+        ResponseHolder response = null;
         URI lra;
 
         if (clientID == null) {
@@ -278,28 +271,26 @@ public class NarayanaLRAClient implements Closeable {
         try {
             String encodedParentLRA = parentLRA == null ? "" : URLEncoder.encode(parentLRA.toString(), "UTF-8");
 
-            aquireConnection();
-
             response = getTarget().path(startLRAUrl)
                     .queryParam(TIMELIMIT_PARAM_NAME, Duration.of(timeout, unit).toMillis())
                     .queryParam(CLIENT_ID_PARAM_NAME, clientID)
                     .queryParam(PARENT_LRA_PARAM_NAME, encodedParentLRA)
                     .request()
-                    .post(Entity.text(""));
+                    .post();
 
             // validate the HTTP status code says an LRAInfo resource was created
             if (isUnexpectedResponseStatus(response, Response.Status.CREATED)) {
-                LRALogger.i18NLogger.error_lraCreationUnexpectedStatus(response.getStatus(), response);
+                LRALogger.i18NLogger.error_lraCreationUnexpectedStatus(response.getStatus(), response.toString());
                 throwGenericLRAException(null, INTERNAL_SERVER_ERROR.getStatusCode(),
                         "LRA start returned an unexpected status code: " + response.getStatus());
                 return null;
             }
 
             // validate that there is an LRAInfo response header holding the LRAInfo id
-            Object lraObject = Current.getLast(response.getHeaders().get(LRA_HTTP_CONTEXT_HEADER));
+            Object lraObject = response.getLastHeader(LRA_HTTP_CONTEXT_HEADER);
 
             if (lraObject == null) {
-                LRALogger.i18NLogger.error_nullLraOnCreation(response);
+                LRALogger.i18NLogger.error_nullLraOnCreation(response.toString());
                 throwGenericLRAException(null, INTERNAL_SERVER_ERROR.getStatusCode(), "LRA creation is null");
                 return null;
             }
@@ -311,7 +302,7 @@ public class NarayanaLRAClient implements Closeable {
             Current.push(lra);
 
         } catch (UnsupportedEncodingException e) {
-            LRALogger.i18NLogger.error_cannotCreateUrlFromLCoordinatorResponse(response, e);
+            LRALogger.i18NLogger.error_cannotCreateUrlFromLCoordinatorResponse(response.toString(), e);
             throwGenericLRAException(null, INTERNAL_SERVER_ERROR.getStatusCode(), e.getMessage());
             return null;
         } catch (Exception e) {
@@ -325,8 +316,6 @@ public class NarayanaLRAClient implements Closeable {
             }
 
             return null;
-        } finally {
-            releaseConnection(response);
         }
 
         // check that the lra is active
@@ -364,22 +353,16 @@ public class NarayanaLRAClient implements Closeable {
     }
 
     public void leaveLRA(URI lraId, String body) throws WebApplicationException {
-        Response response = null;
+        ResponseHolder response = null;
 
-        try {
-            aquireConnection();
+        response = getTarget().path(String.format(leaveFormat, getLRAId(lraId.toString())))
+                .request()
+                .header(LRA_HTTP_CONTEXT_HEADER, lraId)
+                .put(body, MediaType.TEXT_PLAIN);
 
-            response = getTarget().path(String.format(leaveFormat, getLRAId(lraId.toString())))
-                    .request()
-                    .header(LRA_HTTP_CONTEXT_HEADER, lraId)
-                    .put(Entity.entity(body, MediaType.TEXT_PLAIN));
-
-            if (Response.Status.OK.getStatusCode() != response.getStatus()) {
-                LRALogger.i18NLogger.error_lraLeaveUnexpectedStatus(response.getStatus(), response);
-                throwGenericLRAException(null, response.getStatus(), "");
-            }
-        } finally {
-            releaseConnection(response);
+        if (Response.Status.OK.getStatusCode() != response.getStatus()) {
+                LRALogger.i18NLogger.error_lraLeaveUnexpectedStatus(response.getStatus(), response.toString());
+            throwGenericLRAException(null, response.getStatus(), "");
         }
     }
 
@@ -530,8 +513,55 @@ public class NarayanaLRAClient implements Closeable {
         return 1;
     }
 
+/*
+    TODO delete when the TCK is updated
+
+ This code (which replaces getStatus below) has two issues:
+  - it incurs an unneccessary performance hit due to the extra network call.
+  - it does not guarantee that an LRA will be valid when the JAX-RS method executes since it can be
+ closed at any time during the method invocation. The only way to guarantee validity of an LRA during a method invocation is
+ to enlist with it and any attempt to join with an inactive LRA will fail with 412 and that is good enough.
+
+ Therefore remove the following invalid TCK tests:
+    TckLRATypeTests.mandatoryWithClosedLRA
+    TckLRATypeTests.requiredWithClosedLRA
+
+    then we can remove this method (the tracking issue is https://github.com/eclipse/microprofile-lra/issues/242)
+    */
+    public boolean isActive(URI lraId) {
+        ResponseHolder status = new RequestBuilder(lraId)
+                .request()
+                .header("Accept", MediaType.APPLICATION_JSON)
+                .get();
+
+        if (status.getStatus() == Response.Status.NOT_FOUND.getStatusCode()) {
+            return false;
+        } else {
+            String info = status.getResponseString();
+            String[] props = info.replaceAll("\"", "").replaceAll("}", "").split(",");
+            Map<String, String> nvps = new HashMap<>();
+            for (String nv : props) {
+                String[] nvp = nv.split(":");
+                nvps.put(nvp[0], nvp[1]);
+            }
+
+
+            boolean topLevel = Boolean.valueOf(nvps.get("topLevel"));
+            String lraStatus = nvps.get("status");
+
+            if (lraStatus.equals(LRAStatus.Active.name())) {
+                return true;
+            }
+
+            // a nested LRA can still be cancelled/closed even if it ahs closed/cancelled
+            // BTW there is no test that uses it so it is dead code
+            // still "effecively" active
+            return !topLevel && (lraStatus.equals(LRAStatus.Cancelled.name()) || lraStatus.equals(LRAStatus.Closed.name()));
+        }
+    }
+
     public LRAStatus getStatus(URI uri, boolean effectivelyActive) throws WebApplicationException {
-        Response response;
+        ResponseHolder response;
         URL lraId;
 
         try {
@@ -545,21 +575,11 @@ public class NarayanaLRAClient implements Closeable {
         }
 
         try {
-            aquireConnection();
-
-            WebTarget target = getTarget().path(getLRAId(lraId.toString())).path("status");
-
-            if (effectivelyActive) {
-                target = target.queryParam("effectivelyActive", true);
-            }
-
-            response = target
+            response = getTarget().path(getLRAId(lraId.toString())).path("status")
+                    .queryParam("effectivelyActive", effectivelyActive)
                     .request()
-                    .accept(MediaType.TEXT_PLAIN_TYPE)
                     .get();
         } catch (Exception e) {
-            releaseConnection(null);
-
             LRALogger.i18NLogger.error_cannotAccesCorrdinatorWhenGettingStatus(base, lraId, e);
             throwGenericLRAException(uri,
                     Response.Status.INTERNAL_SERVER_ERROR.getStatusCode(),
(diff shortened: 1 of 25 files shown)


rts/lra/lra-test/lra-test-basic/pom.xml
```diff
@@ -53,6 +59,8 @@
             <lra.coordinator.port>${lra.coordinator.port}</lra.coordinator.port>
           </systemPropertyVariables>
           <redirectTestOutputToFile>false</redirectTestOutputToFile>
+          <!-- temporarily disable these tests (and track it via a JBTM -->
```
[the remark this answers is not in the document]
mmusgrov: I updated the comment with the jira (JBTM-3211).
I also included an extra commit in the PR.

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
```diff
@@ -530,8 +513,55 @@ private static int checkMethod(Map<String, String> paths,
...
+  - it incurs an unneccessary performance hit due to the extra network call.
+  - it does not guarantee that an LRA will be valid when the JAX-RS method executes since it can be
+ closed at any time during the method invocation. The only way to guarantee validity of an LRA during a method invocation is
+ to enlist with it and any attempt to join with an inactive LRA will fail with 412 and that is good enough.
+
+ Therefore remove the following invalid TCK tests:
+    TckLRATypeTests.mandatoryWithClosedLRA
+    TckLRATypeTests.requiredWithClosedLRA
+
+    thne we can remove this method
```
tomjenkinson: Is that issue reported upstream?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -776,8 +776,7 @@ private void abortLRA() {
             // just update the status for now and wait for the external LRA close and cancel from there instead
             status = LRAStatus.Cancelling;
```
tomjenkinson: Are you certain this assumption is safe as it is just in memory state? What if another process happened to load the same LRA from disk? Would it be worth writing out this state change or can it be reliably calculated during rehydration?

mmusgrov: It can be reliably calculated at re-hydrate time. The relevant checks are:
https://github.com/jbosstm/narayana/blob/master/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java#L86

and

https://github.com/jbosstm/narayana/blob/master/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java#L252

I raised https://issues.jboss.org/browse/JBTM-3212 to make sure it is implemented correctly.

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
mmusgrov: I added https://github.com/eclipse/microprofile-lra/issues/242

rts/lra/lra-proxy/api/pom.xml
```diff
@@ -42,12 +42,14 @@
...
+<!--
         <dependency>
             <groupId>org.jboss.spec.javax.ws.rs</groupId>
-            <artifactId>jboss-jaxrs-api_2.0_spec</artifactId>
-            <version>${version.jaxrs.api}</version>
+            <artifactId>jboss-jaxrs-api_2.1_spec</artifactId>
+            <version>${version.jaxrs-api}</version>
             <scope>provided</scope>
         </dependency>
+-->
```
xstefank: This is commented out. Is it intentional?

rts/lra/lra-service-base/src/main/java/io/narayana/lra/LRAHttpClient.java
```diff
@@ -0,0 +1,64 @@
...
+public class LRAHttpClient {
+    public static final long PARTICIPANT_TIMEOUT = 1; // number of seconds to wait for requests
+
+    private static LRAHttpClient client = new LRAHttpClient();
+
+    private FutureRequestExecutionService futureRequestExecutionService;
+
+    private LRAHttpClient() {
+        HttpClient httpClient = HttpClientBuilder.create().setMaxConnPerRoute(5).build();
+        ExecutorService executorService = Executors.newFixedThreadPool(5);
```
xstefank: just an idea: maybe we can make these values (maxConnPerRoute and fixedThreadPool) configurable in the future?

rts/lra/lra-service-base/src/main/java/io/narayana/lra/LRAHttpClient.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public static LRAHttpClient getClient() {
+        return client;
+    }
+
+    public ResponseHolder asyncRequest(HttpRequestBase httpMethod, long timelimit, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
+        HttpRequestFutureTask<ResponseHolder> futureTask = submitFutureTask(httpMethod);
+
+        return futureTask.get(timelimit, unit);
```
xstefank: this seems to be a blocking call, maybe you can rename the method to just `request`? This seems confusing from the external point of view.

rts/lra/lra-service-base/src/main/java/io/narayana/lra/ResponseHolder.java
```diff
@@ -0,0 +1,84 @@
...
+            return null;
+        }
+    }
+
+    public boolean hasEntity() {
+        return responseString != null;
+    }
+
+    public String getLastHeader(String headerName) {
+        return httpResponse.getLastHeader(headerName).getValue();
```
xstefank: possible NPE?

rts/lra/lra-service-base/src/main/java/io/narayana/lra/ResponseHolder.java
```diff
@@ -0,0 +1,84 @@
...
+    public String getResponseString() {
+        return responseString;
+    }
+
+    public URI getLocationHeaderAsURI() {
+        try {
+            return locationHeader != null ? new URI(locationHeader) : null;
+        } catch (URISyntaxException e) {
+            if (LRALogger.logger.isInfoEnabled()) {
+                LRALogger.logger.infof("LRARecord.doEnd missing Location header on ACCEPTED response %s failed: %s",
```
xstefank: This error seems out of context in such generic class as ResponseHolder.

rts/lra/lra-service-base/src/main/java/io/narayana/lra/ResponseHolder.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public boolean hasEntity() {
+        return responseString != null;
+    }
+
+    public String getLastHeader(String headerName) {
+        return httpResponse.getLastHeader(headerName).getValue();
+    }
+
+    public String readEntity(Class<String> stringClass) {
```
xstefank: parameter is not used

rts/lra/lra-service-base/src/main/java/io/narayana/lra/ResponseHolder.java
```diff
@@ -0,0 +1,84 @@
...
+        HttpEntity entity = httpResponse.getEntity();
+
+        return entity.toString();
```
xstefank: possible NPE?

rts/lra/lra-test/lra-test-basic/pom.xml
```diff
@@ -53,6 +59,8 @@
             <lra.coordinator.port>${lra.coordinator.port}</lra.coordinator.port>
           </systemPropertyVariables>
           <redirectTestOutputToFile>false</redirectTestOutputToFile>
+          <!-- temporarily disable these tests (and track it via a JBTM-3211 -->
+          <skipTests>true</skipTests>
```
xstefank: Can you elaborate a little on why these tests need to be temporarily disabled? Please be curious to not release new Narayana version with these tests disabled as it may now already introduce regressions.

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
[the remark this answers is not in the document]
mmusgrov: It is already there (I linked the issue number in response to Tom's query)

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
```diff
@@ -530,8 +513,55 @@ private static int checkMethod(Map<String, String> paths,
...
+            boolean topLevel = Boolean.valueOf(nvps.get("topLevel"));
+            String lraStatus = nvps.get("status");
+
+            if (lraStatus.equals(LRAStatus.Active.name())) {
+                return true;
+            }
+
+            // a nested LRA can still be cancelled/closed even if it ahs closed/cancelled
+            // BTW there is no test that uses it so it is dead code
+            // still "effecively" active
```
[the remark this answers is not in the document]
mmusgrov: don't know but I added a break point and it never fired.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -776,8 +776,7 @@ private void abortLRA() {
             // just update the status for now and wait for the external LRA close and cancel from there instead
```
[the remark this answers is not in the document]
mmusgrov: okay

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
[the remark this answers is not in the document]
xstefank: yeah, sorry, I was writing review whole day and missed the last update.

rts/lra/lra-proxy/api/pom.xml
mmusgrov: okay

rts/lra/lra-test/lra-test-basic/pom.xml
tomjenkinson: @xstefank it is described in https://issues.jboss.org/browse/JBTM-321

@mmusgrov thanks for the link - would this be better on the [LRA](https://github.com/jbosstm/narayana/tree/LRA) feature branch unless we could release without it?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
tomjenkinson: Thanks

rts/lra/lra-service-base/src/main/java/io/narayana/lra/LRAHttpClient.java
tomjenkinson: I agree with @xstefank - unless there is a good reason for them to be fixed I think they should be configurable. Is there an LRAEnvironmentBean, if not please can you raise a JBTM to add one and mention these values so we don't miss them

mmusgrov: Perhaps in the future but let's wait for when we add metrics and non functional testing

rts/lra/lra-service-base/src/main/java/io/narayana/lra/LRAHttpClient.java
tomjenkinson: Interesting observation, I expect the request itself is async. requestWithTimeout might be a more accurate name? Although I imagine the request itself is not cancelled so the request will be served at some point on the server side.

@mmusgrov what assumptions are made when the timelimit is expired but the request has not returned yet?

mmusgrov: Okay

rts/lra/lra-service-base/src/main/java/io/narayana/lra/ResponseHolder.java
mmusgrov: I don't think so. Only a non spec compliant (or malicious) coordinator would omit the header.

rts/lra/lra-service-base/src/main/java/io/narayana/lra/ResponseHolder.java
tomjenkinson: Good catch. I also just spotted that a few other log messages in tryDoEnd are attribued to this. It will be clearer to just remove the LRARecord.doEnd prefix I think.

rts/lra/lra-test/lra-test-basic/pom.xml
xstefank: @tomjenkinson JBTM-3211 doesn't go into details neither. I am just curious about what isn't working.

rts/lra/lra-service-base/src/main/java/io/narayana/lra/ResponseHolder.java
mmusgrov: Okay

rts/lra/lra-service-base/src/main/java/io/narayana/lra/ResponseHolder.java
mmusgrov: okay

rts/lra/lra-service-base/src/main/java/io/narayana/lra/ResponseHolder.java
mmusgrov: Good catch

rts/lra/lra-test/lra-test-basic/pom.xml
mmusgrov: Okay I have marked it as a blocker

rts/lra/lra-service-base/src/main/java/io/narayana/lra/LRAHttpClient.java
mmusgrov: So to track it I created JBTM-3213

rts/lra/lra-service-base/src/main/java/io/narayana/lra/LRAHttpClient.java
mmusgrov: @tomjenkinson the request may well be still in progress but the caller is responsible for retrying the request - we make no assumptions about once only semantics so this is not an issue.

rts/lra/lra-test/lra-test-basic/pom.xml
mmusgrov: It should be easy to fix (I was mainly running against the TCK and I discovered the failure when I did a full build yesterday morning) but I wanted to get some eyes on the PR. But I don't see why it needs to block this PR (after all NarayanaLRAClient effectively became internal after we pulled the java client API from the spec). I will fix it now if it is a problem.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
tomjenkinson: Would you agree that https://issues.jboss.org/browse/JBTM-3212 is a blocker? Otherwise this process could be deciding to cancel the LRA and a second process that starts and loads from the object store might decide the complete the transaction?

rts/lra/lra-service-base/src/main/java/io/narayana/lra/LRAHttpClient.java
tomjenkinson: Thanks, that works for me

rts/lra/lra-service-base/src/main/java/io/narayana/lra/LRAHttpClient.java
tomjenkinson: Thanks

rts/lra/lra-test/lra-test-basic/pom.xml
tomjenkinson: I agree, if NarayanaLRAClient is internal it seems fine to me to leave it on this branch. (I assume we don't have quickstarts for NarayanaLRAClient still?)

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
tomjenkinson: Thanks, Mike, I think that should be added into the comment in the code as you did with JBTM-3211 below

rts/lra/lra-test/lra-test-basic/pom.xml
xstefank: Agreed, no need to block this PR for it. I just wanted to bring it up as I am not sure whether for instance, Helidon exposes this functionality. And there are also tests for non-root JAX-RS participants IIRC.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
mmusgrov: Well it is no different from the current code but I agree it is a blocker - I will fix it.

rts/lra/lra-test/lra-test-basic/pom.xml
mmusgrov: I took a look and found the issue. The new RequestBuilder class I added was buggy (calling builder.path(p1).path(p2) didn't append p2 correctly). So thanks to both of you for picking up on this issue.

Also Matej Kralik is using our client so it would be a bit rude if we didn't fix asap

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
mmusgrov: Done

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
mmusgrov: But I think it is better if I fix it in a separate PR. Is that okay @tomjenkinson

mmusgrov:
@tomjenkinson and @xstefank I have pushed fixes for your comments and the PR is ready for another review.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
tomjenkinson: Thanks Mike

rts/lra/lra-service-base/src/main/java/io/narayana/lra/RequestBuilder.java
```diff
@@ -0,0 +1,163 @@
...
+    private URIBuilder builder;
+    private boolean async;
+    private long timeout;
+    private TimeUnit timeUnit;
+
+    public RequestBuilder(URI baseURI) {
+        builder = new URIBuilder(baseURI);
+        paths = new ArrayList<>();
+
+        Collections.addAll(paths, strip(strip(baseURI.getPath())).split("/"));
```
xstefank: Nested strip invocation seems unnecessary. But maybe I am missing something?

rts/lra/lra-service-base/src/main/java/io/narayana/lra/RequestBuilder.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    public RequestBuilder async(long timeout, TimeUnit timeUnit) {
+        this.async = true;
+        this.timeout = timeout;
+        this.timeUnit = timeUnit;
+        return this;
+    }
+
+    private HttpRequestBase setContent(HttpEntityEnclosingRequestBase req, String content, String mediaType) {
```
xstefank: mediaType is ignored. This seems to be important.

rts/lra/lra-service-base/src/main/java/io/narayana/lra/RequestBuilder.java
mmusgrov: okay

rts/lra/lra-service-base/src/main/java/io/narayana/lra/RequestBuilder.java
mmusgrov: We always call it with MediaType.TEXT_PLAIN. But I want to keep the API calls as close as possible to what we had before. If you like I can check for an unsupported media type by replacing it with:

```java
    private HttpRequestBase setContent(HttpEntityEnclosingRequestBase req, String content, String mediaType) {
        if (content != null) {
            if (!mediaType.equals(MediaType.TEXT_PLAIN)) {
                throw new IllegalArgumentException("RequestBuilder: only content type mediaType is supported");
            }
            req.setEntity(new StringEntity(content, HTTP.DEF_CONTENT_CHARSET));
        }

        return req;
    }
```

mmusgrov:
@xstefank if you have no more change requests I will push a commit for those last two comments?

rts/lra/lra-service-base/src/main/java/io/narayana/lra/RequestBuilder.java
xstefank: What API are you referring to, please? I think the cleanest would be to remove the parameter if it's not being used as it may lead to false assumptions if someone won't read the body of the method in the future. But it's not that critical so if you want to keep it this way we can refactor later.

xstefank:
@mmusgrov no, with these changes it's ok from my side and I can approve.

rts/lra/lra-service-base/src/main/java/io/narayana/lra/RequestBuilder.java
mmusgrov: There are multiple HTTP client implementations:

-  java.net.http.HttpClient (only available in JDK 11)
- resteasy client API
- apache HttpClient (which resteasy uses under the covers)

Let's keep the interface to the HTTP client as generic as possible until we make the move to JDK 11.

mmusgrov:
> @mmusgrov no, with these changes it's ok from my side and I can approve.

Thanks for you reviews @tomjenkinson and @xstefank - you both suggested some good improvements.

PR #1518, opened 2019-10-15
JBTM-2867 JBTM-3195 Synchronize access to the state

tomjenkinson:
https://issues.jboss.org/browse/JBTM-2867
https://issues.jboss.org/browse/JBTM-3195

MAIN XTS !AS_TESTS !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !PERF NO_WIN !RTS !TOMCAT JACOCO !LRA !XTS

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/CacheStore.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/CacheStore.java
@@ -738,11 +738,11 @@ class AsyncStore extends Thread // keep priority same as app. threads
                 }
                 catch (ObjectStoreException ex)
                 {
-                    ex.printStackTrace();
+                    tsLogger.i18NLogger.warn_could_not_handle_objectstore(ex);
                 }
                 catch (Exception ex)
                 {
-                    ex.printStackTrace();
+                    tsLogger.i18NLogger.warn_could_not_handle_objectstore_generic(ex);
                 }
             }
         }
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/CacheStore.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/CacheStore.java
@@ -344,15 +344,18 @@ class AsyncStore extends Thread // keep priority same as app. threads
         StoreElement toAdd = new StoreElement(store, workType, objUid, tName,
                 state, ft);
 
-        getList(objUid).addFirst(toAdd);
 
-        synchronized (_workList)
-        {
-            if (state != null)
-                _currentCacheSize += state.size();
+            LinkedList list = getList(objUid);
+            synchronized (list) {
+                list.addFirst(toAdd);
+            }
+            synchronized (_workList) {
 
-            _numberOfEntries++;
-        }
+                if (state != null)
+                    _currentCacheSize += state.size();
+
+                _numberOfEntries++;
+            }
 
         return true;
     }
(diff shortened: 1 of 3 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/CacheStore.java
```diff
@@ -460,84 +464,84 @@ public final int removeWriteState (Uid objUid, int ft)
...
-        {
-            elements = getList(objUid).toArray();
-        }
-        catch (Exception ex)
-        {
-            elements = null;
+        LinkedList list = getList(objUid);
+        synchronized (list) {
+            try {
+                elements = list.toArray();
```
ochaloup: Would not be better to check the "wrong" state in a different way than catching an `Exception`?

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/CachedTest.java
```diff
@@ -71,11 +71,17 @@ public void run()
                 Thread.yield();
                 
                 if (s != null) {
-                    if (participantStore.remove_committed(u, TYPE))
+                    if (participantStore.remove_committed(u, TYPE)) {
                         passed = true;
+                    } else {
+                        System.err.println("Could not remove "+u);
```
ochaloup: I think logger should be used instead of the `System.err.println`

tomjenkinson: I would agree if it were a statement I was introducing.

That statement was already in the code and not actually added by this patch. I will look to add a second commit against https://issues.jboss.org/browse/JBTM-2880

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/CacheStore.java
```diff
@@ -373,63 +376,64 @@ public final int removeState (Uid objUid, int ft)
...
+            } catch (Exception ex) {
+                ex.printStackTrace();
```
tomjenkinson: @ochaloup This one is a new stack trace, I will change it in this commit

tomjenkinson: I have changed this one now

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/CachedTest.java
tomjenkinson: I just realised, this is in the test sorry. I think it is fine to keep it as it is

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/CacheStore.java
tomjenkinson: I wanted to keep the patch as close to the original intention as I could. I do not actually expect an exception here anymore. What I saw was that the list through an array out of bounds exception but I believe that was because the list.toArray was happening concurrent with a list.addFirst (IIRC) - we could likely remove the catch entirely now.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/CachedTest.java
ochaloup: right, I missed. sorry.

PR #1519, opened 2019-10-21
JBTM-3204 Upgrade LRA TCK recovery handling to accommodate specificat…

xstefank:
https://issues.jboss.org/browse/JBTM-3204

### DIFF
--- a/rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
+++ b/rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
@@ -1,0 +1,70 @@
+package io.narayana.lra.arquillian.spi;
+
+import io.narayana.lra.LRAConstants;
+import io.narayana.lra.client.NarayanaLRAClient;
+import org.eclipse.microprofile.lra.tck.service.spi.LraRecoveryService;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.ClientBuilder;
+import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.Response;
+import java.net.URI;
+
+@ApplicationScoped
+public class NarayanaLRARecovery implements LraRecoveryService {
+
+    @Override
+    public void triggerRecovery() {
+        recoverLRAs(
+            System.getProperty(NarayanaLRAClient.LRA_COORDINATOR_HOST_KEY, "localhost"),
+            Integer.parseInt(System.getProperty(NarayanaLRAClient.LRA_COORDINATOR_PORT_KEY, "8080")));
+    }
+
+    @Override
+    public void triggerRecovery(URI lraId) {
+        String host = lraId.getHost();
+        int port = lraId.getPort();
+        if (!recoverLRAs(host, port, lraId)) {
+            // first recovery scan probably collided with periodic recovevery which started
+            // before the test execution so try once more
+            recoverLRAs(host, port, lraId);
+        }
+    }
+
+    /**
+     * Invokes LRA coordinator recovery REST endpoint and returns whether the recovery of intended LRAs happended
+     *
+     * @param host the LRA coordinator host address
+     * @param port the LRA coordinator port
+     * @param lraIds the LRA ids of the LRAs that are intended to be recovered
+     * @return true all intended LRAs recovered, false otherwise
+     */
+    private boolean recoverLRAs(String host, int port, URI... lraIds) {
+        // trigger a recovery scan
+        Client recoveryCoordinatorClient = ClientBuilder.newClient();
+
+        try {
+            String recoveryCoordinatorUrl = String.format("http://%s:%d/%s/recovery",
+                host, port, LRAConstants.RECOVERY_COORDINATOR_PATH_NAME);
+            WebTarget recoveryTarget = recoveryCoordinatorClient.target(URI.create(recoveryCoordinatorUrl));
+
+            // send the request to the recovery coordinator
+            Response response = recoveryTarget.request().get();
+            String json = response.readEntity(String.class);
+            response.close();
+
+            for (URI lraId : lraIds) {
+                if (json.contains(lraId.toASCIIString())) {
+                    // intended LRA didn't recover
+                    return false;
+                }
+            }
+
+            return true;
+        } finally {
+            recoveryCoordinatorClient.close();
+        }
+
+    }
+}
--- a/rts/lra/lra-test/lra-test-arquillian-extension/pom.xml
+++ b/rts/lra/lra-test/lra-test-arquillian-extension/pom.xml
@@ -21,6 +21,11 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
+      <groupId>org.eclipse.microprofile.lra</groupId>
+      <artifactId>microprofile-lra-tck</artifactId>
+      <version>${version.microprofile.lra.tck}</version>
+    </dependency>
+    <dependency>
       <groupId>org.jboss.narayana.rts</groupId>
       <artifactId>narayana-lra</artifactId>
       <version>${project.version}</version>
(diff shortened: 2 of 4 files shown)


xstefank:
@tomjenkinson @mmusgrov I changed the condition to actually check if the LRA is present (it was wrong) however, testForget will always fail because in the first status invocations check the LRA is still active (returning 500) so we will invoke the recovery twice (correct behavior) but the test expects only one invocation. I've submitted PR to the specification to fix this.

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
```diff
@@ -0,0 +1,67 @@
...
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.ClientBuilder;
+import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.Response;
+import java.net.URI;
+
+@ApplicationScoped
+public class NarayanaLRARecovery implements LraRecoveryService {
```
mmusgrov: Can't we change the camel case to LRARecoveryService

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
```diff
@@ -0,0 +1,67 @@
...
+
+        try {
+            String recoveryCoordinatorUrl = String.format("http://%s:%d/%s/recovery",
+                host, port, LRAConstants.RECOVERY_COORDINATOR_PATH_NAME);
+            WebTarget recoveryTarget = recoveryCoordinatorClient.target(URI.create(recoveryCoordinatorUrl));
+
+            // send the request to the recovery coordinator
+            Response response = recoveryTarget.request().get();
+            String json = response.readEntity(String.class);
+            LRALogger.logger.error(lraId);
```
mmusgrov: What is the error here.

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
```diff
@@ -0,0 +1,67 @@
...
+            LRALogger.logger.error(json);
```
mmusgrov: Why is there a second error printed.

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
xstefank: forgotten test logs, sorry

xstefank:
@mmusgrov ready for review again. Please note that snapshot will probably no longer be available. Can you please merge this one and I will work asap on bringing the LRA branch up to date.

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
mmusgrov: What is the decision here? It looks like you changed something but the class name is still  NarayanaLraRecovery

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
```diff
@@ -0,0 +1,67 @@
...
+            response.close();
+
+            return lraId != null && json.contains(lraId.toASCIIString());
```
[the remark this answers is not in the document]
mmusgrov: The code still looks obscure and a reader would still like to know:

1. If you call recovery without an LRA what is the meaning of returning false from this method.
2. If you call recovery without an LRA why do you need to call recovery twice.

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
xstefank: I must have misunderstood you. LraRecoveryService is coming from spec, I will revert this back to NarayanaLRAREcovery then and change the spec later in a single sweep over all classes.

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLraRecovery.java
```diff
@@ -0,0 +1,64 @@
...
+    @Override
+    public void triggerRecovery(URI lraId) {
+        invokeRecovery(lraId, lraId.getHost(), lraId.getPort());
+    }
+
+    private void invokeRecovery(URI lraId, String host, int port) {
+        if (invokeRecoveryCoordinatorEndpoint(lraId, host, port)) {
+            // first recovery scan probably collided with periodic recovevery which started
+            // before the test execution so try once more
+            invokeRecoveryCoordinatorEndpoint(null, host, port);
```
mmusgrov: Why is the first invocation made with a non null LRA id and the second one uses a null one. I think both invocations need the LRA id to be passed in.

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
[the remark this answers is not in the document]
xstefank: ok, I will pass it to both invocations and just ignore the return value in the second call.

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
mmusgrov: Changing the spec might be more of a problem. But I would at least prefer our implementation to be consistent in how we interpret Camel Case.

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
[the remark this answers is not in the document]
mmusgrov: The javadoc says "true if the expected LRA recovered" which does not reflect what the code does. This makes the code in `invokeRecovery(URI lraId, String host, int port)` look wrong.

I think it's best to avoid the overloading of the calls to `invokeRecoveryCoordinatorEndpoint` since that is what is making the code appear confusing. Just call the `invokeRecoveryCoordinatorEndpoint` method from both of the interface methods - that way it will be clear to the reader what the intention is.

Also the interface methods should be responsible for deciding whether or not to trigger recovery. The `triggerRecovery(URI lraId)` can return a boolean to indicate whether or not lraId was recovered and the `triggerRecovery()` should either return a void or a count of recovering LRAs or a list of them. But my preference would be to return a void.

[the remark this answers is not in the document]
mmusgrov: Or returning count has the benefit of letting the caller know whether there is anything left to recover which is more useful that void.

xstefank:
@mmusgrov changed to direct invocations + fixed javadoc. TCK is still passing locally.

PR #1521, opened 2019-10-23
JBTM-3214 Add a method to the BeanPopulator to allow a bean to be set…

mmusgrov:
https://issues.jboss.org/browse/JBTM-3214

MAIN XTS !AS_TESTS !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !PERF NO_WIN !RTS !TOMCAT JACOCO !LRA !XTS

### DIFF
--- a/common/classes/com/arjuna/common/internal/util/propertyservice/BeanPopulator.java
+++ b/common/classes/com/arjuna/common/internal/util/propertyservice/BeanPopulator.java
@@ -174,6 +174,28 @@ public class BeanPopulator
         }
     }
 
+    /**
+     * Attempt to set a named bean instance.
+     *
+     * [Please note that this method is not guaranteed to exist longer term since it is provided in an internal API.
+     * In a later release a public API will be provided to support similar behaviour].
+     *
+     * If the name is not already associated with a bean instance (or is mapped to null) then the name will be associated
+     * with the supplied bean and the method will return null, else the method returns the current instance
+     * associated with the name.
+     *
+     * Once a bean is associated with a name any other configuration mechanism (such as XML property files and
+     * system properties) will be ignored. Therefore this method can be useful if the caller wishes to avoid the
+     * need to parse XML property files.
+     *
+     * @param name the name of the instance to set
+     * @param beanInstance the instance to be associated with the name
+     * @return the previous instance associated with the specified name, or null if there was no mapping for the name.
+     */
+    public static Object setBeanInstanceIfAbsent(String name, Object beanInstance) {
+        return beanInstances.putIfAbsent(name, beanInstance);
+    }
+
     public static String printBean(Object bean) {
         StringBuffer buffer = new StringBuffer();
         printBean(bean, buffer);


mmusgrov:
> Actually I have few questions about enhancing the API this way.
>
> * the JBTM-3214 says `code avoids the use of reflection (and possibly xml parsing)`. Here we talk about reflection to access the bean setters based on the property name, right? I'm just interested if we know much slower is the reflection against the direct access. Is it the part which slows down the startup?

250 % faster.

> * Avoiding the xml parsing is mentioned as questionable. I think it's right that the XML parsing i started only when the `getDefaultInstance` is called. But it needs to be ensured that all beans are set up before such call is invoked.

The additional method avoids the need to parse a property file when the bean is set directly using the new method because the bean will be in the map.

> * what is idea about setting up the bean. It's about to create the bean with defaults defined in the Narayana code base and not taken in to account the `jbossts-properties.xml`? How the user may then change the default behaviour of the Narayana?

Take a look at the linked JIRA

> * If that would be something we do not consider necessary should be the `jbossts-properties.xml` part of the published Narayana artifact? I mean defaults will be defined in codebase from now.

The only change is that a external framework such as quarkus can set the beans directly and in this case the defaults in the jbossts-properties.xml will not be used. That is the point of the patch - quarkus provides its own defaults.

> * And that's which makes me thinking if we should not redesign a little bit the properties processing. When this is a trouble for some environments what about a different approach? I think the API abstraction should be clear and coupled while easy to extend. Adding this method makes a way to change the things from the middle.

No this is not a refactor - it is a change to simplify usage by external frameworks.

>   ** What about to define somewhere at the top two strategies. One for standaloe where properties may be configured with `xml` descriptor and the second which will returns just bare bone beans without any configuration done and it will be up to the integrator to change them?

This PR is not going to redesign or refactor the existing properties mechanism, rather it provides a mechanism for external frameworks to avoid our use of reflection in the parsing code.

>
> Two points about the code
>
> * the call uses `beanInstances.putIfAbsent` and the name of the method should mention this `ifAbsent` as well

It is a setter and normally we don't add this kind of verbosity to method names. At most such a comment would apply to javadoc comments.

> * I think the methods are ordered in purpose from `public` ones to `private` ones. It would be good to do the reordering and put this method to the `public` part of the class.

I am not aware of any such recommendation in our style guidelines.

mmusgrov:
If it is a refactor or a redesign of our properties config mechanism that you are wanting then that is a valid thing to do for version 6 and it should be addressed as a whole. This PR is satisfying an immediate need and should not have to wait for a more expansive approach if indeed our priorities would allow that.

Your PR comments would be better moved to a separate JIRA so can we do that so that we can move on with this PR?

ochaloup:
> > Actually I have few questions about enhancing the API this way.
> > 
> > * the JBTM-3214 says `code avoids the use of reflection (and possibly xml parsing)`. Here we talk about reflection to access the bean setters based on the property name, right? I'm just interested if we know much slower is the reflection against the direct access. Is it the part which slows down the startup?
> 
> 250 % faster.
Ok, I see, great.

> 
> > * Avoiding the xml parsing is mentioned as questionable. I think it's right that the XML parsing i started only when the `getDefaultInstance` is called. But it needs to be ensured that all beans are set up before such call is invoked.
> 
> The additional method avoids the need to parse a property file when the bean is set directly using the new method because the bean will be in the map.
Agree. Just for the xml parsing would not be applied then all the beans have to be setup correctly. If a bean is omitted to be set then the xml parsing is processed.

> 
> > * what is idea about setting up the bean. It's about to create the bean with defaults defined in the Narayana code base and not taken in to account the `jbossts-properties.xml`? How the user may then change the default behaviour of the Narayana?
> 
> Take a look at the linked JIRA
I'm sorry but I can't find any information in the linked jira. What if the user wants to change the recovery interval?

> 
> > * If that would be something we do not consider necessary should be the `jbossts-properties.xml` part of the published Narayana artifact? I mean defaults will be defined in codebase from now.
> 
> The only change is that a external framework such as quarkus can set the beans directly and in this case the defaults in the jbossts-properties.xml will not be used. That is the point of the patch - quarkus provides its own defaults.
I see the point, agree.

> 
> > * And that's which makes me thinking if we should not redesign a little bit the properties processing. When this is a trouble for some environments what about a different approach? I think the API abstraction should be clear and coupled while easy to extend. Adding this method makes a way to change the things from the middle.
> 
> No this is not a refactor - it is a change to simplify usage by external frameworks.
When refactoring is done it's a good chance to make it clean from the API user perspective. Some more changes could not harm the idea of refactoring and providing functionality.

> 
> > ** What about to define somewhere at the top two strategies. One for standalone where properties may be configured with `XML` descriptor and the second which will returns just bare bone beans without any configuration done and it will be up to the integrator to change them?
> 
> This PR is not going to redesign or refactor the existing properties mechanism, rather it provides a mechanism for external frameworks to avoid our use of reflection in the parsing code.
I see. The redesign is the wrong word. I mean having a clear mechanism to define how the configuration is defined. For external the externals providers (which can be tomcat, wfly, quarkus...) we can use only programmatic way and not permitting the xml to be parsed. For the standalone usage, the way to configure will be the xml file.

> 
> > Two points about the code
> > 
> > * the call uses `beanInstances.putIfAbsent` and the name of the method should mention this `ifAbsent` as well
> 
> It is a setter and normally we don't add this kind of verbosity to method names. At most such a comment would apply to javadoc comments.
I think this is about good naming and API comprehensibility.

> 
> > * I think the methods are ordered in purpose from `public` ones to `private` ones. It would be good to do the reordering and put this method to the `public` part of the class.
> 
> I am not aware of any such recommendation in our style guidelines.
Yes, this is about a style the class started and about hints used in Java community.

common/classes/com/arjuna/common/internal/util/propertyservice/BeanPopulator.java
```diff
@@ -418,4 +418,8 @@ private static boolean isPositive(String value) {
     private static boolean isNegative(String value) {
         return "NO".equalsIgnoreCase(value) || "OFF".equalsIgnoreCase(value) || "FALSE".equalsIgnoreCase(value);
     }
+
+    public static void setBeanInstance(String name, Object beanInstance) {
+        beanInstances.putIfAbsent(name, beanInstance);
```
tomjenkinson: Taking into account the comment from @ochaloup , I think if this is not going to force set the API and as it is an internal API I would expect an exception or boolean to report if the bean was set.

ochaloup:
> > > > Actually I have few questions about enhancing the API this way.
> > > > 
> > > > * the JBTM-3214 says `code avoids the use of reflection (and possibly xml parsing)`. Here we talk about reflection to access the bean setters based on the property name, right? I'm just interested if we know much slower is the reflection against the direct access. Is it the part which slows down the startup?
> > > 
> > > 
> > > 250 % faster.
> > 
> > 
> > Ok, I see, great.
> > > > * Avoiding the xml parsing is mentioned as questionable. I think it's right that the XML parsing i started only when the `getDefaultInstance` is called. But it needs to be ensured that all beans are set up before such call is invoked.
> > > 
> > > 
> > > The additional method avoids the need to parse a property file when the bean is set directly using the new method because the bean will be in the map.
> > 
> > 
> > Agree. Just for the xml parsing would not be applied then all the beans have to be setup correctly. If a bean is omitted to be set then the xml parsing is processed.
> 
> That is not correct. If any bean is not set directly then the other strategies are used. Calling the new method will only by pass the the other mechanisms for the particular bean.

 What I understand under the point `If any bean is not set directly then the other strategies are used.` is that if the bean is not directly defined then the XML is parsed, env and system variables are considered. Or do you mean something else here?

> 
> > > > * what is idea about setting up the bean. It's about to create the bean with defaults defined in the Narayana code base and not taken in to account the `jbossts-properties.xml`? How the user may then change the default behaviour of the Narayana?
> > > 
> > > 
> > > Take a look at the linked JIRA
> > 
> > 
> > I'm sorry but I can't find any information in the linked jira. What if the user wants to change the recovery interval?
> 
> The config mechanisms are still in operation unless the caller sets one of the beans. So to change the recovery interval he either sets an instance of RecoveryEnvironmentBean or he just uses the existing mechanism (xml properties or system properties).

 Ok, if I understand the point then until the Quarkus provides the `Config` property for the recovery interval there is no way to change it for the Quarkus deployment. That's fair.

> 
> > > > * If that would be something we do not consider necessary should be the `jbossts-properties.xml` part of the published Narayana artifact? I mean defaults will be defined in codebase from now.
> > > 
> > > 
> > > The only change is that a external framework such as quarkus can set the beans directly and in this case the defaults in the jbossts-properties.xml will not be used. That is the point of the patch - quarkus provides its own defaults.
> > 
> > 
> > I see the point, agree.
> > > > * And that's which makes me thinking if we should not redesign a little bit the properties processing. When this is a trouble for some environments what about a different approach? I think the API abstraction should be clear and coupled while easy to extend. Adding this method makes a way to change the things from the middle.
> > > 
> > > 
> > > No this is not a refactor - it is a change to simplify usage by external frameworks.
> > 
> > 
> > When refactoring is done it's a good chance to make it clean from the API user perspective. Some more changes could not harm the idea of refactoring and providing functionality.
> 
> I have not done any refactoring. I have provided a single method that allows the caller to set one of our beans.
> 
> > > > ** What about to define somewhere at the top two strategies. One for standalone where properties may be configured with `XML` descriptor and the second which will returns just bare bone beans without any configuration done and it will be up to the integrator to change them?
> > > 
> > > 
> > > This PR is not going to redesign or refactor the existing properties mechanism, rather it provides a mechanism for external frameworks to avoid our use of reflection in the parsing code.
> > 
> > 
> > I see. The redesign is the wrong word. I mean having a clear mechanism to define how the configuration is defined. For external the externals providers (which can be tomcat, wfly, quarkus...) we can use only programmatic way and not permitting the xml to be parsed. For the standalone usage, the way to configure will be the xml file.
> 
> I said earlier: "If it is a refactor or a redesign of our properties config mechanism that you are wanting then that is a valid thing to do for version 6 and it should be addressed as a whole. This PR is satisfying an immediate need and should not have to wait for a more expansive approach if indeed our priorities would allow that.

 My sentiment is that every change should not be breaking what was already done.
 My observation of the current code is that the API of the `BeanPopulator` intentionally did not provide a method like set `setBeanInstance` as the overall handling of factories below should provide the bean.
 That's what I mean when I said it accesses the concept in the middle instead of changing it.

> 
> Your PR comments would be better moved to a separate JIRA so can we do that so that we can move on with this PR?
> 
> > > > Two points about the code
> > > > 
> > > > * the call uses `beanInstances.putIfAbsent` and the name of the method should mention this `ifAbsent` as well
> > > 
> > > 
> > > It is a setter and normally we don't add this kind of verbosity to method names. At most such a comment would apply to javadoc comments.
> > 
> > 
> > I think this is about good naming and API comprehensibility.
> > > > * I think the methods are ordered in purpose from `public` ones to `private` ones. It would be good to do the reordering and put this method to the `public` part of the class.
> > > 
> > > 
> > > I am not aware of any such recommendation in our style guidelines.
> > 
> > 
> > Yes, this is about a style the class started and about hints used in Java community.
> 
> But I am not aware of any such recommendation in our style guidelines.

 Those are ideas from the recognized persons in the Java field like https://www.amazon.com/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882.

ochaloup:
Ok. I'm still not convinced about this change.

My personal belief is that if there is a chance it's good to things in way that do not break API or concept.

I consider I could be wrong in understanding so let me rephrase my main point here.

Adding the method for the one particular bean is breaking the whole concept of the bean configuration.
If the developer won't be taking care enough he can easily get to the situation where part of the configuration is defined explicitly via bean creation in the code while another part is loaded from the XML.

Such behaviour would not be intuitive for me.


My thinking is for having is a single mechanism that defines - all beans are for configuration from the XML (then reflection and xml parsing are done) or all beans are configurable with an external provider (no reflection and xml parsing).

Is that idea understandable or should I try better? If it's understandable what is your standpoint about it?

common/classes/com/arjuna/common/internal/util/propertyservice/BeanPopulator.java
mmusgrov: Okay, I can return the previous bean value if it had already been set

tomjenkinson: If it doesn't take effect some feedback needs to be given

ochaloup: agree. I don't like the exceptions to be used for reporting something which is "expected" so using the boolean that returns if the operation succeeded is what I meant. The javadoc should be added for that too.

tomjenkinson: Perhaps it is not expected that it would ever reach this state and returning the error would provide a meaningful condition for the calling code to react to.

e.g throw new Exception("This bean was already configured so your config will be ignored");

tomjenkinson: but if we can internally reconcile the problem then a boolean would be more efficient

tomjenkinson: another option is to reconfigure based on the new values in the beanInstance.

E.g. 
Object foo = beanInstances.putIfAbsent(name, beanInstance);
// overwrite values from foo with those from beanInstance

But in that case I think the method should return Object

mmusgrov:
> Ok. I'm still not convinced about this change.
> 
> My personal belief is that if there is a chance it's good to things in way that do not break API or concept.

And that should be done in a comprehensive manner (and discussed perhaps on the forums too) and a new JIRA should be raised to address it.
This JIRA addresses an immediate need and the wider discussion around how we configure Narayana can wait for a major release.

> 
> I consider I could be wrong in understanding so let me rephrase my main point here.
> 
> Adding the method for the one particular bean is breaking the whole concept of the bean configuration.

It is not one particular bean. It provides a mechanism for frameworks that have there own configuration mechanisms to supply config without having to force them to use XML.

> If the developer won't be taking care enough he can easily get to the situation where part of the configuration is defined explicitly via bean creation in the code while another part is loaded from the XML.

And that is fine although he will loose the benefit of avoiding XML parsing. The target framework is Quarkus and that is fully under our control.

> 
> Such behaviour would not be intuitive for me.

Then the existing status quo will also be counter intuitve since we allow properties to be set via Java system properties and via XML properties files.

> 
> My thinking is for having is a single mechanism that defines - all beans are for configuration from the XML (then reflection and xml parsing are done) or all beans are configurable with an external provider (no reflection and xml parsing).

We already provide a flexible mechanism (system properties and XML properties files). Your thinking is fine for the next major release.

mmusgrov:
> Reading some of @ochaloup comments I am wondering if instead of editing BeanPopulator we should put the setters on https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/arjPropertyManager.java etc too. Without seeing the intended usage of BeanPopulator I can't be sure but I note it is in the internal package and so not something business logic so directly access.

Yes, the bean populator is in an internal package and is being used to integrate Narayana into Quarkus. I would prefer not to be changing public facing APIs unless it is absolutely neccessary, ie I don't think we should be doing any significant refactoring until 6 when we can address the wider issue of how we configure properties.

> 
> So I am thinking we perhaps need the setDefaultInstance in BeanPopulator plus changes to *PropertyManager. For example, calls like this need to work as they do today https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/arjPropertyManager.java#L52 and so we will need to either change BeanPopulator to be able to set them, or change the *PropertyManagers to have a local cache before hitting BeanPopulator.

We need a more comprehensive discussion about property config in Narayana and we should really be enlisting input from the community. It is not something we would want to do in a point release.

common/classes/com/arjuna/common/internal/util/propertyservice/BeanPopulator.java
```diff
@@ -174,6 +174,19 @@ public static void configureFromProperties(Object bean, String instanceName, Pro
         }
     }
 
+    /**
+     * Directly set a bean overriding any previous config. This method of setting a particular config
```
ochaloup: This does not sound right as the bean does not override the previous state. I assume the bean is used only when there was no previous state before.

ochaloup:
>> I consider I could be wrong in understanding so let me rephrase my main point here.
>>
>> Adding the method for the one particular bean is breaking the whole concept of the bean configuration.
>
> It is not one particular bean. It provides a mechanism for frameworks that have there own configuration mechanisms to supply config without having to force them to use XML.

 What does mean mechanism for frameworks? I could be wrong in the interpretation of the code but I understand it in way of XML parsing is done, system and environmental properties are used to set properties in beans.

ochaloup:
I agree with Mike that I would prefer not changing the public api as what the `arjPropertyManager` provides.

common/classes/com/arjuna/common/internal/util/propertyservice/BeanPopulator.java
tomjenkinson: I think it would be helpful to Narayana developers to add a note that this is not guaranteed to exist long term but as it is internal it is not essential

mmusgrov: I updated the javadoc to more accurately reflect what the method does. I also added Tom's caveat about the method being internal and subject to change (JBTM-3215).

PR #1523, opened 2019-10-28
JBTM-3161 Include a no-args public constructor

mmusgrov:
https://issues.jboss.org/browse/JBTM-3161

!MAIN XTS !AS_TESTS !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !PERF NO_WIN !RTS !TOMCAT JACOCO  !XTS


mmusgrov:
> Can you just make the commit message mentioning the class name of the class where the constructor is present, please?

That seems a bit unorthodox. You can use various git tools to search through commit histories.

xstefank:
> > Can you just make the commit message mentioning the class name of the class where the constructor is present, please?
> 
> That seems a bit unorthodox. You can use various git tools to search through commit histories.

True. But the commit message seems to me a little too generic in this case. However, the code change is right thus I approved.

PR #1526, opened 2019-11-03
JBTM-3217 Take Compensate and Complete timeout information from the L…

xstefank:
…RA annotation in the same class

https://issues.jboss.org/browse/JBTM-3217

### DIFF
--- a/rts/lra/narayana-lra/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
+++ b/rts/lra/narayana-lra/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
@@ -118,6 +118,7 @@ public class ServerLRAFilter implements ContainerRequestFilter, ContainerRespons
         LRA transactional = method.getDeclaredAnnotation(LRA.class);
         URI lraId;
         URI newLRA = null;
+        Long timeout = null;
 
         URI suspendedLRA = null;
         URI incommingLRA = null;
@@ -142,6 +143,10 @@ public class ServerLRAFilter implements ContainerRequestFilter, ContainerRespons
             if (cancel0n.length != 0) {
                 containerRequestContext.setProperty(CANCEL_ON_PROP, cancel0n);
             }
+
+            if (transactional.timeLimit() != 0) {
+                timeout = Duration.of(transactional.timeLimit(), transactional.timeUnit()).toMillis();
+            }
         }
 
         boolean endAnnotation = method.isAnnotationPresent(Complete.class)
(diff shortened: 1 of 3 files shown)


rts/lra/narayana-lra/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -142,6 +143,10 @@ public void filter(ContainerRequestContext containerRequestContext) {
             if (cancel0n.length != 0) {
                 containerRequestContext.setProperty(CANCEL_ON_PROP, cancel0n);
             }
+
+            if (transactional.timeLimit() != 0) {
```
mmusgrov: The original code  that you cut and pasted was clearer than your change:

``` java
        long timeLimit = transactional.timeLimit();
        ChronoUnit timeUnit = transactional.timeUnit();

        ... Duration.of(timeLimit, timeUnit).toMillis();
```

Also the original code was fine so why the updates?

rts/lra/narayana-lra/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -464,8 +463,9 @@ private void resumeTransaction(URI lraId) {
         // nothing to do
     }
 
-    private String getCompensatorId(URI lraId, URI baseUri, String path) {
-        Map<String, String> terminateURIs = NarayanaLRAClient.getTerminationUris(resourceInfo.getResourceClass(), baseUri, path);
+    private String getCompensatorId(URI lraId, URI baseUri, String path, Long timeout) {
```
mmusgrov: So this is the only functional change made in the PR.

rts/lra/narayana-lra/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
xstefank: I can change this back, but recomputing timeout every time it's accessed is unnecessary.

rts/lra/narayana-lra/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
xstefank: I don't understand what this comment should mean.

rts/lra/narayana-lra/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: I meant that the having local variables for `timeLimit` and `timeUnit` improved readablity.

I realise that since we now need to pass the timeout into `getTerminationUris` it makes sense to cache the timeout value (so don't change it back to a method call). But it is fine if you want to keep the change and I will approve the PR as is.

rts/lra/narayana-lra/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: Sorry, it was just an observation and incorrect. The other functional change is to `getTerminatorURIs` where you had to delete the code that read the participant timeout annotation. Please ignore.

PR #1530, opened 2019-11-26
JBTM-3112 Respect LRA deadlines during recovery.

mmusgrov:
https://issues.jboss.org/browse/JBTM-3212
https://issues.jboss.org/browse/JBTM-3222

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !AS_TESTS !TOMCAT !JACOCO !mysql !postgres !db2 !oracle !RTS !MAIN !LRA

This PR adds recovery testing for timed out LRAs

### DIFF
--- a/rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
+++ b/rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
@@ -63,7 +63,6 @@ import javax.ws.rs.POST;
 import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
 import javax.ws.rs.WebApplicationException;
-import javax.ws.rs.client.Client;
 import javax.ws.rs.container.Suspended;
 import javax.ws.rs.core.Link;
 import javax.ws.rs.core.MediaType;
@@ -122,8 +121,6 @@ public class NarayanaLRAClient implements Closeable {
     private static final String LINK_TEXT = "Link";
 
     private URI base;
-    private Client client;
-    private boolean connectionInUse;
 
     private static URI defaultCoordinatorURI;
 
@@ -141,11 +138,7 @@ public class NarayanaLRAClient implements Closeable {
      */
     public NarayanaLRAClient() throws URISyntaxException {
         if (defaultCoordinatorURI != null) {
-            try {
                 init(defaultCoordinatorURI);
-            } catch (MalformedURLException e) {
-                throw new URISyntaxException(defaultCoordinatorURI.toString(), e.getMessage());
-            }
         } else {
             init("http",
                     System.getProperty(LRA_COORDINATOR_HOST_KEY, "localhost"),
@@ -186,8 +179,8 @@ public class NarayanaLRAClient implements Closeable {
         init(coordinatorUri);
     }
 
-    private void init(URI coordinatorUri) throws URISyntaxException, MalformedURLException {
-        init(coordinatorUri.toURL().getProtocol(), coordinatorUri.getHost(), coordinatorUri.getPort());
+    private void init(URI coordinatorUri) {
+        setCoordinatorURI(coordinatorUri);
     }
 
     private void setCoordinatorURI(URI uri) {
@@ -208,18 +201,31 @@ public class NarayanaLRAClient implements Closeable {
         return lraId == null ? null : lraId.replaceFirst(".*/([^/?]+).*", "$1");
     }
 
+    /*
+     * strip the uid of the LRA from the URI to obtain the coordinator endpoint
+     */
+    private static URI removeLRAId(URI lraId) throws URISyntaxException {
+        if (lraId == null) {
+            return null;
+        }
+
+        String ascii = lraId.toASCIIString();
+        String id = ascii.replaceFirst(".*/([^/?]+).*", "$1");
+
+        id = ascii.substring(0, ascii.length() - id.length());
+
+        return new URI(id);
+    }
+
     private RequestBuilder getTarget() {
         return new RequestBuilder(base);
     }
 
     public void setCurrentLRA(URI coordinatorUri) {
-        URL url = null;
-
         try {
-            url = coordinatorUri.toURL();
-            init(coordinatorUri);
-        } catch (URISyntaxException | MalformedURLException e) {
-            LRALogger.i18NLogger.error_invalidCoordinatorUrl(url, e);
+            init(removeLRAId(coordinatorUri));
+        } catch (URISyntaxException e) {
+            LRALogger.i18NLogger.error_invalidCoordinatorId(coordinatorUri.toASCIIString(), e);
             throwGenericLRAException(coordinatorUri, BAD_REQUEST.getStatusCode(), e.getMessage());
         }
     }
@@ -768,11 +774,7 @@ public class NarayanaLRAClient implements Closeable {
             URI nextLRA = Current.peek();
 
             if (nextLRA != null) {
-                try {
-                    init(nextLRA);
-                } catch (URISyntaxException | MalformedURLException ignore) {
-                    // the validity of the URI was checked when we added it to Current
-                }
+                init(nextLRA);
             }
         }
     }
@@ -835,7 +837,6 @@ public class NarayanaLRAClient implements Closeable {
     }
 
     public void close() {
-        client.close();
     }
 
     private void throwGenericLRAException(URI lraId, int statusCode, String message) throws WebApplicationException {
--- a/rts/lra/lra-coordinator-jar/pom.xml
+++ b/rts/lra/lra-coordinator-jar/pom.xml
@@ -1,4 +1,6 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?><project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.jboss.narayana.rts</groupId>
@@ -12,35 +14,26 @@
     <packaging>jar</packaging>
 
     <properties>
-        <narayana.version>${project.version}</narayana.version>
+        <ipv6.server.jvm.args>
+            -Djboss.bind.address=[::1] -Djboss.bind.address.management=[::1] -Djboss.bind.address.unsecure=[::1]
+            -Djava.net.preferIPv4Stack=false -Djava.net.preferIPv6Addresses=true
+        </ipv6.server.jvm.args>
+        <modular.jdk.args/>
+        <modular.jdk.props/>
+        <!-- the resource path of the coordinator - must match the deployment URL (see arquillian.xml) -->
+        <lra.coordinator.path>lra-coordinator/lra-coordinator</lra.coordinator.path>
+        <jvm.args.other>
+            -Dlra.coordinator.path=${lra.coordinator.path}
+        </jvm.args.other>
+        <server.jvm.args>${jvm.args.other} ${jvm.args.byteman} ${jvm.args.memory} ${jvm.args.debug} ${jvm.args.modular}</server.jvm.args>
 
-        <failOnMissingWebXml>false</failOnMissingWebXml>
-        <narayana.nodename.property>CoreEnvironmentBean.nodeIdentifier</narayana.nodename.property>
-        <thorntail.http.port>8080</thorntail.http.port>
+        <server.port.offset>100</server.port.offset>
+        <server2.jvm.args>
+            -Djboss.socket.binding.port-offset=${server.port.offset} ${jvm.args.other} ${jvm.args.memory} ${jvm.args.modular}
+        </server2.jvm.args>
     </properties>
 
-    <dependencyManagement>
-        <dependencies>
-            <dependency>
-                <groupId>io.quarkus</groupId>
-                <artifactId>quarkus-bom</artifactId>
-                <version>${version.quarkus}</version>
-                <type>pom</type>
-                <scope>import</scope>
-            </dependency>
-        </dependencies>
-    </dependencyManagement>
-
-    <build>
-        <finalName>lra-coordinator</finalName>
-    </build>
-
     <dependencies>
-        <dependency>
-            <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-smallrye-openapi</artifactId>
-        </dependency>
-
         <dependency>
             <groupId>org.eclipse.microprofile.lra</groupId>
             <artifactId>microprofile-lra-api</artifactId>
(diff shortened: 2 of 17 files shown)


rts/lra/lra-coordinator-jar/pom.xml
```diff
@@ -12,35 +14,26 @@
     <packaging>jar</packaging>
 
     <properties>
-        <narayana.version>${project.version}</narayana.version>
+        <ipv6.server.jvm.args>
+            -Djboss.bind.address=[::1] -Djboss.bind.address.management=[::1] -Djboss.bind.address.unsecure=[::1]
```
xstefank: why do we need jboss specific sys props when this will be used in Quarkus based coordinator? Is it best to keep them in the shared module?

rts/lra/lra-coordinator-jar/pom.xml
```diff
@@ -12,35 +14,26 @@
...
+            -Dlra.coordinator.path=${lra.coordinator.path}
+        </jvm.args.other>
+        <server.jvm.args>${jvm.args.other} ${jvm.args.byteman} ${jvm.args.memory} ${jvm.args.debug} ${jvm.args.modular}</server.jvm.args>
 
-        <failOnMissingWebXml>false</failOnMissingWebXml>
-        <narayana.nodename.property>CoreEnvironmentBean.nodeIdentifier</narayana.nodename.property>
-        <thorntail.http.port>8080</thorntail.http.port>
+        <server.port.offset>100</server.port.offset>
+        <server2.jvm.args>
+            -Djboss.socket.binding.port-offset=${server.port.offset} ${jvm.args.other} ${jvm.args.memory} ${jvm.args.modular}
```
xstefank: I can see now that you are targetting jboss for the deployment of this jar. Why, please? Isn't this jar also be used in the Quarkus coordinator runtime?

rts/lra/lra-coordinator-jar/pom.xml
```diff
@@ -56,37 +49,211 @@
...
+        <dependency>
+            <groupId>org.jboss.byteman</groupId>
+            <artifactId>byteman</artifactId>
+            <version>${version.org.jboss.byteman}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.narayana.rts</groupId>
+            <artifactId>narayana-lra</artifactId>
+            <version>5.10.1.Final-SNAPSHOT</version>
```
xstefank: ${project.version}?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -259,6 +273,25 @@ public boolean restore_state(InputObjectState os, int ot) {
...
+            /*
+             * If the time limit has already been reached then the difference between now and the scheduled
+             * abort time will be negative. Since scheduling a task with a negative time will run it immediately
+             * we must ensure that the setTimeLimit call is placed after the state has been fully re-hydrated.
+             */
+            if (finishTime != null) {
+                long ttl = ChronoUnit.MILLIS.between(LocalDateTime.now(ZoneOffset.UTC), finishTime);
+
+                if (ttl <= 0) {
+                    System.out.printf("TIMER has expired since last reload%n");
```
xstefank: Should it be better to log these statements instead of system.out?

rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/LRACoordinatorTestCase.java
```diff
@@ -0,0 +1,259 @@
...
+    @After
+    public void after() {
+        stopContainer();
+        super.after();
+    }
+
+    /**
+     * Test that an LRA which times out while there is no running coordinator is cancelled
+     * when a coordinator is restarted
+     * @throws URISyntaxException if the LRA or revovery URIs are invalid (should never happen)
```
xstefank:
```suggestion
     * @throws URISyntaxException if the LRA or recovery URIs are invalid (should never happen)
```

rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/LRACoordinatorTestCase.java
```diff
@@ -0,0 +1,259 @@
...
+        assertEquals("LRA listener should have been told that the final state of the LRA was cancelled",
+                LRAStatus.Cancelled.name(), listenerStatus);
+    }
+
+    /**
+     * Test that an LRA which times out while there is no running coordinator is cancelled
+     * when a coordinator is restarted.
+     *
+     * Test that an LRA which times out after a coordinator is restarted after a crash is still active
+     * @throws URISyntaxException if the LRA or revovery URIs are invalid (should never happen)
```
xstefank:
```suggestion
     * @throws URISyntaxException if the LRA or recovery URIs are invalid (should never happen)
```

rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/LRACoordinatorTestCase.java
```diff
@@ -0,0 +1,259 @@
...
+     */
+    @Test
+    public void testRecovery2() throws URISyntaxException {
+        startContainer(null);
+
+        // start an LRA with a long timeout to validate that timed LRAs do not finish early during recovery
+        URI longLRA = lraClient.startLRA(null, "Long Timeout Recovery Test", LONG_TIMEOUT, ChronoUnit.MILLIS);
+        // start an LRA with a short timeout to validate that timed LRAs that time out when the coordinator is unavailable are cancelled
+        URI shortLRA = lraClient.startLRA(null, "Long Timeout Recovery Test", SHORT_TIMEOUT, ChronoUnit.MILLIS);
```
xstefank: different clientId can help with debugging in future

rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/TestBase.java
```diff
@@ -0,0 +1,207 @@
...
+
+        Config config = new Config();
+        String javaVmArguments = System.getProperty("server.jvm.args");
+        config.add("javaVmArguments", javaVmArguments);
+        containerController.start(COORDINATOR_CONTAINER);
+    }
+
+    void stopContainer() {
+        if (containerController.isStarted(COORDINATOR_CONTAINER)) {
+            LRALogger.logger.info("STOPPING CONTAINER%n");
```
xstefank: infof?

rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/TestBase.java
```diff
@@ -0,0 +1,207 @@
...
+
+            deployer.undeploy(COORDINATOR_DEPLOYMENT);
+
+            containerController.stop(COORDINATOR_CONTAINER);
+            containerController.kill(COORDINATOR_CONTAINER);
+        }
+    }
+
+    int recover() throws URISyntaxException {
+        try (Response response = ClientBuilder.newClient().target(new URI(recoveryUrl))
```
xstefank: Is the created client also closed in try-with-resources?

rts/lra/lra-coordinator-jar/pom.xml
tomjenkinson: +1

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
```diff
@@ -275,6 +275,7 @@ public Response startLRA(
         }
 
         String coordinatorUrl = String.format("%s%s", context.getBaseUri(), COORDINATOR_PATH_NAME);
+
```
tomjenkinson: formatting only

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/JaxRsActivator.java
```diff
@@ -10,7 +10,7 @@
 // mark the war as a JAX-RS archive
 @ApplicationPath("/")
 @OpenAPIDefinition(
-    info = @Info(title = "LRA Coordinator", version = "1.0"),
+    info = @Info(title = "LRA Coordinator", version = "1.0-RC1"),
```
tomjenkinson: This looks like something we will have to watch out for. I wonder if how we can make this more prominent that it needs to be updated with LRA coordinator versions?

tomjenkinson: I appreciate this concern is not introduced by this PR but is something to follow up on

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
tomjenkinson: Good catch, this should be using the logger framework

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
```diff
@@ -67,7 +67,23 @@
 
     public Transaction getTransaction(URI lraId) throws NotFoundException {
         if (!lras.containsKey(lraId)) {
+            String uid = getUid(lraId);
+
+            // try comparing on uid since differnt URIs can map to the same resource
+            // (eg localhost versus 127.0.0.1 versus :1 etc)
+            for (Transaction lra : lras.values()) {
```
tomjenkinson: I think this is perhaps best recorded as a different JBTM - but thanks for the comment as it explains the motiviation.

rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/TestBase.java
tomjenkinson: I was going to mention that printf are probably OK in the tests. But I think it is a bit confusing that there is a mix of printf and the logger. I think that using the LRA internal logger could be confusing in the output

rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/TestBase.java
tomjenkinson: I am not even sure if [Client](https://docs.oracle.com/javaee/7/api/javax/ws/rs/client/Client.html) is `AutoCloseable`. If it was I would guess it could need something like the following (assuming it would compile) to work:
```try (Client client = ClientBuilder.newClient(); Response resource = client.target(...```

File: rts/lra/lra-test/lra-test-tck/pom.xml [hunk 13]
```diff
@@ -126,6 +110,5 @@
                 <lra.tck.suite.thorntail.trace.params>-Dthorntail.logging=TRACE</lra.tck.suite.thorntail.trace.params>
             </properties>
         </profile>
-
```
tomjenkinson: formatting

rts/lra/lra-coordinator-jar/pom.xml
tomjenkinson: My assumption is that the byteman had some role in this. I would say I do not expect the container (JBoss, Quarkus) would impact the specific behaviour we are trying to test - @mmusgrov please can you confirm?

It would be useful to note this choice in the pom below where the main transfer of quarkus dependencies to WFLY is done.

xstefank:
> @xstefank I thought the strategy was that the LRA branch was just for the features that rely on LRA GA or RC2? We can still make changes on the main branch when the change would not rely on the temporary artifacts.
> 
> That said I think we could request an RC2 if there is divergence going on as we don't want to spend a lot of time reconciling the changes.

Yes, but if we are changing LRA subsystem in any way there will be conflicts. Changes are made for different reasons but over the same codebase.

rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/TestBase.java
xstefank: but I still close it usually when I type rest client invocations. But you are right it's not AutoClosable. Not sure if this it is necessary to close it.

tomjenkinson: Agreed, I was not suggesting not to close it - just trying to develop the thread. If it's not autocloseable then I think we just need the try-with-resources wrapped with a try/finally block

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
```diff
@@ -768,11 +774,7 @@ private void endLRA(URI lra, boolean confirm) throws WebApplicationException {
             URI nextLRA = Current.peek();
 
             if (nextLRA != null) {
-                try {
-                    init(nextLRA);
-                } catch (URISyntaxException | MalformedURLException ignore) {
-                    // the validity of the URI was checked when we added it to Current
-                }
+                init(nextLRA);
```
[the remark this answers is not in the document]
mmusgrov: I did not change this behaviour in this PR. But I will add some narayana specific tests to cover it and you can track our progress on issue https://issues.jboss.org/browse/JBTM-3223

rts/lra/lra-coordinator-jar/pom.xml
mmusgrov: We test the coordinator jar against a WildFly container.

rts/lra/lra-coordinator-jar/pom.xml
mmusgrov: The LRA coordinator must be able to run on any JAX-RS container (although we have recently introduced a  dependency on CDI). The rationale is that the coordinator is more than just an MP-LRA implementation (we do have other non MP users of this software module).

Secondly testing with different containers provides assurances that we are not locked into any particular container.

And the byteman angle is that Narayana tests (XTS, RTS etc) already use the byteman plus JBossAS plus Arquillian combination so it makes good sense to continue with that approach (additionally we can then leverage techniques and fixes used in these other modules).

rts/lra/lra-coordinator-jar/pom.xml
mmusgrov: +1

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
mmusgrov: +1

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/JaxRsActivator.java
mmusgrov: I created a final constant (in the same file) to give some degree of prominence to it. Now that you have mentioned it I thing I will remember to update it as we produce new LRA releases.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
mmusgrov: +1 (and to the other occurrences of printf in this PR)

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
mmusgrov: Okay I added a new JIRA https://issues.jboss.org/browse/JBTM-3222 and updated the PR description

rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/LRACoordinatorTestCase.java
mmusgrov: +1

rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/LRACoordinatorTestCase.java
mmusgrov: +1

rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/TestBase.java
mmusgrov: +1 I removed all system printf's and fixed this format string on the logger

rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/TestBase.java
mmusgrov: The call that closes the underlying connections/resources is either when the response is closed or when the response entity is fully read (which is what the above code does in its `try with resources code block`).

The spec has this to say about the JAX-RS client API close method:

> Calling this method effectively invalidates all resource targets produced by the client instance. Invoking any method on such targets once the client is closed would result in an IllegalStateException being thrown.

So since I am not using any of those targets after the call finishes (because the client is implicitly created) I don't need to call close on it.

rts/lra/lra-coordinator-jar/pom.xml
```diff
@@ -56,37 +49,211 @@
             <artifactId>arjuna</artifactId>
             <version>${project.version}</version>
             <exclusions>
-                 <exclusion>
-                     <groupId>sun.jdk</groupId>
-                     <artifactId>jconsole</artifactId>
-                 </exclusion>
-             </exclusions>
+                <exclusion>
```
tomjenkinson: formatting?

rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/TestBase.java
tomjenkinson: How about this statement "Client instances must be properly closed before being disposed to avoid leaking resources." from https://docs.oracle.com/javaee/7/api/javax/ws/rs/client/Client.html

mmusgrov:
> Just a note, but shouldn't we focus all LRA work on the LRA branch? In this case this a pretty big PR and potentially a lot of conflicts with the work done in the LRA branch.

No I don't think so. The LRA branch was created in order to avoid dependencies on snapshot builds and for tracking changes to the MP-LRA spec and TCK. My change affects the LRA implementation only which is more than just an MP-LRA implementation. Also since this change works fine with the spec version RC1 it is fine to apply the fix to the master branch (in fact that is the best place for it since users will use our master branch (not the LRA branch which is a moving target) so they will get the fix.

mmusgrov:
@tomjenkinson and @xstefank can you both take another look please.

rts/lra/lra-coordinator-jar/pom.xml
tomjenkinson: That seems reasonable. 

I assume two things:
1. There is no conflict on the LRA branch with a test that perhaps expects quarkus there?
2. This does in any way revert JBTM-3207 / https://github.com/jbosstm/narayana/commit/56e63d201cef2d2dc27598014546199a9004826c

rts/lra/lra-coordinator-jar/pom.xml
tomjenkinson: I think we should resolve this part of the conversation in order to keep the thread below in line

tomjenkinson: @xstefank hopefully you can re-open if you disagree it is not directly related to the discussion below

rts/lra/lra-test/lra-test-tck/pom.xml
mmusgrov: I missed this one and adding it back in again will trigger a full rebuild.
But I would also say that since this is pre release code removing the empty line does not obfuscate the intent of the PR and does not impact anyone wanting to diff the code.

mmusgrov:
> > Just a note, but shouldn't we focus all LRA work on the LRA branch? In this case this a pretty big PR and potentially a lot of conflicts with the work done in the LRA branch.
> 
> No I don't think so. The LRA branch was created in order to avoid dependencies on snapshot builds and for tracking changes to the MP-LRA spec and TCK. My change affects the LRA implementation only which is more than just an MP-LRA implementation. Also since this change works fine with the spec version RC1 it is fine to apply the fix to the master branch (in fact that is the best place for it since users will use our master branch (not the LRA branch which is a moving target) so they will get the fix.

I just checked and the LRA branch only has two small commits on it which I doubt will cause any conflicts - I don't mind doing the merge when we have an RC2. But I still think we are at least a couple of weeks away from RC2 so if we can get more (all even) of the MP-LRA issues resolved then we could do an RC2.

rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/TestBase.java
mmusgrov: Maybe later on in the project we can do some resource utilisation testing and performance optimisations but let's do it as part of a more comprehensive analysis of the non functional behaviour of the system.

rts/lra/lra-coordinator-jar/pom.xml
mmusgrov: I created (and fixed) JBTM-3207 since I did not want to force the coordinator to *only* run on Quarkus. That fix moved the sources into a separate maven artefact called `lra-coordinator-jar`. This way we can run the coordinator on any JAX-RS (and CDI) capable container. The wildfly references in the pom are for (crashrec) testing only. 

The quarkus artefact is called `lra-coordinator` and it pulls in the coordinator sources by depending on `lra-coordinator-jar`. The TCK still runs with the quarkus runner.

mmusgrov: The tests should not rely on any particular container and if they do then the test should be updated to be container agnostic (modulo JAX-RS and CDI).

rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/TestBase.java
mmusgrov: Okay, it is just a test but fair enough. If that is the only pending issue then I will push a fix for this comment?

rts/lra/lra-coordinator-jar/pom.xml
xstefank: ok. I was just curious if JBoss specific properties will be propagated to the quarkus LRA coordinator. But if it's only for tests it should be ok.

rts/lra/lra-coordinator-jar/pom.xml
xstefank: thanks for the explanation. Ok from my side now.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -720,61 +759,85 @@ private int lraStatusToHttpStatus() {
...
             return Response.Status.PRECONDITION_FAILED.getStatusCode();
         }
 
-        if (timeLimit > 0) {
-            if (finishTime != null) {
-                // check whether the new time limit is less than the current one
-                LocalDateTime ft = LocalDateTime.now().plusNanos(timeLimit * 1000000);
+        if (finishTime != null) {
+            // check whether the new time limit is less than the current one
+            LocalDateTime ft = LocalDateTime.now().plusNanos(timeLimit * 1000000);
```
xstefank: sorry I missed this previously. This is missing `ZoneOffset.UTC`.

rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/TestBase.java
```diff
@@ -0,0 +1,213 @@
...
+    @Before
+    public void before() throws URISyntaxException, MalformedURLException {
+        LRALogger.logger.debugf("Starting test %s", testName);
+
+        lraClient = new NarayanaLRAClient(new URI(coordinatorUrl));
+    }
+
+    @After
+    public void after() {
+        LRALogger.logger.debugf("Finised test %s", testName);
```
xstefank:
```suggestion
        LRALogger.logger.debugf("Finished test %s", testName);
```

rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/TestBase.java
```diff
@@ -0,0 +1,213 @@
...
+    int recover() throws URISyntaxException {
+        Client client = ClientBuilder.newClient();
+
+        try (Response response = client.target(new URI(recoveryUrl))
+                .request()
+                .get()) {
+
+            Assert.assertEquals("Unexpected status from reovery call:", 200, response.getStatus());
+
+            // the result wil be a List<LRAStatusHolder> of recovering LRAs but we just need the count
```
xstefank:
```suggestion
            // the result will be a List<LRAStatusHolder> of recovering LRAs but we just need the count
```

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
tomjenkinson: Thanks Mike, I think it needs to be in a separate commit

rts/lra/lra-coordinator-jar/pom.xml
```diff
@@ -1,5 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?><project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
```
tomjenkinson: formatting

rts/lra/lra-coordinator-jar/pom.xml
```diff
@@ -1,5 +1,7 @@
...
     <modelVersion>4.0.0</modelVersion>
+
```
tomjenkinson: formatting

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
mmusgrov: I pulled out the changes and moved them into a separate commit.

rts/lra/lra-coordinator-jar/pom.xml
mmusgrov: Since this is pre release code this change does not obfuscate the intent of the PR and does not impact anyone wanting to diff the code. After release one it would make sense to minimise the occasion reformatting of blocks.

mmusgrov:
> There are still a few JAX-RS Client not closed in LRACoordinatorTestCase but if that's a design decision than ignore this message.

It was deliberate (though not a "design decision"). This is a surefire test case and the JVM is torn down after each test so even if there were consequences to not closing the client it would have no effect on the test. Test cases are different from production code and the goal is coverage and simplicity (I do agree that closing responses is important since network resources live in the kernel).

But that said, I have made the changes anyway.

rts/lra/lra-coordinator-jar/pom.xml
mmusgrov: +1

rts/lra/lra-coordinator-jar/pom.xml
mmusgrov: +1

rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/TestBase.java
mmusgrov: +1

rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/TestBase.java
mmusgrov: +1

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
mmusgrov: +1

PR #1536, opened 2019-12-05
JBTM-3229 Allow to enlist AfterLRA listeners after the LRA enters the ending phase

xstefank:
https://issues.redhat.com/browse/JBTM-3229

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !AS_TESTS !TOMCAT !JACOCO !mysql !postgres !db2 !oracle !RTS !MAIN LRA

This PR builds on top of the https://github.com/jbosstm/narayana/pull/1535 so it can be updated separately but we can still verify that full TCK is passing with all changes. We can merge either one of them first and I will adjust the other if needed.

PR #1535, opened 2019-12-05
JBTM-3228 Nested LRA participants should have Forget methods called w…

xstefank:
…hen the parent LRA is closed

https://issues.redhat.com/browse/JBTM-3228

!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !AS_TESTS !TOMCAT !JACOCO !mysql !postgres !db2 !oracle !RTS !MAIN LRA

There will be one failing TCK test `TckContextTests.testAfterLRAEnlistmentDuringClosingPhase` for which fix will be coming in a separate JBTM. I also included a small code clean up of duplicated method setLraService in LRARecord.

PR #1537, opened 2019-12-06
JBTM-3228 call forget on nested participants

mmusgrov:
https://issues.redhat.com/browse/JBTM-3228

There should only be 2 TCK failures:

1. TckContextTests.testAfterLRAEnlistmentDuringClosingPhase is being fixed by PR #1536 
2. TckContextTests.testParentContextAvailable fails because this PR adds new behaviour so testParentContextAvailable also should check that nested participants are asked to forget when the parent is closed.

### DIFF
--- a/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
+++ b/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
@@ -268,7 +268,7 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
             if (isCompelete() || completeURI == null) {
                 status = ParticipantStatus.Completed;
 
-                return TwoPhaseOutcome.FINISH_OK; // the participant has already completed
+                return parentId != null ? forget(null) : TwoPhaseOutcome.FINISH_OK; // the participant has already completed
             }
 
             endPath = completeURI;  // we are going to ask the participant to complete
@@ -286,6 +286,10 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
             int twoPhaseOutcome = retryGetEndStatus(endPath, compensate);
 
             if (twoPhaseOutcome != -1) {
+                if (parentId != null && status == ParticipantStatus.Completed) {
+                    return forget(null);
+                }
+
                 return twoPhaseOutcome;
             }
         } else {
@@ -296,6 +300,8 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
             // the local invocation was not made so fallback to using JAX-RS
             Client client = ClientBuilder.newClient();
             WebTarget target = client.target(URI.create(endPath.toASCIIString()));
+            LRALogger.logger.debugf("LRARecord.tryDoEnd calling PUT on %s with parent: %s",
+                endPath, parentId == null ? "null" : parentId);
 
             try {
                 // ask the participant to complete or compensate
@@ -303,8 +309,6 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
                         .header(LRA_HTTP_CONTEXT_HEADER, lraId.toASCIIString())
                         .header(LRA_HTTP_PARENT_CONTEXT_HEADER, parentId) // make the context available to participants
                         .header(LRA_HTTP_RECOVERY_HEADER, recoveryURI.toASCIIString())
-                        .property(LRA_HTTP_CONTEXT_HEADER, lraId) // make the context available to the jaxrs filters
-                        .property(LRA_HTTP_PARENT_CONTEXT_HEADER, parentId) // make the context available to jaxrs filters
                         .async();
 
                 Future<Response> asyncResponse = getAsyncResponse(target, PUT.class.getName(), asyncInvoker, compensatorData);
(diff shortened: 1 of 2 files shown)


rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -286,6 +286,10 @@ private int tryDoEnd(boolean compensate) {
             int twoPhaseOutcome = retryGetEndStatus(endPath, compensate);
 
             if (twoPhaseOutcome != -1) {
+                if (parentId != null && status == ParticipantStatus.Completed) {
```
xstefank: will the execution ever get here? Because this method will be called again through recovery with compensate == false and the LRARecord is still completed so why do you not finish on line 271 again?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -286,6 +286,10 @@ private int tryDoEnd(boolean compensate) {
...
+                    return forget(null);
```
xstefank: also if there is no status method in the participant the retryGetEndStatus returns -1 so this seems that it will never get executed.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -386,6 +393,10 @@ private int tryDoEnd(boolean compensate) {
 
         updateStatus(compensate);
 
+        if (parentId != null && status == ParticipantStatus.Completed) {
```
xstefank: parentId is set in the constructor, so won't this be executed even when the nested LRA is completed (and parent is still active)?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -454,9 +465,9 @@ private int retryGetEndStatus(URI endPath, boolean compensate) {
                 // since this method is called from the recovery thread do not block
                 AsyncInvoker asyncInvoker = target.request()
                         .header(LRA_HTTP_CONTEXT_HEADER, lraId.toASCIIString())
+                        .header(LRA_HTTP_PARENT_CONTEXT_HEADER, parentId)
                         .property(LRA_HTTP_PARENT_CONTEXT_HEADER, parentId) // make the context available to participants
                         .header(LRA_HTTP_RECOVERY_HEADER, recoveryURI.toASCIIString())
-                        .property(LRA_HTTP_CONTEXT_HEADER, lraId)  // make the context available to the jaxrs filters
```
xstefank: this is LRA id, isn't this something that filters really rely on?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -849,4 +832,55 @@ public void setLraService(LRAService lraService) {
...
+            close = true;
+            client = ClientBuilder.newClient();
+        }
+
+        LRALogger.logger.debugf("LRARecord.forget calling DELETE on %s with parent: %s",
+                    forgetURI, parentId == null ? "null" : parentId);
+
+        try {
+            // let the participant know he can clean up
+            WebTarget target2 = client.target(forgetURI);
```
xstefank: we can now rename the variables to drop the suffix index

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
mmusgrov: That is an excellent point. Let me go back and take a look at your original PR ...

mmusgrov:
Closing (in favour of PR #1535 ) since the correct place to drive the forget calls is at the level of the Transaction.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
mmusgrov: Yes it can if the participant has Completed or Compensated or if a local transaction has already Closed or Cancelled.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
mmusgrov: But there might be a status method and it might have been called and returned Completed.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
mmusgrov: +1

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
mmusgrov: +1

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
mmusgrov: I added a check to test whether the close call is because the parent was asked to close.

PR #1540, opened 2019-12-19
JBTM-3238 Coordinator#completeNestedLRA and Coordinator#compensateNes…

xstefank:
…tedLRA should return ParticipantStatus

https://issues.redhat.com/browse/JBTM-3238

### DIFF
--- a/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
+++ b/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
@@ -365,13 +365,13 @@ public class Coordinator {
     @PUT
     @Path("nested/{NestedLraId}/complete")
     public Response completeNestedLRA(@PathParam("NestedLraId") String nestedLraId) {
-        return endLRA(toURI(nestedLraId), false, true);
+        return Response.ok(mapToParticipantStatus(endLRA(toURI(nestedLraId), false, true)).name()).build();
     }
 
     @PUT
     @Path("nested/{NestedLraId}/compensate")
     public Response compensateNestedLRA(@PathParam("NestedLraId") String nestedLraId) {
-        return endLRA(toURI(nestedLraId), true, true);
+        return Response.ok(mapToParticipantStatus(endLRA(toURI(nestedLraId), true, true)).name()).build();
     }
 
     @PUT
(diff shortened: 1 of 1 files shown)


xstefank:
@tomjenkinson yes. A simple change, I just missed these two methods on my side.

mmusgrov:
@tomjenkinson This PR satisfies an MP-LRA spec change and there is already a TCK test for it.

mmusgrov:
@xstefank Is it okay to merge this PR?

mmusgrov:
@tomjenkinson thanks for the question. You are correct, none of the TCK tests in 1.0-RC1 would uncover this bug.
@xstefank we need a test for this change. Are there any tests that show that the original behaviour produced incorrect behaviour. Would the test 
`testForgetCalledForNestedParticipantsWhenParentIsClosed` that you added to the LRA branch exercise this bug?

xstefank:
@mmusgrov @tomjenkinson I checked the availability of test for this issue, however, it isn't going to be that easy. I ran into this issue while testing something different and I am not able to reproduce the error again. The code that can run into this issue is in [1] which is hard to hit since INTERNAL_SERVER_ERROR on our Coordinator side is hard to reproduce. I can try to write a test for the previous fix which was fixing the return value of nested Status calls -- this should be easier as it only means that nested LRA should get into Cancelling/Closing state so the parent LRA will call its status. However, I will open a new issue for this and I will not make it a priority right now if you don't mind. There are some more prompt issues at hand on my side. However, return values should be fixed even if we can only conditionally run into this issue I think. Would this be an acceptable resolution?

[1] https://github.com/jbosstm/narayana/blob/master/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java#L342

mmusgrov:
The status code that is being returned is clearly following the wrong model and the fix is straight forward. Personally I think not having a test for this is okay and we can add it when we expose the status method via the client API that will be introduced in MP-LRA 1.x

So @tomjenkinson I am going to approve this PR - but let us know if you feel we can't merge without the test?

mmusgrov:
>   . @mmusgrov I also noted the comment that 412 was probably not the correct code, is that comment stale or if not can we get a JBTM for that too please?

@tomjenkinson The comment is referring to the HTTP status code for an LRA but since we deferred the LRA Client API to a later release we can pick any status code that makes sense in a REST setting. When/if we do an client API we should choose/debate which status code to use.

mmusgrov:
@tomjenkinson The JBTM for the test is https://issues.redhat.com/browse/JBTM-3241

PR #1541, opened 2020-01-02
JBTM-3228: Nested LRA participants should have Forget methods called …

xstefank:
…when the parent LRA is closed

https://issues.redhat.com/browse/JBTM-3228

### DIFF
--- a/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
+++ b/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
@@ -35,7 +35,6 @@ import io.narayana.lra.logging.LRALogger;
 import org.eclipse.microprofile.lra.annotation.LRAStatus;
 import org.eclipse.microprofile.lra.annotation.ParticipantStatus;
 
-import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.PUT;
 import javax.ws.rs.WebApplicationException;
@@ -492,35 +491,11 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
                                     compensate, target.getUri(), status);
 
                             if (forgetURI != null) {
-                                try {
-                                    // let the participant know he can clean up
-                                    WebTarget target2 = client.target(forgetURI);
-                                    AsyncInvoker asyncInvoker2 = target2.request()
-                                            .header(LRA_HTTP_CONTEXT_HEADER, lraId.toASCIIString())
-                                            .property(LRA_HTTP_PARENT_CONTEXT_HEADER, parentId) // make the context available to participants
-                                            .header(LRA_HTTP_RECOVERY_HEADER, recoveryURI.toASCIIString())
-                                            .property(LRA_HTTP_CONTEXT_HEADER, lraId)  // make the context available to the jaxrs filters
-                                            .async();
-
-                                    Future<Response> asyncResponse2 = getAsyncResponse(
-                                            target2, DELETE.class.getName(), asyncInvoker2, compensatorData);
-
-                                    if (asyncResponse2.get(PARTICIPANT_TIMEOUT, TimeUnit.SECONDS).getStatus() ==
-                                            Response.Status.OK.getStatusCode()) {
-                                        return TwoPhaseOutcome.FINISH_OK;
-                                    }
-                                } catch (Exception e) {
-                                    if (LRALogger.logger.isInfoEnabled()) {
-                                        LRALogger.logger.infof("LRARecord.doEnd forget URI %s is invalid (%s)",
-                                                forgetURI, e.getMessage());
-                                    }
-
-                                    // TODO write a test to ensure that recovery only retries the forget request
-                                    return TwoPhaseOutcome.HEURISTIC_HAZARD; // force recovery to keep retrying
-                                } finally {
-                                    Current.pop();
+                                if (invokeForget(client, forgetURI, lraId, parentId, recoveryURI)) {
+                                    return TwoPhaseOutcome.FINISH_OK;
+                                } else {
+                                    return TwoPhaseOutcome.HEURISTIC_HAZARD;
                                 }
-
                             } else {
                                 LRALogger.logger.warnf(
                                         "LRARecord.doEnd(%b) LRA: %s: cannot forget %s: missing forget URI",
@@ -550,6 +525,38 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
         }
 
         return -1;
+    }
+
+    public static boolean invokeForget(Client client, URI forgetURI, URI lraId, URI parentId, URI recoveryURI) {
+        Response response = null;
+
+        try {
+            // let the participant know he can clean up
+            WebTarget target = client.target(forgetURI);
+            Future<Response> asyncResponse = target.request()
+                .header(LRA_HTTP_CONTEXT_HEADER, lraId.toASCIIString())
+                .header(LRA_HTTP_RECOVERY_HEADER, recoveryURI.toASCIIString())
+                .header(LRA_HTTP_PARENT_CONTEXT_HEADER, parentId) // make the context available to participants
+                .property(LRA_HTTP_CONTEXT_HEADER, lraId)  // make the context available to the jaxrs filters
+                .async()
+                .delete();
+
+            response = asyncResponse.get(PARTICIPANT_TIMEOUT, TimeUnit.SECONDS);
+
+            return response.getStatus() == Response.Status.OK.getStatusCode();
+        } catch (Exception e) {
+            if (LRALogger.logger.isInfoEnabled()) {
+                LRALogger.logger.infof("LRARecord.doEnd forget URI %s is invalid (%s)",
+                    forgetURI, e.getMessage());
+            }
+
+            // TODO write a test to ensure that recovery only retries the forget request
+            return false; // force recovery to keep retrying
+        } finally {
+            if (response != null) {
+                response.close();
+            }
+        }
     }
 
     private Future<Response> getAsyncResponse(WebTarget target, String method, AsyncInvoker asyncInvoker, String compensatorData) {
@@ -814,7 +821,7 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
         return 0;
     }
 
-    public URI getRecoveryCoordinatorURI() {
+    public URI getRecoveryURI() {
         return recoveryURI;
     }
 
@@ -844,6 +851,10 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
 
     public void setLraService(LRAService lraService) {
         this.lraService = lraService;
+    }
+
+    public URI getForgetURI() {
+        return forgetURI;
     }
 
     public URI getEndNotificationUri() {
--- a/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
+++ b/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
@@ -79,7 +79,7 @@ public class Transaction extends AtomicAction {
     private URI parentId;
     private String clientId;
     private List<LRARecord> pending;
-    private List<AfterLRAListener> afterLRAListeners;
+    private List<EndLRAListener> endLRAListeners;
     private LRAStatus status;
     private String responseData;
     private LocalDateTime startTime;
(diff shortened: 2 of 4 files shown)


xstefank:
@tomjenkinson this PR is fixing `TckContextTests#testForgetCalledForNestedParticipantsWhenParentIsClosed`

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -900,13 +892,63 @@ public boolean afterLRANotification() {
         return notifiedAll;
     }
 
-    static class AfterLRAListener {
-        public AfterLRAListener(URI notificationURI, URI recoveryId) {
+    private abstract static class EndLRAListener {
```
tomjenkinson: Please can you explain a little of the motivation of the hierarchy of the EndLRAListener, AfterLRAListener and ForgetLRAListener. 

I am wondering about whether some further symmetry in the structure of the implementation of the call(Client) method could be achieved? As an example they appear to diverge slightly in the exception handling (typed exceptions vs Exception)?

xstefank: the JAX-RS calls they are making are different at least AfterLRA is PUT vs Forget is DELETE. There are also some differences in the passed headers - we don't pass LRA id to AfterLRA because a new LRA can be started on these methods but instead, we are passing LRA ENDED header. Also, the log messages in case of exceptions differ. I don't want to rule out that any further alignment can be made but for this PR I would prefer to just stick with refactoring.

xstefank:
@mmusgrov do you think the refactoring is wrong? or why is the refactoring wrong? I tried this in several ways and we really don't need to care about keeping the whole LRARecord. In my opinion, Forget calls in this case are more similar to the AfterLRA calls (they are delivered after the nested LRA is closed) so I plugged them in the afterLRA cycle.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -814,7 +821,7 @@ public int compareTo(AbstractRecord other) {
         return 0;
     }
 
-    public URI getRecoveryCoordinatorURI() {
+    public URI getRecoveryURI() {
```
tomjenkinson: I am wondering what the need for renaming this method was?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -148,11 +148,12 @@ public boolean save_state(OutputObjectState os, int ot) {
...
             }
 
-            if (afterLRAListeners == null) {
+            if (endLRAListeners == null) {
                 os.packInt(0);
             } else {
-                os.packInt(afterLRAListeners.size());
-                for (AfterLRAListener participant : afterLRAListeners) {
+                os.packInt(endLRAListeners.size());
+                for (EndLRAListener participant : endLRAListeners) {
```
tomjenkinson: Doesn't the LRARecord pack these already: https://github.com/mmusgrov/narayana/blob/LRA/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java#L679 - although I am not then sure why the current LRA branch code also packs the AfterLRA elements: https://github.com/jbosstm/narayana/blob/LRA/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java#L156

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
xstefank: just because the method name was misleading. It is actually returning `recoveryURI` which is the participant ID not the recovery coordinator URI (which is included in the `recoveryURI` but it's not it).

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
xstefank: yes, afterLRAlisteners even now in the master are handled in the Transaction class, not in LRARecord. My solution for forgetting nested completed participants (this PR) is building on top of it which is the reason for the refactoring.

@

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
tomjenkinson: Ah OK, thanks for confirming. I agree it is misleading to include coordinator in the method - please can you add it to a separate commit/JBTM. I think the use of it [here](https://github.com/jbosstm/narayana/blob/cc46def573a1a0692bd531aaa0cc39c7a4b52a41/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java#L429) should have the assigned variable name altered too. /cc @mmusgrov

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -850,46 +867,21 @@ URI getParentId() {
...
-                        response = null;
-                    } catch (WebApplicationException | InterruptedException | ExecutionException | TimeoutException e) {
+                    EndLRAListener participant = listeners.next();
+                    if (participant.call(client)) {
+                        listeners.remove();
+                    } else {
                         notifiedAll = false;
-
-                        if (LRALogger.logger.isInfoEnabled()) {
-                            LRALogger.logger.infof("Could not notify AfterLRA listener at %s (%s)", participant.notificationURI, e.getMessage());
```
tomjenkinson: I wonder why this logging was removed?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
tomjenkinson: I see, thanks

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
tomjenkinson: Thanks @xstefank it's not strictly related to this PR but I would like to know why there is save/restore of the AfterLRA in LRARecord and Transaction /cc @mmusgrov

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -492,35 +491,11 @@ private int retryGetEndStatus(URI endPath, boolean compensate) {
...
-                                    }
-
-                                    // TODO write a test to ensure that recovery only retries the forget request
-                                    return TwoPhaseOutcome.HEURISTIC_HAZARD; // force recovery to keep retrying
-                                } finally {
-                                    Current.pop();
+                                if (invokeForget(client, forgetURI, lraId, parentId, recoveryURI)) {
+                                    return TwoPhaseOutcome.FINISH_OK;
+                                } else {
+                                    return TwoPhaseOutcome.HEURISTIC_HAZARD;
```
tomjenkinson: Are there any other locations where failure to invokeForget should cause a HEURISTIC_HAZARD. For instance over [here](https://github.com/mmusgrov/narayana/blob/d88fe2fcef356a333946a02d282f849e0f2094f4/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java#L271) it seems it was proposed to return the HEURISTIC_HAZARD but in the current patch forget is not called at all (at least from the [similar](https://github.com/jbosstm/narayana/blob/9380ad2ffdfb0716aae6ab743a4b6dca81a3d97d/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java#L261) place, perhaps it is called elsewhere by the refactor but then I am not sure how the refactor reports the H_H?)

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
xstefank: sure, I will add a new commit and JBTM.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
xstefank: there might be even a bigger issue with this approach (afterLRA listeners in Transaction) which would, unfortunately, invalidate even this PR :). I am working on a test case that will help.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
xstefank: This code was just refactored to the `call` method in AfterLRAListener - https://github.com/jbosstm/narayana/pull/1541/files/9380ad2ffdfb0716aae6ab743a4b6dca81a3d97d#diff-57ec69ca0d35bbed0ad018339b532a7dR930

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
xstefank: there is no need because already completed nested participants with forget methods are tracked in the same manner as afterLRA listeners - only in Transaction. So LRARecord will no longer be invoked. 

This leads me again to the issue that I realized because of your comments and I think that even afterLRA listeners may be doing something unexpected in the moment. I will report this in a new JBTM (as this PR is not the right place to track it) when I'll have the test ready.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
tomjenkinson: Thanks!

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
tomjenkinson: Sorry, I see that now thanks!

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
xstefank: @tomjenkinson @mmusgrov FYI I created https://issues.redhat.com/browse/JBTM-3239. Let's continue the discussion in the issue, please.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
xstefank: moved to https://issues.redhat.com/browse/JBTM-3240 as I will probably close this PR depending on the resolution of https://issues.redhat.com/browse/JBTM-3239

mmusgrov:
> ... In my opinion, Forget calls in this case are more similar to the AfterLRA calls (they are delivered after the nested LRA is closed) so I plugged them in the afterLRA cycle.

Okay that's fair:- either both behaviours should be implemented in the LRARecord or both should be done at the Transaction level. I will take another look at the PR.

mmusgrov:
What is the rationale for opening 2 PR against this JBTM (pr/1540 and pr/1541)

xstefank:
@mmusgrov this one is JBTM-3228; https://github.com/jbosstm/narayana/pull/1540 is JBTM-3238.

xstefank:
@mmusgrov however, can you please wait with the review until we decide on the solution of https://issues.redhat.com/browse/JBTM-3239?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
tomjenkinson: Thanks @xstefank - if you continue in here please rebase on master now that https://github.com/jbosstm/narayana/pull/1543 is merged

mmusgrov:
> @mmusgrov this one is JBTM-3228; #1540 is JBTM-3238.

Those are he same JIRA - we normally have one PR per JBTM (otherwise it gets confusing for the reviewer).

mmusgrov:
@xstefank Can this one be closed now that we have #1591 ?

xstefank:
@mmusgrov yes, I forgot about this one.

PR #1542, opened 2020-01-06
Add test for JBTM-3239

xstefank:
issue: https://issues.redhat.com/browse/JBTM-3239

This PR aims to LRA branch because the test uses the new APIs added to the LRA TCK (LRAMetricService) with the vision that it can be one day moved to the specification TCK and to prevent replication. However, the issue affects the master branch too and the fix must be backported to the master branch.

### DIFF
--- a/rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/FailingParticipantCallsIT.java
+++ b/rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/FailingParticipantCallsIT.java
@@ -1,0 +1,86 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2019, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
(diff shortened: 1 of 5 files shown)


rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/FailingParticipantCallsIT.java
```diff
@@ -0,0 +1,86 @@
...
+            .get();
+
+        Assert.assertEquals(200, response.getStatus());
+        Assert.assertTrue(response.hasEntity());
+
+        URI lra = URI.create(response.readEntity(String.class));
+
+        new NarayanaLRARecovery().waitForEndPhaseReplay(lra);
+
+        Assert.assertEquals(3,
```
xstefank: The reason for why we checking for 3 repeated calls instead of just two is https://issues.redhat.com/browse/JBTM-3163. However, the test will be correct even after the JBTM-3163 is fixed because waitForEndPhaseReplay(lra) will do the recovery twice (however unnecessary so I will fix this test after this issue is fixed)

xstefank:
I will include this test with the fix for JBTM-3239

PR #1547, opened 2020-01-09
JBTM-3243 Use PostConstruct/PreDestroy for ApplicationScoped bean (instead of Observer)

mmusgrov:
https://issues.redhat.com/browse/JBTM-3243

### DIFF
--- a/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
+++ b/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
@@ -61,7 +61,7 @@ public class LRAService {
     private Map<URI, Transaction> recoveringLRAs = new ConcurrentHashMap<>();
     private Map<URI, ReentrantLock> locks = new ConcurrentHashMap<>();
 
-    private static Map<String, String> participants = new ConcurrentHashMap<>();
+    private Map<String, String> participants = new ConcurrentHashMap<>();
     private LRARecoveryModule lraRecoveryModule;
 
     public Transaction getTransaction(URI lraId) throws NotFoundException {
@@ -188,13 +188,9 @@ public class LRAService {
             lras.remove(lraId);
         }
 
-        if (recoveringLRAs.containsKey(lraId)) {
-            recoveringLRAs.remove(lraId);
-        }
+        recoveringLRAs.remove(lraId);
 
-        if (locks.containsKey(lraId)) {
-            locks.remove(lraId);
-        }
+        locks.remove(lraId);
     }
 
     public void updateRecoveryURI(URI lraId, String compensatorUrl, String recoveryURI, boolean persist) {
@@ -392,7 +388,7 @@ public class LRAService {
         assert lraRecoveryModule == null;
 
         if (LRALogger.logger.isDebugEnabled()) {
-            LRALogger.logger.debugf("LRAServicve.enableRecovery%n");
+            LRALogger.logger.debugf("LRAServicve.enableRecovery");
         }
 
         lraRecoveryModule = new LRARecoveryModule(this);
@@ -411,13 +407,16 @@ public class LRAService {
      */
     @PreDestroy
     void disableRecovery() {
+        if (lraRecoveryModule != null) {
 
-        Implementations.uninstall();
-        RecoveryManager.manager().removeModule(lraRecoveryModule, false);
-        lraRecoveryModule = null;
+            RecoveryManager.manager().removeModule(lraRecoveryModule, false);
+            Implementations.uninstall();
 
-        if (LRALogger.logger.isDebugEnabled()) {
-            LRALogger.logger.debugf("LRAServicve.disableRecovery%n");
+            lraRecoveryModule = null;
+
+            if (LRALogger.logger.isDebugEnabled()) {
+                LRALogger.logger.debugf("LRAServicve.disableRecovery");
+            }
         }
     }
 }
--- a/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
+++ b/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
@@ -35,10 +35,9 @@ import io.narayana.lra.coordinator.domain.model.Transaction;
 
 import org.eclipse.microprofile.lra.annotation.LRAStatus;
 
+import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
 import javax.enterprise.context.ApplicationScoped;
-import javax.enterprise.context.Destroyed;
-import javax.enterprise.context.Initialized;
-import javax.enterprise.event.Observes;
 import javax.ws.rs.InternalServerErrorException;
 import javax.ws.rs.NotFoundException;
 import javax.ws.rs.WebApplicationException;
@@ -387,10 +386,9 @@ public class LRAService {
 
     /**
      * When the deployment is loaded register for recovery
-     *
-     * @param init a javax.servlet.ServletContext
      */
-    void enableRecovery(@Observes @Initialized(ApplicationScoped.class) Object init) {
+    @PostConstruct
+    void enableRecovery() {
         assert lraRecoveryModule == null;
 
         if (LRALogger.logger.isDebugEnabled()) {
@@ -410,11 +408,9 @@ public class LRAService {
 
     /**
      * When the deployment is unloaded unregister for recovery
-     *
-     * @param init a javax.servlet.ServletContext
      */
-    void disableRecovery(@Observes @Destroyed(ApplicationScoped.class) Object init) {
-        assert lraRecoveryModule != null;
+    @PreDestroy
+    void disableRecovery() {
 
         Implementations.uninstall();
         RecoveryManager.manager().removeModule(lraRecoveryModule, false);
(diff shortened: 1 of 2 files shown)


mmusgrov:
@tomjenkinson I will remove the commented out import statement in the source file before I merge. I don't want to push a change yet since that would trigger a retest and you may have other change requests.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
```diff
@@ -62,7 +61,7 @@
     private Map<URI, Transaction> recoveringLRAs = new ConcurrentHashMap<>();
     private Map<URI, ReentrantLock> locks = new ConcurrentHashMap<>();
 
-    private static Map<String, String> participants = new ConcurrentHashMap<>();
+    private Map<String, String> participants = new ConcurrentHashMap<>();
```
tomjenkinson: I guess this is separate to the core JBTM in this PR?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
```diff
@@ -410,18 +404,18 @@ void enableRecovery(@Observes @Initialized(ApplicationScoped.class) Object init)
...
-    void disableRecovery(@Observes @Destroyed(ApplicationScoped.class) Object init) {
-        assert lraRecoveryModule != null;
+    @PreDestroy
+    void disableRecovery() {
+        if (lraRecoveryModule != null) {
 
-        Implementations.uninstall();
-        RecoveryManager.manager().removeModule(lraRecoveryModule, false);
-        lraRecoveryModule = null;
+            Implementations.uninstall();
```
tomjenkinson: I note that this is currently a no-op but I wonder if a separate JBTM should be removing from the RecordTypeManager tp reverse the install?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
```diff
@@ -410,18 +404,18 @@ void enableRecovery(@Observes @Initialized(ApplicationScoped.class) Object init)
...
+            RecoveryManager.manager().removeModule(lraRecoveryModule, false);
+            lraRecoveryModule = null;
 
-        if (LRALogger.logger.isDebugEnabled()) {
-            LRALogger.logger.debugf("LRAServicve.disableRecovery%n");
+            if (LRALogger.logger.isDebugEnabled()) {
+                LRALogger.logger.debugf("LRAServicve.disableRecovery%n");
```
tomjenkinson: I am not sure what the %n refers to here. I think it would be fine to correct the typo given it is debug code.

rts/lra/lra-coordinator-jar/pom.xml
```diff
@@ -103,6 +103,12 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>jakarta.enterprise</groupId>
+            <artifactId>jakarta.enterprise.cdi-api</artifactId>
```
tomjenkinson: I don't think this dependency should be required now? Or at least javax/annotation is not provided directly inside it. If a dependency is required I suspect it would be on one from https://search.maven.org/artifact/jakarta.annotation/jakarta.annotation-api

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
```diff
@@ -189,13 +188,9 @@ public void remove(String state, URI lraId) {
             lras.remove(lraId);
         }
 
-        if (recoveringLRAs.containsKey(lraId)) {
-            recoveringLRAs.remove(lraId);
-        }
+        recoveringLRAs.remove(lraId);
```
tomjenkinson: was this part a general tidy up or was it related to JBTM-3243? if general I think it needs a separate issue?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
```diff
@@ -189,13 +188,9 @@ public void remove(String state, URI lraId) {
...
 
-        if (locks.containsKey(lraId)) {
-            locks.remove(lraId);
-        }
+        locks.remove(lraId);
```
tomjenkinson: was this part a general tidy up or was it related to JBTM-3243? if general I think it needs a separate issue?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
mmusgrov: Well the bean is ApplicationScoped so using a static makes little difference.

tomjenkinson: Agreed, it looks right not to be static but is it related to https://issues.redhat.com/browse/JBTM-3243 is what I was getting at.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
mmusgrov: It wasn't introduced by the PR but I agree it's a good opportunity to correct it.

rts/lra/lra-coordinator-jar/pom.xml
mmusgrov: +1

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
mmusgrov: Really! Perhaps if I remember I will include it in a general tidy up sometime.

tomjenkinson: It's not caused by https://issues.redhat.com/browse/JBTM-3243 is it? But as our LRA implementation is not formally released perhaps just a separate commit would suffice?

mmusgrov: The problem with having to raise a JBTM against every trivial issue on a code base that is tracking an evolving pre-release spec is that is slows down progress to delivering a first cut of the s/w

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
mmusgrov: ditto

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
tomjenkinson: yeah, I tend to agree for the tidy up stuff. a separate commit though would be reasonable as it's not right to include in a commit with a JBTM that does not apply, WDYT?

mmusgrov:
> As well as the open discussions, I think a comment on why no test is provided would be useful.
> 
> Assuming all those can be readily reconciled, the premise of the fix works for me.

No test is provided because the issue is not present when we run against the Thorntail and WildFly containers (and these are the containers we use for Narayana testing).

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
tomjenkinson: I have revised on https://github.com/jbosstm/narayana/pull/1547#discussion_r365351134

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
mmusgrov: I provided an implementation for `Implementations.uninstall()`

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
mmusgrov: Well I think so since the PR is all about getting the semantics of an ApplicationScoped bean correct and using statics in such a bean is counter intuitive.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/Implementations.java
```diff
@@ -48,17 +48,20 @@ public int getType() {
...
-
-            _added = true;
+        if (typeMap == null) {
+            typeMap = new LRACompensatorMap();
+            RecordTypeManager.manager().add(typeMap);
         }
     }
 
     public static synchronized void uninstall() {
+        if (typeMap != null) {
```
mmusgrov: @tomjenkinson This is the implementation for Implementations.uninstall()

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
mmusgrov: Sure, a separate commit in the same PR would work for me (not in this one though since I reverted the change).

mmusgrov:
> Actually, I realised the commit message is wrong

Good spot, thanks

PR #1549, opened 2020-01-16
JBTM-3229 Fix up previous commit

mmusgrov:
https://issues.redhat.com/browse/JBTM-3229

The test for this is in the eclipse repo: https://github.com/eclipse/microprofile-lra/pull/267

### DIFF
--- a/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
+++ b/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
@@ -23,6 +23,7 @@ package io.narayana.lra.coordinator.domain.service;
 
 import com.arjuna.ats.arjuna.AtomicAction;
 import com.arjuna.ats.arjuna.coordinator.ActionStatus;
+import io.narayana.lra.LRAConstants;
 import io.narayana.lra.coordinator.domain.model.LRAData;
 import io.narayana.lra.logging.LRALogger;
 import com.arjuna.ats.arjuna.recovery.RecoveryManager;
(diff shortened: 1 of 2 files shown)


xstefank:
@mmusgrov thanks for catching this. You need to also update the LRA API and TCK snapshot dependencies as the old ones have been deleted. This if statement [1] probably needs to be adjusted also because of the order of the if statement is probably wrong. We should first check for after LRA listener and then only add participant if the tx is not recovering. But I would keep your current PR to catch the missing LRA URI faster.

[1] https://github.com/jbosstm/narayana/blob/LRA/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java#L571

mmusgrov:
There was just one test failure (org.eclipse.microprofile.lra.tck.TckContextTests.testForgetCalledForNestedParticipantsWhenParentIsClosed(TckContextTests.java:244)

Martin is working on that one so this PR should be good to merge now. @tomjenkinson can you take a look please.

The if statement that Martin referred to in his comment is fine and does not need changing.

xstefank:
@mmusgrov @tomjenkinson I can add a TCK test (since it will be really similar to the one that is enlisting LRA listener).

mmusgrov:
> The fix looks good to me, I would ask if a test for the functionality can be provided? It may well be that test should be added to the LRA TCK in which case a link to the issue number would be useful.

There is already a test for it (https://github.com/eclipse/microprofile-lra/pull/267/files#diff-3eeea840a3958c). Tests in the narayana code base should not duplicate the tests in the TCK (only tests that exercise our particular implementation).

mmusgrov:
> Thanks @mmusgrov , I assume our tests fail without this PR then, is that correct?

@tomjenkinson  The test that fails is something that Martin is working on. This PR is address a defect in the state model implementation. The test that shows the issue is in the microprofile-lra TCK in the eclipse repo.

PR #1550, opened 2020-01-17
JBTM-3231 Run both crash rec tests against the same server

mmusgrov:
https://issues.redhat.com/browse/JBTM-3231

There are two lra coordinator crash recovery tests but the second one often fails because the (arquillian) server cannot be started. I have not been able to duplicate the problem on my laptop so as an attempted workaround I have moved each test into its own junit test suite.

### DIFF
--- a/rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/LRACoordinatorRecovery1TestCase.java
+++ b/rts/lra/lra-coordinator-jar/src/test/java/io/narayana/lra/coordinator/LRACoordinatorRecovery1TestCase.java
@@ -53,14 +53,10 @@ import javax.ws.rs.core.Response;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.time.temporal.ChronoUnit;
 
 import static io.narayana.lra.coordinator.LRAListener.LRA_LISTENER_ACTION;
-import static io.narayana.lra.coordinator.LRAListener.LRA_LISTENER_KILL;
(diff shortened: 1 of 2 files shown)


rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
```diff
@@ -840,4 +840,8 @@ private void throwGenericLRAException(URI lraId, int statusCode, String message)
         throw new WebApplicationException(Response.status(statusCode)
                 .entity(String.format("%s: %s", lraId, message)).build());
     }
+
+    public boolean suspendLRA(URI lraId) {
```
tomjenkinson: This does not seem to be referenced elsewhere in the PR?

rts/lra/lra-coordinator-jar/pom.xml
```diff
@@ -186,9 +186,6 @@
                                 <lra-participant.port>9080</lra-participant.port>
                                 <server.jvm.args>${server.jvm.args}</server.jvm.args>
                             </systemPropertyVariables>
-                            <excludes>
-                              <exclude>AnnotationsTestCase.java</exclude>
```
tomjenkinson: @mmusgrov is this tidy up? If so can it be added to a separate commit (same PR is fine)

mmusgrov: I will delete the tidy up

mmusgrov: @tomjenkinson Try that latest push instead. The PR has passed 4 times (zero failures) so I think this fix looks positive.

PR #1559, opened 2020-02-02
JBTM 3239 Ensure post LRA actions are retried in the presence of failures

mmusgrov:
https://issues.redhat.com/browse/JBTM-3239

!LRA !MAIN !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF !NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

I have disabled the test for now while the PR is under review.

### DIFF
--- a/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
+++ b/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
@@ -36,6 +36,7 @@ import io.narayana.lra.ResponseHolder;
 import io.narayana.lra.logging.LRALogger;
 import org.eclipse.microprofile.lra.annotation.LRAStatus;
 import org.eclipse.microprofile.lra.annotation.ParticipantStatus;
+import org.eclipse.microprofile.lra.annotation.ws.rs.LRA;
 
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.client.AsyncInvoker;
@@ -85,12 +86,15 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
     private LRAService lraService;
     private ParticipantStatus status;
     private boolean accepted;
+    private Transaction lra;
 
     public LRARecord() {
     }
 
-    LRARecord(LRAService lraService, String lraId, String linkURI, String compensatorData) {
+    LRARecord(Transaction lra, LRAService lraService, String lraId, String linkURI, String compensatorData) {
         super(new Uid());
+
+        this.lra = lra;
 
         try {
             // if compensateURI is a link parse it into compensate,complete and status urls
@@ -119,6 +123,7 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
 
             this.lraId = new URI(lraId);
             this.parentId = lraService.getTransaction(this.lraId).getParentId();
+            this.status = ParticipantStatus.Active;
 
             this.lraService = lraService;
             this.participantPath = linkURI;
@@ -129,6 +134,10 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
             LRALogger.i18NLogger.error_invalidFormatToCreateLRARecord(lraId, linkURI);
             throw new WebApplicationException(lraId + ": Invalid LRA id: " + e.getMessage(), BAD_REQUEST);
         }
+    }
+
+    public void setLRA(Transaction lra) {
+        this.lra = lra;
     }
 
     String getParticipantPath() {
@@ -254,17 +263,22 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
     private int tryDoEnd(boolean compensate) {
         URI endPath;
 
+        if (isFinished()) {
+            return atEnd(status == ParticipantStatus.FailedToComplete || status == ParticipantStatus.FailedToCompensate
+                    ? TwoPhaseOutcome.FINISH_ERROR : TwoPhaseOutcome.FINISH_OK);
+        }
+
         if (ParticipantStatus.Compensating.equals(status)) {
             compensate = true;
         }
 
         if (compensateURI == null) {
-            return TwoPhaseOutcome.FINISH_OK;
+            return atEnd(TwoPhaseOutcome.FINISH_OK);
         }
 
         if (compensate) {
             if (isCompensated()) {
-                return TwoPhaseOutcome.FINISH_OK; // the participant has already compensated
+                return atEnd(TwoPhaseOutcome.FINISH_OK); // the participant has already compensated
             }
 
             endPath = compensateURI; // we are going to ask the participant to compensate
@@ -273,7 +287,7 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
             if (isCompelete() || completeURI == null) {
                 status = ParticipantStatus.Completed;
 
-                return TwoPhaseOutcome.FINISH_OK; // the participant has already completed
+                return atEnd(TwoPhaseOutcome.FINISH_OK); // the participant has already completed
             }
 
             endPath = completeURI;  // we are going to ask the participant to complete
@@ -291,7 +305,7 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
             int twoPhaseOutcome = retryGetEndStatus(endPath, compensate);
 
             if (twoPhaseOutcome != -1) {
-                return twoPhaseOutcome;
+                return atEnd(twoPhaseOutcome);
             }
         } else {
             httpStatus = tryLocalEndInvocation(endPath); // see if participant is in the same JVM
@@ -322,7 +336,7 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
 
                 if (httpStatus == Response.Status.GONE.getStatusCode()) {
                     updateStatus(compensate);
-                    return TwoPhaseOutcome.FINISH_OK; // the participant must have finished ok but we lost the response
+                    return atEnd(TwoPhaseOutcome.FINISH_OK); // the participant must have finished ok but we lost the response
                 }
 
                 if (response.hasEntity()) {
@@ -338,8 +352,8 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
                 httpStatus == Response.Status.INTERNAL_SERVER_ERROR.getStatusCode()) {
             // the body should contain a valid ParticipantStatus
             try {
-                return reportFailure(compensate, endPath.toASCIIString(),
-                        ParticipantStatus.valueOf(responseData).name());
+                return atEnd(reportFailure(compensate, endPath.toASCIIString(),
+                        ParticipantStatus.valueOf(responseData).name()));
             } catch (IllegalArgumentException ignore) {
                 // ignore the body and let recovery discover the status of the participant
             }
@@ -361,18 +375,88 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
                  * so if we were to return FINISH_ERROR recovery would not replay the log.
                  * To force the record to be eligible for recovery we return a heuristic hazard.
                  */
-                return TwoPhaseOutcome.HEURISTIC_HAZARD;
+                return atEnd(TwoPhaseOutcome.HEURISTIC_HAZARD);
             }
 
             status = ParticipantStatus.Completing; // recovery will figure out the status via the status url
 
-            return TwoPhaseOutcome.FINISH_ERROR;
+            return atEnd(TwoPhaseOutcome.FINISH_ERROR);
         }
 
         updateStatus(compensate);
 
         // if the the request is still in progress (ie accepted is true) let recovery finish it
-        return accepted ? TwoPhaseOutcome.HEURISTIC_HAZARD : TwoPhaseOutcome.FINISH_OK;
+        return atEnd(accepted ? TwoPhaseOutcome.HEURISTIC_HAZARD : TwoPhaseOutcome.FINISH_OK);
+    }
+
+    boolean isFinished() {
+        // nested participants must still be able to compensate even if they are closed
+        if (compensateURI == null) {
+            return afterURI != null;
+        }
+
+        switch (status) {
+            case Completed:
+                /* FALLTHRU */
+            case FailedToComplete:
+                return parentId == null; // completed nested LRAs must remain cancellable
+            case Compensated:
+                /* FALLTHRU */
+            case FailedToCompensate:
+                return true;
+            default:
+                return false;
+        }
+    }
+
+    private boolean afterLRARequest(URI target, String payload) {
+        try {
+            RequestBuilder builder = new RequestBuilder(target)
+                    .request()
+                    .header(LRA.LRA_HTTP_RECOVERY_HEADER, recoveryURI.toASCIIString());
+
+            if (target.equals(afterURI)) {
+                builder.header(LRA.LRA_HTTP_ENDED_CONTEXT_HEADER, lra.getId().toASCIIString());
+            } else {
+                builder.header(LRA.LRA_HTTP_CONTEXT_HEADER, lra.getId().toASCIIString());
+            }
+
+            builder.async(PARTICIPANT_TIMEOUT, TimeUnit.SECONDS);
+
+            ResponseHolder response = target.equals(forgetURI) ? builder.delete() : builder.put(payload, MediaType.TEXT_PLAIN);
+
+            if (response.getStatus() == 200) {
+                return true;
+            }
+        } catch (WebApplicationException e) {
+            if (LRALogger.logger.isInfoEnabled()) {
+                LRALogger.logger.infof("Could not notify URI at %s (%s)", target, e.getMessage());
+            }
+        }
+
+        return false;
+    }
+
+    private int atEnd(int res) {
+        // Only run the post LRA actions if both the LRA and participant are in an end state
+        // check the participant first since it will have been removed from one of the lists
+        if (!isFinished() || !lra.isFinished()) {
+            if (afterURI != null) {
+                return TwoPhaseOutcome.HEURISTIC_HAZARD;
+            }
+
+            return res;
+        }
+
+        // run post LRA actions
+        if (afterURI == null || afterLRARequest(afterURI, lra.getLRAStatus().name())) {
+            afterURI = null;
+
+            // the post LRA actions succeeded so remove the participant from the intentions list otherwise retry
+            return TwoPhaseOutcome.FINISH_OK;
+        }
+
+        return TwoPhaseOutcome.HEURISTIC_HAZARD;
     }
 
     private void updateStatus(boolean compensate) {
@@ -445,6 +529,15 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
 
                 // 200 and 410 are the only valid response code for reporting the participant status
                 if (response.getStatus() == Response.Status.GONE.getStatusCode()) {
+                    /*
+                     * The specification states (in section 3.2.10. Reporting the status of a participant):
+                     * If the participant has already responded successfully to an @Compensate or @Complete method
+                     * invocation then it MAY report 410 Gone HTTP status code
+                     *
+                     * This means that if the participant was asked to compensate then it has now compensated, or
+                     * if the participant was asked to complete then it has now completed.
+                     */
+                    status = compensate ? ParticipantStatus.Compensated : ParticipantStatus.Completed;
                     return TwoPhaseOutcome.FINISH_OK;
                 } else if (response.getStatus() == Response.Status.ACCEPTED.getStatusCode()) {
                     return TwoPhaseOutcome.HEURISTIC_HAZARD;
--- a/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
+++ b/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
@@ -30,8 +30,6 @@ import com.arjuna.ats.arjuna.coordinator.BasicAction;
 import com.arjuna.ats.arjuna.coordinator.RecordList;
 import com.arjuna.ats.arjuna.coordinator.RecordListIterator;
 import com.arjuna.ats.arjuna.coordinator.RecordType;
-import io.narayana.lra.RequestBuilder;
-import io.narayana.lra.ResponseHolder;
 import io.narayana.lra.logging.LRALogger;
 import com.arjuna.ats.arjuna.state.InputObjectState;
 import com.arjuna.ats.arjuna.state.OutputObjectState;
@@ -39,10 +37,8 @@ import com.arjuna.ats.internal.arjuna.thread.ThreadActionData;
 
 import io.narayana.lra.coordinator.domain.service.LRAService;
 import org.eclipse.microprofile.lra.annotation.LRAStatus;
-import org.eclipse.microprofile.lra.annotation.ws.rs.LRA;
 
 import javax.ws.rs.WebApplicationException;
-import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
 
@@ -56,7 +52,6 @@ import java.time.LocalDateTime;
 import java.time.ZoneOffset;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
@@ -66,7 +61,6 @@ import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantLock;
 
-import static io.narayana.lra.LRAHttpClient.PARTICIPANT_TIMEOUT;
 
 public class Transaction extends AtomicAction {
     private static final String LRA_TYPE = "/StateManager/BasicAction/TwoPhaseCoordinator/LRA";
@@ -75,7 +69,6 @@ public class Transaction extends AtomicAction {
     private URI parentId;
     private String clientId;
     private List<LRARecord> pending;
-    private List<AfterLRAListener> afterLRAListeners;
     private LRAStatus status;
     private String responseData;
     private LocalDateTime startTime;
@@ -157,16 +150,6 @@ public class Transaction extends AtomicAction {
             } else {
                 os.packBoolean(true);
                 os.packString(status.name());
-            }
-
-            if (afterLRAListeners == null) {
-                os.packInt(0);
-            } else {
-                os.packInt(afterLRAListeners.size());
-                for (AfterLRAListener participant : afterLRAListeners) {
-                    os.packString(participant.notificationURI.toASCIIString());
-                    os.packString(participant.recoveryId.toASCIIString());
-                }
             }
         } catch (IOException e) {
             return false;
@@ -261,17 +244,6 @@ public class Transaction extends AtomicAction {
             startTime = os.unpackBoolean() ? LocalDateTime.ofInstant(Instant.ofEpochMilli(os.unpackLong()), ZoneOffset.UTC) : null;
             finishTime = os.unpackBoolean() ? LocalDateTime.ofInstant(Instant.ofEpochMilli(os.unpackLong()), ZoneOffset.UTC) : null;
             status = os.unpackBoolean() ? LRAStatus.valueOf(os.unpackString()) : null;
-
-            int cnt = os.unpackInt();
-
-            if (cnt == 0) {
-                afterLRAListeners = null;
-            } else {
-                afterLRAListeners = new ArrayList<>();
-                for (int i = 0; i < cnt; i++) {
-                    afterLRAListeners.add(new AfterLRAListener(new URI(os.unpackString()), new URI(os.unpackString())));
-                }
-            }
 
             /*
              * If the time limit has already been reached then the difference between now and the scheduled
@@ -374,6 +346,21 @@ public class Transaction extends AtomicAction {
                 (status.equals(LRAStatus.Cancelling) || status.equals(LRAStatus.Closing));
     }
 
+    boolean isFinished() {
+        switch (status) {
+            case Closed:
+                /* FALLTHRU */
+            case Cancelled:
+                /* FALLTHRU */
+            case FailedToClose:
+                /* FALLTHRU */
+            case FailedToCancel:
+                return true;
+            default:
+                return false;
+        }
+    }
+
     private int cancelLRA() {
         return end(true);
     }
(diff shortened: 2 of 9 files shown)


mmusgrov:
> @mmusgrov the only failing test in the CI is the newly added `FailingParticipantCallsIT.testFailingAfterLRA` - http://narayanaci1.eng.hst.ams2.redhat.com/job/btny-pulls-narayana/PROFILE=LRA,jdk=jdk8.latest,label=linux/882/console

I haven't seen that one fail on my laptop so I'll need to investigate.

mmusgrov:
@tomjenkinson the last run passed. The LRA for the run that failed is still in the process of "LRAStatus.Closing" and the test only does one recovery pass (which ought to be sufficient). NB the test always passes on my laptop. So I am thinking that it is just a case of timing on CI is slower. I am going to push a change that extends the time we wait for participant callbacks. It is currently 1 second so I will try 10 seconds to help it pass CI. Note that having short timeouts is fine since one of the recovery passes will usually finish the LRA.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -445,6 +530,8 @@ private int retryGetEndStatus(URI endPath, boolean compensate) {
 
                 // 200 and 410 are the only valid response code for reporting the participant status
                 if (response.getStatus() == Response.Status.GONE.getStatusCode()) {
+                    // 412 implies that the participant did the correct thing and cleaned up so infer the status
```
tomjenkinson: This comment seems unassertive to me. Does 412 "imply" or "require" that the participant did this thing. Also "infer" could imply an element of doubt, is it possible to express as "determine". Are we certain of the outcome here or are there cases where valid responses could be misinterpreted?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -519,6 +477,10 @@ private int doEnd(boolean cancel) {
             status = toLRAStatus(res);
         }
 
+        // it would be less invasive to allow recoovery to run post LRA actions
```
tomjenkinson: Perhaps the TCK could have an RFE raised?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
```diff
@@ -173,7 +173,8 @@ public void finished(Transaction transaction, boolean fromHierarchy) {
             // the LRA is top level or it's a nested LRA that was closed by a
             // parent LRA (ie when fromHierarchy is true) then it's okay to forget about the LRA
 
-            if (transaction.afterLRANotification()) {
+            if (!transaction.hasPendingActions()) {
+                // this call is only required to clean up cached LRAs. Follow up with a JBTM to avoid this
```
tomjenkinson: Does the JBTM exist yet? It would be useful to reference the actual JBTM in the comment

mmusgrov:
> @mmusgrov please can you clarify how the status can be determined to be "LRAStatus.Closing" when receiving 1 back from https://github.com/jbosstm/narayana/pull/1559/files#diff-9f2aa1dd7328a722f2cd7e794e184e4eR65.

I cannot find that. Are you looking at a preexisting test in the same directory (namely NonRootLRAParticipantIT)

> 
> Also, the test assumes it will get called three times, why do we return INTERNAL_SERVER_ERROR twice in FailingAfterLRAListener#afterLRA?

The test is checking that the implementation will keep retrying the post LRA notification whenever it returns anything other than 200 (the spec says "If the method annotated with @AfterLRA returns an unexpected HTTP status or never reaches the caller then the implementation MUST invoke the same method again."

The test is coded to fail the post LRA notification three times and then succeed:
https://github.com/jbosstm/narayana/pull/1559/files#diff-9f2aa1dd7328a722f2cd7e794e184e4eR55

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
mmusgrov: I reworded the comment and referenced the specification

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
mmusgrov: This PR runs agains MP-LRA RC1 but I notice that the latest snapshot does run a recovery pass now. Therefore I have reworded the comment.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
mmusgrov: I raised [JBTM-3250](https://issues.redhat.com/browse/JBTM-3250) and reworded the comment.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -361,18 +375,89 @@ private int tryDoEnd(boolean compensate) {
...
+            default:
+                return false;
+        }
+    }
+
+    private boolean request(URI target, Transaction lra, String payload) {
+        try {
+            RequestBuilder builder = new RequestBuilder(target)
+                    .request()
+                    .header(LRA.LRA_HTTP_ENDED_CONTEXT_HEADER, lra.getId().toASCIIString())
```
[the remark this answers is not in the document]
mmusgrov: Okay, I fixed the choice of headers depending upon whether or not we are running the AfterLRA notification.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -361,18 +375,89 @@ private int tryDoEnd(boolean compensate) {
...
+            case Compensated:
+                /* FALLTHRU */
+            case FailedToCompensate:
+                return true;
+            default:
+                return false;
+        }
+    }
+
+    private boolean request(URI target, Transaction lra, String payload) {
```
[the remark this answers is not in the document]
mmusgrov: I used request so that the same method could replace the other REST calls in the class. But I will reword the method name since I have not done that in this PR.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -361,18 +375,89 @@ private int tryDoEnd(boolean compensate) {
...
+            if (LRALogger.logger.isInfoEnabled()) {
+                LRALogger.logger.infof("Could not notify URI at %s (%s)", target, e.getMessage());
+            }
+        }
+
+        return false;
+    }
+
+    private int atEnd(int res) {
+        URI nested = compensateURI != null ? extractParentLRA(compensateURI) : null;
```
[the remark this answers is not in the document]
mmusgrov: The method understands the narayana LRA format for nested LRAs acting as participants. But since I only need the parent in order to deliver the forget notification and since we don't have a test for that in LRA RC1 (you added it after that in issue 208 / pr 258) I have removed the logic. We can add it back in again when we add an implementation of the MP-LRA test [TckContextTests#testForgetCalledForNestedParticipantsWhenParentIsClosed](https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckContextTests.java#L218)

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -361,18 +375,89 @@ private int tryDoEnd(boolean compensate) {
...
+            if (afterURI != null || notifyForget) {
+                return TwoPhaseOutcome.HEURISTIC_HAZARD;
+            }
+
+            return res;
+        }
+
+        // run post LRA actions
+        boolean r1 = afterURI == null || request(afterURI, lra, lra.getLRAStatus().name());
+        boolean r2 = !notifyForget || request(forgetURI, lra, lra.getLRAStatus().name());
```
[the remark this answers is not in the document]
mmusgrov: Okay

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -532,12 +494,87 @@ private int doEnd(boolean cancel) {
...
+    }
+
+    protected boolean isInEndState() {
+        if (status == LRAStatus.Cancelling && allFinished(heuristicList, failedList)) {
+            status = LRAStatus.Cancelled;
+        } else if (status == LRAStatus.Closing && allFinished(heuristicList, failedList)) {
+            status = LRAStatus.Closed;
+        }
+
+        switch (status) {
```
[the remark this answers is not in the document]
mmusgrov: Thanks

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -532,12 +494,87 @@ private int doEnd(boolean cancel) {
...
+    protected void checkParticipant(RecordList participants) {
+        RecordListIterator i = new RecordListIterator(participants);
+        AbstractRecord r;
+
+        while ((r = i.iterate()) != null) {
+            if (r instanceof LRARecord) {
+                LRARecord rec = (LRARecord) r;
+
+                rec.setLraService(getLraService());
+                rec.setLRA(this);
```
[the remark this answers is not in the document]
mmusgrov: This value is set by the recovery module during each recovery pass so it does not need to be durable.

rts/lra/lra-service-base/src/main/java/io/narayana/lra/LRAHttpClient.java
```diff
@@ -19,7 +19,7 @@
 import java.util.concurrent.TimeoutException;
 
 public class LRAHttpClient {
-    public static final long PARTICIPANT_TIMEOUT = 1; // number of seconds to wait for requests
+    public static final long PARTICIPANT_TIMEOUT = 10; // number of seconds to wait for requests
```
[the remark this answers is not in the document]
mmusgrov: Well 1 second should be more than enough but the new test on the existing CI infrastructure intermittently fails because the 1 second time limit is breached. I have not studied your test in detail to find out why the callbacks sometimes take longer than 1 second (or whether the test is making certain assumptions) but I made it 10 s for expediency. Note that the shorter timelimit should work if the tests run sufficient recovery passes (but I did not spend any time investigating).

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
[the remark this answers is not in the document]
xstefank: it seems that you've removed this line. LRARecord already should have parentId in the field so no need to extract it.

xstefank:
it is targetted for 3 calls because afterLRA notifications were delivered twice - https://issues.redhat.com/browse/JBTM-3163. I am not sure if this is still occurring with this PR. But as Mike mentioned it shouldn't influence the test. The recovery is just replayed two times.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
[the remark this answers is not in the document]
xstefank: also is it needed to save Transaction in LRARecord? there are individual fields already present (lraId, recoveryId, parentId) so only thing additional thing needed for afterLRA is the outcome of the LRA.

rts/lra/lra-service-base/src/main/java/io/narayana/lra/LRAHttpClient.java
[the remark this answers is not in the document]
mmusgrov: I have changed it to 2 seconds. When we take the latest MP LRA snapshot we can go back to 1 second since the latest TCK runs a recovery pass which would pick it up. The  RC1 TCK doesn't retry you see (also see my [earlier comment](https://github.com/jbosstm/narayana/pull/1559#issuecomment-581549705)).

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/resource/FailingAfterLRAListener.java
```diff
@@ -0,0 +1,67 @@
...
+    @LRA(LRA.Type.REQUIRED)
+    public String doWorkInLRA(@HeaderParam(LRA.LRA_HTTP_CONTEXT_HEADER) URI lraId) {
+        return lraId.toASCIIString();
+    }
+
+    @PUT
+    @Path("after-lra")
+    @AfterLRA
+    public Response afterLRA(@HeaderParam(LRA.LRA_HTTP_ENDED_CONTEXT_HEADER) URI endedLRAId) {
+        if (afterLRACounter.getAndIncrement() < 2) {
```
tomjenkinson: @mmusgrov please can you add a reference to https://issues.redhat.com/browse/JBTM-3163 here so we know why it is doing this twice. IIUC the [comment](https://github.com/jbosstm/narayana/pull/1559#issuecomment-582071967) from @xstefank then if that is fixed then [this](https://github.com/jbosstm/narayana/pull/1559/files#diff-f1f5ec6c4099230d8d191d68c62295cdR100) line will probably start failing as afterLRA will only be called twice.

xstefank: @tomjenkinson no, it shouldn't start failing. The recovery will just be replayed two times.

tomjenkinson: If I understand you correctly, once https://issues.redhat.com/browse/JBTM-3163 is fixed instead of two notifications for one recovery we could go down to "getAndIncrement < 1"

Am I right in thinking therefore that https://github.com/jbosstm/narayana/pull/1559/files#diff-6ba4dd5973413b54d0299d8935a156c1R54-R56 is not actually required or used in this PR? If so I think we should remove that part? I will create a discussion localised on that line for this.

tomjenkinson: Actually I see the code may well apply if the scan is in progress

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
```diff
@@ -0,0 +1,95 @@
...
+        LRALogger.logger.info("LRA " + lraId + "has finished the recovery " + counter);
+    }
+
+    public boolean waitForEndPhaseReplay(URI lraId) {
+        String host = lraId.getHost();
+        int port = lraId.getPort();
+        if (!recoverLRAs(host, port, lraId)) {
+            // first recovery scan probably collided with periodic recovevery which started
+            // before the test execution so try once more
+            return recoverLRAs(host, port, lraId);
```
tomjenkinson: @xstefank @mmusgrov perhaps we could turn off automated recovery for this test if we want more predictability

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/resource/FailingAfterLRAListener.java
xstefank: It is possible to move it to < 1 after https://issues.redhat.com/browse/JBTM-3163 is resolved, but not required. 

Sorry, the second link doesn't probably take me where you've intended. But if it's NarayanaLRARecovery then that code is correct. The two invocations there are about something different -- namely the clash with periodic recovery which may sometime happen. That is a different issue which was decided that it will not be resolved some time ago.

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
xstefank: @tomjenkinson AfterLRA can be replayed several times successfully. The test is just testing that it will be eventually replayed, but the specification allows it to be replayed even 4 or 5 times (unbounded). So I think the test is correct.

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/resource/FailingAfterLRAListener.java
tomjenkinson: I understand this better now. Currently we need 3 total calls to this method  (two from the normal cleanup, one (though will receive 2) from recovery)

In the future we need 2 notifications (one from the normal cleanup, one from recovery) so will be able to change this line to `if (afterLRACounter.getAndIncrement() < 1)` after JBTM-2863 is solved. As such it needs the comment and a reference to the bit that needs cleanup adding to the JBTM

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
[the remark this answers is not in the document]
mmusgrov: ... But since I only need the parent in order to deliver the forget notification and since we don't have a test for that in LRA RC1 (you added it after that in issue 208 / pr 258) I have removed the logic. We can add it back in again when we add an implementation of the MP-LRA test TckContextTests#testForgetCalledForNestedParticipantsWhenParentIsClosed

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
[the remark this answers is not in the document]
mmusgrov: Not sure what you mean. But the LRA is set dynamically on each recovery pass so there is no need to make it durable.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
tomjenkinson: Thanks @mmusgrov I note the previous comment indicated invasiveness of calling this here but timeliness now? TBH I don't quite understand why they are considered "Post" actions as it seems to be the phase that actually completes the LRA?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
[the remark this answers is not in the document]
xstefank: the passed transaction (which is passed on every recovery pass to every record) is used only in afterLRA and nested forget calls (post LRA actions). But LRARecord already has fields for lraId, parentId, recoveryId. So is it really necessary to pass the Transaction object on each pass?

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/resource/FailingAfterLRAListener.java
mmusgrov: I am not clear on what this discussion is about. The test says `afterLRACounter.getAndIncrement() < 2` so it will return a 500 error status 2 times followed by a 200 status. The test is verifying that the LRA implementation will keep attempting the AfterLRA notification until it receives an HTTP 200 OK response. So in all we expect 2+1=3 notifications. The test FailingParticipantCallsIT asserts that the callback was invoked 3 times.

On the issue of JBTM-3163 that should have been incidentally resolved by the PR but it seems irrelevant to this PR.

mmusgrov:
> > > @mmusgrov please can you clarify how the status can be determined to be "LRAStatus.Closing" when receiving 1 back from https://github.com/jbosstm/narayana/pull/1559/files#diff-9f2aa1dd7328a722f2cd7e794e184e4eR65.
> >
> >
> > I cannot find that. Are you looking at a preexisting test in the same directory (namely NonRootLRAParticipantIT)
>
> My question was in conjunction with your [comment](https://github.com/jbosstm/narayana/pull/1559#issuecomment-581549705) about the status being `LRAStatus.Closing', this output from the failed job:

Okay I should have said that the AfterLRA notifications were not delivered in a timely manner. I sumised this based on my knowledge of the code and the fact that the test always passed on my laptop but failed on CI. So I concluded that 1 second wasn't long enough on our CI. I have since changed the time to 2 seconds.

>
> ```
> java.lang.AssertionError: expected:<3> but was:<1>
>       at io.narayana.lra.arquillian.FailingParticipantCallsIT.testFailingAfterLRA(FailingParticipantCallsIT.java:100)
> ```

I surmised that the missing two calls were due to the test finishing early. It only does one recovery pass and if that pass failed due to 1 second being too short a time then the test fails. Changing the time the implentation waits for the callback to be delivered from 1 to 2 seconds fixed it. Alternatively one could hypothesise that the tardiness on CI is intermittent so replaying recovery may have resolved it too.

>
> And inspection of where that value of "1" could come [from](https://github.com/jbosstm/narayana/pull/1559/files#diff-9f2aa1dd7328a722f2cd7e794e184e4eR65).
>

Yes it was 1 because one of the callbacks did succeed but the test is looking for three of them.

> My understanding is an LRA has a status of [`LRAStatus.Closing`](https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/LRAStatus.java#L64) only until all participants have one of the completed states. This is relevant here as the `@AfterLRA'` resource indicated it had been called once. Although I seem to remember `@AfterLRA` were to be classed as participants, I can find in the spec: `[Note that the resource does not need to be a participant in order to receive this notification](https://github.com/eclipse/microprofile-lra/blob/master/spec/src/main/asciidoc/microprofile-lra-spec.adoc).' so it's confusing to me.

The spec does not require listeners (the resource has an @AfterLRA) to also be participants (the resource has an @Compesate).

>
> Getting clarification on this doesn't need to hold up the PR being merged as the issue at hand is about retrying notifications which seems to be happening. I do want to clarify if at the time of the CI failure I was looking at the status of the LRA was indeed `LRAStatus.Closing` or not. If it was (and `@AfterLRA' are participants) then we need to make sure we get the spec text updated to clarify that the are participants.

So the spec is clear on this point (so does not need an modification).

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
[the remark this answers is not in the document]
mmusgrov: It is used in the following line: https://github.com/jbosstm/narayana/pull/1559/files#diff-0f226ae6945188e648e485306754f30cR443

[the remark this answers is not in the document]
xstefank: ok, I see. Thanks.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
mmusgrov: Invasive in the sense that it bulks up the number of changes in the PR. I changed it to timeliness since it is a better functional argument for it.

mmusgrov: The are called "Post LRA" actions since they are performed after the LRA has reached a terminal state.

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/resource/FailingAfterLRAListener.java
tomjenkinson: The discussion is fundamentally related to the "2" in your formula "2+1=3", in other words why we have the need to return the ISE twice rather than just once.

The reason I understand that the test originally needed to skip "2" calls before was because of a duplicate notification (JBTM-3163) so the two calls would have been from the same routine execution in Narayana ([this](https://issues.redhat.com/browse/JBTM-3163?focusedCommentId=13764034&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13764034)?) and so returning OK to the second notification would not have tested the call was repeated.

The premise of the observation is related to why the test not just say:
```
if (afterLRACounter.getAndIncrement() < 1) {
     return Response.status(Response.Status.INTERNAL_SERVER_ERROR).build();
}
return Response.ok().build();
```

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
tomjenkinson: Sure, but under test conditions in Narayana we can control this to reduce the number of retries. If we turned off background recovery and triggered the recovery manually you would only need one call to recoverLRAs(host, port, lraId).

However, given the recovery manager is (I assume) running in a service similar to in WildFly then actually this is not necessarily possible. What might be possible is to delay the background recovery thread from starting for some arbitrarily long period using https://github.com/jbosstm/narayana/blob/1e17caa16b5efb9b16cdb2d8489a1b93722ebdd5/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/RecoveryEnvironmentBean.java#L47

xstefank: @tomjenkinson thanks, I didn't know about this. We can use I think. But let's leave that to some other PR/issue, please.

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/resource/FailingAfterLRAListener.java
tomjenkinson: @mmusgrov I think this is the last area to get clarity on. Do we need to return ISE on the first 2 calls to `afterLRA` or could it just be one now? And what is the plan to resolve JBTM-3163, will it be linked to this PR somehow?

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
tomjenkinson: Yeah, seems reasonable as the test is valid either way I think as we are getting the extra notifications.

mmusgrov: We already [tweak that config option](https://github.com/jbosstm/narayana/blob/master/rts/lra/lra-test/pom.xml#L118). You could change it now on the LRA branch or just wait for the merge into master. I don't think there is any particular urgency.

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/resource/FailingAfterLRAListener.java
mmusgrov: Well okay I changed it to 2 to get faster test turnaround time on testing.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -361,18 +375,88 @@ private int tryDoEnd(boolean compensate) {
...
+        if (!isFinished() || !lra.isFinished()) {
+            if (afterURI != null) {
+                return TwoPhaseOutcome.HEURISTIC_HAZARD;
+            }
+
+            return res;
+        }
+
+        // run post LRA actions
+        if (afterURI == null || afterLRARequest(afterURI, lra, lra.getLRAStatus().name())) {
```
tomjenkinson: @mmusgrov please can you expand on the situation where afterURI can be null at this point and what the value of res could be? 

If afterURI is null we will return TwoPhaseOutcome.FINISH_OK but res might not always be that from what I can see in above calls to atEnd(int) - including at least one call that might pass in HEURISTIC_HAZARD. Maybe in those cases afterURI could never be null? But then what about the one which uses a value from retryGetEndStatus or reportFailure

mmusgrov: This bit of code will only run if it is time to run the post LRA actions. If those actions succeed then we are finished with this record and we return FINISH_OK which removes this record from the intentions list.

In more detail:

1. If the status passed to the atEnd routine is HEURISTIC_HAZARD then the previous check (line 443) `if (!isFinished() || !lra.isFinished()) {` will have been true so the bit of code you have highlighted will not be reached and we return HH or the res value depending on whether or not afterURI is null.
2. The reportFailure route will always return TwoPhaseOutcome.FINISH_OK (since the spec only requires us to log failures). But do note that we have JBTM-3247 "Failed LRA records are reported but they not kept" which when fixed will keep those records (and the fix for that should take account of your current comment).
3. The other route you mention is retryGetEndStatus then:
   - if this returns HH then it is covered by point 1) above
   - if it returns FE (FINISH_ERROR) then it ought to have been covered by point 2) above where failures should be reported but the log can be removed. Although I do notice in this routine that the failure will not be reported using a logging statement but that is part of the fix for JBTM-3247 (keep failed records).

tomjenkinson: Thanks Mike. Regarding the missing logging you mentioned "will not be reported using a logging statement" - do you mean this particular detection of the LRAStatus being FailedToClose or FailedToCancel was not logged but another one would have been or that it might not have been reported at all? If the latter, it would seem different to https://issues.redhat.com/browse/JBTM-3247

Please can you add the conditions necessary to remember to verify JBTM-3247 text so we remember to check them?

mmusgrov: I mean not logged at all. I will send all failed outcomes through the reportFailure() route so that they will always be reported on the console. But note that JBTM-3247 keeps the failed records and they are available for inspection using the RecoveryCoordinator resource in the same way that we allow inspection of recovering LRAs. At least that is the way I have resolved JBTM-3247. But my fix for JBTM-3247 requires this current PR so can I have your approval so that I can raise a PR for it.

tomjenkinson: Thanks Mike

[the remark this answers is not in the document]
tomjenkinson: @mmusgrov is there is some reason the private method would need the parameters?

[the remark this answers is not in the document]
mmusgrov: It can use the field

PR #1560, opened 2020-02-19
JBTM-3231 LRA testrecovery2 fails

mayankkunwar:
Here is the link for the issue : https://issues.redhat.com/browse/JBTM-3231

LRA !MAIN !AS_TESTS !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS !TOMCAT !JACOCO

### DIFF
--- a/rts/lra/lra-coordinator/pom.xml
+++ b/rts/lra/lra-coordinator/pom.xml
@@ -51,6 +51,7 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>${surefire-plugin.version}</version>
                 <configuration>
+                    <redirectTestOutputToFile>false</redirectTestOutputToFile>
                     <systemProperties>
                         <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
                     </systemProperties>
--- a/rts/lra/lra-test/pom.xml
+++ b/rts/lra/lra-test/pom.xml
@@ -23,7 +23,12 @@
     <properties>
         <lra.coordinator.exec.plugin.phase>none</lra.coordinator.exec.plugin.phase> <!-- not starting coordinator by default, need -Parq -->
         <lra.coordinator.port>8080</lra.coordinator.port>
-        <lra.coordinator.trace.params></lra.coordinator.trace.params>
+        <lra.coordinator.log.console.disable></lra.coordinator.log.console.disable>
+        <lra.coordinator.log.file.enable></lra.coordinator.log.file.enable>
+        <lra.coordinator.log.file.path></lra.coordinator.log.file.path>
+        <lra.coordinator.log.file.level></lra.coordinator.log.file.level>
+        <lra.coordinator.log.file.module.arjuna></lra.coordinator.log.file.module.arjuna>
+        <lra.coordinator.log.file.module.narayana></lra.coordinator.log.file.module.narayana>
         <lra.coordinator.debug.params></lra.coordinator.debug.params> <!-- has content when -Ddebug.coordinator is specified -->
         <lra.coordinator.debug.port>8787</lra.coordinator.debug.port>
         <lra.coordinator.jar.path>${project.basedir}/../lra-coordinator/target/lra-coordinator-runner.jar</lra.coordinator.jar.path>
@@ -75,7 +80,17 @@
                 </property>
             </activation>
             <properties>
-                <lra.coordinator.trace.params>-Dthorntail.logging=TRACE</lra.coordinator.trace.params>
+                <!-- # Disabling console trace level logging (By default logger level is INFO)-->
+                <lra.coordinator.log.console.disable>-Dquarkus.log.console.enable=false</lra.coordinator.log.console.disable>
+                <!--# Enabling file trace logging-->
+                <lra.coordinator.log.file.enable>-Dquarkus.log.file.enable=true</lra.coordinator.log.file.enable>
+                <lra.coordinator.log.file.level>-Dquarkus.log.file.level=TRACE</lra.coordinator.log.file.level>
+                <!--# Send output to a trace.log file under the target/failsafe-reports/ directory-->
+                <lra.coordinator.log.file.path>-Dquarkus.log.file.path=${project.build.directory}/failsafe-reports/lra-coordinator-trace.log</lra.coordinator.log.file.path>
+                <!--# Setting categories to TRACE level-->
+                <lra.coordinator.log.file.module.arjuna>-Dquarkus.log.category."com.arjuna".level=TRACE</lra.coordinator.log.file.module.arjuna>
+                <lra.coordinator.log.file.module.narayana>-Dquarkus.log.category."io.narayana.lra".level=TRACE</lra.coordinator.log.file.module.narayana>
+
             </properties>
         </profile>
 
@@ -114,7 +129,12 @@
                                         <argument>${java.home}/bin/java</argument>
                                         <argument>-Djava.net.preferIPv4Stack=true</argument>
                                         <argument>${lra.coordinator.debug.params}</argument>
-                                        <argument>${lra.coordinator.trace.params}</argument>
+                                        <argument>${lra.coordinator.log.console.disable}</argument>
+                                        <argument>${lra.coordinator.log.file.enable}</argument>
+                                        <argument>${lra.coordinator.log.file.level}</argument>
+                                        <argument>${lra.coordinator.log.file.path}</argument>
+                                        <argument>${lra.coordinator.log.file.module.arjuna}</argument>
+                                        <argument>${lra.coordinator.log.file.module.narayana}</argument>
                                         <argument>-DRecoveryEnvironmentBean.periodicRecoveryPeriod=10</argument>
                                         <argument>-DRecoveryEnvironmentBean.recoveryBackoffPeriod=5</argument>
                                         <argument>-DObjectStoreEnvironmentBean.objectStoreDir=${project.build.directory}/lra_txn_logs</argument>
--- a/rts/lra/pom.xml
+++ b/rts/lra/pom.xml
@@ -36,6 +36,9 @@
 
         <test.logs.to.file>false</test.logs.to.file>
         <skipITs>true</skipITs> <!-- integration tests are disabled by default, use profile to run 'arq' -->
+
+        <thorntail.test.logging.category.arjuna></thorntail.test.logging.category.arjuna>
+        <thorntail.test.logging.category.narayana></thorntail.test.logging.category.narayana>
     </properties>
 
     <repositories>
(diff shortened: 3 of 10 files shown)


mayankkunwar:
Okay, sure @tomjenkinson I will follow it from next time.

rts/lra/lra-test/pom.xml
```diff
@@ -75,7 +75,8 @@
                 </property>
             </activation>
             <properties>
-                <lra.coordinator.trace.params>-Dthorntail.logging=TRACE</lra.coordinator.trace.params>
+<!--                <lra.coordinator.trace.params>-Dthorntail.logging=TRACE</lra.coordinator.trace.params>-->
```
ochaloup: Could you please remove this line here? I think it's not necessary here anymore when the coordinator runs as a quarkus application.

scripts/hudson/narayana.sh
```diff
@@ -56,6 +56,7 @@ function init_test_options {
     [ x"$CODE_COVERAGE_ARGS" != "x" ] || CODE_COVERAGE_ARGS=""
     [ $ARQ_PROF ] || ARQ_PROF=arq	# IPv4 arquillian profile
     [ $IBM_ORB ] || IBM_ORB=0
+    [ $ENABLE_TRACE_LOGS ] || ENABLE_TRACE_LOGS=" -Ptrace.thorntail.test -Ptrace.lra.coordinator"
```
ochaloup: Maybe this variable should be named more precisely? What I mean is that this `ENABLE_TRACE_LOGS` is meant to work really only for the `rts/lra` maven module. It's not beneficial for any other Narayana maven module. Could be rather something like `ENABLE_LRA_TRACE_LOGS`?

rts/lra/lra-test/pom.xml
mayankkunwar: Okay, I will remove this.

scripts/hudson/narayana.sh
mayankkunwar: Okay, I will the changes.

mayankkunwar:
> I can see how trace logging to the console is enabled but I cannot see how send the output to a file?
> I know we do not currently have any logging documentation but please could you add a comment to one of the poms indicating how to enable logging to a file.

@mmusgrov I think logging to a file depends on where the narayana is deployed, like here in this case, it is used by Quarkus

ochaloup:
> > @mmusgrov I think logging to a file depends on where the narayana is deployed, like here in this case, it is used by Quarkus

@mayankkunwar 
The logging to file is controlled by configuration of surefire/failsafe plugin. For the `rts/lra/pom.xml` can be used the `-Dtest.logs.to.file=true` (see https://github.com/jbosstm/narayana/blob/master/rts/lra/pom.xml#L109)
Maybe defining that property as default for the Narayana CI jobs is not a bad idea.

mayankkunwar:
> > > @mmusgrov I think logging to a file depends on where the narayana is deployed, like here in this case, it is used by Quarkus
> 
> @mayankkunwar
> The logging to file is controlled by configuration of surefire/failsafe plugin. For the `rts/lra/pom.xml` can be used the `-Dtest.logs.to.file=true` (see https://github.com/jbosstm/narayana/blob/master/rts/lra/pom.xml#L109)
> Maybe defining that property as default for the Narayana CI jobs is not a bad idea.

@ochaloup 
Oh! yes I totally forgot about this. Thanks

mmusgrov:
@ochaloup The intention was to be able to enable trace logging for the CI job only and to be able to capture the resulting logs for later analysis. Otherwise we only log at info level.

ochaloup:
@mmusgrov I see. And only for clarification the CI job should log for `TRACE` for LRA container and for the test client side as well, right?

Then if I understand it right @mayankkunwar the basic configuration even for the client should be `INFO`. That would probably means that the `java.util.logging.config.file` should  be configurable via some `-D` option as well (and not hardcoded as it's now to be default with `FINEST`).

rts/lra/lra-coordinator/pom.xml
```diff
@@ -51,6 +51,7 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>${surefire-plugin.version}</version>
                 <configuration>
+                    <redirectTestOutputToFile>${test.logs.to.file}</redirectTestOutputToFile>
```
ochaloup: please, remove this line. The `lra-coordinator` has no tests included under the directory structure and this is not needed.

rts/lra/pom.xml
```diff
@@ -101,6 +104,12 @@
               <groupId>org.sonatype.plugins</groupId>
               <artifactId>nexus-staging-maven-plugin</artifactId>
             </plugin>
+
+<!--            This plugin is used for file logging, when the logging is enabled. But logging is only enabled when CI job is executed.
```
ochaloup: I would vote for removing this comment completely as for anybody who understands the `maven` it's superfluous.
If you want to comment then don't say things like "this plugin is used for file logging" which is just not true.
Maybe you can define some property definitions like `test.logs.to.file` but it's still a bit dangerous (by my perspective) to say about default value etc. as those are things which change.

rts/lra/pom.xml
```diff
@@ -36,6 +36,9 @@
 
         <test.logs.to.file>false</test.logs.to.file>
         <skipITs>true</skipITs> <!-- integration tests are disabled by default, use profile to run 'arq' -->
+
+        <thorntail.test.logging></thorntail.test.logging>
+        <java.util.file.logging.level></java.util.file.logging.level>
```
ochaloup: please, remove this property completely. From what I know (if I'm not badly mistaken somehow) the value of `ConsoleHandler` has first no impact to be passed as `-D` command line property. Second has for sure no impact to thorntail logging setup.

rts/lra/lra-test/pom.xml
```diff
@@ -75,7 +80,17 @@
...
             </activation>
             <properties>
-                <lra.coordinator.trace.params>-Dthorntail.logging=TRACE</lra.coordinator.trace.params>
+<!--                # Disabling console logging-->
+                <lra.coordinator.log.console.disable>-Dquarkus.log.console.enable=false</lra.coordinator.log.console.disable>
+<!--                # Enabling file logging-->
+                <lra.coordinator.log.file.enable>-Dquarkus.log.file.enable=true</lra.coordinator.log.file.enable>
+                <lra.coordinator.log.file.level>-Dquarkus.log.file.level=TRACE</lra.coordinator.log.file.level>
+<!--                # Send output to a trace.log file under the target/failsafe-reports/ directory-->
+                <lra.coordinator.log.file.path>-Dquarkus.log.file.path=target/failsafe-reports/trace.log</lra.coordinator.log.file.path>
```
ochaloup: please rename the `trace.log` to `lra-coordinator-trace.log` or similar

rts/lra/lra-test/pom.xml
```diff
@@ -75,7 +80,17 @@
                 </property>
             </activation>
             <properties>
-                <lra.coordinator.trace.params>-Dthorntail.logging=TRACE</lra.coordinator.trace.params>
+<!--                # Disabling console logging-->
+                <lra.coordinator.log.console.disable>-Dquarkus.log.console.enable=false</lra.coordinator.log.console.disable>
+<!--                # Enabling file logging-->
+                <lra.coordinator.log.file.enable>-Dquarkus.log.file.enable=true</lra.coordinator.log.file.enable>
+                <lra.coordinator.log.file.level>-Dquarkus.log.file.level=TRACE</lra.coordinator.log.file.level>
+<!--                # Send output to a trace.log file under the target/failsafe-reports/ directory-->
```
ochaloup: please, use the same indentation for the `<!--` as for the rest of the upper and lower lines

rts/lra/lra-test/pom.xml
ochaloup: I would recommend to use to use property `${project.build.directory}/` rather than the `target/` constant

rts/lra/lra-test/pom.xml
```diff
@@ -23,7 +23,12 @@
     <properties>
         <lra.coordinator.exec.plugin.phase>none</lra.coordinator.exec.plugin.phase> <!-- not starting coordinator by default, need -Parq -->
         <lra.coordinator.port>8080</lra.coordinator.port>
-        <lra.coordinator.trace.params></lra.coordinator.trace.params>
+        <lra.coordinator.log.console.disable></lra.coordinator.log.console.disable>
```
ochaloup: I'm a bit concerned about so many log properties. Would you mind to merge it under just one? E.g. `lra.coordinator.log.params` or similar? And then put all the configuration on one line rather than splitting it over multiple.

But this could be only my personal feeling on having the properties too bloated.

rts/lra/lra-coordinator/pom.xml
mayankkunwar: okay

rts/lra/pom.xml
mayankkunwar: Okay, will remove this comment

rts/lra/pom.xml
mayankkunwar: okay

rts/lra/lra-test/pom.xml
mayankkunwar: will do the same

rts/lra/lra-test/pom.xml
mayankkunwar: okay

rts/lra/lra-test/pom.xml
mayankkunwar: okay

rts/lra/lra-test/pom.xml
mayankkunwar: okay, will make the changes

mayankkunwar: I tried merging them, but after that functionality stopped working. So for now I will leave this as it is.

rts/lra/pom.xml
```diff
@@ -170,6 +177,18 @@
...
 
+        <profile>
+            <id>trace.thorntail.test</id>
+            <activation>
+                <property>
+                    <name>trace.thorntail.test</name>
+                </property>
+            </activation>
+            <properties>
+                <thorntail.test.logging>-Dthorntail.logging=TRACE</thorntail.test.logging>
```
mmusgrov: We should only log narayana specific messages which can be done by appending the log category to the property name. So this line should be
```
<thorntail.test.logging>-Dthorntail.logging.com.arjuna=TRACE -Dthorntail.logging.io.narayana.lra=TRACE</thorntail.test.logging>
```

There is a similar line in TCK pom:
```
<lra.tck.suite.thorntail.trace.params>-Dthorntail.logging.com.arjuna=TRACE -Dthorntail.logging.io.narayana.lra=TRACE</lra.tck.suite.thorntail.trace.params>
```

BTW why do we need to repeat this property with a different name. Can they be combined.

mayankkunwar: Okay, I will change the property to "-Dthorntail.logging.com.arjuna=TRACE"

In TCK there is same property repeated because maybe in case, we want to enable, only TCK trace logs. But for now it is not the case, so I will remove the duplication part and will apply same property to both places, i.e. "thorntail.test.logging"

PR #1561, opened 2020-02-20
[JBTM-3257] not throwing NPE when request command on recovery trigger is

ochaloup:
https://issues.redhat.com/browse/JBTM-3257
https://issues.redhat.com/browse/JBTM-3259

MAIN AS_TESTS

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/logging/arjunaI18NLogger.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/logging/arjunaI18NLogger.java
@@ -1192,9 +1192,9 @@ public interface arjunaI18NLogger {
 	@LogMessage(level = WARN)
 	public void warn_recovery_AtomicActionRecoveryModule_3(Uid arg0, @Cause() Throwable arg1);
 
-	@Message(id = 12292, value = "Connection - IOException", format = MESSAGE_FORMAT)
+	@Message(id = 12292, value = "Connection error on running a work on the socket. IOException: {0}", format = MESSAGE_FORMAT)
 	@LogMessage(level = WARN)
-	public void warn_recovery_Connection_1();
+	public void warn_recovery_Connection_1(IOException cause);
 
 	@Message(id = 12293, value = "Setting timeout exception.", format = MESSAGE_FORMAT)
 	@LogMessage(level = WARN)
@@ -1605,6 +1605,14 @@ public interface arjunaI18NLogger {
 	@Message(id = 12399, value = "Problem in doWork of CacheStore trying to perform state management", format = MESSAGE_FORMAT)
 	@LogMessage(level = WARN)
 	void warn_could_not_handle_objectstore_generic(@Cause Exception ex);
+
+	@Message(id = 12400, value = "Cannot terminate the recovery manager as the implementation is not known. Could be the recovery manager was not started yet?", format = MESSAGE_FORMAT)
+	String get_recovery_manager_implementation_is_not_set();
+
+	@Message(id = 12401, value = "There is already started a recovery manager in mode ''{0}'' which is different from the requested mode ''{1}''." +
+			"If you consider starting in a different mode then first terminate the currently running recovery manager.", format = MESSAGE_FORMAT)
+	String get_recovery_manager_already_started_in_different_mode(int startedRecoveryManagerMode, int modeProvidedToMethodParameter);
+
     /*
         Allocate new messages directly above this notice.
           - id: use the next id number in numeric sequence. Don't reuse ids.
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/logging/arjunaI18NLogger.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/logging/arjunaI18NLogger.java
@@ -27,6 +27,7 @@ import static org.jboss.logging.Logger.Level.WARN;
 import static org.jboss.logging.Logger.Level.TRACE;
 import static org.jboss.logging.annotations.Message.Format.MESSAGE_FORMAT;
 
+import java.io.IOException;
 import java.net.ServerSocket;
 import java.net.Socket;
 
@@ -1379,9 +1380,9 @@ public interface arjunaI18NLogger {
 	@LogMessage(level = WARN)
 	public void warn_recovery_WorkerService_1(@Cause() Throwable arg0);
 
-	@Message(id = 12339, value = "IOException", format = MESSAGE_FORMAT)
+	@Message(id = 12339, value = "Cannot read line from the socket. IOException: {0}", format = MESSAGE_FORMAT)
 	@LogMessage(level = WARN)
-	public void warn_recovery_WorkerService_2();
+	public void warn_recovery_WorkerService_2(IOException cause);
 
 	@Message(id = 12340, value = "RecoveryManager scan scheduled to begin.", format = MESSAGE_FORMAT)
 	@LogMessage(level = INFO)
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/RecoveryManager.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/RecoveryManager.java
@@ -133,7 +133,7 @@ public class RecoveryManager
 	else
 	{
 	    if (_recoveryManager.mode() != mode)
-		throw new IllegalArgumentException();
+		  throw new IllegalArgumentException(tsLogger.i18NLogger.get_recovery_manager_already_started_in_different_mode(_recoveryManager.mode(), mode));
 	}
 
 	return _recoveryManager;
@@ -442,7 +442,8 @@ public class RecoveryManager
                     }
                     catch(InterruptedException interruptedException)
                     {
-                        // do nothing
+                        System.err.println("The retry attempt was interrupted. Exiting...");
+                        throw new IllegalStateException(interruptedException);
                     }
                     manager = manager();
                 }
@@ -483,7 +484,7 @@ public class RecoveryManager
     private final void checkState ()
     {
         if (_theImple == null)
-            throw new IllegalStateException();
+            throw new IllegalStateException(tsLogger.i18NLogger.get_recovery_manager_implementation_is_not_set());
     }
 
     private RecoveryManagerImple _theImple = null;
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/RecoveryManager.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/RecoveryManager.java
@@ -209,6 +209,7 @@ public class RecoveryManager
 
         _theImple.stop(async);
         _theImple = null;
+        _recoveryManager = null;
     }
 
     /**
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/Listener.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/Listener.java
@@ -242,8 +242,8 @@ public class Listener extends Thread
            }
            try {
                wait();
-           } catch (InterruptedException e) {
-               // ignore
+           } catch (InterruptedException ie) {
+               tsLogger.logger.debug("Waiting for connection close was interrupted", ie);
            }
        }
       
@@ -252,6 +252,7 @@ public class Listener extends Thread
        try {
            this.join();
        } catch (InterruptedException ie) {
+           tsLogger.logger.debug("Waiting to ensure the listener thread has exited was interrupted", ie);
        }
    }
 
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/WorkerService.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/WorkerService.java
@@ -92,7 +92,7 @@ public class WorkerService implements Service
 		if (!isAsync)
 		{
             synchronized (this) {
-                if (doWait) {
+                if (doWait && _periodicRecovery.getMode() != PeriodicRecovery.Mode.TERMINATED) {
                     // ok, the periodic recovery thread cannot have finished responding to the last scan request
                     // so it is safe to wait. if we delivered the request while the last scan was still going
                     // then it will have been ignored but that is ok.
@@ -102,6 +102,9 @@ public class WorkerService implements Service
 		    }
 		    catch (Exception ex)
 		    {
+                if(tsLogger.logger.isTraceEnabled()) {
+                    tsLogger.logger.trace("Waiting for recovery scan to complete finished with an exception", ex);
+                }
                 tsLogger.i18NLogger.info_recovery_WorkerService_4();
 		    }
                 }
@@ -126,8 +129,8 @@ public class WorkerService implements Service
 
         out.flush();
 	}
-	catch (IOException ex) {
-        tsLogger.i18NLogger.warn_recovery_WorkerService_2();
+	catch (IOException ioe) {
+        tsLogger.i18NLogger.warn_recovery_WorkerService_2(ioe);
     }
 	catch ( Exception ex ) {
         tsLogger.i18NLogger.warn_recovery_WorkerService_1(ex);
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/WorkerService.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/WorkerService.java
@@ -63,7 +63,7 @@ public class WorkerService implements Service
 	{
 	    String request = in.readLine();
 
-        if (request.equals("PING"))
+        if ("PING".equals(request))
         {
             out.println("PONG");
         }
--- a/ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/recovery/RecoveryDriverUnitTest.java
+++ b/ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/recovery/RecoveryDriverUnitTest.java
@@ -41,25 +41,25 @@ public class RecoveryDriverUnitTest
     @Test
     public void testInvalid () throws Exception
     {
-        RecoveryDriver rd = new RecoveryDriver(0, "foobar");
+        RecoveryDriver rd = new RecoveryDriver(0, "non-existent-hostname");
         
         try
         {
             rd.asynchronousScan();
-            
-            fail();
+
+            fail("Recovery driver asynchronously calls to a non-existent host:port at 'non-existent-hostname:0'. Failure is expected.");
         }
-        catch (final Exception ex)
+        catch (final Exception expected)
         {
         }
         
         try
         {
             rd.synchronousScan();
-            
-            fail();
+
+            fail("Recovery driver synchronously calls to a non-existent host:port at 'non-existent-hostname:0'. Failure is expected.");
         }
-        catch (final Exception ex)
+        catch (final Exception expected)
         {
         }
     }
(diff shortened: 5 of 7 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/WorkerService.java
```diff
@@ -63,7 +63,7 @@ public void doWork (InputStream is, OutputStream os) throws IOException
 	{
 	    String request = in.readLine();
 
-        if (request.equals("PING"))
+        if ("PING".equals(request))
```
tomjenkinson: Would it be possible to add a test for this please? I think the description of https://issues.redhat.com/browse/JBTM-3257 indicates that the scan would have started but if I understand where this condition can occur it is before the instruction is read. I wonder why an IOException is not coming out of in.readLine?

ochaloup: ok. investigation was not easy but I assume I got the reason. see my comment below

ochaloup:
@tomjenkinson I found that the reason why I got the {{NPE}} was not because of abruptly closing the connection as it's but because there was not called flush (on the client side) before the connection was closed (I changed the description of the initial issue to cover the cause better JBTM-3257).
As there is send a message and the client don't flush the server side reader waits to get all data to provide the {{String}} (https://github.com/jbosstm/narayana/blob/5.10.4.Final/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/WorkerService.java#L64).
But the socket is closed meanwhile and the line is then {{null}} and NPE is thrown.
The test which shows this is here
https://github.com/ochaloup/narayana/commit/65e56fc545b2d6d85579533c03a923a266a39642
But there is no way to really check for the NPE as it's only a message in test log. The behaviour is the same. The socket is closed and NPE is only about being showed.

During the time of investigating I found other issue which I consider more troublesome (JBTM-3259).

The PR contains fix for two issues now and some refactoring to get more information from the error states.

ochaloup:
Re-running the tests for the recent changes as I found an error in the test for abrupt terminating of the recovery manager.

ochaloup:
@mmusgrov @tomjenkinson as I changed the code recently (like two days ago) - is there something you would be concerned about or it's ok to go into the codebase?

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/RecoveryManager.java
```diff
@@ -441,7 +442,8 @@ public static void main (String[] args)
                     }
                     catch(InterruptedException interruptedException)
                     {
-                        // do nothing
+                        System.err.println("The retry attempt was interrupted. Exiting...");
```
mmusgrov: Why no log message?

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/Listener.java
```diff
@@ -242,16 +242,17 @@ public synchronized void stopListener()
            }
            try {
                wait();
-           } catch (InterruptedException e) {
-               // ignore
+           } catch (InterruptedException ignoredIe) {
+               tsLogger.logger.debug("Waiting for connection close was interrupted", ignoredIe);
```
mmusgrov: It is not ignored anymore so the name needs updating too.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/Listener.java
```diff
@@ -242,16 +242,17 @@ public synchronized void stopListener()
...
            }
        }
       
        // make sure this listener thread has exited before we return
 
        try {
            this.join();
-       } catch (InterruptedException ie) {
+       } catch (InterruptedException ignoredIe) {
+           tsLogger.logger.debug("Waiting to ensure the listener thread has exited was interrupted", ignoredIe);
```
mmusgrov: rename ignoredIe

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/recovery/RecoveryDriverUnitTest.java
```diff
@@ -46,20 +46,20 @@ public void testInvalid () throws Exception
         try
         {
             rd.asynchronousScan();
-            
-            fail();
+
+            fail("Recovery driver asynchronously calls a non-existing host and port 'foobar:0'. Failure is expected.");
```
mmusgrov: What does `foobar:0` signify?

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/recovery/RecoverySocketUnitTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+    @BeforeClass
+    public static void getInitialState() {
+        socketRecoveryListenerInitialState = recoveryPropertyManager.getRecoveryEnvironmentBean().isRecoveryListener();
+        periodicRecoveryPeriodInitialState =  recoveryPropertyManager.getRecoveryEnvironmentBean().getPeriodicRecoveryPeriod();
+        recoveryBackoffPeriodInitialState = recoveryPropertyManager.getRecoveryEnvironmentBean().getRecoveryBackoffPeriod();
+    }
+
+    @AfterClass
+    public static void returnBackInitialState() {
```
mmusgrov: The naming is odd. How about resetInitialState instead?

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/recovery/RecoverySocketUnitTest.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    @Before
+    public void enableRecoveryListener() throws InterruptedException{
+        recoveryPropertyManager.getRecoveryEnvironmentBean().setRecoveryListener(true);
+        recoveryPropertyManager.getRecoveryEnvironmentBean().setPeriodicRecoveryPeriod(1);
+        recoveryPropertyManager.getRecoveryEnvironmentBean().setRecoveryBackoffPeriod(1);
+        recoveryManager = RecoveryManager.manager();
+        // wait for the socket would be established
+        Thread.sleep(50);
```
mmusgrov: Will you add more comment about why the test is waiting?

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/recovery/RecoverySocketUnitTest.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    @After
+    public void returnStateOfRecoveryListener() {
```
mmusgrov: What is it returning to. More standard naming conventions would be just terminateRecoveryListener or stopRecoveryListener?

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/logging/arjunaI18NLogger.java
```diff
@@ -1604,6 +1605,14 @@ public void warn_objectstore_JDBCImple_over_max_image_size(int imageSize,
 	@Message(id = 12399, value = "Problem in doWork of CacheStore trying to perform state management", format = MESSAGE_FORMAT)
 	@LogMessage(level = WARN)
 	void warn_could_not_handle_objectstore_generic(@Cause Exception ex);
+
+	@Message(id = 12400, value = "Cannot terminate the recovery manager as the implementation is not known. Could be the recovery manager was not started yet?", format = MESSAGE_FORMAT)
+	String get_recovery_manager_implementation_is_not_set();
```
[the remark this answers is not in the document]
ochaloup: please, take a look at the annotations the code. This is not a log message but a message used for exception. The `get_` prefix corresponds with the format of file in such case.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/Listener.java
ochaloup: I think it's ignored as it's just logged and only under the debug level. But I can rename it.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/RecoveryManager.java
ochaloup: It's because this printing is part of the `main` class which is invoked from the command line. I think it's more logical to print this message to STDERR rather than to log.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/recovery/RecoveryDriverUnitTest.java
ochaloup: From my understanding it's a non-existent `hostname:port` sequence. I think it's good point to rename it to be more descriptive.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/recovery/RecoverySocketUnitTest.java
ochaloup: +1, thanks

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/recovery/RecoverySocketUnitTest.java
ochaloup: +1, thanks

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/recovery/RecoverySocketUnitTest.java
ochaloup: Thanks. Good point. This is my omission from the prior test setup.
The recovery manager startup is synchronous and the waiting should not be necessary here at all.

ochaloup:
@mmusgrov Thanks for the comments. I updated the PR + added some explanation above.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/logging/arjunaI18NLogger.java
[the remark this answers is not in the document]
mmusgrov: Okay

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/WorkerService.java
tomjenkinson: Sorry to go back in time so far by when you refer to a comment to look at below please can you indicate the one I should look at?

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/recovery/RecoverySocketUnitTest.java
```diff
@@ -0,0 +1,165 @@
...
+            // streams to and from the RecoveryManager listener
+            BufferedReader fromServer = new BufferedReader(new InputStreamReader(connectorSocket.getInputStream(), StandardCharsets.UTF_8));
+            PrintWriter toServer = new PrintWriter(new OutputStreamWriter(connectorSocket.getOutputStream(), StandardCharsets.UTF_8));
+
+            toServer.println(RecoveryDriver.PING);
+            toServer.flush();
+            String stringResponse = fromServer.readLine();
+            assertEquals("Expecting the correct response string for command " + RecoveryDriver.PING, RecoveryDriver.PONG, stringResponse);
+        } catch (final SocketTimeoutException stex) {
+            failOnSocketTimeout(stex, RecoveryDriver.PING);
```
tomjenkinson: Is there a test where we expect the socket to time out and we verify that?

ochaloup: No, but I can add it. Just note that the socket which times out is the one on the client side as the remote socket is the server socket which waits for a new connections.

ochaloup:
@tomjenkinson I squashed the refactoring commit and added the socket timeout test

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/WorkerService.java
ochaloup: @tomjenkinson I meant the one where I summarized my effort. It's this one: https://github.com/jbosstm/narayana/pull/1561#issuecomment-589639235
But as you went through the code already I assume there is not many new information in that comment.

tomjenkinson: Thanks @ochaloup I am wondering which of the tests will make the variable request be equal to null? Is that scenario able to be created in the unit tests using WorkerService? Something like create the PrintWriter client side and then just close the socket and somehow verify that the WorkerService did not NPE when it got a null https://github.com/jbosstm/narayana/pull/1561/files#diff-133ca2f8c1781ec3f48d3b73b2060a23L64

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/recovery/RecoverySocketUnitTest.java
```diff
@@ -143,6 +142,35 @@ public void socketScanVerbose() throws Exception {
         }
     }
 
+    @Test
+    public void socketScanTerminateAbruptly() throws Exception {
+        try (Socket connectorSocket = getSocket()) {
+            PrintWriter toServer = new PrintWriter(new OutputStreamWriter(connectorSocket.getOutputStream(), StandardCharsets.UTF_8));
+            toServer.println(RecoveryDriver.SCAN);
+            toServer.flush();
+            recoveryManager.terminate();
```
tomjenkinson: I wonder if byteman could ensure that this terminated occurs at the right time to detect that the scan has started as is being interrupted?

Also, should we be checking the outout from the read side to make sure the scan took place?

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/recovery/RecoverySocketUnitTest.java
```diff
@@ -143,6 +142,35 @@ public void socketScanVerbose() throws Exception {
...
+        } catch (final SocketTimeoutException stex) {
+            failOnSocketTimeout(stex, RecoveryDriver.SCAN);
+        }
+    }
+
+    @Test
+    public void socketTimeout() throws Exception {
```
tomjenkinson: Was this expected in here or in JBTM-3257?

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/recovery/RecoverySocketUnitTest.java
ochaloup: The terminate needs to occur before the `SCAN` command returns back any value. Otherwise the test would not verifies about the abrupt termination.

Sure Byteman could be used. If you prefer that I can try to add some byteman test for this.
I personally prefer here the standard "blackbox" test which verifies if the termination works or has some issue.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/recovery/RecoverySocketUnitTest.java
ochaloup: It's a general test on the socket connection so it could be here but it could be there. It does not hit the specific fixed issue. The socket times out as the input is not finished. This behaviour is the same before the fixes and after the fixes as well.
I can move it if it's preferred.

tomjenkinson: Given your answer, I think the thing to consider is if it was part of the investigation of JBTM-3259 or JBTM-3257 and may ever want backporting or whether it is more applicable to have a separate commit as it is not directly related to either? If an additional commit is necessary I think it could be prefixed it as "ADDITIONAL TEST ONLY:" (like we tend to do for "BUILD SCRIPT ONLY:" and "RELEASE SCRIPT ONLY:") /cc @mmusgrov

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/recovery/RecoverySocketUnitTest.java
tomjenkinson: I will close this discussion to move it closer to where I would most like to explore

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/WorkerService.java
tomjenkinson: @ochaloup Testing getting a null on the request is the key thing I am trying to get to for JBTM-3257. Which of the tests results in that?

I was thinking that a test could open the socket and then close the socket (without sending anything) and it might result in `request` from `String request = in.readLine();` being null?

If you do need to send "SCAN" as described in https://issues.redhat.com/browse/JBTM-3257 then perhaps rather than `toServer.println` you could use https://docs.oracle.com/javase/7/docs/api/java/io/PrintWriter.html#print(java.lang.String) rather than println but I don't know of a way to abruptly kill the socket in Java so perhaps that is not possible / predictable in a platform neutral manner?

ochaloup: @tomjenkinson 
test which you ask for is here: https://github.com/ochaloup/narayana/commit/65e56fc545b2d6d85579533c03a923a266a39642
...but it's not possible to be really tested. The trouble is that NPE is thrown but NPE is left just as an error in test log. NPE is at a thread which just disappears and does not influence any functionality.

I didn't add the test into the test suite as it's not easily possible to find the NPE happens or not. And from my perspective it's not an important test. The functionality is not harmed. On socket timeout the client does not wait for any information. So the NPE is only printed to log. While no information is delivered wrongly.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/recovery/RecoverySocketUnitTest.java
ochaloup: @tomjenkinson I have really not much an idea what you strive for. I have had the commits separated to three where two were for issues, one was for the maintenance.
Then I was asked to squash them.

tomjenkinson: Hi @ochaloup sorry! I guess I didn't understand exactly what exactly the third commit covered. Let me try to understand the other part of the thread first and I will come back here.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/WorkerService.java
tomjenkinson: I think this is the part I am not sure of most. What to do about that test as that is the part that really covers testing of JBTM-3257 IIUC. I will consider further.

tomjenkinson: @ochaloup please can you take a look at https://github.com/ochaloup/narayana/pull/8 ?

ochaloup: @tomjenkinson wow, that's nice. Thanks! I will add the test to PR and I will try to reformat the commits. I hope in a good order :-)

ochaloup:
@tomjenkinson thanks a lot for the fixed test!

The PR contains 4 commits. Let's see how it works for you now:question: :sunglasses:

PR #1564, opened 2020-03-05
JBTM-3262  improve string serialization

jhalliday:
https://issues.redhat.com/browse/JBTM-3262

!AS_TESTS !TOMCAT !JACOCO !LRA !BLACKTIE
!PERF  !QA_JTA !RTS !QA_JTS_OPENJDKORB !XTS !QA_JTS_JDKORB !QA_JTS_JACORB

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
@@ -33,6 +33,7 @@ package com.arjuna.ats.arjuna;
 
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.nio.charset.StandardCharsets;
 import java.util.Enumeration;
 import java.util.Hashtable;
 import java.util.concurrent.locks.ReentrantLock;
@@ -646,7 +647,7 @@ public class StateManager
         {
             // pack the marker first.
 
-            os.packString(StateManager.marker);
+            os.packStringBytes(StateManager.markerBytes);
 
             /*
              * Only pack something if there is a transaction. Otherwise the
@@ -1399,4 +1400,5 @@ public class StateManager
     private ReentrantLock mutex = new ReentrantLock();
 
     private static final String marker = "#ARJUNA#";
+    private static final byte[] markerBytes = marker.getBytes(StandardCharsets.UTF_8);
 }
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/state/OutputBuffer.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/state/OutputBuffer.java
@@ -358,26 +358,26 @@ public class OutputBuffer
         if (!_valid)
             throw new IOException(tsLogger.i18NLogger.get_state_OutputBuffer_10());
 
-        int sz = 0;
-        String dummy = null;
+        if(s == null) {
+            packInt(0);
+        } else {
+            packStringBytes( s.getBytes(StandardCharsets.UTF_8) );
+        }
+    }
 
-        if (s != null)
-        {
-            sz = s.length() + 1;
-            dummy = s + '\0';
-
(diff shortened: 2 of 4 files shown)


jhalliday:
I split this over two commits. First one is a required bug fix, the second is an optional perf enhancement

nmcl:
Would be good to get a summary of what changes and why because when I check the git code changes it's hard to see the wood for the trees given the sheer number of changes which appear to be solely due to formatting (maybe tabs versus spaces?)

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
```diff
@@ -641,12 +642,12 @@ protected void packHeader (OutputObjectState os, Header hdr)
...
         Uid txId = ((hdr == null) ? null : hdr.getTxId());
         Uid processUid = ((hdr == null) ? null : hdr.getProcessId());
-        
+
         try
         {
             // pack the marker first.
 
-            os.packString(StateManager.marker);
+            os.packStringBytes(markerBytes);
```
mmusgrov: What is the performance difference between packing the bytes directly versus packing the string - is it that the string copy (`os.packString(StateManager.marker)`) is avoided?

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/state/OutputBuffer.java
```diff
@@ -358,26 +358,26 @@ public final synchronized void packString (String s) throws IOException
...
         _valid = false;
 
-        if (sz > 0)
-        {
-            byte[] bytes = dummy.getBytes(StandardCharsets.UTF_8);
-            _output.write(bytes, 0, bytes.length);
-            realign(bytes.length);
-        }
+        _output.write(bytes, 0, bytes.length);
+        _output.writeByte(0);
```
mmusgrov: Just a question - did the original code add space for the string delimiter but not actually pack it into the output buffer? How did you uncover this bug?

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
jhalliday: The calculation that transforms the String's internal representation (char[]) to the encoded form (byte[]) need not be repeated for every single StateManager instance (which is MANY), in addition to which you avoid the gc churn and copies to/from the short-lived byte[] that's needed by the encoder. The actual packing once you have the byte[] is then the same either way. The encoding is a trivial overhead on tx that hit the network, but worth a 1-2% on in-memory cases. May not have bothered, except it's easy to piggyback the optimization onto the bug fix.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/state/OutputBuffer.java
jhalliday: The original works by creating a new string with the terminator appended as a charater, then sizing/encoding that in a single step, rather than treating the string and the terminator as separate elements. The latter is more efficient, as it avoids creating an additional String object.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
ochaloup: maybe a matter of taste but would not be better to use `StateManager.markerBytes` when static variable is accessed here?

mmusgrov:
TESTIT

mmusgrov:
@jhalliday the automated test failed to resolve io.thorntail:config-api:jar:1.6.1 from https://repo1.maven.org/maven2/io/thorntail/config-api/1.6.1/ due to 503 http status. My guess is that it's caused by a transient comms failure. Do you mind if I resubmit the PR `MAIN profile test`?

jhalliday:
I just slipped in another optimization, so it will need to go again anyhow.

PR #1565, opened 2020-03-10
JBTM-3265 run a second recovery pass

mmusgrov:
https://issues.redhat.com/browse/JBTM-3265

### DIFF
--- a/rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/FailingParticipantCallsIT.java
+++ b/rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/FailingParticipantCallsIT.java
@@ -90,7 +90,10 @@ public class FailingParticipantCallsIT {
             }
         }
 
-        new NarayanaLRARecovery().waitForRecovery(lra);
+        NarayanaLRARecovery recoveryHelper = new NarayanaLRARecovery();
+
+        recoveryHelper.waitForRecovery(lra);
+        recoveryHelper.waitForRecovery(lra);
 
         try {
             response = client.target(UriBuilder.fromUri(baseURL.toExternalForm())


rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/FailingParticipantCallsIT.java
```diff
@@ -90,7 +90,10 @@ public void testFailingAfterLRA() {
             }
         }
 
-        new NarayanaLRARecovery().waitForRecovery(lra);
+        NarayanaLRARecovery recoveryHelper = new NarayanaLRARecovery();
+
+        recoveryHelper.waitForRecovery(lra);
+        recoveryHelper.waitForRecovery(lra);
```
xstefank: May I ask why this fixes the issue? `waitForRecovery` should keep replaying the end phase until the LRA is finished. Is something not expected returned from the first pass? In that case, we should probably fix the NarayanaLRARecovery.

mmusgrov:
The test fails because the callback from the coordinator to the listener is not sent/received in time (we have coded a hard limit of 1 second on callbacks). What happens is that the coordinator sends the callback and either the listener does not respond within the second or the response is not received within a second.

LRA guarantees that it will keep retrying until the listener acknowledges receipt of the callback. Hence the second recovery pass.

Note that the coordinator is correctly issuing the callback but I am unable to reproduce this failure on my laptop (even when I apply heavy load to the CPU) which lends wait to it being a straight forward CI resourcing problem.

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/FailingParticipantCallsIT.java
mmusgrov: The test fails because the callback from the coordinator to the listener is not sent/received in time (we have coded a hard limit of 1 second on callbacks). What happens is that the coordinator sends the callback and either the listener does not respond within the second or the response is not received within a second.

LRA guarantees that it will keep retrying until the listener acknowledges receipt of the callback. Hence the second recovery pass.

Note that the coordinator is correctly issuing the callback but I am unable to reproduce this failure on my laptop (even when I apply heavy load to the CPU) which lends wait to it being a straight forward CI resourcing problem.

mmusgrov: If I could duplicate the failure on my laptop then I could debug the issue more precisely. This particular PR is theoretical fix based on code inspection.

mmusgrov:
@ochaloup this bug is a test issue (not an implementation one). Although the fix to the test isn't ideal I will say that it is just a test and it is urgent since it is regularly breaking our CI. If we do see subsequent failures then we can revisit how the test is written later on (ie the sooner we commit the fix the sooner we should see CI passing again).

ochaloup:
@mmusgrov I'm sorry for my late review. The bigger urgency is caused as well by me by not looking it sooner.
But  I really don't like this speculative

ochaloup:
Ouch, I'm sorry, I didn't want to close/reopen. My fault on commenting.

mmusgrov:
@ochaloup can you raise a PR containing your fix and I will close this one.

mmusgrov:
Closing in favour of pr #1567

PR #1569, opened 2020-03-17
JBTM-3260 On publishing a new Narayana release there should be created Quarkus PR to update Narayana version

mayankkunwar:
https://issues.redhat.com/browse/JBTM-3260

MAIN

### DIFF
--- a/narayana-release-process.sh
+++ b/narayana-release-process.sh
@@ -143,6 +143,33 @@ then
 fi
 cd ..
 
+# Raising PR for quarkus to upgrade Narayana to current version.
+QRKSISSUE = WFLYISSUE
+if [ ! -d "quarkus" ]
+then
+  (git clone git@github.com:jbosstm/quarkus.git -o jbosstm; cd quarkus; git remote add upstream git@github.com/quarkusio/quarkus.git)
+fi
+cd quarkus
+git fetch jbosstm
+git branch | grep $QRKSISSUE
+if [[ $? != 0 ]]
+then
+  git fetch upstream;
+  git checkout -b ${QRKSISSUE}
+  git reset --hard upstream/master
+  cd /bom/runtime/
+  CURRENT_VERSION_IN_QRKS=`grep 'narayana.version>' pom.xml | cut -d \< -f 2|cut -d \> -f 2`
+  sed -i "s/narayana.version>$CURRENT_VERSION_IN_QRKS/narayana.version>$CURRENT/g" pom.xml
+  cd ../../
+  git add /bom/runtime/pom.xml
+  git commit -m "${QRKSISSUE} Upgrade Narayana to $CURRENT"
+  git push --set-upstream jbosstm ${QRKSISSUE}
+  git checkout 5_BRANCH
+  git reset --hard jbosstm/5_BRANCH
+  xdg-open https://github.com/jbosstm/quarkus/pull/new/$QRKSISSUE &
+fi
+cd ..
+
 cd ~/tmp/narayana/$CURRENT/sources/documentation/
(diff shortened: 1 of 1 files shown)


narayana-release-process.sh
```diff
@@ -143,6 +138,33 @@ then
 fi
 cd ..
 
+# Raising PR for quarkus to upgrade Narayana to current version.
+QRKSISSUE = WFLYISSUE
```
ochaloup: I'm not sure if it's good to go with the same issue name with Quarkus as it was for WFLY

narayana-release-process.sh
```diff
@@ -143,6 +138,33 @@ then
...
+fi
+cd quarkus
+git fetch jbosstm
+git branch | grep $QRKSISSUE
+if [[ $? != 0 ]]
+then
+  git fetch upstream;
+  git checkout -b ${QRKSISSUE}
+  git reset --hard upstream/master
+  cd /bom/runtime/
```
ochaloup: I think this won't work as you go with abs path

narayana-release-process.sh
```diff
@@ -143,6 +138,33 @@ then
...
+  CURRENT_VERSION_IN_QRKS=`grep 'narayana.version>' pom.xml | cut -d \< -f 2|cut -d \> -f 2`
+  sed -i "s/narayana.version>$CURRENT_VERSION_IN_QRKS/narayana.version>$CURRENT/g" pom.xml
+  cd ../../
+  git add /bom/runtime/pom.xml
```
ochaloup: I assume this won't work (abs path used)

narayana-release-process.sh
```diff
@@ -143,6 +138,33 @@ then
...
+  git commit -m "${QRKSISSUE} Upgrade Narayana to $CURRENT"
+  git push --set-upstream jbosstm ${QRKSISSUE}
+  git checkout 5_BRANCH
```
ochaloup: I assume there is no `5_BRANCH` in quarkus repo. I think the `master` should be used.

ochaloup:
@mayankkunwar ok, I'm putting this PR **on hold** as @xstefank pointed me out to the `dependabot`. Maybe we may just add the Narayana dependency to the quarkus config dependabot file (https://github.com/quarkusio/quarkus/blob/master/.dependabot/config.yml). Just I'm not sure about rules of such thing.

Please, do not proceed with the changes that I suggested and let's talk about the bot tomorrow - if it's a feasible option for us.

narayana-release-process.sh
mayankkunwar: I am using the same issue name because when narayana-release-process.sh is called, WFLYISSUE is coming as third argument, and I was not knowing from where, when and who calls this file. Otherwise I would have used the fourth argument for QRKSISSUE.

narayana-release-process.sh
mayankkunwar: locally this was working, but anyways it is always better to use full path, so I will change it as requested.

narayana-release-process.sh
mayankkunwar: I will refactor this as per request

narayana-release-process.sh
mayankkunwar: okay, will change it to master.

mayankkunwar:
@ochaloup okay, we can have a discussion on the same before moving forward.

mayankkunwar:
I am going to close this PR, as we tried to solve this problem with a better alternative solution, by adding narayana dependencies to quarkus dependabot config.xml.
Here is the PR link: https://github.com/quarkusio/quarkus/pull/7977

PR #1562, opened 2020-03-02
Lra

mmusgrov:
https://issues.redhat.com/browse/JBTM-3261

This PR brings the LRA branch up to date with master (and with the latest microprofile-lra snapshot build).

There will be two failures:
1. One is caused by a bug in the MP-LRA TCK (https://github.com/eclipse/microprofile-lra/pull/281)
  Remark: the PR also includes a fix for this issue
2. The other is known (TckContextTests#testForgetCalledForNestedParticipantsWhenParentIsClosed) and is not yet implemented

I am preparing a fix for 2

LRA !MAIN !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF !NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

PR #1571, opened 2020-03-24
Lra.merge merge master into the LRA branch

mmusgrov:
https://issues.redhat.com/browse/JBTM-3261

This PR brings the LRA branch up to date with master (as of March 2nd) and with the microprofile-lra snapshot (1.0-20200302.071313-701).  It replaces PR #1562 which I deliberately deleted while attempting a rebase against current master. Note that this new PR introduced a new failure (TckRecoveryTests.testCancelWhenParticipantIsRestarted) which needs investigating.

There is still a need to take the latest LRA snapshot build and the latest Narayana master - if we take the current PR (after running CI on it) then there will be fewer changes to resolve.

There are two failures:

[ERROR]   TckContextTests.testForgetCalledForNestedParticipantsWhenParentIsClosed:244 testForgetCalledForNestedParticipantsWhenParentIsClosed: resource should have called forget for the nested LRA expected:<1> b
ut was:<0>                                                                                                                                                                                                         
[ERROR]   TckRecoveryTests.testCancelWhenParticipantIsRestarted:134 After LRA with Cancelled status should have been called. Was 0 expected:<200> but was:<500>                                                    

The first one is known since it is not yet implemented.
The second one was not present in the other PR (#1562 ) so needs investigating.
and 

LRA !MAIN !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF !NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle


rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -549,23 +612,23 @@ public LRARecord enlistParticipant(URI coordinatorUrl, String participantUrl, St
...
+            setTimeLimit(timeLimit);
+
+            return p;
+        } else if (isRecovering() && p.getCompensator() == null && p.getEndNotificationUri() != null) {
+            // the participant is an AfterLRA listener so manually add it to heuristic list
+            heuristicList.putRear(p);
+            updateState();
 
             return p;
         } else if (isRecovering() && p.getCompensator() == null && p.getEndNotificationUri() != null) {
```
xstefank: duplicated if branching, probably a merge/rebase conflict

mmusgrov:
> so I think the cleaner solution would be to rebase LRA branch on top of the current master now (resolving these conflicts rather). I expect the RC2 will be released in a quite short period of time (only one LRA issue is left open on 1.0 milestone).

Okay I will have a go at rebasing the LRA branch onto master instead, after which we can drop the LRA branch. Then we can go back to tracking LRA snapshot dependencies on a monthly basis which, given that RC2 is imminent, should be manageable.

mmusgrov:
I am closing this PR in favour of rebasing onto master instead (see Martin's comment above https://github.com/jbosstm/narayana/pull/1571#pullrequestreview-381757764).

Martin has kindly offered to take on the rebase and will raise a PR in due course.

PR #1574, opened 2020-03-27
[JBTM-3270] tracking persisted REST-AT participant 

ochaloup:
for removal not showing warning in the server log

https://issues.redhat.com/browse/JBTM-3270

MAIN RTS AS_TESTS
!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !TOMCAT !JACOCO !LRA

### DIFF
--- a/rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
+++ b/rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
@@ -44,6 +44,11 @@ public final class RecoveryManager {
 
     private final Map<String, ParticipantDeserializer> deserializers = new ConcurrentHashMap<String, ParticipantDeserializer>();
 
+    /**
+     * A map of participants persisted in the object store, it maps 'Participant Id' to 'Uid'
+     */
+    private final Map<String, Uid> persistedParticipants = new ConcurrentHashMap<>();
+
     private RecoveryManager() {
     }
 
@@ -82,6 +87,8 @@ public final class RecoveryManager {
             final Uid uid = new Uid(participantInformation.getId());
 
             recoveryStore.write_committed(uid, PARTICIPANT_INFORMATION_RECORD_TYPE, state);
+            // to identify the uid from the participant persisted into the object store in order to delete it later
+            persistedParticipants.put(participantInformation.getId(), uid);
         } catch (Exception e) {
             LOG.warn("Failure while persisting participant information.", e);
         }
@@ -93,12 +100,15 @@ public final class RecoveryManager {
         }
 
         final RecoveryStore recoveryStore = StoreManager.getRecoveryStore();
-        final Uid uid = new Uid(participantInformation.getId());
+        String participantId = participantInformation.getId();
 
-        try {
-            recoveryStore.remove_committed(uid, PARTICIPANT_INFORMATION_RECORD_TYPE);
-        } catch (ObjectStoreException e) {
-            LOG.warn("Failure while removing participant information from the object store.", e);
+        if(persistedParticipants.get(participantId) != null) {
+            try {
+                recoveryStore.remove_committed(new Uid(participantId), PARTICIPANT_INFORMATION_RECORD_TYPE);
+                persistedParticipants.remove(participantId);
+            } catch (ObjectStoreException ose) {
+                LOG.warn("Failure while removing participant information from the object store.", ose);
+            }
         }
     }
 
@@ -156,6 +166,8 @@ public final class RecoveryManager {
                         if (participantInformation != null) {
                             ParticipantsContainer.getInstance().addParticipantInformation(
                                     participantInformation.getId(), participantInformation);
+                            // adding entry to map listing all persisted participants into the object store (to be removed later)
+                            persistedParticipants.put(participantInformation.getId(), uid);
                         }
                     }
                 }
@@ -198,6 +210,8 @@ public final class RecoveryManager {
 
         final ParticipantInformation participantInformation = new ParticipantInformation(id, applicationId, recoveryUrl,
                 participant, status);
+        // the participant was loaded from object store, make sure that the map consisting all the persisted ids knows about it
+        persistedParticipants.put(participantInformation.getApplicationId(), new Uid(participantInformation.getId()));
 
         if (!synchronizeParticipantUrlWithCoordinator(participantInformation)) {
             try {


rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
```diff
@@ -44,6 +45,11 @@
 
     private final Map<String, ParticipantDeserializer> deserializers = new ConcurrentHashMap<String, ParticipantDeserializer>();
 
+    /**
+     * A map of participants saved in the persistent store. &lt;Participant Id :: Uid&gt;
```
mmusgrov: What does `<Participant Id :: Uid>` mean. It also starts a new sentence making it more confusing.

Is it suggesting that the variable is

```java
* A map of participant ids to Uids for keeping track of which participants have been saved to the recovery store.
```

Also if you are going to make changes to this PR I would suggest using the same term for the store throughout (it currently contains three different names: persistent store, object store and recovery store)

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
```diff
@@ -93,12 +101,15 @@ public void removeParticipantInformation(final ParticipantInformation participan
...
-        try {
-            recoveryStore.remove_committed(uid, PARTICIPANT_INFORMATION_RECORD_TYPE);
-        } catch (ObjectStoreException e) {
-            LOG.warn("Failure while removing participant information from the object store.", e);
+        if(committedParticipants.get(participantId) != null) {
+            try {
+                recoveryStore.remove_committed(new Uid(participantId), PARTICIPANT_INFORMATION_RECORD_TYPE);
+                committedParticipants.remove(participantId);
+            } catch (ObjectStoreException e) {
+                LOG.warn("Failure while removing participant information from the object store.", e);
```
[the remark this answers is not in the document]
ochaloup: I think the record should not be removed from the map if the `remove_committed` fails. It means that the object store was not updated. It's important to ensure to really cleaning the object store in the next attempt. The warning sounds correct here as it's needed to warn the user that the participant could not be removed because of some internal failure, probably on the storage.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
ochaloup: Agree, thanks for the note. I updated the comments.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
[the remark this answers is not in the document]
mmusgrov: The record will not be removed. However subsequent attempts to remove it will not be attempted (if we failed once then it is likely to continually fail). The way you have it coded is that if the remove fails the warning will appear in the log on every recovery pass. A single WARNING in the log should be sufficient for notifying the user about this error.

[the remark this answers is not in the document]
ochaloup: I really don't agree on this strategy. When a record is not removed from the log then it's the Narayana strategy to inform the user about the consistent failure.
My idea is to really let the warning being printed on every recovery pass.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
```diff
@@ -44,6 +44,11 @@
 
     private final Map<String, ParticipantDeserializer> deserializers = new ConcurrentHashMap<String, ParticipantDeserializer>();
 
+    /**
+     * A map of participants persisted in the object store, it maps 'Participant Id' to 'Uid'
+     */
+    private final Map<String, Uid> committedParticipants = new ConcurrentHashMap<>();
```
tomjenkinson: Would persistedParticipants be clearer (though I do accept the store API is calling it committed)? Or are these really committed from a transactional perspective?

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
```diff
@@ -81,6 +86,8 @@ public void persistParticipantInformation(final ParticipantInformation participa
             final OutputObjectState state = getParticipantInformationOutputState(participantInformation);
             final Uid uid = new Uid(participantInformation.getId());
 
+            // to identify the uid from the participant persisted into the object store in order to delete it later
+            committedParticipants.put(participantInformation.getId(), uid);
```
tomjenkinson: I note this is going in the map before the store is updated

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
```diff
@@ -93,12 +100,15 @@ public void removeParticipantInformation(final ParticipantInformation participan
         }
 
         final RecoveryStore recoveryStore = StoreManager.getRecoveryStore();
-        final Uid uid = new Uid(participantInformation.getId());
```
tomjenkinson: A general question, how do we reach removeParticipantInformation if the participant was not property written to disk before. Do we need to differentiate on rollback vs other conditions?

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
ochaloup: This was intentional originally as I was thinking that it's safer to try to try a remove an non-existing record from the object store than does not try it at all. I was thinking on some issue being thrown after the record was persisted but the call was not yet returned to this point.
Then the `persistedParticipants` would not be aware.
But when I'm rethinking now it seems to me it's safe to move it after the persistent call.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
ochaloup: +1

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
ochaloup: If I understand correctly what you ask about then the call to `removeParticipantInformation` is just called on these type of "actions" from the `ParticipantResource` (https://github.com/jbosstm/narayana/blob/5.10.4.Final/rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantResource.java#L276)
So any call on rollback, commit, forget may cause to call here. There is no differentiation from where the method is called from.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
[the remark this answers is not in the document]
mmusgrov: Normally we only continually report recovery failures when we are expecting recovery to happen and repeating warnings for that purpose is acceptable. But failing to delete file will most likely fail forevermore and the user only needs to be told about it once.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
tomjenkinson: It follows a little to the conversation @mmusgrov and you are having. It relates particularly to whether a failure to remove could be transient or not

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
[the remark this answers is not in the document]
tomjenkinson: @mmusgrov - is it not possible to have a transient storage problem though? Perhaps we need to understand better _why_ the remove is failing. But that might be better in some object store API change, like a return type that determines if the failure is expected to be transient or not. 

Your proposed change (in the first comment of this thread) does not look to be rendering fully to me (or Ondra already took that in). How do you propose to suppress subsequent warnings? I am not aware of a precedent to do that in Narayana so far. Doesn't the recovery manager report repeated failures to recover things? But if there are other examples of that let's evaluate as repeating the message at WARN at least might not be so useful (perhaps trace) and might be covered by wording the message that the failure will not be re-reported?

BTW, given the current state of the patch (in that we only put it in the map when write_committed passes or it is recovered/recreated, I can't guess why remove_committed would throw an exception now but maybe there is something in the recover/recreated paths that would do that?

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
```diff
@@ -82,6 +87,8 @@ public void persistParticipantInformation(final ParticipantInformation participa
             final Uid uid = new Uid(participantInformation.getId());
 
             recoveryStore.write_committed(uid, PARTICIPANT_INFORMATION_RECORD_TYPE, state);
+            // to identify the uid from the participant persisted into the object store in order to delete it later
+            persistedParticipants.put(participantInformation.getId(), uid);
         } catch (Exception e) {
             LOG.warn("Failure while persisting participant information.", e);
```
tomjenkinson: @mmusgrov is it more or is the failure to write to disk of a participant just logged but ignored? Is that safe?

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
ochaloup: I probably don't understand the point. Failure to remove causes there is found the record in the next recovery round and the attempt to rollback/commit will be retried.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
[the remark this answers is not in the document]
ochaloup: @mmusgrov I don't agree that user should be informed about the failure of the storage just once. When the storage is not cleaned the e.g. rollback operation will be retried. Probably with some error. Additional information that there is the trouble with removal is important part of the context.

@tomjenkinson  the `remove_committed` is not expected to throw an exception in the current state of the patch. It just logs a warning that there is not a way to remove the record.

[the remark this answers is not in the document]
tomjenkinson: @ochaloup I guess Mike was trying to prevent was some outcome a bit like the following (and this is total pseudo log statements and pseudo code):

```
while true {
  // rest of the algorithm, then
  "Failure while removing participant information from the object store" with a caused by of some permanent error like "File does not exist"
}
```

What I am not sure of whether that could ever be the case. And if it could, rather than suppressing the logging alone it would perhaps be better to consider what can cause the condition and deal with it in the algorithm to give a break out clause?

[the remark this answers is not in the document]
tomjenkinson: > @tomjenkinson the `remove_committed` is not expected to throw an exception in the current state of the patch. It just logs a warning that there is not a way to remove the record.

Responding to this separate. I was not clear enough, let me rephrase: given the current state of the patch, I can't guess why there would be problem to remove the record but maybe there is something in the recover/recreated paths that would cause there to be a problem to remove it somehow?

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
tomjenkinson: I will resolve this as we are discussing something related in the parallel thread

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
```diff
@@ -93,12 +100,15 @@ public void removeParticipantInformation(final ParticipantInformation participan
...
 
         final RecoveryStore recoveryStore = StoreManager.getRecoveryStore();
-        final Uid uid = new Uid(participantInformation.getId());
+        String participantId = participantInformation.getId();
 
-        try {
-            recoveryStore.remove_committed(uid, PARTICIPANT_INFORMATION_RECORD_TYPE);
-        } catch (ObjectStoreException e) {
-            LOG.warn("Failure while removing participant information from the object store.", e);
+        if(persistedParticipants.get(participantId) != null) {
```
tomjenkinson: @ochaloup is there any need to synchronise in case a thread doing write_participant and persistedParticipants.put could execute concurrently with a thread checking if something is in the map? I don't know the answer but did you assess that?

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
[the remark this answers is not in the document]
ochaloup: @tomjenkinson ok. maybe I understand, at least partially.

@tomjenkinson @mmusgrov  First, for the understanding of the changes, please, consult the source code. Where the method is called from and what is the context it's called from.
It's quite possible I misunderstand something but if we are talking from start in some semi-real context it's something which I'm not able to follow and which makes my work pretty hard.

@tomjenkinson the failure of the removal is the same from any part of the code that uses the object store. There is some IO failure or some internal storage failure that makes impossible to remove the record. It's something which can't be handled by algorithm. And in case there is some adjustment possible then it has to be handled by the `ObjectStore` API in arjunacore not here in the `rts/at` module.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
ochaloup: yes, I did. There is used the `ConcurrentHashMap`. If something is checking (e.g. recovery) if something is in map, it may not find it but it will find it in the next cycle.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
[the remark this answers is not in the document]
tomjenkinson: Thanks for clarifying @ochaloup. If the patch introduces no likelihood of causing the remove_committed to fail then I think changes to dealing with non-recoverable remove_committed errors should be dealt with as an RFE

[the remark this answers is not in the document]
mmusgrov: ot possible to have a transient storage problem though? Perhaps we need to understand better _why_ the remove is failing. But that might be better in some object store API change, like a return type that determines if the failure is expected to be transient or not.
> 
> Your proposed change (in the first comment of this thread) does not look to be rendering fully to me (or Ondra already took that in). How do you propose to suppress subsequent warnings? I am not aware of a precedent to do that in Narayana so far. Doesn't the recovery manager report repeated failures to recover things? But if there are other examples of that let's evaluate as repeating the message at WARN at least might not be so useful (perhaps trace) and might be covered by wording the message that the failure will not be re-reported?
> 
> BTW, given the current state of the patch (in that we only put it in the map when write_committed passes or it is recovered/recreated, I can't guess why remove_committed would throw an exception now but maybe there is something in the recover/recreated paths that would do that?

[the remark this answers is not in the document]
mmusgrov: I am adding my approval of this PR on the following basis:

- XTS does the same as what's in this PR.
- The error should may only ever occur each time an InboundBridgeManager is created (after a server restart) and after a transaction is finished.
- As far as I can see this new code is not ran by Periodic Recovery so my concern about filling the logs on each recovery pass is not a valid concern.

But I'd still argue that my points made in the previous comment are still valid but not sufficient to block this PR.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
mmusgrov: Customers normally log at WARN level. Do you want it logged at ERROR instead or do you expect the code to recover from disk failures by retrying?

mmusgrov:
The job output for the AS_TESTS failure is no longer present. But the primary test for this PR is the RTS profile which did pass.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
ochaloup: I found I missed this comment. But the WARN level is considered because of the retrying happens and there is the hope the disk failure will be recovered.

PR #1585, opened 2020-04-02
JBTM-3276 SlotStore implementation.

jhalliday:
This is part one, with the in-memory backend.
https://issues.redhat.com/browse/JBTM-3276

CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF !LRA !NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

I have set the PR to only test MAIN. I will investigate the AS_TEST, JACOCO and LRA Jenkin's axis failures.

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/RAMSlots.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/RAMSlots.java
@@ -27,7 +27,7 @@ package com.arjuna.ats.internal.arjuna.objectstore.slot;
  */
 public class RAMSlots {
 
-    public final byte[][] slots;
+    private final byte[][] slots;
 
     public RAMSlots(SlotStoreEnvironmentBean config) {
         slots = new byte[config.getNumberOfSlots()][];
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/RAMSlots.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/RAMSlots.java
@@ -1,0 +1,47 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2020, Red Hat, Inc. and/or its affiliates,
+ * and individual contributors as indicated by the @author tags.
+ * See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ * This copyrighted material is made available to anyone wishing to use,
+ * modify, copy, or redistribute it subject to the terms and conditions
+ * of the GNU Lesser General Public License, v. 2.1.
+ * This program is distributed in the hope that it will be useful, but WITHOUT A
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+ * PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more details.
+ * You should have received a copy of the GNU Lesser General Public License,
+ * v.2.1 along with this distribution; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA  02110-1301, USA.
+ *
+ * (C) 2020,
+ * @author Jonathan Halliday (jonathan.halliday@redhat.com)
+ */
+package com.arjuna.ats.internal.arjuna.objectstore.slot;
+
+/**
+ * Trivial in-memory backend for the SlotStore, useful for benchmarking but not much else.
+ *
+ * @author Jonathan Halliday (jonathan.halliday@redhat.com), 2020-03
+ */
+public class RAMSlots {
+
+    public final byte[][] slots;
+
+    public RAMSlots(SlotStoreEnvironmentBean config) {
+        slots = new byte[config.getNumberOfSlots()][];
+    }
+
+    public void write(int slot, byte[] data, boolean sync) {
+        slots[slot] = data;
+    }
+
+    public byte[] read(int slot) {
+        return slots[slot];
+    }
+
+    public void clear(int slot, boolean sync) {
+        slots[slot] = null;
+    }
+}
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStore.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStore.java
@@ -203,20 +203,21 @@ public class SlotStore {
     /**
      * Return all keys in the store having the same typename and (matching or unknown) state as the provided key.
      *
-     * @param slotStoreKey a template key to match in the search. Uid part is ignored, typename and state are used in matching
+     * @param templateKey a template key to match in the search. Uid part is ignored, typename and state are used in matching
      * @return an array, possible empty but non-null, of matching keys.
      */
-    public SlotStoreKey[] getMatchingKeys(SlotStoreKey slotStoreKey) {
+    public SlotStoreKey[] getMatchingKeys(SlotStoreKey templateKey) {
 
-        List<SlotStoreKey> keys = new ArrayList<>();
+        List<SlotStoreKey> matchingKeys = new ArrayList<>();
 
-        for (SlotStoreKey key : slotIdIndex.keySet()) {
-            if (key.getTypeName().equalsIgnoreCase(slotStoreKey.getTypeName()) &&
-                    (key.getStateStatus() == StateStatus.OS_UNKNOWN || key.getStateStatus() == slotStoreKey.getStateStatus())) {
-                keys.add(key);
+        for (SlotStoreKey candidateKey : slotIdIndex.keySet()) {
+            if (candidateKey.getTypeName().equalsIgnoreCase(templateKey.getTypeName()) &&
+                    // OS_UNKNOWN in the template acts as a wildcard.
+                    (templateKey.getStateStatus() == StateStatus.OS_UNKNOWN || candidateKey.getStateStatus() == templateKey.getStateStatus())) {
+                matchingKeys.add(candidateKey);
             }
         }
 
-        return keys.toArray(new SlotStoreKey[0]);
+        return matchingKeys.toArray(new SlotStoreKey[0]);
     }
 }
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStore.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStore.java
@@ -1,0 +1,222 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2020, Red Hat, Inc. and/or its affiliates,
+ * and individual contributors as indicated by the @author tags.
+ * See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ * This copyrighted material is made available to anyone wishing to use,
+ * modify, copy, or redistribute it subject to the terms and conditions
+ * of the GNU Lesser General Public License, v. 2.1.
+ * This program is distributed in the hope that it will be useful, but WITHOUT A
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+ * PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more details.
+ * You should have received a copy of the GNU Lesser General Public License,
+ * v.2.1 along with this distribution; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA  02110-1301, USA.
+ *
+ * (C) 2020,
+ * @author Jonathan Halliday (jonathan.halliday@redhat.com)
+ */
+package com.arjuna.ats.internal.arjuna.objectstore.slot;
+
+import com.arjuna.ats.arjuna.logging.tsLogger;
+import com.arjuna.ats.arjuna.objectstore.StateStatus;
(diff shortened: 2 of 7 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStore.java
```diff
@@ -0,0 +1,222 @@
...
+     * @param config The configuration parameters for the instance
+     * @throws IOException if the required backing storage can't be initialized.
+     */
+    public SlotStore(SlotStoreEnvironmentBean config) throws IOException {
+
+        this.config = config;
+
+        // unused for now, but eventually we'll have a disk backend...
+        File storeDir = new File(config.getStoreDir());
+        if (!storeDir.exists() && !storeDir.mkdirs()) {
```
mmusgrov, line 68: Can the file exist if the directory hierarchy hasn't been created?

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStore.java
```diff
@@ -0,0 +1,222 @@
...
+            throw new IOException("record not found for " + key);
+        }
+        byte[] data = slots.read(slotId);
+        // it's possible, though unlikely, for the data to be null here due to a concurrent remove
+        if (data == null) {
+            throw new IOException("record not found for " + key);
+        }
+        InputBuffer inputBuffer = new InputBuffer(data);
+        SlotStoreKey.unpackFrom(inputBuffer);
+
```
mmusgrov, line 107: Do you ignore the return value because you just want to validate inputBuffer. If so why not add an isValid() method? Is it because `SlotStore` is only ever accessed via the adaptor (`SlotStoreAdaptor`), in which case validation as a side effect of unpacking is fine.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStoreAdaptor.java
```diff
@@ -0,0 +1,349 @@
...
+        }
+
+        SlotStoreKey key = new SlotStoreKey(null, typeName, matchState);
+
+        SlotStoreKey[] matchingKeys = store.getMatchingKeys(key);
+
+        OutputObjectState buffer = new OutputObjectState();
+
+        try {
+            if (matchingKeys != null) {
```
mmusgrov, line 81: Superfluous test for null because matchingKeys cannot be null because SlotStoreAdaptor#getMatchingKeys(key) always returns a valid array.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStoreAdaptor.java
```diff
@@ -0,0 +1,349 @@
...
+        return true;
+    }
+
+    @Override
+    public boolean allObjUids(String typeName, InputObjectState foundInstances) throws ObjectStoreException {
+        if (tsLogger.logger.isTraceEnabled()) {
+            tsLogger.logger.trace("SlotStore.allObjUids(" + typeName + ")");
+        }
+
+        return allObjUids(typeName, foundInstances, StateStatus.OS_UNKNOWN);
```
mmusgrov, line 104: I'd have expected this method to return all Uids regardless of their StateStatus value. But since it only returns `OS_UNKNOWN` ones will you include some javadoc for the method signature.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStoreAdaptor.java
```diff
@@ -0,0 +1,349 @@
...
+     * @return <code>true</code> if no errors occurred, <code>false</code>
+     * otherwise.
+     */
+    @Override
+    public boolean allTypes(InputObjectState foundTypes) throws ObjectStoreException {
+        if (tsLogger.logger.isTraceEnabled()) {
+            tsLogger.logger.trace("SlotStore.allTypes()");
+        }
+
+        String[] knownTypes = store.getKnownTypes(); // may contain trailing null elements
```
mmusgrov, line 120: The comment is confusing because the array of `knownTypes` is populated in the method `getKnownTypes` by the call `types.add(key.getTypeName());` but the added typeName can never be null (because the `SlotStoreKey` constructor dereferences it when checks for "/").

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStoreAdaptor.java
```diff
@@ -0,0 +1,349 @@
...
+
+        OutputObjectState buffer = new OutputObjectState();
+
+        try {
+            for (String typeName : knownTypes) {
+                if (typeName == null) {
+                    continue;
+                }
+
+                if (typeName.startsWith("/")) {
```
mmusgrov, line 135: Remark: ArjunaCore does not document that typename hierarchies are delimited by slashes
Remark: could you not leverage java.nio.files.Path to split the type name for you instead of using the following tricky code

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/RAMSlots.java
```diff
@@ -0,0 +1,47 @@
...
+package com.arjuna.ats.internal.arjuna.objectstore.slot;
+
+/**
+ * Trivial in-memory backend for the SlotStore, useful for benchmarking but not much else.
+ *
+ * @author Jonathan Halliday (jonathan.halliday@redhat.com), 2020-03
+ */
+public class RAMSlots {
+
+    public final byte[][] slots;
```
mmusgrov, line 30: My IDE is complaining that many of your access modifiers are too permissive. I know the slot store is private to SlotStoreAdaptor but restricting the modifier will stop IDE's from complaining.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStore.java
```diff
@@ -0,0 +1,222 @@
...
+            return false;
+        }
+
+        slots.write(slotId, data, config.isSyncWrites());
+        slotIdIndex.put(key, slotId);
+
+        Integer previousSlot = slotIdIndex.put(key, slotId);
+
+        // If it's a rewrite, we need to release the older version's slot
+        if (previousSlot != null) {
```
mmusgrov, line 169: So now there are two entries for the same state. Can periodic recovery kick in at this point and attempt to recover both?

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStore.java
jhalliday: The dir may be created in one run and its existence validated on the next, since it doesn't get deleted when the server shuts down. Wouldn't be a lot of use for recovery if it did! That code shouldn't really be in the RAM backed version at all, but the path also serves as the store name so it didn't get cleanly factored out.

first run of a server, exist check fails, so mkdirs runs and must succeed
subsequent runs of a server, exists check passes, so mkdirs doesn't run as it's shortcut out

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStore.java
jhalliday: The data contains the serialized key needed to rebuild the slotIdIndex on recovery, plus the serialized record. The read operation already has the key - that's the thing passed as an arg to the function. So there is no point deserializing the copy that's in the data. But we need to skip over it to reach the record state we do need. Perhaps I should reverse the packing order so the key trails the record, but as reads are rare anyhow it may not be worth the optimization

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStoreAdaptor.java
jhalliday: D'oh

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStore.java
mmusgrov: Of course, thanks for the clarification.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStoreAdaptor.java
jhalliday: Arg. That's actually a bug in getMatchingKeys - if should indeed return all by using
slotStoreKey.getStateStatus() == StateStatus.OS_UNKNOWN
rather than
key.getStateStatus() == StateStatus.OS_UNKNOWN
I should probably call them lookFor and candidateMatch or something else more distinctive.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStoreAdaptor.java
jhalliday: oops. That's a copy-paste from the Hornetq version, which can do odd things because it's backed by a ConcurrentMap. You're right, it doesn't apply here.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStoreAdaptor.java
jhalliday: ArjunaCore uses leading delimiters in some calls and not in others, considering the typename to be equivalent with or without, which works ok when backed by the file system as the file path code fixes them up when concatenating, but not when backed by non-fs things like the string indexing here. Basicall "/foo".equals("foo") is assumed by the core to be true even though it's not, because in practice new File(baseDIr, "/foo").equals( new File(baseDIr, "foo") ) is true.
Took an age to track that down when building the Hornetq version. It's another thing that got copied in and could be tweaked, but probably isn't worth the effort. More complex parsing is overkill as we don't care about / within the typename, only at the start.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/RAMSlots.java
jhalliday: The machine ain't the boss of me. Buy yeah, true.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStore.java
jhalliday: There is a time window where internally that's true, just as in wider recovery we may still have a log for a tx that has actually completed and isn't in the resource managers any more.
In the case of SlotStore it is (will be, when I write it...) indeterminate which state is used, which is fine so log as exactly one is. That's taken care of by the index, which only ever holds one even if two underlying slots are used. Periodic Recovery (in the ArjunaCore sense) sees only one, as it's going via the index. Recovery internally at the SlotStore level (one-off at startup time) will see both and insert only one into the index.

mmusgrov: Thanks for a thorough answer.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStoreAdaptor.java
mmusgrov: I was asking more about using java.nio.files.Path to split the type name instead of your hand rolled complex parsing. But if you prefer to keep the complex version then fine and we will maintain it for you ;)

mmusgrov: Actually using java.nio.* is more complex and slower. Please ignore my comment.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStoreAdaptor.java
mmusgrov: @jhalliday I was also expecting a javadoc entry for this method since it only return only OS_UNKNOWN states, why not all states and whatever the answer is shouldn't the javadoc say why?

jhalliday: It returns all states - OS_UNKNOWN functions as a wildcard in this case. I think that started out as a quirk of the FileStore and then spread, it's not documented behaviour of the RecoveryStore interface, but I'm happy to make it so. The javadoc in the adaptor is mostly copy/paste of the interface javadoc just to help me remember how things should work, which is why that method didn't get any.

PR #1590, opened 2020-04-08
JBTM-3273 -> Avoid implementing InvocationHandler in jtaLogger

mayankkunwar:
https://issues.redhat.com/browse/JBTM-3273

!MAIN
!QA_JTS_JACORB
!QA_JTS_JDKORB
!QA_JTS_OPENJDKORB
AS_TESTS JDK11
!PERF
!XTS
!CORE
!BLACKTIE
!RTS
!mysql
!postgres
!db2
!oracle
!TOMCAT
!JACOCO
!LRA

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/RecoveryRequired.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/RecoveryRequired.java
@@ -1,0 +1,14 @@
+package com.arjuna.ats.jta.logging;
+
+public class RecoveryRequired {
+
+    private static boolean recoveryProblems;
+
+    public static boolean isRecoveryProblems() {
+        return recoveryProblems;
+    }
+
+    public static void setRecoveryProblems(boolean recoveryProblems) {
+        RecoveryRequired.recoveryProblems = recoveryProblems;
+    }
+}
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/jtaLogger.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/jtaLogger.java
@@ -33,47 +33,22 @@ package com.arjuna.ats.jta.logging;
 
 import org.jboss.logging.Logger;
 
-import java.lang.reflect.InvocationHandler;
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
-import java.lang.reflect.Proxy;
-
-public class jtaLogger implements InvocationHandler
+public class jtaLogger
 {
     public static final Logger logger = Logger.getLogger("com.arjuna.ats.jta");
-    public static final jtaI18NLogger i18NLogger = (jtaI18NLogger) Proxy.newProxyInstance(
-            jtaI18NLogger.class.getClassLoader(),
-            new Class[] { jtaI18NLogger.class },
-            new jtaLogger(Logger.getMessageLogger(jtaI18NLogger.class, "com.arjuna.ats.jta")));
+
+    /**
+     * For jtaI18NLogger, if any method, prefixed with <b>"warn_recovery"</b> is called then,
+     * <b>RecoverRequired</b> class variable i.e. "recoveryProblem" should be set as <b>true</b>.
+     * */
+    public static final jtaI18NLogger i18NLogger = Logger.getMessageLogger(jtaI18NLogger.class, "com.arjuna.ats.jta");
 
     private jtaI18NLogger jtaI18NLoggerImpl;
-    private static boolean recoveryProblems;
 
     private jtaLogger() {
     }
 
     private jtaLogger(jtaI18NLogger logger) {
         jtaI18NLoggerImpl = logger;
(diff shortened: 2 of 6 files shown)


mayankkunwar:
Since the changes are not what was expected, till then I will put this PR on Hold.

mayankkunwar:
@ochaloup could you please review the code.

mayankkunwar:
Thanks for the approval. I will merge it now

ochaloup:
@mmusgrov @mayankkunwar  I'm sorry for not getting to this sooner. I will try better next time.
Still I consider there is at least one pitfall in the concurrency handling. I consider the JBTM-3273 as not fully finished yet. I need to check this in more details and I will let you know back.

mayankkunwar:
Okay @ochaloup .. Thanks

ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/RecoveryRequired.java
```diff
@@ -0,0 +1,14 @@
+package com.arjuna.ats.jta.logging;
```
ochaloup, line 1: I'm not sure about the placing the class under the `logging` package. As it informs about recovery processing it makes more sense to me to put it under `recovery` package. What was your reasoning about placing it here?

ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/RecoveryRequired.java
```diff
@@ -0,0 +1,14 @@
...
+
+public class RecoveryRequired {
```
ochaloup, line 3: What is your reasoning for naming the class as `RecoveryRequired`?

ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/RecoveryRequired.java
```diff
@@ -0,0 +1,14 @@
...
+
+    private static boolean recoveryProblems;
```
ochaloup, line 5: As this is a flag which is used to mark if a recovery problem occurs and this flag could be accessed from different threads and value could be set from the different threads then the code has to(!) consider that. The property should be of type atomic.

ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/jtaLogger.java
```diff
@@ -33,47 +33,22 @@
...
+public class jtaLogger
 {
     public static final Logger logger = Logger.getLogger("com.arjuna.ats.jta");
-    public static final jtaI18NLogger i18NLogger = (jtaI18NLogger) Proxy.newProxyInstance(
-            jtaI18NLogger.class.getClassLoader(),
-            new Class[] { jtaI18NLogger.class },
-            new jtaLogger(Logger.getMessageLogger(jtaI18NLogger.class, "com.arjuna.ats.jta")));
+
+    /**
+     * For jtaI18NLogger, if any method, prefixed with <b>"warn_recovery"</b> is called then,
```
ochaloup, line 41: I think this is a good comment but I consider it should be put into the class `jtaI18Logger` as it's the place where new methods for logging are added (and where the methods are named). I think when it's here it's better chance to be missed.

ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/RecoveryRequired.java
mayankkunwar: Since this class is only used by jtaLogger for now, therefore I thought of placing it under `com.arjuna.ats.jta.logging` package would be fine for now. Maybe later if any other package requires it, then we could move it under ` recovery ` package?

ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/RecoveryRequired.java
mayankkunwar: I thought it would add meaning to the action, that user has to take after he gets the result.

ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/RecoveryRequired.java
mayankkunwar: Sorry I missed this thought. Thanks for pointing this out. I will make it thread safe.

ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/jtaLogger.java
mayankkunwar: Sure, I will move the comments to jtaI18Logger

ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/RecoveryRequired.java
mayankkunwar: I will move it to `recovery` package

PR #1596, opened 2020-04-14
JBTM-3286 Expose OpenAPI and Swagger definitions for LRA coordinator

xstefank:
https://issues.redhat.com/browse/JBTM-3286

!MAIN !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF LRA !NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

PR #1599, opened 2020-04-15
CI SCRIPT: on testing using -fae maven option

ochaloup:
MAIN LRA XTS RTS AS_TESTS  TOMCAT
!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !PERF NO_WIN !JACOCO

### DIFF
--- a/scripts/hudson/narayana.sh
+++ b/scripts/hudson/narayana.sh
@@ -432,7 +432,7 @@ function osgi_tests {
 function xts_as_tests {
   echo "#-1. XTS AS Integration Test"
   cd ${WORKSPACE}/jboss-as
-  ./build.sh -f testsuite/integration/xts/pom.xml -B -Pxts.integration.tests.profile -Dversion.org.jboss.narayana=${NARAYANA_CURRENT_VERSION} "$@" test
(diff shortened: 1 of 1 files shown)


mmusgrov:
@ochaloup Why aren't we applying this change to the other profiles too?

ochaloup:
@mmusgrov The change was applied based on the Martin' request on PR https://github.com/jbosstm/narayana/pull/1596
I think adding the `-fae` option to some of the other axis of the script makes sense. I'm going to to so.

ochaloup:
I'm merging this PR as it was agreed that `-fae` option is desirable.
I started the tests for reason to see if I didn't make some error in the `sh` commands. There are the CI failures but all what I can see are caused by some functional instability but not with the wrong `sh` script starting them (which was this PR changeset about).

* AS_TESTS failed on not being able to catch the starting app server with the cli command. Something I'm trying to understand whole this week.
* RTS failed because of the maven central http request failed
* LRA failed because of the known issue which will be fixed in TCK

PR #1603, opened 2020-04-17
JBTM-3247 Do not delete LRA records on the failed list

mmusgrov:
https://issues.redhat.com/browse/JBTM-3291

LRA MAIN !AS_TESTS !QA_JTA !JACOCO !QA_JTS_JACORB !TOMCAT !RTS !XTS !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

This PR changes the LRA implementation such that we now retain failure records (previously we reported the failure and then deleted them).

The OpenAPI doc for reading the records justifies their retention by saying:

"Failure records are vital pieces of data needed to aid failure tracking and analysis and are retained for inspection."

And I raised [JBTM-3297](https://issues.redhat.com/browse/JBTM-3297) to move these records to another part of the store so that standard LRA recovery processing does not need to read them.

### DIFF
--- a/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/RecoveryCoordinator.java
+++ b/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/RecoveryCoordinator.java
@@ -21,6 +21,7 @@
  */
 package io.narayana.lra.coordinator.api;
 
+import io.narayana.lra.coordinator.domain.model.LRAData;
 import io.narayana.lra.coordinator.domain.model.LRAStatusHolder;
 import io.narayana.lra.coordinator.domain.service.LRAService;
 import io.narayana.lra.logging.LRALogger;
@@ -36,6 +37,7 @@ import org.jboss.logging.Logger;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
+import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.NotFoundException;
 import javax.ws.rs.PUT;
@@ -53,6 +55,7 @@ import java.util.List;
 
 import static io.narayana.lra.LRAConstants.RECOVERY_COORDINATOR_PATH_NAME;
 import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
+import static javax.ws.rs.core.Response.Status.PRECONDITION_FAILED;
 
 @ApplicationScoped
 @Path(RECOVERY_COORDINATOR_PATH_NAME)
@@ -151,5 +154,68 @@ public class RecoveryCoordinator {
         content = @Content(schema = @Schema(type = SchemaType.ARRAY, implementation = LRAStatusHolder.class)))
     public List<LRAStatusHolder> getRecoveringLRAs() {
         return lraService.getAllRecovering(true);
+    }
+
+    @GET
+    @Path("failed")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Operation(summary = "List failed Long Running Actions",
+            description = "Returns LRAs that have failed. " +
+                    " Failure records are vital pieces of data needed to aid failure tracking and analysis " +
+                    " and are retained for inspection.")
+    @APIResponse(responseCode = "200",
+            content = @Content(schema = @Schema(type = SchemaType.ARRAY, implementation = LRAStatusHolder.class)))
+    public List<LRAStatusHolder> getFailedLRAs() {
+        return lraService.getFailedLRAs();
+    }
+
+    @DELETE
+    @Path("{LraId}")
+    @Operation(summary = "Remove the log for a failed LRA")
+    @APIResponses({
+            @APIResponse(responseCode = "204",
+                    description = "If the LRA log was successfully removed"),
+            @APIResponse(responseCode = "412",
+                    description = "If the LRA is not in an end state (in which case the response " +
+                            "entity will indicate the current state at the time of the request)"),
+            @APIResponse(responseCode = "412",
+                    description = "If the input LRA does not correspond to a valid URI (in which case the " +
+                            "response entity will contain the error message)"),
+            @APIResponse(responseCode = "500",
+                    description = "If the attempt to remove the LRA log failed. This return code does not " +
+                            "discriminate between a failure at the log storage level or if the log did not exist)")
+    })
+    public Response deleteFailedLRA(
+            @Parameter(name = "LraId", description = "The unique identifier of the LRA", required = true)
+            @PathParam("LraId")String lraId) throws NotFoundException {
+        URI lra;
+
+        try {
+            lra = new URI(lraId);
+
+            // verify that the LRA is not still being processed
+            // will throw NotFoundException if it's unknown or is failed (to be caught and processed in the catch block)
+            LRAData lraData = lraService.getLRA(lra);
+
+            // 412 the LRA is not in an end state (return 412 and the actual status of the LRA)
+            return Response.status(PRECONDITION_FAILED).entity(lraData.getStatus()).build();
+        } catch (NotFoundException e) {
+            // the LRA has finished and, if it corresponds to a failure record, it is safe to delete it
+            if (lraService.removeLog(lraId)) {
+                // 204 the log for the LRA was successfully removed
+                return Response.noContent().build(); // return 204
+            }
+
+            // 500 remove log failed (or it was not present)
+            return Response.status(INTERNAL_SERVER_ERROR).build();
+        } catch (URISyntaxException e) {
+            // 412 the user provided URI was invalid
+            if (LRALogger.logger.isDebugEnabled()) {
+                LRALogger.logger.debugf("%s#deleteLRA: %s: %s",
+                        getClass().getCanonicalName(), lraId, e.getMessage());
+            }
+
+            return Response.status(PRECONDITION_FAILED).entity(String.format("%s: %s", lraId, e.getMessage())).build();
+        }
     }
 }
--- a/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
+++ b/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
@@ -235,7 +235,7 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
     // - all compensators compensated ok
     // This participant can infer which possibility happened since it will have been told to complete or compensate
     public int topLevelAbort() {
-        return doEnd(true);
+        return doEnd(lra.isCancel());
     }
 
     @Override
@@ -245,7 +245,7 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
 
     @Override
     public int topLevelCommit() {
-        return doEnd(false);
+        return doEnd(lra.isCancel());
     }
 
     private int doEnd(boolean compensate) {
@@ -352,7 +352,7 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
                 httpStatus == Response.Status.INTERNAL_SERVER_ERROR.getStatusCode()) {
             // the body should contain a valid ParticipantStatus
             try {
-                return atEnd(reportFailure(compensate, endPath.toASCIIString(),
+                return atEnd(reportFailure(compensate, endPath,
                         ParticipantStatus.valueOf(responseData).name()));
             } catch (IllegalArgumentException ignore) {
                 // ignore the body and let recovery discover the status of the participant
@@ -367,20 +367,9 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
                         endPath, httpStatus);
             }
 
-            if (compensate) {
-                status = ParticipantStatus.Compensating; // recovery will figure out the status via the status url
-
-                /*
-                 * We are mapping compensate onto Abort. TwoPhaseCoordinator uses presumed abort
-                 * so if we were to return FINISH_ERROR recovery would not replay the log.
-                 * To force the record to be eligible for recovery we return a heuristic hazard.
-                 */
-                return atEnd(TwoPhaseOutcome.HEURISTIC_HAZARD);
-            }
-
-            status = ParticipantStatus.Completing; // recovery will figure out the status via the status url
-
-            return atEnd(TwoPhaseOutcome.FINISH_ERROR);
+            // recovery will figure out the status via the status url
+            status = compensate ? ParticipantStatus.Compensating : ParticipantStatus.Completing;
+            accepted = true;
         }
 
         updateStatus(compensate);
@@ -407,6 +396,11 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
             default:
                 return false;
         }
+    }
+
+
+    boolean isFailed() {
+        return status == ParticipantStatus.FailedToCompensate || status == ParticipantStatus.FailedToComplete;
     }
 
     private boolean afterLRARequest(URI target, String payload) {
@@ -438,6 +432,12 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
     }
 
     private int atEnd(int res) {
+        if (parentId != null
+                && (status == ParticipantStatus.Completed || status == ParticipantStatus.FailedToComplete)) {
+            // completed nested participants must remain compensatable
+            return TwoPhaseOutcome.HEURISTIC_HAZARD; // ask to be called again
+        }
+
         // Only run the post LRA actions if both the LRA and participant are in an end state
         // check the participant first since it will have been removed from one of the lists
         if (!isFinished() || !lra.isFinished()) {
@@ -467,15 +467,14 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
         }
     }
 
-    private int reportFailure(boolean compensate, String endPath, String failureReason) {
+    private int reportFailure(boolean compensate, URI endPath, String failureReason) {
         status = compensate ? ParticipantStatus.FailedToCompensate : ParticipantStatus.FailedToComplete;
 
         LRALogger.logger.warnf("LRARecord: participant %s reported a failure to %s (cause %s)",
-                endPath, compensate ? COMPENSATE_REL : COMPLETE_REL, failureReason);
+                endPath.toASCIIString(), compensate ? COMPENSATE_REL : COMPLETE_REL, failureReason);
 
-        // permanently failed so tell recovery to ignore us in the future. TODO could move it to
-        // another list for reporting
-        return TwoPhaseOutcome.FINISH_OK;
+        // permanently failed so ask recovery to ignore us in the future.
+        return TwoPhaseOutcome.FINISH_ERROR;
     }
 
     private int retryGetEndStatus(URI endPath, boolean compensate) {
@@ -508,7 +507,7 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
                         return TwoPhaseOutcome.HEURISTIC_HAZARD;
                     case FailedToClose:
                     case FailedToCancel:
-                        return TwoPhaseOutcome.FINISH_ERROR;
+                        return reportFailure(compensate, endPath, "unknown");
                     default:
                         return TwoPhaseOutcome.HEURISTIC_HAZARD;
                 }
@@ -539,7 +538,10 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
                      */
                     status = compensate ? ParticipantStatus.Compensated : ParticipantStatus.Completed;
                     return TwoPhaseOutcome.FINISH_OK;
-                } else if (response.getStatus() == Response.Status.ACCEPTED.getStatusCode()) {
+                } else if (response.getStatus() == Response.Status.ACCEPTED.getStatusCode() ||
+                        Response.Status.Family.familyOf(response.getStatus()).equals(
+                            Response.Status.Family.SERVER_ERROR)) {
+                    // these response codes indicate that the implementation should retry later
                     return TwoPhaseOutcome.HEURISTIC_HAZARD;
                 } else if (response.getStatus() == Response.Status.OK.getStatusCode() &&
                         response.getResponseString() != null) {
@@ -562,14 +564,13 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
                                     compensate, endPath, status);
 
                             if (forgetURI != null) {
-                                return forget() ? TwoPhaseOutcome.FINISH_OK : TwoPhaseOutcome.HEURISTIC_HAZARD;
+                                if (!forget()) {
+                                    // we will retry the forget on the next recovery cycle
+                                    return TwoPhaseOutcome.HEURISTIC_HAZARD;
+                                }
                             }
 
-                            if (compensate) {
-                                return TwoPhaseOutcome.FINISH_OK;
-                            }
-
-                            return TwoPhaseOutcome.FINISH_OK;
+                            return reportFailure(compensate, endPath, "Unknown");
                         default:
                             return TwoPhaseOutcome.HEURISTIC_HAZARD;
                     }
--- a/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
+++ b/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
@@ -55,6 +55,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
@@ -121,8 +122,7 @@ public class Transaction extends AtomicAction {
 
     public boolean save_state(OutputObjectState os, int ot) {
         if (!super.save_state(os, ot)
-                || !save_list(os, ot, pendingList)
-                || !save_list(os, ot, preparedList)) {
+                || !save_list(os, ot, pendingList)) { // other lists are maintained in BasicAction
             return false;
         }
 
@@ -230,9 +230,33 @@ public class Transaction extends AtomicAction {
     public boolean restore_state(InputObjectState os, int ot) {
 
         if (!super.restore_state(os, ot)
-                || !restore_list(os, ot, pendingList)
-                || !restore_list(os, ot, preparedList)) {
+                || !restore_list(os, ot, pendingList)) { // other lists are maintained in BasicAction
             return false;
+        }
+
+        // restore_state may have put failed records onto the prepared list so move them back again:
+        for (AbstractRecord rec = preparedList.peekFront(); rec != null; rec = preparedList.peekNext(rec)) {
+            if (rec instanceof LRARecord) {
+                LRARecord p = (LRARecord) rec;
+
+                if (p.isFailed()) {
+                    boolean moveRec = true;
+                    AbstractRecord r;
+
+                    preparedList.remove(p);
+
+                    // put it back on the failedList if it isn't already on it
+                    for (r = failedList.peekFront(); r != null; r = failedList.peekNext(r)) {
+                        if (p.equals(r)) {
+                            break;
+                        }
+                    }
+
+                    if (r == null) {
+                        failedList.putFront(p);
+                    }
+                }
+            }
         }
 
         uid = get_uid().fileStringForm();
@@ -350,6 +374,18 @@ public class Transaction extends AtomicAction {
                 (status.equals(LRAStatus.Cancelling) || status.equals(LRAStatus.Closing));
     }
 
+
+    public boolean isCancel() {
+        switch (status) {
+            case Cancelling: /* FALLTHRU */
+            case Cancelled: /* FALLTHRU */
+            case FailedToCancel:
+                return true;
+            default:
+                return false;
+        }
+    }
+
     boolean isFinished() {
         switch (status) {
             case Closed:
@@ -428,13 +464,16 @@ public class Transaction extends AtomicAction {
 
                     // repopulate the pending list TODO it won't neccessarily be present during recovery
                     pendingList = new RecordList();
+                    preparedList = new RecordList();
+                    failedList = new RecordList();
 
-                    pending.forEach(r -> pendingList.putRear(r));
+                    moveTo(pendingList, preparedList);
+                    moveTo(heuristicList, preparedList);
 
                     updateState(LRAStatus.Cancelling);
 
-                    super.phase2Abort(true);
-//                res = super.Abort();
+                    // call commit since the abort route does not save the failed list
+                    super.phase2Commit(true);
 
                     res = status();
 
@@ -447,11 +486,10 @@ public class Transaction extends AtomicAction {
         } else {
             if (cancel || status() == ActionStatus.ABORT_ONLY) {
                 // compensators must be called in reverse order so reverse the pending list
-                int sz = pendingList == null ? 0 : pendingList.size();
-
-                for (int i = sz - 1; i > 0; i--) {
-                    pendingList.putRear(pendingList.getFront());
-                }
+                preparedList = new RecordList();
+                failedList = new RecordList();
+                moveTo(pendingList, preparedList);
+                moveTo(heuristicList, preparedList);
 
                 // tell each participant that the LRA canceled
                 updateState(LRAStatus.Cancelling);
@@ -461,7 +499,8 @@ public class Transaction extends AtomicAction {
                     heuristicList = new RecordList();
                 }
 
-                super.phase2Abort(true); // this route to abort forces a log write on failures and heuristics
+                // call commit since the abort route does not save the failed list
+                super.phase2Commit(true);
                 res = super.status();
             } else {
                 // tell each participant that the LRA closed ok
@@ -480,14 +519,15 @@ public class Transaction extends AtomicAction {
             }
         }
 
-        if (getSize(heuristicList) != 0 || getSize(failedList) != 0) {
+        if (getSize(heuristicList) != 0) {
             status = cancel ? LRAStatus.Cancelling : LRAStatus.Closing;
+        } else if (getSize(failedList) != 0) {
+            status = cancel ? LRAStatus.FailedToCancel : LRAStatus.FailedToClose;
         } else if (getSize(pendingList) != 0 || getSize(preparedList) != 0) {
             status = LRAStatus.Closing;
         } else {
             status = toLRAStatus(res);
         }
-
 
         if (!isRecovering()) {
             if (lraService != null) {
@@ -499,17 +539,18 @@ public class Transaction extends AtomicAction {
 
         finishTime = LocalDateTime.now(ZoneOffset.UTC);
 
+        deactivate(); // ensure that the new status is persisted
+
         return res;
     }
 
     protected void runPostLRAActions() {
-        // if there are no more heuristics or failures then update the status of the LRA
-        if (isInEndState()) {
+        // if there are no more heuristics then update the status of the LRA
+        if (isInEndState() && heuristicList != null && heuristicList.size() != 0) {
             if (preparedList == null) {
                 preparedList = new RecordList();
             }
             moveTo(heuristicList, preparedList);
-            moveTo(failedList, preparedList);
             checkParticipant(preparedList);
             super.phase2Commit(true);
         }
@@ -566,10 +607,11 @@ public class Transaction extends AtomicAction {
     }
 
     protected boolean isInEndState() {
+        // update the status first by checking for heuristics and failed participants
         if (status == LRAStatus.Cancelling && allFinished(heuristicList, failedList)) {
-            status = LRAStatus.Cancelled;
+            status = ((failedList != null) && (failedList.size() == 0)) ? LRAStatus.Cancelled : LRAStatus.FailedToCancel;
         } else if (status == LRAStatus.Closing && allFinished(heuristicList, failedList)) {
-            status = LRAStatus.Closed;
+            status = ((failedList != null) && (failedList.size() == 0)) ? LRAStatus.Closed : LRAStatus.FailedToClose;
         }
 
         return isFinished();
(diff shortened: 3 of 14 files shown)


mmusgrov:
@ochaloup There was one MP-LRA TCK test failure which is unrelated to my changes. I propose to investigate it independently of this PR. So I would say that this PR is ready for review.   

> > [ERROR] joinWithOneResourceSameMethodTwiceWithCancel(org.eclipse.microprofile.lra.tck.TckTests)  Time elapsed: 16.723 s  <<< FAILURE!
> java.lang.AssertionError: joinWithOneResourceSameMethodTwiceWithCancel:: resource should have compensated with no completions
> 	at org.eclipse.microprofile.lra.tck.TckTests.joinWithOneResource(TckTests.java:544)
> 	at org.eclipse.microprofile.lra.tck.TckTests.joinWithOneResourceSameMethodTwiceWithCancel(TckTests.java:462)

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/RecoveryCoordinator.java
```diff
@@ -152,4 +153,41 @@ public String replaceCompensator(
...
+            @Parameter(name = "LraId", description = "The unique identifier of the LRA", required = true)
+            @PathParam("LraId")String lraId) throws NotFoundException {
+        try {
+            if (lraService.removeLog(new URI(lraId))) {
+                return Response.ok().build();
+            }
+
+            return Response.status(INTERNAL_SERVER_ERROR).build();
+        } catch (URISyntaxException e) {
+            return Response.status(INTERNAL_SERVER_ERROR).entity(String.format("%s: %s", lraId, e.getMessage())).build();
```
ochaloup, line 190: I think it would be good to log the full exception stacktrace when asked by more verbose logging.
It's handy in ivestigation on what's happening in the system.
I think when `TRACE` level is enabled the exception with stacktrace and probably some information what was the reason should be logged in to the coordinator log.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -228,11 +228,35 @@ private boolean restore_list(InputObjectState os, int ot, RecordList list) {
...
+                LRARecord p = (LRARecord) rec;
+
+                if (p.isFailed()) {
+                    boolean moveRec = true;
+                    AbstractRecord r;
+
+                    preparedList.remove(p);
+
+                    // put it back on the failedList if it isn't already on it
+                    for (r = failedList.peekFront(); r != null; r = failedList.peekNext(r)) {
```
ochaloup, line 247: If I'm not mistaken by looking into the code the `failedList` (and neither the `preparedList`) are guarded by some lock or other synchronization guard (while the `RecordList` is not meant to be thread safe). As I'm not fully in-depth into the logic my point is if this code is safe when executed from different threads (if it may happen)?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -207,6 +207,80 @@ public void getRecoveringLRAs(Map<URI, Transaction> lras) {
...
+                            Transaction lra = new Transaction(lraService, new Uid(uid));
+
+                            lra.activate(); // or just parse lraId extracting the uid
+
+                            if (uid.fileStringForm().equals(lra.getUid())) {
+                                return _recoveryStore.remove_committed(uid, lra.type());
+                            }
+                        }
+                    } catch (Exception ex) {
+                        moreUids = false;
```
ochaloup, line 234: Could we log the exception reason and the stacktrace under `TRACE` log level, please?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -207,6 +207,80 @@ public void getRecoveringLRAs(Map<URI, Transaction> lras) {
...
+                    }
+                }
+            }
+        } catch (ObjectStoreException e) {
+            if (LRALogger.logger.isInfoEnabled()) {
```
ochaloup, line 239: Could we log the exception reason and the stacktrace under TRACE log level, please?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -207,6 +207,80 @@ public void getRecoveringLRAs(Map<URI, Transaction> lras) {
...
+                        if (uid.equals(Uid.nullUid())) {
+                            moreUids = false;
+                        } else {
+                            Transaction lra = new Transaction(lraService, new Uid(uid));
+
+                            lra.activate();
+                            lras.put(lra.getId(), lra);
+                        }
+                    } catch (Exception ex) {
+                        moreUids = false;
```
ochaloup, line 267: Could we log the exception reason and the stacktrace under TRACE log level, please?

ochaloup:
And I agree with moving the failed `joinWithOneResourceSameMethodTwiceWithCancel` to a separate issue. I was checking this on my laptop and it seems it's an intermittent failure. (Where intermittent failures are the bad ones :-/)

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/RecoveryCoordinator.java
mmusgrov: +1

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
mmusgrov: These lists are managed by ArjunaCore and BasicAction - things like save and restore state and commit etc are controlled from BasicAction which have there own synchronisation control.

I can create a JIRA that calls for stress testing of the implementation to try to tease out any dormant issues in this area?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mmusgrov: +1

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mmusgrov: +1

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mmusgrov: +1

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/FailedLRAIT.java
```diff
@@ -0,0 +1,232 @@
...
+            }
+        }
+    }
+
+    // ask the recovery coordinator to delete its log for an LRA
+    private int removeFailedLRA(URI lra) throws URISyntaxException, UnsupportedEncodingException {
+        Response response = null;
+        String recoveryUrl = String.format("http://%s:%d/%s",
+                lra.getHost(), lra.getPort(), LRAConstants.RECOVERY_COORDINATOR_PATH_NAME);
+        String txId = URLEncoder.encode(lra.toASCIIString(), "UTF-8");
```
xstefank, line 220: I would expect to also see the operation getFailedLRAs and removeFailed(URI) in NarayanaLRAClient. WDYT?

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/resource/LRAParticipantWithStatusURI.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @GET
+    @Path(FORGET_COUNT_PATH)
+    public int getForgetCount() {
+        return forgetCount.get();
+    }
+
+    @PUT
+    @Path("compensate")
+    @Produces(MediaType.APPLICATION_JSON)
```
xstefank, line 82: it produces Enums name, so it should be probably text/plain.

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/resource/LRAParticipantWithStatusURI.java
```diff
@@ -0,0 +1,125 @@
...
+    public Response compensateWork(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId)
+            throws NotFoundException {
+        compensateCount.incrementAndGet();
+
+        return Response.status(500).entity(ParticipantStatus.FailedToCompensate.name()).build();
+    }
+
+    @PUT
+    @Path("complete")
+    @Produces(MediaType.APPLICATION_JSON)
```
xstefank, line 93: also here I would suggest text/plain

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/resource/LRAParticipantWithStatusURI.java
```diff
@@ -0,0 +1,125 @@
...
+    @Complete
+    public Response completeWork(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId)
+            throws NotFoundException {
+        completeCount.incrementAndGet();
+
+        return Response.status(500).entity(ParticipantStatus.FailedToComplete.name()).build();
```
xstefank, line 99: Don't Narayana retry on 500 status code? In the spec we use 200 and only the response body (here correct FailedToComplete) to distinguish failed states.

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/resource/LRAParticipantWithStatusURI.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    @DELETE
+    @Path("delete")
+    @Forget
+    public Response foget() {
```
xstefank, line 105:
```suggestion
    public Response forget() {
```

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/FailedLRAIT.java
```diff
@@ -0,0 +1,232 @@
...
+            fail("lra not in failed list");
+        }
+    }
+
+    private URI invokeInTransaction(String resourcePath, int expectedStatus) {
+        Response response = null;
+
+        try {
+            response = client.target(UriBuilder.fromUri(baseURL.toExternalForm())
+                    .path(LRAParticipantWithStatusURI.LRA_PARTICIPANT_PATH)
```
xstefank, line 145: This method is called also in tests that should test LRAParticipantWithoutStatusURI so this value can't be hardcoded if the method is to be reused.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/RecoveryCoordinator.java
```diff
@@ -152,4 +153,41 @@ public String replaceCompensator(
...
+    public List<LRAStatusHolder> getFailedLRAs() {
+        return lraService.getFailedLRAs();
+    }
+
+    @DELETE
+    @Path("{LraId}")
+    @Operation(summary = "Remove the log for an LRA")
+    @APIResponses({
+            @APIResponse(responseCode = "500",
+                    description = "The LRA was not recognised"),
```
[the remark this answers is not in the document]
mmusgrov: +1

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/RecoveryCoordinator.java
```diff
@@ -152,4 +153,41 @@ public String replaceCompensator(
...
+            @APIResponse(responseCode = "200",
+                    description = "If the LRA was successfully removed",
+                    content = @Content(schema = @Schema(implementation = Boolean.class)))
```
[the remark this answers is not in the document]
mmusgrov: Oops

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/RecoveryCoordinator.java
```diff
@@ -152,4 +153,41 @@ public String replaceCompensator(
...
+    })
+    public Response deleteLRA(
+            @Parameter(name = "LraId", description = "The unique identifier of the LRA", required = true)
+            @PathParam("LraId")String lraId) throws NotFoundException {
+        try {
+            if (lraService.removeLog(new URI(lraId))) {
+                return Response.ok().build();
```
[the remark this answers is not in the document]
mmusgrov: Returning a resource, or more correctly a link to it, that no longer exists seems wrong in a REST context. How about just returning `204 No Content` with an empty body.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -245,7 +245,7 @@ public int topLevelOnePhaseCommit() {
 
     @Override
     public int topLevelCommit() {
-        return doEnd(false);
+        return doEnd(lra.isCancel());
```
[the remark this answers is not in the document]
mmusgrov: We always commit the transaction (at the level of BasicAction) which then calls commit on each record enlisted with the BasicAction. The private method called LRARecord#doEnd needs to know whether the LRA is cancelling or closing, hence the query to ask the LRA whether it is cancelling or not.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -367,20 +367,9 @@ private int tryDoEnd(boolean compensate) {
...
-                 */
-                return atEnd(TwoPhaseOutcome.HEURISTIC_HAZARD);
-            }
-
-            status = ParticipantStatus.Completing; // recovery will figure out the status via the status url
-
-            return atEnd(TwoPhaseOutcome.FINISH_ERROR);
+            // recovery will figure out the status via the status url
+            status = compensate ? ParticipantStatus.Compensating : ParticipantStatus.Completing;
+            accepted = true;
```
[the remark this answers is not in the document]
mmusgrov: That would be useful. Since we'd need to expose that information in the JSON responses can we have a new JIRA for that?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -562,14 +564,13 @@ private int retryGetEndStatus(URI endPath, boolean compensate) {
                                     compensate, endPath, status);
 
                             if (forgetURI != null) {
-                                return forget() ? TwoPhaseOutcome.FINISH_OK : TwoPhaseOutcome.HEURISTIC_HAZARD;
+                                if (!forget()) {
+                                    // we will retry the forget on the next recovery cycle
```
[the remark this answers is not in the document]
mmusgrov: The code has just tried the forget call which has failed (`if (!forget())`) on the previous line so we need to make sure the forget call is retried later so we return a HEURISTIC_HAZARD)

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/FailedLRAIT.java
mmusgrov: Hmm, that's a good point, thanks.

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/FailedLRAIT.java
mmusgrov: Currently we only use NarayanaLRAClient internally for the participant JAX-RS filter and some of our tests.
The API to the coordinator is the OpenAPI doc.

So there would be no consumers of the two methods that you are proposing. But I do agree that these two methods would be useful in a client API at the spec level.

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/resource/LRAParticipantWithStatusURI.java
mmusgrov: Okay

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/resource/LRAParticipantWithStatusURI.java
mmusgrov: Okay

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/resource/LRAParticipantWithStatusURI.java
mmusgrov: That is only if the method returns an invalid enum value. The actual meaning of the relevant codes is:

200 means the resource has completed (https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java#L116)

500 means the resource has failed to complete (https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java#L129)

If any other code is returned (or, in the 500 case, the body does not correspond to a valid state)  then the implementation SHOULD either keep retrying or attempt to discover the status (https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java#L163)

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/resource/LRAParticipantWithStatusURI.java
mmusgrov: +1

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/RecoveryCoordinator.java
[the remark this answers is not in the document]
xstefank: This is just from my experience, DB provides etc also do this. But ok 204 can do also.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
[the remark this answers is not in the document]
xstefank: https://issues.redhat.com/browse/JBTM-3299

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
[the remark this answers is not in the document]
xstefank: ah, my miss. Thanks. You can resolve this.

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/FailedLRAIT.java
xstefank: I would expect narayana users to use the client directly. If they won't be spec only of course. Should we take this in another JIRA or will you update this PR?

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/resource/LRAParticipantWithStatusURI.java
xstefank: My mistake, thanks. This can be resolved.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/RecoveryCoordinator.java
mmusgrov: Although this stack trace will only be showing Resteasy internals since the exception is saying that the user passed an invalid URI and therefore the stacktrace will not provide any useful information (just noise).

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
ochaloup: I think the stress testing is good but not necessary for the initial insight. If there is the code which can be accessed from different threads and is not synchronized we should be thinking about the circumstances.
As I'm looking into this once again with the understanding the code is invoked from the `BasicAction` I can see that the lists are used in the `Transaction` processing which is the instance of the thread local. That way the processing should be bound to the one thread and the concurrent access of the lists would not be possible.
I"m resolving this doubt here now.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/RecoveryCoordinator.java
ochaloup: I think information that the uri syntax was wrong with details why it was wrong could be still usable for a detailed analysis.

mmusgrov: Oh yes definitely. All I am saying is that the stacktrace will not provide any additional information. I am adding trace level stack traces for the other places you ask for though.

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/FailedLRAIT.java
mmusgrov: We don't require a new JIRA. Clients will use the REST api.

mmusgrov: Okay I will add it

mmusgrov: We don't have any methods in NarayanaLRAClient that return collections of LRAs so let's take that to another JIRA if you don't mind.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/RecoveryCoordinator.java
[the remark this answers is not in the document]
mmusgrov: I made the change but reordered the APIResponse annotations (with 204 first)

mmusgrov:
@xstefank and @ochaloup I have resolved both of your comments.

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/FailedLRAIT.java
xstefank: ah sure because we didn't change it since the spec version. However, I think this is a good idea to add all operations exposed by coordinator to the programmatic API also. Later when we again introduce client API in the spec we will just have more methods in narayana client so users can chose to inject spec version or narayana one. Please create a new JBTM, we should add also other methods that the coordinator exposes as JAX-RS endpoints. This can be resolved.

mmusgrov: https://issues.redhat.com/browse/JBTM-3303

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -177,7 +152,10 @@ private void processTransactionsStatus() {
                         doRecoverTransaction(currentUid);
                     }
                 } catch (ObjectStoreException e) {
-                    if (LRALogger.logger.isInfoEnabled()) {
+                    if (LRALogger.logger.isTraceEnabled()) {
+                        LRALogger.logger.tracef("failed to access transaction store %s: %s",
```
ochaloup, line 156: This won't print the stacktrace as it uses the `tracef` definition of `tracef(String format, Object... params)` but it needs to be used the `tracef(Throwable t, String format, Object... params)`
Then printing "manually" the exception message with `e.getMessage()` is not necessary as the message will be automatically printed by logger when exception `e` is passed as `Throwable`.

It could be like

```
LRALogger.logger.tracef(e, "failed to access transaction store %s", currentUid);
```

And maybe one more question on the message. Is the `currentUid` meaning "an access of the transaction store". Should not be the message informs about not able to accessing some LRA UID or similar?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -207,7 +185,103 @@ public void getRecoveringLRAs(Map<URI, Transaction> lras) {
...
+
+            lra.activate(); // or just parse lraId extracting the uid
+
+            if (uid.fileStringForm().equals(lra.getUid())) {
+                try {
+                    success.set(_recoveryStore.remove_committed(uid, lra.type()));
+                } catch (ObjectStoreException e) {
+                    if (LRALogger.logger.isTraceEnabled()) {
+                        LRALogger.logger.tracef(
+                                "LRARecoverModule: Object store exception while removing record %s",
```
ochaloup, line 203: as above, please use the `tracef` method `tracef(Throwable t, String format, Object... params)`

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -207,7 +185,103 @@ public void getRecoveringLRAs(Map<URI, Transaction> lras) {
...
+        }
+    }
+
+    private boolean getUids(InputObjectState aa_uids) {
+        synchronized (this) {
+            try {
+                return _recoveryStore.allObjUids(_transactionType, aa_uids);
+            } catch (ObjectStoreException e) {
+                if (LRALogger.logger.isTraceEnabled()) {
+                    LRALogger.logger.tracef(
```
ochaloup, line 237: the same as above, please use the `tracef(Throwable t, String format, Object... params)`

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -207,7 +185,103 @@ public void getRecoveringLRAs(Map<URI, Transaction> lras) {
...
+                Uid uid = new Uid(uids.unpackBytes());
+
+                if (uid.equals(Uid.nullUid())) {
+                    return;
+                }
+
+                consumer.accept(uid);
+            } catch (IOException e) {
+                if (LRALogger.logger.isTraceEnabled()) {
+                    LRALogger.logger.tracef(
```
ochaloup, line 270: the same as above, please use the `tracef(Throwable t, String format, Object... params)`

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/FailedLRAIT.java
mmusgrov: Fixed with the last commit

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/RecoveryCoordinator.java
mmusgrov: Fixed with the last commit

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mmusgrov: Fixed with the last commit

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mmusgrov: Fixed with the last commit

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mmusgrov: Fixed with the last commit

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/RecoveryCoordinator.java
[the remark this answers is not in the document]
mmusgrov: Fixed with the last commit

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/RecoveryCoordinator.java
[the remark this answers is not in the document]
mmusgrov: Fixed with the last commit

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/resource/LRAParticipantWithStatusURI.java
mmusgrov: Fixed with the last commit

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/resource/LRAParticipantWithStatusURI.java
mmusgrov: Fixed with the last commit

rts/lra/lra-test/lra-test-basic/src/test/java/io/narayana/lra/arquillian/resource/LRAParticipantWithStatusURI.java
mmusgrov: Fixed with the last commit

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mmusgrov: 1. The ObjectStore API does not say under what circumstances the exception is thrown since it depends upon which underlying store is in use and the reasons for a failure are varied and we certainly don't want to start adding logging logic for the different store types. Hence the generic message.

2. The log messages are for the benefit of the support organisation to figure out where the issue lies. The actual message text (excluding the values of variables) is less relevant (provided an engineer can identify where in the code the message was generated from).

3. Including the exception message in the trace message means that users who aren't interested in the call stack can still read the log messages and see what the precise reason for the error was (without having to scan down to the stacktrace).

But I did fix the log message text.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mmusgrov: fixed with the latest push

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mmusgrov: fixed with the latest push

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mmusgrov: fixed with the latest push

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
ochaloup: Just for the technical merit I would mention that user don't need to scan down to the stacktrace to see the error message as it's always logged on the first line. For more details he needs to really go down to cause clauses but he has to go through exception stacktrace always for it.
Not a thing which I consider important.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/RecoveryCoordinator.java
```diff
@@ -152,4 +153,48 @@ public String replaceCompensator(
...
+    @APIResponses({
+            @APIResponse(responseCode = "204",
+                    description = "If the LRA was successfully removed"),
+            @APIResponse(responseCode = "404",
+                    description = "The LRA was not recognised"),
+            @APIResponse(responseCode = "500",
+                    description = "If the input LRA does not correspond to a valid URI " +
+                            "(in which case the response entity will contain the error message) " +
+                            "or if the attempt to remove the LRA failed")
+    })
```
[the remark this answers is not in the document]
mmusgrov: So what are you asking for?
Are you saying that we should throw an error if the record is not in the failed state. If so I will reword the API and update the implementation.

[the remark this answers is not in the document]
xstefank: 412 precondition failed would be more appropriate I think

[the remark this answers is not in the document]
mmusgrov: So are you saying that we should return 412 if the record is not in a failed state. If so I will reword the API and update the implementation (since in progress LRAs should not be deleted)?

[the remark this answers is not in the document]
xstefank: @mmusgrov yes, exactly. Thanks.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -508,7 +507,7 @@ private int retryGetEndStatus(URI endPath, boolean compensate) {
                         return TwoPhaseOutcome.HEURISTIC_HAZARD;
                     case FailedToClose:
                     case FailedToCancel:
-                        return TwoPhaseOutcome.FINISH_ERROR;
+                        return reportFailure(compensate, endPath, "unknown");
```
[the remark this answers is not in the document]
xstefank: @mmusgrov this will be used in the log message on line https://github.com/jbosstm/narayana/pull/1603/files#diff-0f226ae6945188e648e485306754f30cR473. So "LRARecord: participant %s reported a failure to %s (cause unknown)" but we know that it is already in failed state, so I would rather say (cause already in failed state) but it's not that important.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -562,14 +564,13 @@ private int retryGetEndStatus(URI endPath, boolean compensate) {
...
+                                    return TwoPhaseOutcome.HEURISTIC_HAZARD;
+                                }
                             }
 
-                            if (compensate) {
-                                return TwoPhaseOutcome.FINISH_OK;
-                            }
-
-                            return TwoPhaseOutcome.FINISH_OK;
+                            return reportFailure(compensate, endPath, "Unknown");
```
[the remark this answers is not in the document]
xstefank: sorry, that was my mistake, you're right about the call handling. But we should include here that it was a participant response that caused the failure state. But again not that essential, just an idea.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/RecoveryCoordinator.java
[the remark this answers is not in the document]
mmusgrov: Thanks, now I know how to code it up.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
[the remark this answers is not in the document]
mmusgrov: Let's revisit these reasons when we have better reporting of the reasons why participants fail to finish cleanly.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/RecoveryCoordinator.java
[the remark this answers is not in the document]
mmusgrov: @xstefank I added OpenAPI APIResponse annotations for 204, 412, 412 and 500 (and I had to make some updates to LRARecoveryModule so this code can distinguish between them).

PR #1607, opened 2020-04-22
CI SCRIPTS: stabilize AS_TESTS axis, see more at WFLY-13385

ochaloup:
Solution for our CI environment on the unstable AS_TESTS axis.
See more at https://issues.redhat.com/browse/WFLY-13385

AS_TESTS
!MAIN !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS !TOMCAT !JACOCO !LRA


ochaloup:
@mmusgrov I had to declined my fix in the AS testsuite for WFLY-13385 as it was told me it should be configured in our CI. This change is for providing the default CLI timeout longer. For now it seems it works and the jobs passed.

ochaloup:
Interestingly enough there is a different error in the test run than we were used to up to now. Kind of, I think this PR fix is desirable as the failure is connected to JMX and some other communication error.
I'm not sure if it's intermittent, a real error of the current WFLY master head or another env issue we could experience regularly.

PR #1608, opened 2020-04-23
[JBTM-3302] adding information about checkstyle rules to PR template

ochaloup:
https://issues.redhat.com/browse/JBTM-3302

MAIN CORE
!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS !AS_TESTS !TOMCAT !JACOCO !LRA !DB_TESTS mysql db2 postgres oracle

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -374,7 +374,7 @@
               <excludes>**/*$logger.java,**/*$bundle.java</excludes>
               <useFile/>
               <!-- checkstyle is set to be run on particular modules by disabling skip -->
-              <skip>true</skip>
+              <skip>${checkstyle.skip}</skip>
           </configuration>
           <dependencies>
               <dependency>
@@ -406,8 +406,11 @@
   </dependencies>
 
   <properties>
-    <!-- IP stack configs. -->
+    <!-- Checkstyle is disabled by default, but it could be enabled for default by a particular module -->
+    <checkstyle.skip>true</checkstyle.skip>
+    <!-- The surefire/failsafe test execution logs goes to a file by default -->
     <testLogToFile>true</testLogToFile>
+    <!-- IP stack configs. -->
     <jvm.args.ip>-Djava.net.preferIPv4Stack=true -Djava.net.preferIPv6Addresses=false</jvm.args.ip>
     <maven.build.timestamp.format>yyyy/MMM/dd HH:mm</maven.build.timestamp.format>
     <buildproperty.date>${maven.build.timestamp}</buildproperty.date>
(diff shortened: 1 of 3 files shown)


ochaloup:
@tomjenkinson @mmusgrov regarding the PR template - would not be more useful if we define the axis for CI not in "positive" form as it's now (btw. there is `NO_WIN` which is negative) but in "negative" form where the committer has to enable the axis relevant for the PR?

I mean to have there
```
!MAIN !CORE !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS !AS_TESTS !TOMCAT !JACOCO !LRA !DB_TESTS mysql db2 postgres oracle
```

ochaloup:
@tomjenkinson I meant this rather as a kick off question. I was asking you about your opinion first as I wanted to understand the reasoning. As you tend to have default being to test everything I'm not going to open any discussion.

ochaloup:
@mmusgrov I missed that point in the jira description, sorry.

I changed the PR. But I think using `checkstyle.skip` parameter.
The checkstyle mvn plugin has only the config element `<skip>` so we need to use negation and maven does not have the boolean negation support. That way we should use  `checkstyle.skip` instead of the `checkstyle`.

pom.xml
```diff
@@ -374,7 +374,7 @@
               <excludes>**/*$logger.java,**/*$bundle.java</excludes>
               <useFile/>
               <!-- checkstyle is set to be run on particular modules by disabling skip -->
-              <skip>true</skip>
+              <skip>${checkstyle.skip}</skip>
```
mmusgrov, line 377: If the user does not provide a value for checkstyle.skip this defaults to true (at least it did when I tested it).

ochaloup:
@mmusgrov that's strange. The `CORE` axis passed (I'm letting it executed once again to check) and if the checkstyle would be enabled e.g. the ArjunaCore will start failing.
It works fine on my computer.
Where and how do you experience this behaviour?

pom.xml
ochaloup: Please, give me some points how I can reproduce the test failure you can see.

mmusgrov: Hmm, I cannot reproduce it now, not sure what I did before. Sorry about that.
But good call about using the plugin property name (namely checkstyle.skip)

ochaloup: Thank you. I'm going to merge this.

PR #1609, opened 2020-04-23
JBTM-3293 Add lra-coordinator-war module

xstefank:
https://issues.redhat.com/browse/JBTM-3293

### DIFF
--- a/narayana-full/src/main/assembly/bin.xml
+++ b/narayana-full/src/main/assembly/bin.xml
@@ -213,6 +213,11 @@
 			<outputDirectory>rts/lra/</outputDirectory>
 		</file>
 		<file>
+			<source>../rts/lra/lra-coordinator-war/target/lra-coordinator.war</source>
+			<fileMode>0644</fileMode>
+			<outputDirectory>rts/lra/</outputDirectory>
+		</file>
+		<file>
 			<source>../rts/lra/lra-coordinator-thorntail/target/lra-coordinator-thorntail.jar</source>
 			<fileMode>0644</fileMode>
 			<outputDirectory>rts/lra/</outputDirectory>
--- a/rts/lra/lra-coordinator-war/pom.xml
+++ b/rts/lra/lra-coordinator-war/pom.xml
@@ -1,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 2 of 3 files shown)


ochaloup:
@xstefank So the approach goes intentionally in direction of having the war to be built but not released it to nexus, right?
As I now tend to think that it should be published somehow.

ochaloup:
@xstefank it's possible (https://search.maven.org/remotecontent?filepath=org/jboss/narayana/rts/narayana-lra/5.10.4.Final/narayana-lra-5.10.4.Final.jar) but I agree it's not a standard way.
Normally the application is downloaded from somewhere else. We could just, for example, publish it via github and link it at the Narayana web page.
In fact, the LRA should be probably integrated to WildFly as a galleon feature pack and then the war will not make more sense, right?

Anyway, by default the mvn modules are not published and the `pom.xml` needs to be updated with the `release` profile: https://github.com/jbosstm/narayana/blob/5.10.4.Final/rts/lra/lra-coordinator-jar/pom.xml#L241

xstefank:
@ochaloup @mmusgrov so what would be preferable from narayana standpoint, please?

ochaloup:
I don't have a clear preference. By me:

1. we need a fix for JBTM-3293 for we can deploy coordinator to WildFly
2. the galleon feature pack will come in the remote future (do you know something more about the plans,schedule... @mmusgrov?)
3. fix is to provide the `war` archive
4. downloading `war` from the maven central is not usual but having that artifact there is not problematic
5. in summary: I think current solution is what you have in this PR + adding the release profile to `pom.xml`

ochaloup:
@xstefank we discussed this with @mmusgrov on team meeting and we agreed (I hope I'm not incorrect here) that having a separate war module which will be published to maven central is the current solution to take.
For me approving this PR it would be needed to provide the `release` profile to `pom.xml`: https://github.com/jbosstm/narayana/blob/5.10.4.Final/rts/lra/lra-coordinator-jar/pom.xml#L241

mmusgrov:
> 2. the galleon feature pack will come in the remote future (do you know something more about the plans,schedule... @mmusgrov?)

I did make a start on it but it still needed quite a bit of work.
My view is that a galleon feature pack is high priority and I am struggling to find time to make any progress on it. Ideally, the work should be passed onto someone else.

> 5. in summary: I think current solution is what you have in this PR + adding the release profile to `pom.xml`

Making it available from central makes sense.
But we also need it in our release zip (https://github.com/jbosstm/narayana/blob/master/narayana-full/src/main/assembly/bin.xml)

mmusgrov:
The PR test failure is unrelated so I am merging this PR. @xstefank thank you for the enhancement.

PR #1611, opened 2020-04-27
CI SCRIPTS: a follow-up fix for stabilizing AS_TESTS axis (WFLY-13385)

ochaloup:
the change of the connection-timeout has to be propagated into jboss-cli.xml before the AS_TESTS starts

AS_TESTS
!MAIN !CORE !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS !TOMCAT !JACOCO !LRA

PR #1615, opened 2020-04-29
CI SCRIPTS: a follow-up fix for stabilizing AS_TESTS axis (WFLY-1338)

ochaloup:
the change of the connection-timeout has to be propagated into the jboss-cli.xml before the tests start

Another attempt around the AS_TESTS axis.

AS_TESTS


ochaloup:
Afther follow-up discussion with Jean Francois he persuaded me that the right approach is one I tried at the https://github.com/jbosstm/narayana/pull/1611 but as I took a wrong configuration options.

It could be my fix does not work well for the stabilization either. Still investigating.

PR #1623, opened 2020-05-18
[JBTM-3316] use LRA TCK timeout factor instead of delays

ochaloup:
https://issues.redhat.com/browse/JBTM-3316

This fix could help in CI instability for `TckTests#timeLimit`

LRA


ochaloup:
The good thing is that it seems that the LRA CI test failure is now limited to one failure instead of two.

```
testCancelWhenParticipantIsUnavailable(org.eclipse.microprofile.lra.tck.TckRecoveryTests)  Time elapsed: 28.349 s  <<< FAILURE!
java.lang.AssertionError: Compensate on restarted service should have been called. Was -1 expected:<200> but was:<500>
	at org.eclipse.microprofile.lra.tck.TckRecoveryTests.testCancelWhenParticipantIsUnavailable(TckRecoveryTests.java:215)
```

I think the fix for this one could be to implement the `LRARecoveryService#waitForCallbacks`. But I will need to investigate further if it's correct suspicion.

PR #1626, opened 2020-05-21
JBTM-3320 LRA forget calls for nested completed participants are not …

xstefank:
…repeated and do not close the LRA

https://issues.redhat.com/browse/JBTM-3320

### DIFF
--- a/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
+++ b/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
@@ -480,7 +480,12 @@ public class Transaction extends AtomicAction {
                     status = toLRAStatus(status());
                 } else {
                     // forget calls for nested participants
-                    forgetAllParticipants();
+                    if (forgetAllParticipants()) {
+                        updateState(LRAStatus.Closed);
+                    } else {
+                        // some forget calls have not been received, we need to repeat them at the next recovery pass
+                        return ActionStatus.H_HAZARD;
+                    }
                 }
             }
         } else {
@@ -690,10 +695,14 @@ public class Transaction extends AtomicAction {
         return findLRAParticipant(participantUrl, true) != null;
     }
 
-    public void forgetAllParticipants() {
-        if (pending != null) {
-            pending.forEach(LRARecord::forget);
+    public boolean forgetAllParticipants() {
+        if (pending == null) {
+            return true;
         }
+
+        pending.removeIf(LRARecord::forget);
+
+        return pending.isEmpty();
     }
 
     private void savePendingList() {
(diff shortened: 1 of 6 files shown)


rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -690,10 +697,21 @@ public boolean forgetParticipant(String participantUrl) {
         return findLRAParticipant(participantUrl, true) != null;
     }
 
-    public void forgetAllParticipants() {
-        if (pending != null) {
-            pending.forEach(LRARecord::forget);
+    public boolean forgetAllParticipants() {
+        boolean notifiedAll = true;
+
+        Iterator<LRARecord> iterator = pending.iterator();
```
ochaloup, line 703: just two ideas
* could be `pending` `null` or it could not be?
* I would, personally and if I don't miss some detail, consider more readable the lamda style like `pending.removeIf(record -> record.forget())`

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -480,7 +480,13 @@ private int doEnd(boolean cancel) {
                     status = toLRAStatus(status());
                 } else {
                     // forget calls for nested participants
-                    forgetAllParticipants();
+                    if (forgetAllParticipants()) {
+                        updateState(LRAStatus.Closed);
+                        return ActionStatus.COMMITTED;
```
mmusgrov, line 485: Do we need to call `runPostLRAActions();`. Have you tested your changes with an AfterLRAListener registered with the nested LRA?

xstefank: AfterLRA notifications should depend on the outcome of forget calls, as nested LRA is not finished before forget calls are delivered. Or did I misunderstood the scenario? I can add such test if needed.

mmusgrov: After your code changes the transaction end logic finishes early and you are skipping a section of code that we currently execute. In particular the code on line 
https://github.com/jbosstm/narayana/blob/0351d68661591df1fd53bd81d741fc7ded362bfb/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java#L520 that was designed to perform the AfterLRA notifications is skipped.

xstefank: sorry @mmusgrov I've just got back to this. So if I remove this return statement then it should be ok right?

xstefank: @mmusgrov actually, it doesn't seem to matter to return here. I've added the test as you've requested.

mmusgrov: I am still uncomfortable why this route through the end() routine is returning early. No other code path returns early so it makes maintaining this method more difficult. We either need to avoid the early return or fully comment why it is safe to do so.

Thanks for the extra test - I didn't really understand what is was doing so some commentary would have been helpful to the both the reviewer and to future maintainers.

PR #1629, opened 2020-06-01
JBTM-3323 LRA server filter is registered twice

xstefank:
https://issues.redhat.com/browse/JBTM-3323

I've decided to remove dynamic registration because I expect that if the user adds the LRA maven dependency then they intend to use LRA and so the filter should be registered.


xstefank:
@mmusgrov I think the rebase is only CI issue I discussed with @ochaloup yesterday. I will force push to retrigger it.

PR #1635, opened 2020-06-08
[SCRIPTS ONLY] increasing AS_TESTS timeout factor for WFLY testsuite being more stable

ochaloup:
AS_TESTS


ochaloup:
After some more testing I'm now not convinced to really increase the timeout factor. The change let the testsuite to pass but it seems it was not the cause. Increasing the timeout factor may stabilize the testsuite a bit but it prolongs the execution time. As it seems the testsuite with changes in 5_BRANCH can run without any error (I took this feeling from the last run of the narayana CI job and from my other test jobs) I'm now closing this PR. If I find it's necessary I would reopen it later.

PR #1637, opened 2020-06-17
[JBTM-3318] tckrecoverytests timeout fixes

ochaloup:
https://issues.redhat.com/browse/JBTM-3318

LRA

### DIFF
--- a/rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/LRAAnnotationAdjuster.java
+++ b/rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/LRAAnnotationAdjuster.java
@@ -1,0 +1,209 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2020, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package io.narayana.lra.arquillian.spi;
+
+import org.eclipse.microprofile.lra.annotation.ws.rs.LRA;
+import org.jboss.logging.Logger;
+
+import javax.ws.rs.core.Response;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.time.temporal.ChronoUnit;
+import java.util.Arrays;
+import java.util.Map;
+
+/**
+ * Helper class to do internal changes of annotations during runtime.
+ */
+public class LRAAnnotationAdjuster {
+    private static final Logger log = Logger.getLogger(LRAAnnotationAdjuster.class);
+    private static final String ANNOTATIONS_FIELD_NAME = "annotations";
+    private static final String ANNOTATION_DATA_METHOD_NAME = "annotationData";
+
+    /**
+     * Take the clazz, check if contains the {@link LRA} annotation.
+     * The LRA annotation is then replaced by wrapped {@link LRAWrapped}.
+     */
+    static void processWithClass(Class<?> clazz) {
+        LRA lraAnnotation = clazz.getDeclaredAnnotation(LRA.class);
+        if (lraAnnotation != null) {
+            LRAAnnotationAdjuster.adjustLRAAnnotation(clazz, lraAnnotation);
+        }
+        Arrays.stream(clazz.getMethods()).forEach(method -> {
+            LRA lraAnnotationMethod = method.getDeclaredAnnotation(LRA.class);
+            if (lraAnnotationMethod != null) {
+                LRAAnnotationAdjuster.adjustLRAAnnotation(method, lraAnnotationMethod);
+            }
+        });
+    }
+
+    /**
+     * Changing the LRA annotation declared on class by wrapping it with {@link LRAWrapped}.
+     */
+    static void adjustLRAAnnotation(Class<?> clazzToLookFor, LRA originalLRAAnnotation) {
+        if (doesJDKDefineAnnotationsField()) {
+            // JDK7 has "annotations" field
+            try {
+                Field annotations = Class.class.getDeclaredField(ANNOTATIONS_FIELD_NAME);
+                annotations.setAccessible(true);
+                @SuppressWarnings("unchecked")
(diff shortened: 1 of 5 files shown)


ochaloup:
LRA test failed on
```
[ERROR] timeLimitWithPreConditionFailed(org.eclipse.microprofile.lra.tck.TckTests)  Time elapsed: 1.753 s  <<< FAILURE!
java.lang.AssertionError: Expected 412 or 410 response
	at org.eclipse.microprofile.lra.tck.TckTests.timeLimitWithPreConditionFailed(TckTests.java:369)
```
the reason is that the LRA test fix -  https://github.com/eclipse/microprofile-lra/pull/313 - is not incorporated in `1.0-M1` tag
this PR adjusts the `timeLimit` by `tck.time.factor` and if the `sleep` at https://github.com/eclipse/microprofile-lra/blob/1.0-M1/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LraResource.java#L431 is not updated then another race condition appears. I changed the factor to `2.0` which means that `timeLimit` vs. sleep time in the bean hopefully works, even without the fix in the TCK.

ochaloup:
@mmusgrov I hope the timeout is now configured properly and I would like to ask for the review.

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/LRAAnnotationAdjuster.java
```diff
@@ -0,0 +1,206 @@
...
+                        (Map<Class<? extends Annotation>, Annotation>) annotations.get(clazzToLookFor);
+                map.put(LRA.class, new LRAWrapped(originalLRAAnnotation));
+            } catch (Exception  e) {
+                e.printStackTrace();
+            }
+        } else {
+            try {
+                // JDK8+ has "annotationData" private method
+                // obtaining reference to private class AnnotationData
+                Method method = Class.class.getDeclaredMethod(ANNOTATION_DATA_METHOD_NAME, null);
```
mmusgrov, line 80: My IDE says this is a varargs param so you don't need to use null.

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/LRAAnnotationAdjuster.java
```diff
@@ -0,0 +1,206 @@
...
+        boolean jdkLower = true;
+        try {
+            Class.class.getDeclaredField(ANNOTATIONS_FIELD_NAME);
+        } catch (NoSuchFieldException ignore) {
+            jdkLower = false;
+        }
+        return jdkLower;
+    }
+
+    private static class LRAWrapped implements LRA {
```
mmusgrov, line 136: Fiendishly clever.

As an aside, we'd need to remember to change this class if the LRA annotation changes in MP-LRA 1.x.

mmusgrov:
@ochaloup I suspect your fix will also resolve [JBTM-3307](https://issues.redhat.com/browse/JBTM-3318) (LRACoordinatorRecovery2TestCase regularly fails on CI). Do you think it is worth adding a commit to this PR that unigores the test.

ochaloup:
@mmusgrov I'm going to assign the JBTM-3307 to me and I'm going to check if the fix solves it as well.

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/LRAAnnotationAdjuster.java
ochaloup: +1

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/LRAAnnotationAdjuster.java
ochaloup: sure. I think that should be fine as this is part of the test which is usually compiled and run. The compilation would fail if the `LRAWrapped` does not implement some LRA method (or implement a different definition).

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/LRAAnnotationAdjuster.java
```diff
@@ -0,0 +1,206 @@
...
+        if (doesJDKDefineAnnotationsField()) {
+            // JDK7 has "annotations" field
+            try {
+                Field annotations = Class.class.getDeclaredField(ANNOTATIONS_FIELD_NAME);
+                annotations.setAccessible(true);
+                Map<Class<? extends Annotation>, Annotation> map =
+                        (Map<Class<? extends Annotation>, Annotation>) annotations.get(clazzToLookFor);
+                map.put(LRA.class, new LRAWrapped(originalLRAAnnotation));
+            } catch (Exception  e) {
+                e.printStackTrace();
```
xstefank, line 74: I think you prefer logging instead of stack traces.

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
```diff
@@ -32,6 +33,25 @@
...
+    private static final Logger log = Logger.getLogger(NarayanaLRARecovery.class);
+
+    /**
+     * A bit of hacking to change the internals of annotations defined in LRA TCK.
+     * There is need to adjust timeout defined on the annotation definition.
+     */
+    static {
+        String[] resourceClassNames = new String[]{
+                "org.eclipse.microprofile.lra.tck.participant.api.LraResource",
+                "org.eclipse.microprofile.lra.tck.participant.api.RecoveryResource"};
```
xstefank, line 45: wouldn't be better to use getName() here?

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/LRAAnnotationAdjuster.java
ochaloup: +1

rts/lra/lra-test/lra-test-arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
ochaloup: not using the `getName()` was intentional. If the extension is used for deployments where these resources are not provided then using of the `Class.getName()` makes the whole `NarayanaLRARecovery`failing as it tries to link a class which is not on classpath.
It's the case e.g. for `lra-test-basic` where those classes are not available.

xstefank: wouldn't it be better then to add LRA TCK dependency to the basic module then to keep track of possible renames in the dependent project? :)

ochaloup: I was investigating here a bit about what's needed to do and the TCK dependency would need to be added to any module which is to use the arquillian extension. I don't want to add such requirements on users.
This is an enhancement for the TCK behaviour (and I agree it's a bit hacky but without definition of long pause time into TCK our AMS CI is currently in troubles to get running properly the LRA tests) and not any other tests.
When there is a failure on loading class (ie. the TCK tests is renamed) then if the test starts to fail it will be needed to investigate the reasons.
By me this loosely coupling of this Arq extenstion to TCK tests is desired here.

PR #1641, opened 2020-06-26
[JBTM-3338] providing thread-safety for CMR configuration

ochaloup:
https://issues.redhat.com/browse/JBTM-3338

MAIN PERF
!AS_TESTS !QA_JTA !TOMCAT
!QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS NO_WIN !RTS !JACOCO !LRA

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
@@ -21,6 +21,7 @@
 package com.arjuna.ats.jta.common;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -94,33 +95,37 @@ public class JTAEnvironmentBean implements JTAEnvironmentBeanMBean
     private volatile String xaResourceRecordWrappingPluginClassName;
     private volatile XAResourceRecordWrappingPlugin xaResourceRecordWrappingPlugin;
 
-	private int asyncCommitPoolSize = 10;
+	private volatile int asyncCommitPoolSize = 10;
 
-    private int orphanSafetyInterval = 20000;
+    private volatile int orphanSafetyInterval = 20000;
     
-    private String commitMarkableResourceTableName = "xids";
+    private volatile String commitMarkableResourceTableName = "xids";
 
-	private Map<String, String> commitMarkableResourceTableNameMap = new HashMap<String, String>();
+	private volatile Map<String, String> commitMarkableResourceTableNameMap
+            = Collections.unmodifiableMap(new HashMap<>());
 
-	private List<String> commitMarkableResourceJNDINames = new ArrayList<String>();
+	private volatile List<String> commitMarkableResourceJNDINames
+            = Collections.unmodifiableList(new ArrayList<>());
 	
-	private boolean performImmediateCleanupOfCommitMarkableResourceBranches = false;
+	private volatile boolean performImmediateCleanupOfCommitMarkableResourceBranches = false;
 	
-	private boolean notifyCommitMarkableResourceRecoveryModuleOfCompleteBranches = true;
+	private volatile boolean notifyCommitMarkableResourceRecoveryModuleOfCompleteBranches = true;
 
-	private int commitMarkableResourceRecordDeleteBatchSize = 30000;
+	private volatile int commitMarkableResourceRecordDeleteBatchSize = 30000;
 
-	private Map<String, Boolean> performImmediateCleanupOfCommitMarkableResourceBranchesMap = new HashMap<String, Boolean>();
+	private volatile Map<String, Boolean> performImmediateCleanupOfCommitMarkableResourceBranchesMap
+            = Collections.unmodifiableMap(new HashMap<>());
 
-	private Map<String, Integer> commitMarkableResourceRecordDeleteBatchSizeMap = new HashMap<String, Integer>();
+	private volatile Map<String, Integer> commitMarkableResourceRecordDeleteBatchSizeMap
+            = Collections.unmodifiableMap(new HashMap<>());
 
 	private static final String defaultTransactionOperationsProviderClassName = ServerVMClientUserTransactionOperationsProvider.class.getName(); 
 	private volatile String userTransactionOperationsProviderClassName = defaultTransactionOperationsProviderClassName;
 	private volatile UserTransactionOperationsProvider userTransactionOperationsProvider = null;
 
-	private boolean strictJTA12DuplicateXAENDPROTOErr = false;
+	private volatile boolean strictJTA12DuplicateXAENDPROTOErr = false;
 
-	private boolean transactionToThreadListenersEnabled = false;
+	private volatile boolean transactionToThreadListenersEnabled = false;
 
 	/**
      * Returns true if subtransactions are allowed.
@@ -1052,9 +1057,7 @@ public class JTAEnvironmentBean implements JTAEnvironmentBeanMBean
 	 * @return the table name map
 	 */
 	public Map<String, String> getCommitMarkableResourceTableNameMap() {
-		synchronized (this) {
 			return commitMarkableResourceTableNameMap;
-		}
 	}
 
 	/**
@@ -1063,19 +1066,16 @@ public class JTAEnvironmentBean implements JTAEnvironmentBeanMBean
 	 * 
 	 * @param commitMarkableResourceTableNameMap
 	 *            The name of the table.
-	 * 
 	 */
 	public void setCommitMarkableResourceTableNameMap(
 			Map<String, String> commitMarkableResourceTableNameMap) {
-		synchronized (this) {
-			if (commitMarkableResourceTableNameMap == null) {
-				this.commitMarkableResourceTableNameMap = new HashMap<String, String>();
-			} else if (!commitMarkableResourceTableNameMap
-					.equals(this.commitMarkableResourceTableNameMap)) {
-				this.commitMarkableResourceTableNameMap = new HashMap<String, String>(
-						commitMarkableResourceTableNameMap);
-			}
-		}
+	    if (commitMarkableResourceTableNameMap == null) {
+            this.commitMarkableResourceTableNameMap =
+                    Collections.unmodifiableMap(new HashMap<>());
+        } else {
+            this.commitMarkableResourceTableNameMap =
+                    Collections.unmodifiableMap(new HashMap<>(commitMarkableResourceTableNameMap));
+        }
 	}
 
 	/**
@@ -1085,9 +1085,7 @@ public class JTAEnvironmentBean implements JTAEnvironmentBeanMBean
 	 * @return The list of JNDI names
 	 */
 	public List<String> getCommitMarkableResourceJNDINames() {
-		synchronized (this) {
-			return new ArrayList<String>(commitMarkableResourceJNDINames);
-		}
+        return commitMarkableResourceJNDINames;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


ochaloup:
The QA_JTA failed because of the connection error on Oracle database (http://narayanaci1.eng.hst.ams2.redhat.com/job/btny-pulls-narayana/PROFILE=QA_JTA,jdk=jdk8.latest,label=linux/177/artifact/qa/testoutput/jdbcresources02_oracle_thin_jndi/JDBCResources02_abstract_Test001/cleanup_output.txt)
```
err: java.sql.SQLException: java.sql.SQLException: Listener refused the connection with the following error:
err: ORA-12518, TNS:listener could not hand off client connection
err: The Connection descriptor used by the client was:
err: (DESCRIPTION=(ADDRESS=(PROTOCOL=tcp)(PORT=1521)(HOST=narayanaci1.***))(CONNECT_DATA=(SID=XE)))
err: 
err: 	at com.arjuna.ats.internal.jdbc.BaseTransactionalDriverXAConnection.createConnection(BaseTransactionalDriverXAConnection.java:36)
err: 	at com.arjuna.ats.internal.jdbc.IndirectRecoverableConnection.createConnection(IndirectRecoverableConnection.java:191)
err: 	at com.arjuna.ats.internal.jdbc.BaseTransactionalDriverXAConnection.getConnection(BaseTransactionalDriverXAConnection.java:139)
err: 	at com.arjuna.ats.internal.jdbc.ConnectionImple.getConnection(ConnectionImple.java:814)
err: 	at com.arjuna.ats.internal.jdbc.ConnectionImple.init(ConnectionImple.java:143)
err: 	at com.arjuna.ats.internal.jdbc.ConnectionImple.<init>(ConnectionImple.java:111)
err: 	at com.arjuna.ats.internal.jdbc.ConnectionManager.create(ConnectionManager.java:147)
err: 	at com.arjuna.ats.jdbc.TransactionalDriver.connect(TransactionalDriver.java:89)
err: 	at java.sql.DriverManager.getConnection(DriverManager.java:664)
err: 	at java.sql.DriverManager.getConnection(DriverManager.java:247)
err: 	at org.jboss.jbossts.qa.JDBCResources02Cleanups.Cleanup01.main(Cleanup01.java:85)
err: Caused by: java.sql.SQLException: Listener refused the connection with the following error:
err: ORA-12518, TNS:listener could not hand off client connection
err: The Connection descriptor used by the client was:
err: (DESCRIPTION=(ADDRESS=(PROTOCOL=tcp)(PORT=1521)(HOST=narayanaci1.***))(CONNECT_DATA=(SID=XE)))
err: 
err: 	at oracle.jdbc.driver.DatabaseError.throwSqlException(DatabaseError.java:112)
err: 	at oracle.jdbc.driver.DatabaseError.throwSqlException(DatabaseError.java:261)
err: 	at oracle.jdbc.driver.T4CConnection.logon(T4CConnection.java:387)
err: 	at oracle.jdbc.driver.PhysicalConnection.<init>(PhysicalConnection.java:441)
err: 	at oracle.jdbc.driver.T4CConnection.<init>(T4CConnection.java:165)
err: 	at oracle.jdbc.driver.T4CDriverExtension.getConnection(T4CDriverExtension.java:35)
err: 	at oracle.jdbc.driver.OracleDriver.connect(OracleDriver.java:801)
err: 	at oracle.jdbc.pool.OracleDataSource.getPhysicalConnection(OracleDataSource.java:297)
err: 	at oracle.jdbc.xa.client.OracleXADataSource.getPooledConnection(OracleXADataSource.java:472)
err: 	at oracle.jdbc.xa.client.OracleXADataSource.getXAConnection(OracleXADataSource.java:159)
err: 	at oracle.jdbc.xa.client.OracleXADataSource.getXAConnection(OracleXADataSource.java:133)
err: 	at com.arjuna.ats.internal.jdbc.BaseTransactionalDriverXAConnection.createConnection(BaseTransactionalDriverXAConnection.java:31)
err: 	... 10 more
```

ochaloup:
hi @tomjenkinson and @mmusgrov would you be so kind and review this PR?

ochaloup:
@tomjenkinson if it does then it's dangerous. If anybody changes the content in the different thread it will lead to inconsistency. In other words this is the scenario which this fix is meant to adjust.

ochaloup:
Oh, I see. Thanks for this point about [the WFLY integration](https://github.com/wildfly/wildfly/blob/master/transactions/src/main/java/org/jboss/as/txn/service/CMResourceService.java#L83-L86). I've just missed that code part.

In which context we need to consider the [WFLY-4500](https://issues.redhat.com/browse/WFLY-4500)? I understand the issue as a fix for a "typo" where a wrong variable was used.
You mean that the `Map` should not be used and there should be used rather a `Set` or `List` and when the JNDI is part of the  list then it will be considered as being set as _`true`_ (aka. in current `Map` context the JNDI may obtain a value of `true` or `false`).

And in fact I'm checking the code the other reason why there was not used the copy-constructor could be the way how the list/map is stored in the CMRRecord (https://github.com/jbosstm/narayana/blob/5.10.5.Final/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/CommitMarkableResourceRecord.java#L113). There is used the `static` variable so the WFLY service stop is directly propagated to the CMRRecord. If there is a copy-constructor then the `static` property would not be reinitialized.

My concerns are not mainly about `volatility`. My concern is danger of using the `HashMap` in multi threaded environment. Let's say the WFLY is reloaded. Then as far as I'm aware the `<service>.stop()` method is invoked. If there is a CMR resource in proccesing meanwhile the map/list/non-volatile value may cause a NPE to be thrown or worse it could provide data in inconsistent state.

ochaloup:
@tomjenkinson ok, just I don't fully catch what should be reported. I mean - if this PR is committed then it makes troubles in the behaviour of the CMR setup. As well the Narayana internal implementation expects mutable CMR setup. From that I assume this PR can't be merged as it is.
If I understand current CMR processing in WildFly it requires `:reload` to get configuration applied. I think the processing of the changes would not be hard to do if the CMRRecord does not use the `static` properties but gets in touch directly with the configuration bean. I don't think there is required some next level of abstraction.


side note: on investigation of CMR handling in WFLY I hit another minor issue: https://issues.redhat.com/browse/WFLY-13626

ochaloup:
@tomjenkinson ok, I see. I understand what you mean by the issue creation. I has been in a parallel universe where I was thinking that this PR needs to address the configuration thread-safety issue and then the potential issues will be solved.

@mmusgrov it's how Tom explained, plus the concurrency in mollification of the shared map (time the {{:reload}} is called) and another thread reading the map (processing on CMR resource is processed in parallel with the app container reload process) may lead to race condition quite easy as well.

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
```diff
@@ -94,33 +95,37 @@
...
-    private String commitMarkableResourceTableName = "xids";
+    private volatile String commitMarkableResourceTableName = "xids";
 
-	private Map<String, String> commitMarkableResourceTableNameMap = new HashMap<String, String>();
+	private volatile Map<String, String> commitMarkableResourceTableNameMap
+            = Collections.unmodifiableMap(new HashMap<>());
 
-	private List<String> commitMarkableResourceJNDINames = new ArrayList<String>();
+	private volatile List<String> commitMarkableResourceJNDINames
+            = Collections.unmodifiableList(new ArrayList<>());
```
mmusgrov, line 108: 1. This looks odd since it creates an empty list which can never be modified.
2. It does not match with how we implement other collections in the bean (xaRecoveryNodes for example) which makes maintenance more difficult.
3. Now every API user that mutates the returned collection has to be re-coded to catch UnsupportedOperationException. [Remark: if we do want immutable collections then we would need to wait until we move to java 9 which does have them].

The technique does not address issue https://issues.redhat.com/browse/WFLY-13633 which I thought was one of the reasons for revising how we configure CMR collections.

The technique has the advantage of avoiding the need to give callers copies of the collection when it hasn't changed. But we would need to quantify whether or not that benefit outweighs the disadvantages.

ochaloup:
@tomjenkinson the performance regression is really strange. I verified the code and ran the perf tests on my machine[1] and I can't see any such degradation (~18%). More strange is that the code changes do not adjust any standard JTA processing but only CMR. I'm going to retest it here.

@mmusgrov Thanks for feedback. I've been looking around the code and I still think my changes make sense. The idea of the changeset is not using copy-constructor for `getters` as it slows down the processing. Instead of it returns a "ready-to-go" collection which is unmodifiable. Having it unmodifiable denies the data changes on the underlying structure.

Regarding your points

ad 1) `This looks odd since it creates an empty list which can never be modified.`
It's the same as constructs above. The getter will have for return an empty collection (when it was not modified before).

ad 2) `It does not match with how we implement other collections`
Yes, I admit it's not implemented as the other collection in the bean. Let's discuss on how to do it in similar fashion while not discarding the performance. I haven't got a better idea yet. Would you have some idea here?
I didn't want to use the copy-constructor as I think it's of creation of a new object on every getter call. It would be nice e..g use something like a copy-on-write map but that does not exist in Java.
If we may consider the performance will be good enough with the copy-constructor at getter then standard list/map can be used.
Would you have some ideas here?

ad 3.1) `Now every API user that mutates the returned collection has to be re-coded to catch `
The backward compatibility point is more than valid - if the unmodifiable collection is used. I think we can either say it's Narayana internals and such change is fine. Or we can add an additional `getter` (with similar but not the same name) and deprecate the current one. The API will contain two methods where the new one can be used in the CMR code while the old one will be still available for the external clients.

ad 3.2) `if we do want immutable collections then we would need to wait until java 9`
That's not a right assumption. Java 9 does not provide any specific immutable collection handling other than having a factory methods which makes easier to construct them. But the point that using a mutable operation causes the `UnsupportedOperationException` being thrown remains.

[1]
```
# 5.10.5.Final
Benchmark               Mode  Cnt       Score      Error  Units
JTAStoreTests.jtaTest  thrpt   25  185425.762 ± 4350.681  ops/s

# 5.10.6.Final-SNAPSHOT with this PR
JTAStoreTests.jtaTest  thrpt   25  182886.243 ± 4056.631  ops/s
```

mmusgrov:
I would prefer not to make ad hoc changes to how we handle these particular bean properties since if we are going to address the performance problems of using copy constructors then we need to do that as part of an overhaul/investigation of our property system. We were already going to do that as part of the Quarkus work so we should just fold this performance requirement into that work.

I also raised the point in my review that the technique used in this PR does not address issue https://issues.redhat.com/browse/WFLY-13633 which I thought was one of the reasons for revising how we configure CMR collections.

Since the PR does not address WFLY-13633 (which is the ultimate reason for this PR) I have not studied all of the comments made above (since the comments do not necessarily impact on WFLY-13633).

ochaloup:
@mmusgrov I'm sorry but I don't consider this as an adhoc changes as I proceeded investigation around this topic. This PR is not perfect and has parts for improvements but it was not meant as an adhoc change in any manner.

The part with with the `unmodifiableList` was meant for futher discussion as I mentioned already. I can imagine e.g. copyonwrite would be better in this case. But unfortunately Java does not provide the implementation of Map. So I used the unmodifiable version that it could be discussed later.
While I agree I could consider the backward compatibility more strictly in first version of my PR it's not hard to be fixed.

I personally think that's this and Quarkus issue are the great chance to investigate, to think and eventually to rework the property system.
My personal opinion is that there are good reasons to think about some redesign (better and easier Quarkus integration, performance, runtime changes without need to restart JVM to improve UX for any runtime Narayana integrates to).

The idea behind this and why this PR solves the issue  WFLY-13633 is that any CMR change in WFLY requires `:reload` operation to be called. That will cause the service will be started again and the setter is called. So the new definitions of the JNDI will be propagated to Narayana which will use it during its CMR processing. So even the call `jtaEnvironmentBean.getValue().getCommitMarkableResourceJNDINames().remove(jndiName)` has not direct implication to the JNDI change the runtime will be effectively using the new jndi definition.

As overall preference is against general changes in handling I'm closing this PR.

PR #1644, opened 2020-07-02
JBTM-3341 Failed LRA is not kept in FailedTo* state in the object store

xstefank:
https://issues.redhat.com/browse/JBTM-3341

### DIFF
--- a/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
+++ b/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
@@ -374,6 +374,10 @@ public class Transaction extends AtomicAction {
                 (status.equals(LRAStatus.Cancelling) || status.equals(LRAStatus.Closing));
     }
 
+    public boolean isFailed() {
+        return status != null &&
+            (status.equals(LRAStatus.FailedToCancel) || status.equals(LRAStatus.FailedToClose));
+    }
 
     public boolean isCancel() {
         switch (status) {
--- a/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
+++ b/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
@@ -179,7 +179,7 @@ public class LRAService {
 
             if (!transaction.hasPendingActions()) {
                 // this call is only required to clean up cached LRAs (JBTM-3250 will remove this cache).
-                remove(transaction.getId());
+                remove(transaction);
             }
         }
     }
@@ -198,6 +198,13 @@ public class LRAService {
         } catch (Exception e) {
             return false; // the uid segment is invalid
         }
+    }
+
+    public boolean remove(Transaction lra) {
+        if (lra.isFailed()) { // persist failed LRA state
+            lra.deactivate();
+        }
+        return remove(lra.getId());
     }
 
     public boolean remove(URI lraId) {
--- a/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/RecoveringLRA.java
+++ b/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/RecoveringLRA.java
@@ -97,6 +97,8 @@ class RecoveringLRA extends Transaction {
                     switch (getLRAStatus()) {
                         case Closed:
                         case Cancelled:
+                        case FailedToClose:
+                        case FailedToCancel:
                             getLraService().finished(this, false);
                             break;
                         default:


rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
```diff
@@ -377,7 +377,7 @@ public Response compensateNestedLRA(@PathParam("NestedLraId") String nestedLraId
     @PUT
     @Path("nested/{NestedLraId}/forget")
     public Response forgetNestedLRA(@PathParam("NestedLraId") String nestedLraId) {
-        lraService.remove(toURI(nestedLraId));
+        lraService.remove(lraService.getTransaction(toURI(nestedLraId)));
```
mmusgrov, line 380: This code calls into LRAService to convert an id into a transaction and then calls back into lraService with the returned transaction. Why not add a method to LRAService that takes the id (and internally converts it):

```java
public boolean remove(URI lra)
```

ochaloup: I think the `LRAService#remove` should work directly with `lraId` and not with `Transaction`.The `remove` method should probably use the LRA `Transaction` instance obtained from the list and not the one that comes in argument
- then this line is currently wrong: https://github.com/jbosstm/narayana/blob/5.10.5.Final/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java#L207

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -374,6 +374,10 @@ public boolean isRecovering() {
                 (status.equals(LRAStatus.Cancelling) || status.equals(LRAStatus.Closing));
     }
 
+    public boolean isFailed() {
+        return status != null &&
```
ochaloup, line 378: Thsis is not a request for a change but just a note "thinking loudly"
Do you know why the LRA `Transaction` instance permits the `status` to be `null`?
I think it could be better to assert that's not null? I can see the code may assign the `status` to `null` during recovery. Maybe some future change could adjust it.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
xstefank: the issue is in the other call of the `remove` method where the recovered transaction object on which we actually call the second phase again and has its state updated accordingly (failedToClose) is not included in the lra list. The LRA taken from the list is in the wrong state.

Probably the better fix would be to include the Recovery LRA object in the lra list directly but that would need to be replaced on every recovery pass as a new recovery LRA is created and so far we didn't run into this issue so I've decided that this is an easier fix.

Let me know how would you like to proceed, please.

ochaloup: @xstefank for me to understand better, the code that we discuss is
```
if (lra.isFailed()) { // persist failed LRA state
  lra.deactivate();
}
```
where the `lra` is deactivated and this call has to be run on a `lra` instance, right?
You suggest that the passed `lra` is not part of the `Coordinator` `lras` when `remove` is called from `LRAService#finished`, correct?
If the `lra` is not included in the lra list (which I assume is meant to be `lras`) then the code I pasted above won't never be called. Then it won't be a problem. 

Probably. So I understand it in a wrong way. Can you explain me a bit more, please?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
mmusgrov: The status cannot be null. We used to use null to indicate that the transaction was active but we have since introduced an explicit state for this (`LRAStatus.Active`). I didn't mention it in my review since it is unrelated to the PR, I was going to clean up the code in one go in a dedicated PR.

xstefank: thanks Mike, that will be better.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
mmusgrov: Just keep the existing  `lraService.remove` method and maybe add a new one that takes a transaction instance. I would advice against refactoring the code at this stage.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
mmusgrov: Where does "the code assign null to the status"? It should not do that.

Checking for status == null is a throwback to when we did not have an `LRAStatus.Active` state so the null checks are superfluous but I would advice against refactoring in the PR - do it in a separate one that does a general clean up of the code.

ochaloup: @mmusgrov https://github.com/jbosstm/narayana/blob/master/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java#L272

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
ochaloup: I understand the issue now and I got the point about the `lra` instance passed to the `remove` method. 
I was trying to understand the processing and I'm not able to come with a better/nicer way. I assume it's fine to be this way.

I don't like code duplication thus I do not prefer the ducplication code into `lraService.remove` method but if it's fine for the rest then ok.

xstefank:
@mmusgrov @ochaloup I just added the remove(URI) method as Mike requested. I believe this is good to go now.

ochaloup:
Ok, I have not more points about this so let's merge it.

PR #1648, opened 2020-07-09
[SCRIPTS ONLY] removing the info on checks on WFLY issues and docker …

ochaloup:
for release 5.9.9.Final

### DIFF
--- a/narayana-release-process.sh
+++ b/narayana-release-process.sh
@@ -1,24 +1,16 @@
 #!/bin/bash
 
 # You can run this script with 0, 1 or 3 arguments.
-# Arguments are then tranformed to env varibales CURRENT, NEXT and WFLYISSUE
+# Arguments are then tranformed to env varibales CURRENT, NEXT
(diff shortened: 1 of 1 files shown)


ochaloup:
hi @tomjenkinson I have few more changes for the release script. I wanted to start with the release and I found there are checks which are not necessary anymore. I want to remove them for not being misguiding (e.g. docker is not necessary to be checked when it's not used etc.)

ochaloup:
I aborted the CI.

narayana-release-process.sh
```diff
@@ -43,10 +35,6 @@ then
   exit
 fi
 
-# Do this early to prevent later interactive need
-docker login docker.io
```
tomjenkinson, line 47: Until we have the next minor we would probably push updates to docker but as we have a 5.10 so we don't need to do this for this branch still but in future we would think when is the right time to remove this step

tomjenkinson: same observation would apply to the WFLYISSUE I suppose

PR #1651, opened 2020-07-20
JBTM-3346 Journal store file count grows unbounded

jhalliday:
change poolSize to have default limit.
upgrade artemis-journal to pull in https://issues.apache.org/jira/browse/ARTEMIS-2837


mmusgrov:
The failures are the same as the ones [reported](https://github.com/jbosstm/narayana/pull/1650#pullrequestreview-451649222) on #1650:  plus an extra failure on the CORE axis: OSGiJTATest org.jboss.arquillian.container.spi.client.container.LifecycleException: Cannot obtain MBean server connection

NB We have been plagued by connectivity issues on the AMS CI for a couple of months now and we are still investigating why the infrastructure (docker based) is unstable.

PR #1656, opened 2020-08-17
[SCRIPTS ONLY] .bat script to use github token instead of user/password

ochaloup:
!CORE !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !XTS !PERF !RTS !AS_TESTS !TOMCAT !JACOCO !LRA

### DIFF
--- a/scripts/hudson/narayana.bat
+++ b/scripts/hudson/narayana.bat
@@ -107,6 +107,6 @@ goto:eof
    for /f "tokens=1,2,3,4 delims=/" %%a in ("%GIT_BRANCH%") do set IS_PULL=%%b&set PULL_NUM=%%c
    if not "%IS_PULL%"=="pull" goto:eof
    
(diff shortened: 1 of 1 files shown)


ochaloup:
@tomjenkinson thanks Tom for fixing the description. The good point from perspective of this PR is that the comment was passed to github comment and it's shown.
@tomjenkinson @mayankkunwar could you, in case, review the change in this PR? I think it can be merged as it is.

The other problem is that the PR fails on blacktie. There is a failure `org.jboss.narayana.blacktie.jatmibroker.xatmi.ConnectionException: Could not create the receiver on: JAVA_Converse: 12`. I think I will check this with @mmusgrov when he is back as here I'm not sure how to grasp the error.

mayankkunwar:
For this PR why does it say, `Tests passed on Windows - Job complete` for `jdk8.latest,label=catelyn/21/` and `jdk11.latest,label=catelyn/21/`, but actually these builds failed. Is this a known issue?

mayankkunwar:
Hi @tomjenkinson I see your point. Thanks for the explanation.

PR #1657, opened 2020-08-17
JBTM-3299 Expose optional additional description of LRARecord status

xstefank:
https://issues.redhat.com/browse/JBTM-3299

### DIFF
--- a/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
+++ b/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
@@ -128,7 +128,7 @@ public class Coordinator {
 
     private static LRAData convert(LRAStatusHolder lra) {
         return new LRAData(lra.getLraId(), lra.getClientId(),
-                lra.getStatus().name(),
+                lra.getStatus().name(), lra.getStatusReason().toString(), lra.getReportedParticipants().report(),
                 lra.isClosed(), lra.isCancelled(), lra.isRecovering(), lra.isActive(), lra.isTopLevel(),
                 lra.getStartTime(), lra.getFinishTime());
     }
--- a/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRAData.java
+++ b/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRAData.java
@@ -7,6 +7,8 @@ public class LRAData {
     private String lraId;
     private String clientId;
     private String status;
+    private String statusReason;
+    private String reportedParticipants;
     private boolean isClosed;
     private boolean isCancelled;
     private boolean isRecovering;
@@ -15,13 +17,15 @@ public class LRAData {
     private long startTime;
     private long finishTime;
 
-    public LRAData(String lraId, String clientId, String status,
+    public LRAData(String lraId, String clientId, String status, String statusReason, String reportedParticipants,
                            boolean isClosed, boolean isCancelled, boolean isRecovering,
                            boolean isActive, boolean isTopLevel,
                            long startTime, long finishTime) {
         this.lraId = lraId;
         this.clientId = clientId;
         this.status = status;
+        this.statusReason = statusReason;
+        this.reportedParticipants = reportedParticipants;
         this.isClosed = isClosed;
         this.isCancelled = isCancelled;
         this.isRecovering = isRecovering;
@@ -41,6 +45,14 @@ public class LRAData {
 
     public String getStatus() {
         return this.status;
+    }
+
+    public String getStatusReason() {
+        return statusReason;
+    }
+
+    public String getReportedParticipants() {
+        return reportedParticipants;
     }
 
     public boolean isClosed() {
@@ -101,6 +113,7 @@ public class LRAData {
                 "lraId='" + lraId + '\'' +
                 ", clientId='" + clientId + '\'' +
                 ", status='" + status + '\'' +
+                ", statusReason='" + statusReason + '\'' +
                 ", isClosed=" + isClosed +
                 ", isCancelled=" + isCancelled +
                 ", isRecovering=" + isRecovering +
--- a/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
+++ b/rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
@@ -376,6 +376,8 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
         updateStatus(compensate);
 
         // if the the request is still in progress (ie accepted is true) let recovery finish it
+        lra.setLRAStatusReason(LRAStatusReason.ParticipantResponse);
+        lra.getReportedParticipants().addParticipant(recoveryURI, LRAStatusReason.ParticipantResponse);
         return atEnd(accepted ? TwoPhaseOutcome.HEURISTIC_HAZARD : TwoPhaseOutcome.FINISH_OK);
     }
 
@@ -440,6 +442,7 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
                 && (status == ParticipantStatus.Completed || status == ParticipantStatus.FailedToComplete)) {
             if (lraService.getLRA(parentId).isActive()) {
                 // completed nested participants must remain compensatable
+                lra.setLRAStatusReason(LRAStatusReason.NestedClose);
                 return TwoPhaseOutcome.HEURISTIC_HAZARD; // ask to be called again
             } else {
                 // the parent is finishing so this is the post LRA invocation
@@ -451,6 +454,7 @@ public class LRARecord extends AbstractRecord implements Comparable<AbstractReco
         // check the participant first since it will have been removed from one of the lists
         if (!isFinished() || !lra.isFinished()) {
             if (afterURI != null) {
+                lra.setLRAStatusReason(LRAStatusReason.AfterLRARequestPhase);
                 return TwoPhaseOutcome.HEURISTIC_HAZARD;
             }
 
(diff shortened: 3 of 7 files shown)


rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRAStatusHolder.java
```diff
@@ -76,6 +78,10 @@ public LRAStatus getStatus() {
         return lraStatus;
     }
 
+    public String getLraStatusReason() {
+        return isRecovering() ? lraStatusReason : "N/A";
```
ochaloup, line 82: I think it could be good to move this `N/A` constant under `LRAStatusReason`.
The change of the text of the reason may be done at one place while the constant is used on the other place as well.

ochaloup: A question: why is exposed the reason only in case the LRA is recovering? I can't get it. Does it mean that the `LRAStatusReason` describes only some failure states?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -1025,4 +1045,14 @@ public boolean hasPendingActions() {
 
         return participants;
     }
+
+    public static final class LRAStatusReason {
```
ochaloup, line 1049: Is the correct name for the `enum "LRAStatusReason" when it seems it gathers only failure states? Is there a expectation that some additional non-error statuses will be added?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
```diff
@@ -128,7 +128,7 @@
 
     private static LRAData convert(LRAStatusHolder lra) {
         return new LRAData(lra.getLraId(), lra.getClientId(),
-                lra.getStatus().name(),
+                lra.getStatus().name(), lra.getLraStatusReason(),
```
mmusgrov, line 131: rename `getLraStatusReason` to `getStatusReason` since the field is already part of an LRA data structure. Similarly for the setter.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -550,10 +554,14 @@ private int retryGetEndStatus(URI endPath, boolean compensate) {
...
-                            Response.Status.Family.SERVER_ERROR)) {
-                    // these response codes indicate that the implementation should retry later
+                } else if (response.getStatus() == Response.Status.ACCEPTED.getStatusCode()) {
+                    // this response code indicate that the implementation should retry later
+                    lra.setLRAStatusReason(Transaction.LRAStatusReason.PARTICIPANT_RESPONSE);
+                    return TwoPhaseOutcome.HEURISTIC_HAZARD;
+                } else if (Response.Status.Family.familyOf(response.getStatus())
+                    .equals(Response.Status.Family.SERVER_ERROR)) {
+                    // participant processing failed, retry later
+                    lra.setLRAStatusReason(Transaction.LRAStatusReason.PARTICIPANT_FAILED_STATUS);
```
mmusgrov, line 564: Won't this status reason overwrite the existing reason (the same comment applies to the other places where the participant record updates the LRA status reason).

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRAStatusHolder.java
```diff
@@ -48,6 +48,7 @@
     String[] failedParticipants;
 
     private LRAStatus lraStatus;
+    private String lraStatusReason;
```
mmusgrov, line 51: Rename: `lraStatusReason` to `statusReason` since the field is already part of an LRA data structure.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRAStatusHolder.java
mmusgrov: Status reasons are constrained to the belong to a fixed set so use a data type (ie don't use free format strings).

Also, why would the reason for being in a particular state ever *not* be applicable? Perhaps you meant "Unknown".

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -151,6 +153,14 @@ public boolean save_state(OutputObjectState os, int ot) {
                 os.packBoolean(true);
                 os.packString(status.name());
             }
+
+            if (statusReason == null) {
```
mmusgrov, line 157: Use an enum to model the fixed finite set of reasons for the LRA to be in a given state. Then you will never need to pack a null value.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -361,6 +373,14 @@ protected void setLRAStatus(int actionStatus) {
         status = toLRAStatus(actionStatus);
     }
 
+    public String getLRAStatusReason() {
```
mmusgrov, line 376: See my previous comment about renaming these methods (to getStatusReason etc)

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
mmusgrov: The set of reasons for an LRA to be in a given state is finite so use an enum to model it. If you really think this structure needs to be extensible (which we should strive to avoid) then introduce a value that can be further embellished (such as `Other` or `Unknown`) by using a secondary data type.

Note that I did not think about whether you have captured the full set of reasons for each state transition in our model (there are reasons why a state is left and reasons why a state is entered so the values need to capture both reasons). If you can wait, I will investigate this tonight (I'm on PTO this week).

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -1025,4 +1045,14 @@ public boolean hasPendingActions() {
 
         return participants;
     }
+
+    public static final class LRAStatusReason {
+        public static final String TIMEOUT = "LRA timeout";
```
mmusgrov, line 1050: Note that you may still include descriptive strings as fields of `Enum` datatypes so you can keep the reason text if you use an Enum data structure to model the reasons for each state transition.

mmusgrov:
The issue description included [a list of strings to be included in the JSON response for participant state changes](https://issues.redhat.com/browse/JBTM-3299?focusedCommentId=14049191&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14049191). I didn't see that being addressed in this PR.

mmusgrov:
> Note that I did not think about whether you have captured the full set of reasons for each state transition in our model (there are reasons why a state is left and reasons why a state is entered so the values need to capture both reasons). If you can wait, I will investigate this tonight.

The list of reasons for LRA state transitions are:

```java
public enum LRAStatusReason {
    StateTransition // triggered by the protocol machine
    ClientRequest, // an external request 
    LRATimeout, // the time limit for the LRA was reached
    ParticipantTimeout, // the time limit for the participant was reached
    Heuristic, // the participant did the opposite of what was requested
}
```

And valid reasons for each LRA state transition are:

  `Active -> Closing` reason: {ClientRequest, StateTransition (for a nested LRA)}
  `Active -> Cancelling` reason: {ClientRequest, LRATimeout, ParticipantTimeout, StateTransition (for a nested LRA)}

  `Closing -> Closed` reason {StateTransition}
  `Closing -> FailedToClose` reason: {Heuristic} // I deliberately excluded ParticipantTimeout

  `Cancelling -> Cancelled` reason: {StateTransition}
  `Cancelling -> FailedToCancel` reason: {Heuristic} // I deliberately excluded ParticipantTimeout

If there is a need then there would be a similar set of reasons for the participant state model but access to the reasons would need be protected since the participants may be separated by organisational boundaries (personally I would prefer just LRAStatusReason).

(I will put the above comments on the JIRA for tracking and discussion purposes).

If the LRA is in a terminal state but:

- there are listeners, then we need a field to report that: `boolean listenersNotified;`
- there are participants which are waiting for the forget notification, then we need a field to report that: `boolean forgetSent;`

but only if there is no `emum ParticipantStatusReason`.

The discovery of the status reason should be performed on demand (I believe all the information is present in the Transaction and LRARecord classes but I would need to double check).

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRAData.java
```diff
@@ -101,6 +113,7 @@ public String toString() {
                 "lraId='" + lraId + '\'' +
                 ", clientId='" + clientId + '\'' +
                 ", status='" + status + '\'' +
+                ", statusReason='" + statusReason + '\'' +
```
ochaloup, line 116: I'm not sure if this was discussed elsewhere - why the `toString` method does not contain the `String` `reportedParticipants`?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRAStatusHolder.java
ochaloup: I see. Good.

mmusgrov:
@xstefank I am putting this PR on hold. When you're ready can you add a comment and I will remove the HOLD label.

xstefank:
@mmusgrov this will need rework, closing it for now.

PR #1658, opened 2020-08-19
[JBTM-3355] harmonize the LRA DTO usage to use only LRAData

ochaloup:
https://issues.redhat.com/browse/JBTM-3355

This is a step before the JBTM-3294 where only one Coordinator LRA DTO object is to be used. The `LRAStatusHolder` is replaced by `LRAData` where is needed and status calls are a bit narrowed.

MAIN LRA
!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS !AS_TESTS !TOMCAT !JACOCO


rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRAData.java
```diff
@@ -97,18 +83,102 @@ public int hashCode() {
...
+    }
+    public static interface BuilderWithHttpStatus {
+        BuilderWithResponseData httpStatus(int httpStatus);
+    }
+    public static interface BuilderWithResponseData {
+        BuilderFinal responseData(String responseData);
+    }
+    public static interface BuilderFinal {
+        LRAData build();
+    }
```
xstefank, line 183: Sorry, but I think that the Builder here is overcomplicated. What would be the reason to keep the order of methods in the builder? Also since all `BuilderWith*` interfaces have only one method which represents next step in the building process then all parameters are required, and so this is just a really complicated way of writing an all args constructor :)

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -145,12 +157,8 @@ public boolean save_state(OutputObjectState os, int ot) {
                 os.packLong(finishTime.toInstant(ZoneOffset.UTC).toEpochMilli());
             }
 
-            if (status == null) {
-                os.packBoolean(false);
-            } else {
-                os.packBoolean(true);
-                os.packString(status.name());
-            }
+            os.packBoolean(true);
```
xstefank, line 160: is boolean before status required?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -269,7 +277,7 @@ public boolean restore_state(InputObjectState os, int ot) {
             clientId = os.unpackString();
             startTime = os.unpackBoolean() ? LocalDateTime.ofInstant(Instant.ofEpochMilli(os.unpackLong()), ZoneOffset.UTC) : null;
             finishTime = os.unpackBoolean() ? LocalDateTime.ofInstant(Instant.ofEpochMilli(os.unpackLong()), ZoneOffset.UTC) : null;
-            status = os.unpackBoolean() ? LRAStatus.valueOf(os.unpackString()) : null;
+            status = os.unpackBoolean() ? LRAStatus.valueOf(os.unpackString()) : LRAStatus.Active;
```
xstefank, line 280: `os.unpackBoolean()` will be always true

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -362,24 +371,27 @@ protected void setLRAStatus(int actionStatus) {
     }
 
     boolean isClosed() {
-        return status != null && status.equals(LRAStatus.Closed);
+        assert status != null;
+        return status.equals(LRAStatus.Closed);
```
xstefank, line 375: you gave me an idea earlier in the PR :) maybe we can rewrite these comparisons to use `==` as they are enums. We can save some processing.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
```diff
@@ -124,45 +123,41 @@ public synchronized ReentrantLock tryLockTransaction(URI lraId) {
...
         if (scan) {
             RecoveryManager.manager().scan();
         }
 
-        return recoveringLRAs.values().stream().map(LRAStatusHolder::new).collect(toList());
+        return recoveringLRAs.values().stream().map(t -> t.getLRAData()).collect(toList());
     }
 
-    public List<LRAStatusHolder> getAllRecovering() {
+    public List<LRAData> getAllRecovering() {
```
xstefank, line 160: we should use unified either Collection or List but I don't see the reason to mix.

xstefank (changes requested):

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
mmusgrov: Agreed, status can never be null.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -145,12 +157,8 @@ public boolean save_state(OutputObjectState os, int ot) {
...
             }
 
-            if (status == null) {
-                os.packBoolean(false);
-            } else {
-                os.packBoolean(true);
-                os.packString(status.name());
-            }
+            os.packBoolean(true);
+            os.packString(status == null ? LRAStatus.Active.name() : status.name());
```
mmusgrov, line 161: status can never be null.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -351,9 +359,10 @@ protected LRAService getLraService() {
     /**
      * return the current status of the LRA
      *
-     * @return null if the LRA is still active (not closing, cancelling, closed or cancelled
+     * @return current LRA status, will not return <code>null</code>
```
mmusgrov, line 362: The text after the comma is not need (and therefore confusing).
status can never be null.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
```diff
@@ -351,9 +359,10 @@ protected LRAService getLraService() {
...
      */
     public LRAStatus getLRAStatus() {
+        assert status != null;
```
mmusgrov, line 365: Agreed, status can never be null. But why do we need to be asserting that.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
mmusgrov: +1

Also `isClosed`, `isCancelled`, `isActive` etc are not required since any caller just needs to call `getStatus()` and then do the test (in fact I think they are unused). Also note that `class Transaction` is internal to the implementation so we have full control over the users of this class.
 
The other methods such as `isFinished` and `isRecovering` are useful since the check multiple states.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
mmusgrov: +1
I would choose `List` since there is a natural ordering for LRAs.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
```diff
@@ -474,4 +463,9 @@ void disableRecovery() {
             }
         }
     }
+
+    private Collection<LRAData> getLraDataByStatus(Map<URI, Transaction> lrasToFilter, LRAStatus status) {
```
mmusgrov, line 467: Do we need to include `Lra` in the method names, it is clear that these methods are collecting LRAs.
But using it in `lrasToFilter` is good since it explains that the map contains LRAs.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -195,7 +195,8 @@ public void getRecoveringLRAs(Map<URI, Transaction> lras) {
 
     /**
      * remove an LRA log record
-     * @param lraUid
+     *
+     * @param lraUid lra id that will be removed from the log record
```
mmusgrov, line 199: LRA is an acronym so it should be capitalised (ie lra -> LRA) but obviously this comment does not apply to java variable names.

mmusgrov (changes requested):

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRAData.java
```diff
@@ -97,18 +83,102 @@ public int hashCode() {
...
+        private long finishTime;
+        private int httpStatus;
+        private String responseData;
+
+        private Builder() {
+            // no public instance
+        }
+
+        public static BuilderWithLraId instanceOf() {
+            return new Builder();
```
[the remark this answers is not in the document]
ochaloup: I want not to require the order of the methods particularly but it's a secondary effect.
What I want is to require that the all the arguments are defined. For the `build()` method could be called and the object is created the developer has to load all data through the builder.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRAData.java
ochaloup: I can remove the `Builder` if you both think it's over-complicated. My position is that the creation of an object with constructor of 8 arguments are pretty messy and builder helps here to put all the values at the right place.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
ochaloup: +1

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
ochaloup: +1

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRAData.java
```diff
@@ -97,18 +83,102 @@ public int hashCode() {
...
+        private long startTime;
+        private long finishTime;
+        private int httpStatus;
+        private String responseData;
+
+        private Builder() {
+            // no public instance
+        }
+
+        public static BuilderWithLraId instanceOf() {
```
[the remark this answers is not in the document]
ochaloup: +1

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRAData.java
```diff
@@ -72,12 +58,12 @@ public long getFinishTime() {
...
-    public long getTimeNow() {
-        return LocalDateTime.now().atZone(ZoneOffset.UTC).toInstant().toEpochMilli();
+    public int getHttpStatus() {
+        return this.httpStatus;
     }
 
-    public ZoneOffset getZoneOffset() {
-        return ZoneOffset.UTC;
+    public String getResponseData() {
+        return this.responseData;
```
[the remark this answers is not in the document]
ochaloup: +1

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
ochaloup: +1, good point, I was thinking to use the `Collection` on `getAll` method for I can use the `Set` to prevent duplication of the records in the result collection. Then it happens I introduced this ambiguity. But I can switch back to `List`.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
```diff
@@ -112,25 +110,23 @@
     @APIResponse(description = "The LRA",
         content = @Content(schema = @Schema(type = SchemaType.ARRAY, implementation = LRAData.class))
     )
-    public List<LRAData> getAllLRAs(
+    public Collection<LRAData> getAllLRAs(
             @Parameter(name = STATUS_PARAM_NAME, description = "Filter the returned LRAs to only those in the give state (see CompensatorStatus)")
             @QueryParam(STATUS_PARAM_NAME) @DefaultValue("") String state) {
-        List<LRAStatusHolder> lras = lraService.getAll(state);
+        LRAStatus requestedLRAStatus = null;
+        if(state != null && !state.isEmpty()) {
```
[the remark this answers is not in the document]
ochaloup: +1

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRAData.java
```diff
@@ -97,18 +83,102 @@ public int hashCode() {
...
-                ", startTime=" + startTime +
-                ", finishTime=" + finishTime +
-                '}';
+        return String.format(
+            "%s {lraId='%s', clientId='%s', status='%s', isTopLevel=%b, isRecovering=%b, startTime=%d, finishTime=%d, httpStatus=%d, responseData='%s'}",
+                this.getClass().getSimpleName(), lraId, clientId, status,
+                isTopLevel, isRecovering, startTime, finishTime, httpStatus, responseData);
+    }
+
+    public static class Builder implements BuilderWithLraId, BuilderWithClientId, BuilderWithStatus, BuilderWithTopLevel,
```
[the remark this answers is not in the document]
ochaloup: I commented that above. I think it's easier to use a builder than a constructor with 8 parameters. This makes all the parameters required.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
ochaloup: assertion is good to demonstrate that the code is designed in way where the `status` won't be null. The developer says that he is aware of that fact and announce it to anybody reading the code later.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
ochaloup: +1,  `getDataByStatus` is fine

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
ochaloup: +1

ochaloup:
@mmusgrov @xstefank I hope I addressed your remarks. Please take a look once again now.

The only remaining review note is the  constructor vs. builder discussion. My personal opinion is that the builder is beneficial but if you both think that it makes the API misleading I will remove it. If you may, let me know.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRAData.java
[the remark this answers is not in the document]
ochaloup: I removed the `responseData` from `LRAData` while I left them in `LRARecord` where it seems they are used.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRAData.java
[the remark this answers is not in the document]
mmusgrov: This is still complicated and difficult for me to understand, I would not like to have to maintain this code.
This is internal class and is for our use only - only the JSON representation is ever exposed (ie it isn't part of any external API).

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
mmusgrov: the null check is superfluous

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
mmusgrov: If we assert program invariant's everywhere they exist then the code would be littered with assertions making the program difficult to read.

Add a comment to the field if you really want to stress that the status can only take values according to a well defined state model

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRAData.java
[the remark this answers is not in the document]
mmusgrov: See my other comment on the complexity of the builder code.

I have a reasonable degree of experience with Java but I still don't really understand this code.

Constructors for DTOs are similar to copy constructors which conventionally take values for all those fields that cannot be calculated.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
ochaloup: agree, I think these are superfluous. I will change this.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRAData.java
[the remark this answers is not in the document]
ochaloup: I don't consider the builder code complex. It's just a Java code. It's a way how Java permits to build the API when it does not permit optional parameters or named ones.
I agree on the point that this is a private class and thus strictly with not a direct benefit.
I will change it for constructor.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/Transaction.java
ochaloup: +1

ochaloup:
@mmusgrov I updated the PR

mmusgrov (approved):

xstefank (approved):

PR #1663, opened 2020-08-24
JBTM-3079 test change for master

tomjenkinson:
…by the recovery manager
https://issues.redhat.com/browse/JBTM-3079

!MAIN !TOMCAT !AS_TESTS !RTS !JACOCO XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF !LRA NO_WIN !DB_TESTS mysql db2 postgres oracle

### DIFF
--- a/txbridge/src/test/java/org/jboss/jbossts/txbridge/tests/inbound/junit/InboundCrashRecoveryTests.java
+++ b/txbridge/src/test/java/org/jboss/jbossts/txbridge/tests/inbound/junit/InboundCrashRecoveryTests.java
@@ -29,8 +29,9 @@ import org.jboss.jbossts.txbridge.tests.inbound.client.TestClient;
 import org.jboss.jbossts.txbridge.tests.inbound.service.TestServiceImpl;
 import org.jboss.jbossts.txbridge.tests.inbound.utility.TestSynchronization;
 import org.jboss.jbossts.txbridge.tests.inbound.utility.TestXAResource;
-
 import org.jboss.jbossts.txbridge.tests.inbound.utility.TestXAResourceRecovered;
+import org.jboss.jbossts.txbridge.tests.inbound.utility.TestXAResourceRecoveryHelper;
+
 import org.jboss.logging.Logger;
 import org.jboss.shrinkwrap.api.Archive;
 import org.junit.*;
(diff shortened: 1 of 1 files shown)


ochaloup (commented):
Based on my investigation I think the test is not needed to be changed. What I think it could be reasonable to add the code as a new test and not removing the original one.

The reason for this update was introduced is that the 4.17 branch fails with the original test while `master` branch passes.

The test is meant to run as -> the TestXAResource is prepared then the test make the Bridge participant waiting. Then recovery is triggered and in version without the fix there is find some pending XAResource and it's rolled-back. The fix makes the inbound txbridge recovery manager not to abort it. In the meantime there is running a participant completion task (`com.arjuna.webservices11.wsat.sei.CoordinatorPortTypeImpl`). This is run by `TaskManager` periodically and trying to finish the XTS participants. The first time the recovery let the task went through prepare phase and on the second time (the recovery is run for the second time) the participant is committed.
The difference to 4.17 is that for 4.17 the resource is committed by the completion task as well but there is taken the TestXAResource while in master it works with TestXAResourceRecovered.

The InboundCrashRecoveryTests checks that commits occurs on the TestXAResoruceRecovered. It's the instance of the TestXAResource (enlisted to the in-flight transaction ) but the TestXAResoruceRecovered is instance taken from the object store - aka. the recovered one. The both instances would point to the same physical participant (e.g. if the participant is database then the recovered and in-flight commit call result of the commit of the work in database).

When the line https://github.com/jbosstm/narayana/blob/5.10.5.Final/txbridge/src/test/java/org/jboss/jbossts/txbridge/tests/inbound/junit/InboundCrashRecoveryTests.java#L276 is changed to TestXAResource then the 4.17 passes.
The reason for this is that the `master` branch contains several more changes about the work with the imported transactions. This is the changeset which is needed to get 4.17 brach to pass the original test:
https://github.com/jbosstm/narayana/compare/jbosstm:4.17...ochaloup:JBTM-3079-3138-2734-2701-3034?expand=1

mmusgrov:
I will not formally review this PR since I am not that familiar with XTS but it certainly looks like the test changes are validating that in flight transactions are correctly being ignored by the bridge recovery manager.

ochaloup:
I'm closing this PR as from discussion from @tomjenkinson we agreed to use both test not to change the current one. I will create a new PR.

PR #1668, opened 2020-08-27
[JBTM-3079][JBTM-2734][JBTM-2701][JBTM-3034] for 4.17

ochaloup:
https://issues.jboss.org/browse/JBTM-3079
https://issues.jboss.org/browse/JBTM-2734
https://issues.jboss.org/browse/JBTM-2701
https://issues.jboss.org/browse/JBTM-3034

XTS JTA_AS_TESTS
!MAIN !CORE !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !PERF NO_WIN !RTS !AS_TESTS !TOMCAT !JACOCO !LRA

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/PeriodicRecovery.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/PeriodicRecovery.java
@@ -814,7 +814,7 @@ public class PeriodicRecovery extends Thread
             }
 
             if (tsLogger.logger.isDebugEnabled()) {
-                tsLogger.logger.debug(" ");
+                tsLogger.logger.debugf("PeriodicRecovery: recovery module '%s' second pass processed", m);
             }
         }
 
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/CommitMarkableResourceRecordRecoveryModule.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/CommitMarkableResourceRecordRecoveryModule.java
@@ -434,6 +434,10 @@ public class CommitMarkableResourceRecordRecoveryModule implements
 
 	@Override
 	public synchronized void periodicWorkSecondPass() {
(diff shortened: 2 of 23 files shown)


ochaloup:
The particular issues are included here because

* `[JBTM-3079] InboundBridge recovery aborts live transactions` : fixes an issue of inbound XTS transaction being rolled-back by recovery despite it's active and not finished by TM
* `JBTM-2734 Make sure that when we recover the transactions we refreshed` : fixes the regression introduced in 4.17.43.Final (it was not in 4.17.42.Final)
* `JBTM-2701 Updated to allow JCA recover call to refresh the list of XIDs` : is required based on the changes from JBTM-2734 for the txbridge crash recovery tests passes. Otherwise the `testCrashTwoLogs` and others fail
* `[JBTM-3034] removing xid from processing when to be committed` : added for the same reason as JBTM-2701
* + few more commits which are required because of the `4.17` branch and backporting reasons

tomjenkinson (approved):
Thanks @ochaloup ! I think we should try to reflect those comments into the relevant JBTM. For instance try to work out which of the two material changes in 4.17.43.Final caused JBTM-2734 so that we can put a "caused by" link in that issue. I am going to add @mmusgrov as a reviewer too but thanks again!

ochaloup:
>  so that we can put a "caused by" link in that issue.

@tomjenkinson that's difficult. I can't define those points as "caused by". Whole this PR is about backporting some functionality from `master` with a great many changes to the older branch of `4.17`. The changes in `master` were done for different reasons and in a different order. These 5 changes are related but are not causing one other (in my understanding).
What I can say is that this is set of changes which interoperates together to make working the subordinate bridge XTS transactions when the subordinate recovery modules are added to the mix of `4.17` branch (https://github.com/jbosstm/narayana/compare/4.17.42.Final..4.17.43.Final).
I can add a comment to the particular issues in this way - which I agree is a really good point- or if you consider this should be done/linked differently then it could be easier if you make those changes as you know how the process works correctly.

mmusgrov (approved):

ochaloup:
Putting `On Hold` as we need to agree if this is the full functionality needed for JBTM-3079.

PR #1670, opened 2020-09-15
[JBTM-3357] Do not cache CMR configuration properties

gaol:
Issue: https://issues.redhat.com/browse/JBTM-3357

MAIN AS_TESTS

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/RecoveryEnvironmentBean.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/RecoveryEnvironmentBean.java
@@ -470,6 +470,19 @@ public class RecoveryEnvironmentBean implements RecoveryEnvironmentBeanMBean
     }
 
     /**
+     * Reloads the RecoveryModule.
+     *
+     * This method will reload the RecoveryModules forcibly.
+     *
+     * @return the copy of list of RecoveryModule instances
+     */
+    public List<RecoveryModule> reloadRecoveryModules() {
+        List<RecoveryModule> instances = ClassloadingUtility.loadAndInstantiateClassesWithInit(RecoveryModule.class, recoveryModuleClassNames);
+        setRecoveryModules(instances);
+        return new ArrayList<>(instances);
+    }
+
+    /**
      * Sets the instances of RecoveryModule.
      * The provided list will be copied, not retained.
      *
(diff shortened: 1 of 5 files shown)


mmusgrov:
Ah you provided a fix. I too was also working on it, I just had test still to do.
I will review what you have provided but whatever the outcome the PR would need to add a test in https://github.com/jbosstm/narayana/tree/master/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable.

mmusgrov:
@gaol Please sign the CLA over here (project name is JBoss Transactions): https://cla.jboss.org/index.seam

mmusgrov:
TESTIT

mmusgrov:
@gaol My fix ensured that all uses of CMR properties were dynamically read from the JTAEnvironmentBean.
I also added methods to JTAEnvironmentBean to avoid property changes interleaving:
``` java
    public void removeCommitMarkableResource(String jndiName) {
    public void addCommitMarkableResource(String jndiName, String tableName, boolean immediateCleanup, int batchSize) {
```

How does your PR relate to my approach?

gaol:
@mmusgrov Sorry again that I don't know you are working on it currently, please go ahead with your approach because I am quite new to this project, and has definitely less knowledge about it. :smile: 

Here are summary of the changes in this PR, just FYI:

* Added `reloadRecoveryModules()` method in `RecoveryEnvironmentBean` and invoke it `PeriodicRecovery.loadModules()` method. this new method re-load the RecoveryModules forcibly, and it will be called when [ArjunaTransactionManagerService](https://github.com/wildfly/wildfly/blob/master/transactions/src/main/java/org/jboss/as/txn/service/ArjunaRecoveryManagerService.java#L135) is started in WildFly server. The `:reload` operation will lead to `ArjunaRecoveryManagerService.start()`, new `CommitMarkableResourceRecordRecoveryModule` instance will be created at this time.
* Change static fields in `CommitMarkableResourceRecordRecoveryModule`, `CommitMarkableResourceRecord`, `TransactionImple` to final instance fields, so that new instances of these classes will pick new values from `JTAEnvironmentBean`

The benefit of using `final` fields is one step less to get the environment information in run-time.

Because any changes in `/subsystem=transactions/commit-markable-resource=*` will lead to `reload-required` state in WildFly server, I think it would be enough to pick the new values after `:reload`.

From what I understand from your description on your fix, I think your approach will make these values all dynamic, and it makes the changes in `JTAEnvironmentBean` won't need `:reload` anymore.

gaol:
suppressed by #1676

PR #1673, opened 2020-09-21
JBTM-3358 Update to MicroProfile LRA 1.0-M2

xstefank:
https://issues.redhat.com/browse/JBTM-3358

MAIN !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF LRA NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRACDIExtension.java
+++ b/rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRACDIExtension.java
@@ -39,6 +39,7 @@ import javax.enterprise.inject.spi.BeanManager;
 import javax.enterprise.inject.spi.Extension;
 import javax.enterprise.util.AnnotationLiteral;
 import java.io.IOException;
+import java.lang.reflect.Modifier;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRACDIExtension.java
```diff
@@ -101,12 +102,12 @@ public void observe(@Observes AfterBeanDiscovery event, BeanManager beanManager)
      * @return Collected methods wrapped in {@link LRAParticipant} class or null if no non-JAX-RS methods have been found
      */
     private LRAParticipant getAsParticipant(ClassInfo classInfo) throws ClassNotFoundException {
-        if (!isLRAParticipant(classInfo)) {
+        Class<?> javaClass = getClass().getClassLoader().loadClass(classInfo.name().toString());
```
ochaloup, line 105: There is a risk that `Class.getClassLoader()` will be null if the class was loaded by bootstrap class loader. Are we sure that this never happens? I understand that this code snippet was already involved in the code base - this PR does not add it. But just wondering.

ochaloup (approved):
Seems fine to me. Hoping that the LRA tests pass and that the `timeLimit` test failures will stop bothering the CI runs.

rts/lra/lra-proxy/api/src/main/java/io/narayana/lra/client/internal/proxy/nonjaxrs/LRACDIExtension.java
xstefank: @ochaloup not sure TBH. Maybe worth exploring in the new JBTM?

ochaloup: +1

mmusgrov: The bootstrap class loader loads system classes but the CDI extension is not a system class so a null check is not necessary.

Even if this were not the case, then why create a JIRA rather than updating the PR by checking for null, the code already handles null values on the next line (but this is really not necessary).

mmusgrov (approved):

ochaloup:
a note: I created the mentioned issue here https://issues.redhat.com/browse/JBTM-3359

PR #1676, opened 2020-09-23
JBTM-3357 Do not cache CMR properties

mmusgrov:
https://issues.redhat.com/browse/JBTM-3357

MAIN !TOMCAT AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF !LRA !NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

This PR replaces #1670
The CMR collection based properties are dependent on each other and need to be set as a group (see JTAEnvironmentBean).

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/CommitMarkableResourceRecord.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/CommitMarkableResourceRecord.java
@@ -41,6 +41,7 @@ import javax.transaction.xa.Xid;
 import com.arjuna.ats.arjuna.exceptions.ObjectStoreException;
 import com.arjuna.ats.internal.jta.resources.XAResourceErrorHandler;
 import com.arjuna.ats.internal.jta.transaction.arjunacore.TransactionSynchronizationRegistryImple;
+import com.arjuna.ats.jta.common.CmrJndiConfig;
 import org.jboss.tm.ConnectableResource;
 import org.jboss.tm.XAResourceWrapper;
 
@@ -112,8 +113,6 @@ public class CommitMarkableResourceRecord extends AbstractRecord {
 	private static CommitMarkableResourceRecordRecoveryModule commitMarkableResourceRecoveryModule;
 	private static final JTAEnvironmentBean jtaEnvironmentBean = BeanPopulator
 			.getDefaultInstance(JTAEnvironmentBean.class);
-	private static final Map<String, String> commitMarkableResourceTableNameMap = jtaEnvironmentBean
-			.getCommitMarkableResourceTableNameMap();
 	private static final String defaultTableName = jtaEnvironmentBean
 			.getDefaultCommitMarkableTableName();
 	private boolean isPerformImmediateCleanupOfBranches = jtaEnvironmentBean
@@ -121,8 +120,6 @@ public class CommitMarkableResourceRecord extends AbstractRecord {
 	private Connection preparedConnection;
 	private static final boolean isNotifyRecoveryModuleOfCompletedBranches = jtaEnvironmentBean
 			.isNotifyCommitMarkableResourceRecoveryModuleOfCompleteBranches();
-	private static final Map<String, Boolean> isPerformImmediateCleanupOfCommitMarkableResourceBranchesMap = jtaEnvironmentBean
-			.getPerformImmediateCleanupOfCommitMarkableResourceBranchesMap();
 
 	static {
 		commitMarkableResourceRecoveryModule = null;
@@ -174,7 +171,9 @@ public class CommitMarkableResourceRecord extends AbstractRecord {
 		this.basicAction = basicAction;
 		heuristic = TwoPhaseOutcome.FINISH_OK;
 
-		String tableName = commitMarkableResourceTableNameMap
+		CmrJndiConfig cmrConfig = jtaEnvironmentBean.getCmrJndiConfig();
+		String tableName = cmrConfig
+				.getCommitMarkableResourceTableNameMap()
 				.get(commitMarkableJndiName);
 		if (tableName != null) {
 			this.tableName = tableName;
@@ -182,7 +181,8 @@ public class CommitMarkableResourceRecord extends AbstractRecord {
 			this.tableName = defaultTableName;
 		}
 
-		Boolean boolean1 = isPerformImmediateCleanupOfCommitMarkableResourceBranchesMap
+		Boolean boolean1 = cmrConfig
+				.getPerformImmediateCleanupOfCommitMarkableResourceBranchesMap()
 				.get(commitMarkableJndiName);
 		if (boolean1 != null) {
 			isPerformImmediateCleanupOfBranches = boolean1;
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
@@ -1182,7 +1182,7 @@ public class JTAEnvironmentBean implements JTAEnvironmentBeanMBean
 	/**
 	 * Alter the default batch size or set to -1 to disable batch deletion of
 	 * CommitMarkableResourceRecord from the database.
-	 * 
+	 *
 	 * @param batchSize
 	 *            -1 to prevent batching.
 	 */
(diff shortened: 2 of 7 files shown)


ArjunaJTA/jta/src/test/resources/standalone-cmr.xml
```diff
@@ -148,6 +148,14 @@
                         <password>sa</password>
                     </security>
                 </datasource>
+                <datasource jndi-name="java:jboss/datasources/ExampleDS2" pool-name="ExampleDS2" enabled="true" use-java-context="true">
```
ochaloup, line 151: I think there is missing attribute `connectable=true`. If so then it will be missing in the `ExampleDS` as well.

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
```diff
@@ -1243,6 +1243,42 @@ public void setNotifyCommitMarkableResourceRecoveryModuleOfCompleteBranches(
...
+            commitMarkableResourceTableNameMap.put(jndiName, tableName);
+            performImmediateCleanupOfCommitMarkableResourceBranchesMap.put(jndiName, immediateCleanup);
+            commitMarkableResourceRecordDeleteBatchSizeMap.put(jndiName, batchSize);
+
+            // use the standard setter for notifyCommitMarkableResourceRecoveryModuleOfCompleteBranches,
+            // commitMarkableResourceRecordDeleteBatchSize and commitMarkableResourceTableName
+        }
+    }
+
+    public CmrJndiConfig getCmrJndiConfig() {
```
ochaloup, line 1273: I can't get the reason of having this new DTO `CmrJndiConfig`. Is that for more easily gathering the CMR data in the user code or is there a different reason?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/CommitMarkableResourceRecord.java
```diff
@@ -112,17 +113,13 @@
 	private static CommitMarkableResourceRecordRecoveryModule commitMarkableResourceRecoveryModule;
 	private static final JTAEnvironmentBean jtaEnvironmentBean = BeanPopulator
 			.getDefaultInstance(JTAEnvironmentBean.class);
-	private static final Map<String, String> commitMarkableResourceTableNameMap = jtaEnvironmentBean
-			.getCommitMarkableResourceTableNameMap();
 	private static final String defaultTableName = jtaEnvironmentBean
 			.getDefaultCommitMarkableTableName();
 	private boolean isPerformImmediateCleanupOfBranches = jtaEnvironmentBean
```
ochaloup, line 118: Is it fine to leave other configuration options for CMR as `static` here?

ochaloup (changes requested):
I have two questions about the code changes.
The "request changes" state of my review concerns mainly the `connectable` attribute for the datasource. If I'm not wrong then it's needed for CMR and we miss it in the tests.

ArjunaJTA/jta/src/test/resources/standalone-cmr.xml
mmusgrov: You are correct. I have added to both datasources (the integration test appears to use CMR both with and without this property).

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
mmusgrov: The properties in the DTO are all dependent on each other and need to be set as a group. Therefore they also need to be read as a group under the protection of the JTAEnvironmentBean lock. The methods `removeCommitMarkableResource`, `addCommitMarkableResource` and `getCmrJndiConfig` all need to be protected by this lock and this is achieved by copying them into the DTO.

BTW the DTO is still caching the collections. I have just pushed an update to the CmrJndiConfig constructor which avoids the caching. Good spot.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/CommitMarkableResourceRecord.java
mmusgrov: The WildFly CMR service does not touch these other properties.

ochaloup (commented):
Changing for a comment but I'm not sure about approving as it is now.

My question about the `isPerformImmediateCleanupOfBranches` relates to the fact that it's configuration which may be changed by `cli` command but has got no effect after `reload` of the server (https://docs.wildfly.org/20/wildscribe/subsystem/transactions/commit-markable-resource/index.html). (I understand that's not the part of the blocker issue but still it's something for consideration at least as a separate issue.)

I don't get the reason for having the `CmrJndiConfig`. For me it seems like adding more code which is not necessary.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/CommitMarkableResourceRecord.java
ochaloup: I see, +1

In general I think it would be good to handle all the CMR properties in the same way and not differently either they are used in WFLY or not. But it's my personal standpoint and I do not insist on it.

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
ochaloup: I see, with the copy constructor this makes sense now.

But I don't like the approach in general. I could be missing something but let me phrase by objections.

* These new methods changes the way how the `JTAEnvironmentBean` has been used. There were no methods which groups some operations so far (if I don't count the tightly coupled attributes with `*ClassNames`). Bean used the setter and getters and that was all. The Narayana machinery of the injecting of properties expects to call the setters and getters.
* The patch seems to be tailored only for the current WFLY issue where we expect only 4 properties to be set up and which are needed to be guarded by lock. But `JTAEnvBean` may be used in different runtimes.
* There is specified no contract of the class how it should be used regards of these new methods. If there is a new integrator then how to get the right way work with CMR properties.
* Despite the lock guards the `add`, `remove` and `get` there is not thread safety ensured. The arrays and maps may be changed in the background by another thread. What thread safety is tried to be ensured by the lock? Should not be lock defined at the caller side for the block? (https://github.com/wildfly/wildfly/blob/master/transactions/src/main/java/org/jboss/as/txn/service/CMResourceService.java#L72-L77)
* If those CMR properties are dependent and they should be changed atomically would not be better to define some (immutable) DTO which gather those values and may be switched in some map/list atomically?

ochaloup (changes requested):
I would have few remarks here.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/CommitMarkableResourceRecord.java
mmusgrov: Caching the per CM resource properties (batch-size, immediate-cleanup, table name) will not cause the data integrity issue that this PR is fixing.

However, you are correct in pointing out that these properties can be set on a per CM resource basis but narayana assumes that they are global properties. I will raise a separate issue to fix this.

ochaloup: I'm fine here. The issue could be good while I agree that these ones are the problematic causing the integrity issue. I meant this like a note and resolving this.

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
mmusgrov: Context: the JTA config bean maintains a list of CM resources, each identified by a unique JNDI name. Currently it does this via a number of collections.

I introduce new methods `add`, `remove` and `get` to manage these collections -  the other getters and setters in the config bean need to be deprecated, ie it is not tailored to the WFLY issue (this addresses your 1st, 2nd and 3rd points). Therefore the CMR propertes are managed as group and my PR does this (this addresses your 5th point).

When a new resource is added we need to update four maps atomically where each map is keyed by the unique resource name (this addresses your 4th point). When a reader gets the config the constructor for the DTO takes a copy of the four collections, the reader can then search the collections for the CM resource it is interested in (with no concern about thread safety).

The only point that I may not have addressed is your statement: `The Narayana machinery of the injecting of properties`. I don't see where that applies to the CMR functionality and why my PR breaks it.

mmusgrov: I think I see part of your concern in point 1 about the `Narayana injection machinery`, that relates to reading the properties from XML files. Let me check our XML parser and see if it can support this new technique.

ochaloup: I understand.

ochaloup (approved):
After off list discussion I understood the design of the fix. I agree it fixes the CMR trouble in WFLY and I'm approving the PR.

I would only point that could be good to mark the appropriate getters/setters as `@Deprecated`.

Still I'm a bit reserved to the solution and my concerns are: the API precedence in the `JTAEnvironmentalBean`, suboptimal performance solution with lock over the whole instance and thread safety threat when the deprecated setters/getters are used.

gaol:
@mmusgrov @ochaloup Would you agree that after this change, the changes in `/subsystem=transactions/commit-markable-resource=*` in WildFly server do not require `:reload` anymore ? Thanks!

mmusgrov:
I've put this on hold since the WFLY issues was fixed by requiring a restart (see PR https://github.com/wildfly/wildfly/pull/13620)

mmusgrov:
@ochaloup is it okay for me to withdraw this PR?

ochaloup:
@mmusgrov yes, sure. this was fixed at different level. we can resurrect the effort later when needed.

PR #1677, opened 2020-09-30
JBTM-3349 Replace Apache HttpClient with JAX-RS client in LRA modules 

xstefank:
https://issues.redhat.com/browse/JBTM-3349

Contains also a fix for https://issues.redhat.com/browse/JBTM-3189.

Please note that this is the minimal set of changes needed to pass all test suites. There are several enhancements that can be implemented on top of it but the time was prioritized.


rts/lra/lra-client/pom.xml
```diff
@@ -57,6 +57,16 @@
             <version>2.5</version>
             <scope>provided</scope>
         </dependency>
+      <dependency>
+        <groupId>org.eclipse.microprofile.rest.client</groupId>
+        <artifactId>microprofile-rest-client-api</artifactId>
+        <version>1.4.1</version>
```
ochaloup, line 63: Do you think we can move the version definition to a maven property?

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
```diff
@@ -569,9 +562,14 @@ public LRAStatus getStatus(URI uri) throws WebApplicationException {
...
         try {
-            response = getTarget().path(getLRAId(lraId.toString())).path("status")
-                    .request()
-                    .get();
+            response = coordinatorClient.getLRAStatus(getLRAId(lraId.toString()));
+        } catch (WebApplicationException wae) {
+            Response waeResponse = wae.getResponse();
+            if (waeResponse.getStatus() == Response.Status.NOT_FOUND.getStatusCode()) {
+                String responseContent = waeResponse.readEntity(String.class);
+                throw new NotFoundException("Failed to get status of LRA id " + lraId
```
ochaloup, line 570: For internationalization purposes we are used to use the error messages defined in `i18Logger` (ie. `LRALogger.i18NLogger.*`). Would be possible to change it that way?

(I understand that it was not done so in the code before but as I hit this in the diff I think it could be good to be changed.)

rts/lra/lra-service-base/src/main/java/io/narayana/lra/LRAData.java
```diff
@@ -8,14 +8,16 @@
...
 public class LRAData {
-    private final String lraId;
-    private final String clientId;
-    private final LRAStatus status;
-    private final boolean isTopLevel;
-    private final boolean isRecovering;
-    private final long startTime;
-    private final long finishTime;
-    private final int httpStatus;
+    private String lraId;
```
ochaloup, line 11: I don't understand the purpose of the change here. The `LRAData` was intentionally made immutable. Why is needed this to be changed?

ochaloup (changes requested):
I like the fix. I have just found two minor request for a change and I have one doubt on changes in `LRAData`.

rts/lra/lra-client/pom.xml
xstefank: sure, that is an oversight on my side, thanks

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
xstefank: I would suggest a new JBTM for this as this is not affecting only this code.

rts/lra/lra-service-base/src/main/java/io/narayana/lra/LRAData.java
xstefank: Used for the JSON (fasterxml) exchanges. FasterXML works with a default constructor and setters/getters still :)

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
ochaloup: +1

rts/lra/lra-service-base/src/main/java/io/narayana/lra/LRAData.java
ochaloup: Ok, I see. I was thinking there will be some reason just I haven't got it from the context. Then this is fine.

ochaloup (approved):

xstefank:
Just to be clear, this only rewrites the Coordinator API to use REST client from the participants. Rewriting Coordinator calls to participants to REST client is not possible as the endpoints differ for each participant and are collected dynamically. If we are to totally remove HttpClient and its associated API even from the coordinator we will need to switch back to plain resteasy client that works with Quarkus REST client extension (bundles both MP REST client and Resteasy JAX-RS client). Should I add this to this PR or create a new JBTM? The changes would be quite different to the issue title so I would prefer a new JBTM if total removal of HttpClient is required.

ochaloup:
@xstefank I would prefer a separate issue/PR for that additional change.

xstefank:
So the question is to @mmusgrov about what is meant by "Quarkus REST client" (extension or MP REST client) in the issue title. I have the required changes almost ready, so it doesn't matter to me whether its a new PR or this one.

ochaloup:
@xstefank ok, then, by me, feel free to add the new set of changes to this PR and we can recheck it.

mmusgrov:
@xstefank Sorry for the delay in looking at this.

I have only skimmed the PR this evening and will take a more detailed look during the week. I did note a couple of things that may interest you though:

1) Can we also have the OpenAPI documentation on the LRACoordinatorClient interface.

2) We also need a RestClient interface for the participants. Then we will be able to replace uses of Apache HTTP Client (which is the primary reason for the JIRA: ie "JBTM-3349 Replace Apache HttpClient with Quarkus REST Client") in the following classes:

    io.narayana.lra.LRAHttpClient
    io.narayana.lra.RequestBuilder
    io.narayana.lra.ResponseHolder

3) The answer to you question about "what is meant by "Quarkus REST client" is:
 the only reason we need this PR is that Quarkus does not support Apache HttpClient. So we need an alternative. That could be java.net.HttpURLConnection, or it could be Java 9 support for HTTP clients, or it could be RestClient basically anything that works with Quarkus - the details are in the linked JIRAs (https://issues.redhat.com/browse/JBTM-3349 and https://issues.redhat.com/browse/JBTM-3189).

mmusgrov (changes requested):
I added two comments on the PR which give my reasons for requesting changes.
Note that I have not fully reviewed the changes. I will do so when my comment is addressed/discussed.

@xstefank if you don't have time to fix the full issue then I don't mind accepting your partial fix and I will finish it off. Let me know how you want to tackle it? But if you do intend to finish it then this PR needs to address the full fix.

mmusgrov:
> @xstefank I would prefer a separate issue/PR for that additional change.

The calls to the participants can still use RestClient (you set the target URL on the builder) and they are needed to fulfil the requirements of the JIRA, ie we don't want a second PR so fix all of the issue in the current PR (I added some notes above in my previous comment).

xstefank:
@mmusgrov inline

> Can we also have the OpenAPI documentation on the LRACoordinatorClient interface.
No. This is not possible because the OpenAPI documents only server API. Basically, the Coordinator OpenAPI doc is the input for the REST client.

> @xstefank if you don't have time to fix the full issue then I don't mind accepting your partial fix and I will finish it off. Let me know how you want to tackle it? But if you do intend to finish it then this PR needs to address the full fix.
I am going to include the changes in this PR soon. I've run into some problems with the TCK runs and closing of clients but I'll refactor around it today.

xstefank:
@ochaloup, @mmusgrov ready for review

rts/lra/lra-client/pom.xml
```diff
@@ -57,6 +57,11 @@
             <version>2.5</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+          <groupId>org.jboss.resteasy</groupId>
+          <artifactId>resteasy-client-microprofile</artifactId>
+          <version>${version.org.jboss.resteasy}</version>
```
ochaloup, line 63: should not be this in scope `provided`?

ochaloup (commented):
For me the current state of PR seems fine to be merged. I have only a question on scope of the `lra-client` dependency.

ochaloup:
@mmusgrov ad https://github.com/jbosstm/narayana/pull/1677#issuecomment-703310966 )
What is the purpose of OpenAPI documentation for REST client? My understanding was that the OpenAPI explains the REST endpoints where the REST client serves in opposite for calling one.

mmusgrov:
> @mmusgrov inline
> 
> > Can we also have the OpenAPI documentation on the LRACoordinatorClient interface.
> > No. This is not possible because the OpenAPI documents only server API. Basically, the Coordinator OpenAPI doc is the input for the REST client.
> 

io.narayana.lra.coordinator.api.Coordinator implements all of the methods of LRACoordinatorClient so there is nothing stopping you from adding `Coordinator implements LRACoordinatorClient` (apart from the name not looking correct).

Not doing this means that we have to maintain the API in two separate places which opens the door to them diverging.

mmusgrov (changes requested):
All you changed was to go back to using JAX-RS for the client invocations, that is what we had originally. Please read the JIRA description (and read my comment on the PR https://github.com/jbosstm/narayana/pull/1677#issuecomment-703310966)

xstefank:
@mmusgrov 

> io.narayana.lra.coordinator.api.Coordinator implements all of the methods of LRACoordinatorClient so there is nothing stopping you from adding Coordinator implements LRACoordinatorClient (apart from the name looking odd).

I believe adding a dependency on coordinator to the client module (lra-client) would pull whole coordinator, even with JAX-RS endpoint transitively when someone would just use the lra-client which is I believe not wanted.

> All you changed was to go back to using JAX-RS for the client invocations, that is what we had originally. Please read the JIRA description (and read my comment on the PR #1677 (comment))

Quarkus REST client extension supports both MP REST client and plain RESTEasy JAX-RS client (it should in both JVM and native mode).

xstefank:
@ochaloup @mmusgrov 

After some discussions, we've reached a consensus that we will include Quarkus REST client compatible MP REST client interface for the LRA coordinator and the RESTEasy provided JAX-RS client for participant calls (as the paths are dynamic). 

Unknowingly I included a partial revert of the fix for https://issues.redhat.com/browse/JBTM-3189 in this PR as I returned to the JAX-RS client invocation. I've run into the same issue as described in [JBTM-3189](https://issues.redhat.com/browse/JBTM-3189) but only with the Thorntail coordinator. The issue is that ForkJoinPool is using a wrong ClassLoader which is fixed/hacked in the Quarkus so we don't see these failures with Quarkus based LRA coordinator. For this reason, I've also included this hack in the lra-coordinator-thorntail module and the TCK now passes without issues.

I've run the TCK and basic tests with this PR and:

* Quarkus coordinator - SUCCESS
* Thorntail coordinator - SUCCESS 
* LRA coordinator WAR deployed on WFLY master - there were some errors but unrelated to the ForkJoinPool issue (different tests) so we should take this in new JBTM

So I believe this PR is good to go. Also, I prepared a fix that will allow Quarkus coordinator to be compiled to native on top of this PR which I'll propose in new JBTM after this PR is merged.

ochaloup (approved):

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
```diff
@@ -238,13 +234,8 @@ public Response startLRA(
                     return Response.status(response.getStatus()).build();
                 }
             } else {
-                ResponseHolder resp = new RequestBuilder(parentLRAUrl)
```
mmusgrov, line 241: You can't just ignore code.
Are there any other places in this PR where you have removed stuff that's required?

xstefank: This was something that I actually thought was an oversight as this code calls the same method os the other if branch ([L235](https://github.com/jbosstm/narayana/pull/1677/files#diff-2c664918d9eca78964a0ea919711c8aaL235)) just through JAX-RS invocation which doesn't make sense if the parent LRA id is not found. Would you prefer I take this in a new JBTM?

mmusgrov: The if part of the statement is a local invocation.
The else part of the statement is a remote invocation.

ie they execute different behaviours so you can't just remove the else part (we need to interact with remote coordinators since this is not a centralised service - we're operating in distributed microservices based environment. 

I am not aware of any bug in the original code so there's no need for a JBTM, unless you have a test case that shows there is a problem with it.

xstefank:
@mmusgrov only fix the code mentioned in your last comment, no other changes.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
mmusgrov: I vaguely recall either a demo or a quickstart that relied on multiple coordinators. Check the quickstarts (in fact I think you or someone else needs to verify that the quickstarts still run with these updates before we attempt merging them).

mmusgrov: In particular the quickstarts rely on thorntail. As we discussed earlier replacing ApacheHttp client calls with JAX-RS thorntail does not work without your ForkJoinClassLoading fix. But running the quickstarts would validate that fix.

mmusgrov:
> @mmusgrov ad [#1677 (comment)](https://github.com/jbosstm/narayana/pull/1677#issuecomment-703310966) )
> What is the purpose of OpenAPI documentation for REST client? My understanding was that the OpenAPI describes the REST endpoints while the REST client serves in opposite direction, for the caller's outgoing requests.

OpenAPI documents the REST API used by our clients and implemented by our coordinator so it is meaningful to clients.
In addition if you change the coordinator you are going to have to change the client API and vice versa and if we don't force the coordinator to implement the client API you cannot enlist the compilers help to keep the two in sync.

xstefank:
@mmusgrov but that was also the case up to date when we had just plain JAX-RS calls and we haven't run into any issues. The REST client is just the declarative syntax of the same invocations. I don't personally believe that can be an issue in the future. We will catch any outdated endpoints in the client in TCK or basic tests.

xstefank:
@ochaloup @mmusgrov I did some testing with all our supported runtimes (Quarus, TT, WFLY) and with two possible solutions to the last issue with this PR -- whether to include or not include MP REST client in the lra-client module.

I've created a rewrite of MP REST client to plain JAX-RS client in https://github.com/xstefank/narayana/tree/jbtm-3349-jaxrs-only and tried to integrate lra-client into individual runtimes with this results:

* Quarkus - still requires `quarkus-rest-client` extension otherwise tries to pick Jersey client.
* TT - works
* WFLY works

Because of https://issues.redhat.com/browse/WFLY-13925 we can't include `resteasy-client-microprofile` dependency directly in the lra-client module. The other possible solution would be to include the REST Client API dependency - `microprofile-rest-client-api`. With this change in the current PR:

* Quarkus - requires also only `quarkus-rest-client`
* TT - requires `microprofile-restclient` fraction
* WFLY - works

So since plain JAX-RS client is a little simpler in the TT scenario I can rewrite this PR according to the linked branch. However, this is really just a little difference in my opinion and I prefer the declarative syntax of MP REST client (also it is still under active development and if we can solve WFLY-13925 then we can add implementation dependency to the lra-client in the same way as we have for resteasy-client right now). @ochaloup @mmusgrov please let me know how to proceed.

xstefank:
@mmusgrov @ochaloup Remove MP REST client. It's only JAX-RS Client everywhere now.

rts/lra/lra-client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
```diff
@@ -0,0 +1 @@
+org.jboss.resteasy.client.jaxrs.ResteasyClientBuilder
```
ochaloup, line 1: is this still needed here? does it help the coordinator to work right?

ochaloup (approved):
For me the changes seem fine. The failed test is the only known one with timing. I started the quickstarts with these changes, I will report back here how it passes.

I would only ask @xstefank - as we discussed the point that the rest client requires special dependency handling on Quarkus then if somebody wants to couple the coordinator with the client under his application then he will need to add that maven dependency to his project for such setup works, right?

rts/lra/lra-client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
xstefank: no, thanks

xstefank:
@ochaloup 
> I would only ask @xstefank - as we discussed the point that the rest client requires special dependency handling on Quarkus then if somebody wants to couple the coordinator with the client under his application then he will need to add that maven dependency to his project for such setup works, right?

The dependency is quarkus-rest-client extension which will set up the RESTEasy client. There is nothing preventing combining coordinator and client in a single app.

rts/lra/lra-client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
xstefank: actually, some other runtimes can still pick this up so I would prefer to leave it here.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
```diff
@@ -238,13 +239,34 @@ public Response startLRA(
...
+                        .get(PARTICIPANT_TIMEOUT, TimeUnit.SECONDS);
+
+                    if (response.getStatus() != Response.Status.OK.getStatusCode()) {
+                        String errMessage = String.format("The coordinator at %s returned an unexpected response: %d",
+                            parentLRAUrl, response.getStatus());
+                        return Response.status(response.getStatus()).entity(errMessage).build();
+                    }
+                } catch (Exception e) {
+                    LRALogger.logger.errorf("Cannot contact the LRA Coordinator at %s", parentLRA);
+                    return Response.status(PRECONDITION_FAILED).entity(e.getMessage()).build();
```
mmusgrov, line 262: This error code seems a bit random. The method signature is already handling it (ie it throws WebApplicationException). Also we don't need to log anything here (see my previous comments about logging errors). If you really did want to log it at trace level then just re-throw the WebApplicationException.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -419,17 +432,21 @@ private boolean afterLRARequest(URI target, String payload) {
...
-            ResponseHolder response = target.equals(forgetURI) ? builder.delete() : builder.put(payload, MediaType.TEXT_PLAIN);
+            Future<Response> responseFuture =  target.equals(forgetURI) ? builder.async().delete()
+                : builder.async().put(Entity.text(payload));
+            Response response = responseFuture.get(PARTICIPANT_TIMEOUT, TimeUnit.SECONDS);
 
             if (response.getStatus() == 200) {
                 return true;
             }
-        } catch (WebApplicationException e) {
+        } catch (Exception e) {
```
mmusgrov, line 442: WebApplicationException is the exception of choice here - since we are using JAX-RS.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -718,8 +745,14 @@ boolean forget() {
                     forgetURI, e.getMessage());
                 // TODO write a test to ensure that recovery only retries the forget request
                 return false; // force recovery to keep retrying
+            } catch (Exception e) {
```
mmusgrov, line 748: Similar comment as above. We are using JAX-RS so we need to use the WebApplicationException since that will already hold the relevant status code.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -718,8 +745,14 @@ boolean forget() {
...
+                LRALogger.logger.infof("LRARecord.forget at '%s' could not reach participant: %s",
```
mmusgrov, line 749: I would use debug here (failure to contact a participant is not particularly noteworthy since it is often an expected outcome in a distributed system).

rts/lra/lra-coordinator-thorntail/src/main/java/ForkJoinClassLoading.java
```diff
@@ -0,0 +1,69 @@
...
+import javax.enterprise.event.Observes;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Originally copied from https://github.com/quarkusio/quarkus/blob/1.8.1.Final/core/deployment/src/main/java/io/quarkus/runner/bootstrap/ForkJoinClassLoading.java
+ */
+@ApplicationScoped
+public class ForkJoinClassLoading {
```
mmusgrov, line 35: I like it, good find :-)

rts/lra/lra-coordinator/pom.xml
```diff
@@ -61,6 +61,10 @@
     </build>
 
     <dependencies>
+        <dependency>
```
mmusgrov, line 64: Not used

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
```diff
@@ -238,13 +239,34 @@ public Response startLRA(
                     return Response.status(response.getStatus()).build();
                 }
             } else {
-                ResponseHolder resp = new RequestBuilder(parentLRAUrl)
+                LRALogger.i18NLogger.error_parentLRANotFound(parentLRAUrl);
```
mmusgrov, line 242: Failure to find a resource in a distributed system is not an error (in fact it is a common occurrence).

mmusgrov (changes requested):

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
mmusgrov: But I do notice that if the LRA cannot be registered with the parent then we should be closing the one that was just started on the previous line. I will raise an issue to get that fixed.

pom.xml
```diff
@@ -459,6 +459,7 @@
     <version.javax.inject>1</version.javax.inject>
     <version.javax.servlet>3.0.1</version.javax.servlet>
     <version.commons-httpclient>3.1-jbossorg-1</version.commons-httpclient>
+    <version.commons-io>2.8.0</version.commons-io>
```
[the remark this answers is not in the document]
xstefank: This is required for the seamless inclusion in Quarkus and TT. The resteasy client doesn't bring this dependency and it fails some transformations in MessageBodyReaders and Writers. I was hesitant to include this also but take it as a short term workaround that we should try to fix in RESTEasy.

pom.xml
```diff
@@ -494,6 +495,7 @@
     <version.org.wildfly.checkstyle-config>1.0.8.Final</version.org.wildfly.checkstyle-config>
     <version.com.github.docker-java>3.0.14</version.com.github.docker-java>
     <version.org.eclipse.microprofile.openapi>1.1.2</version.org.eclipse.microprofile.openapi>
+    <version.org.eclipse.microprofile.rest.client>1.4.1</version.org.eclipse.microprofile.rest.client>
```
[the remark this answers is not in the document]
xstefank: +1

rts/lra/lra-client/pom.xml
```diff
@@ -57,6 +57,16 @@
             <version>2.5</version>
             <scope>provided</scope>
         </dependency>
+      <dependency>
+        <groupId>org.jboss.resteasy</groupId>
+        <artifactId>resteasy-json-binding-provider</artifactId>
```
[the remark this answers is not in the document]
xstefank: same as `commons-io`

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
```diff
@@ -230,6 +226,27 @@ public void setCurrentLRA(URI coordinatorUri) {
         }
     }
 
+    public List<LRAData> getAllLRAs() {
+        Client client = null;
+        try {
+            client = ClientBuilder.newClient();
```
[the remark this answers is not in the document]
xstefank: thanks, in NarayanaLRAClient this is possible to cache.

I'll just note that I've found it difficult to cache the client in the LRARecords as they are persisted and recovered from ObjectStore this is kind of difficult. I would like to investigate more in the future.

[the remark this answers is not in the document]
xstefank: So it's not, probably just synchronization issues and the same client is used concurrently in different enlistments for instance. Worth investigating.

pom.xml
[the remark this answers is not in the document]
mmusgrov: I don't follow, if a dependency is not used it does not need to be included.

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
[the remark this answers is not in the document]
mmusgrov: My comment was
> "Can you move the Client construction code to a separate method so that we can cache it in the future (you don't need to cache it in this PR but we should be doing so in the future)"

ie I am not asking you to cache the client in this PR.

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
```diff
@@ -230,6 +226,27 @@ public void setCurrentLRA(URI coordinatorUri) {
...
+
+            if (response.getStatus() != Response.Status.OK.getStatusCode()) {
+                LRALogger.i18NLogger.error_getAllLRAs(response.getStatus());
+                throw new WebApplicationException(response);
+            }
+
+            return response.readEntity(new GenericType<List<LRAData>>() {});
+        } finally {
+            if (client != null) {
+                client.close();
```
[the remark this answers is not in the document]
xstefank: @mmusgrov sorry, but I would prefer to keep this for now and really investigate the TCK hangs with Client caching in new JBTM. Would that be ok with you?

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
```diff
@@ -358,16 +383,25 @@ public URI joinLRA(URI lraId, Long timelimit,
...
+        Response response = null;
 
-        response = getTarget().path(String.format(leaveFormat, getLRAId(lraId.toString())))
+        try {
+            client = ClientBuilder.newClient();
+
+            response = client.target(base)
+                .path(String.format(LEAVE_PATH, getLRAId(lraId.toString())))
                 .request()
-                .header(LRA_HTTP_CONTEXT_HEADER, lraId)
```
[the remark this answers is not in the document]
xstefank: The header is not consumed on the coordinator, the LRA id is parsed from the path directly.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
xstefank: +1

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
xstefank: The other exceptions come from the Future.get call.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
xstefank: here the WebApplicationException is caught in the previous catch block.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
xstefank: ok

rts/lra/lra-coordinator/pom.xml
xstefank: no, this is used for getting JAX-RS client in the coordinator.

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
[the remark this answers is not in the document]
mmusgrov: You did half of what I asked (getClient()) but not the other half. You either need both changes or neither. The reason I asked to get the client via a method is so that we can cache it in the future. The finally block needs to call a method so that any future strategy for caching the connection will be in one place.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
mmusgrov: Yes I know but the WebApplicationException will include the jax-rs error even for the Future.get call.

xstefank: not sure what you are proposing. The `e.getMessage()` will contain all the info from WebApplicationException in case. Or do you propose to throw WebApplicationException here?

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
mmusgrov: The WebTarget, Invocation.Builder, AsyncInvoker and get on Future<Response> methods we are using all ultimately return a Response object to the caller so you don't need to catch checked exceptions.

pom.xml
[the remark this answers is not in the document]
mmusgrov: @xstefank you have not addressed this comment yet. Also if no other module needs a property then it would need to go in the appropriate pom (this one is the top level narayana pom).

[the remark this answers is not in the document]
xstefank: ok, I'll move the version. This is is both with `resteasy-json-binding-provider` required to make lra-client work with resteasy in Quarkus and TT because they do package jackson MessageBodyReaders and MessageBodyWriters in a separate extensions or in the case of TT in a separate fractions. I thought that this is unnecessary if we can provide the reader/writer manually.

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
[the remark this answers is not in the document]
mmusgrov: Which hangs? I am talking about opening and closing clients using a single method. The reason is as follows:

The docs (https://docs.oracle.com/javaee/7/api/javax/ws/rs/client/package-summary.html) say "... Client instances - configurable, heavy-weight objects that manage the underlying communication infrastructure ...".

That indicates to me that we need to be configuring a fixed minimal set of clients and we then share them thus avoiding the set up and tear down costs. If we put the relevant logic behind a method call then we can more easily decide how we want to do that. The original code (that this PR is reinstating) was already doing that (see pr/1515).

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
mmusgrov: My point is that the original code was fine as it was.

pom.xml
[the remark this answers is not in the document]
mmusgrov: Wait. Is this a runtime dependency? I will try running the test suite on your PR ...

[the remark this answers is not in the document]
xstefank: yes, that is runtime. But it won't be affecting the test suite. Only if you add lra-client dependency to the newly generated Quarkus or TT application.

rts/lra/lra-client/pom.xml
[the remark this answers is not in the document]
mmusgrov: Maybe it's a runtime dependency so I am resolving this conversation.

rts/lra/lra-client/pom.xml
```diff
@@ -57,6 +57,16 @@
...
+        <version>${version.org.jboss.resteasy}</version>
+      </dependency>
+      <dependency>
+        <groupId>commons-io</groupId>
+        <artifactId>commons-io</artifactId>
```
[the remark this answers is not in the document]
mmusgrov: Maybe it's a runtime dependency so I am resolving this conversation.

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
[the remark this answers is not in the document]
mmusgrov: The change I am requesting is a structural change as opposed to a functional one so I am resolving this conversation.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
mmusgrov: I will leave this thread unresolved since I still argue that the original code was fine.

rts/lra/lra-coordinator-jar/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
mmusgrov: I will leave this thread unresolved since you don't need to catch checked exceptions.

rts/lra/lra-coordinator/pom.xml
mmusgrov: Maybe this is a runtime dependency (although not marked as such) so I will resolve this conversation.

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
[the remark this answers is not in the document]
mmusgrov: Since my comments do not effect the functionality I am resolving this conversation.

rts/lra/lra-client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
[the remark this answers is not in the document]
mmusgrov: Well spotted.

mmusgrov (approved):

PR #1688, opened 2020-10-09
[JBTM-3372] Add GH Actions workflow for LRA testing

xstefank:
https://issues.redhat.com/browse/JBTM-3372

!MAIN !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF !LRA NO_WIN DB_TESTS !mysql !db2 !postgres !oracle

Examples:
https://github.com/xstefank/narayana/pull/3
https://github.com/xstefank/narayana/pull/2

Note that this action can't run the tests that depend on the running WFLY. If that is a requirement we need to setup WFLY download in the lra-coordinator-jar module.

### DIFF
--- a/.github/workflows/lra-test.yml
+++ b/.github/workflows/lra-test.yml
@@ -1,0 +1,46 @@
+name: LRA
+
+on:
+  push:
+    branches: [ master ]
+    paths:
+      - "rts/lra/**"
+  pull_request:
+    branches: [ master ]
+    paths:
+      - "rts/lra/**"
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  lra-test:
+    name: LRA test with JDK ${{ matrix.java-version }}
+    timeout-minutes: 120
+    strategy:
+      matrix:
+        java-version: [ 8, 11 ]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-java@v1
+        name: Set up JDK ${{ matrix.java-version }}
+        with:
+          java-version: ${{ matrix.java-version }}
+
+      - uses: actions/cache@v2
+        name: Cache local Maven repository
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+
+      - name: LRA test
+        run: |
+          git config --global user.email "you@example.com"
+          git config --global user.name "Your Name"
+          WORKSPACE=$PWD PROFILE=LRA ./scripts/hudson/narayana.sh
+        env:
+          LRA_TCK_TIMEOUT_FACTOR: 1
+


.github/workflows/lra-test.yml
```diff
@@ -0,0 +1,37 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    - name: Narayana - build
+      run: mvn clean install -DskipTests
```
ochaloup, line 28: maybe `mvn clean install -DskipTests -Pcommunity` and you can skip te second `mvn clean install` down below here

ochaloup (approved):
I like this. Thanks for the looking into this! I have one comment on the way how to run compilation but otherwise fine.

It's good idea to apply this and check the results against our current CI. The WFLY tests are not critical now and we can add the WildFly runs as "a new axis" when the https://issues.redhat.com/browse/JBTM-3343 will be finished.

.github/workflows/lra-test.yml
```diff
@@ -0,0 +1,33 @@
...
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    - name: Narayana - build
```
tomjenkinson, line 27: I think we should change this and the subsequent lines to be something like
`PROFILE=LRA ./scripts/hudson/narayana.sh`

But also, we need to rebase the PR first:
`./scripts/hudson/narayana-rebase.sh`

tomjenkinson (dismissed):

.github/workflows/lra-test.yml
xstefank: @tomjenkinson without clean rebase the GH actions won't run. Please check the example at https://github.com/xstefank/narayana/pull/5. So I don't think narayana-rebase.sh is necessary.

About `PROFILE=LRA ./scripts/hudson/narayana.sh` this job was created precisely to test in different environment with different settings so for instance avoiding `-Dlra.tck.timeout.factor=1.5` from this script. Also GH actions will set PR status directly on PR so no need for comments. It is easier to keep only the command for now. If we are to move the whole CI from Hudson here then it will make sense to reuse the scripts with adjustments. But now it would be just overhead. Sorry, if I'm missing something in the scripts. In that case please point me to the correct configuration and I'll change it in the PR. But now I would like to avoid setting a bunch of environment properties to run a single script that runs essentially single command.

tomjenkinson: Do I understand that this will rebase before running the testing somehow because of GH Actions?

The narayana-rebase.sh actually does a rebase so we can be sure the PR is running with the latest changes.

If that is the case then we don't need the narayana-rebase.sh

.github/workflows/lra-test.yml
```diff
@@ -0,0 +1,33 @@
...
+      run: ./build.sh clean install -Pcommunity -DskipTests -Didlj-enabled=true
+    - name: LRA - test (tck/basic)
+      run: |
+        cd rts/lra/lra-test
```
tomjenkinson, line 31: It seems this is not something in the current narayana.sh? https://github.com/jbosstm/narayana/blob/master/scripts/hudson/narayana.sh

.github/workflows/lra-test.yml
```diff
@@ -0,0 +1,33 @@
...
+        mvn clean install -Parq
```
tomjenkinson, line 32: If you aren't going to use narayana.sh, please can you use the maven in the Narayana repo though still?

I am not sure though whether moving away from narayana.sh is good to be able to keep the GH action and the existing CI running in parallel with the same testing /cc @ochaloup

.github/workflows/lra-test.yml
xstefank: It won't run the actions before you solve all the conflicts. So it basically forces you to rebase. It's true that clean rebase without conflicts could help with many merges in short periods but that will rarely happen I think and we will still be running hudson jobs this is just parallel testing for the timing issues. It won't be easy to set GH actions according to your hudson setup. This would be just a start and if there is a requirement to migrate everything, including scripts, we can do that later in a separate JBTM. 

The scripts require some environment properties that aren't configurable per run in GH actions.

.github/workflows/lra-test.yml
xstefank: There isn't a possibility to run the full test suite. The issue is with the tests in lra-coordinator-jar module because they require full WFLY server which is not downloaded included during the builds. As mentioned above this is just a quick setup mainly for TCK (but also includes basic test suite) so we can catch timing issues in the Hudson CI. We will build on top of this if necessary later. I am not sure if Narayana team would want to migrate all testing to GH actions.

.github/workflows/lra-test.yml
xstefank: I can reuse narayana.sh but that would be only running a single line at https://github.com/jbosstm/narayana/blob/master/scripts/hudson/narayana.sh#L483 if I am not missing anything. This would just require a bunch of env properties to be set in this yaml directly. Do you want me to change to that? Also it would need to branch the command because we don't want timeout adjustment that is included in the Hudson CI.

.github/workflows/lra-test.yml
tomjenkinson: The automatica rebasing of PRs has a purpose in that if an unrelated defect is fixed and merged, it doesn't require the PR author to rebase manually and the project can just re-run the PR.

.github/workflows/lra-test.yml
tomjenkinson: I see, I am not sure if just running the TCK would give that much information. It would be better if the whole of the LRA testing could take place to replicate the existing LRA `PROFILE` in narayana.sh if possible

.github/workflows/lra-test.yml
xstefank: GH actions won't run if the PR is not updated. So it wouldn't have a chance to rebase automatically.

.github/workflows/lra-test.yml
tomjenkinson: In this branch of the discussion I am discussing if we _weren't_ to use narayana.sh. So in this case it would be something like changing the line `mvn clean install -Parg` to (I guess, based on the path): `../../../build.sh clean install -Parq` (or remove the `cd` and go with `./build.sh -f rts/lra/lra-test/pom.xml clean install -Parq`

It's not very clear to me where the `mvn` is coming from in the command as it stands, maybe it's part of `ubuntu-latest`? I think for reproducibility though we should be using the Maven in narayana: https://github.com/jbosstm/narayana/tree/master/tools/maven by calling https://github.com/jbosstm/narayana/blob/master/build.sh

Does that help clarify?

.github/workflows/lra-test.yml
tomjenkinson: I have edited my comment which may have made it clearer what I was getting at before but if not then I wasn't thinking the job would ever run automatically, I more meant that someone from the project team can just re-rerun it. But yes, I understand from you now that it might not be possible to re-run tests when using GitHub actions until there is an update in the code (unless GitHub actions can respond to comments from maintainers maybe?)

.github/workflows/lra-test.yml
xstefank: this won't work with TT I believe as it's documented in https://github.com/jbosstm/narayana/blob/master/scripts/hudson/narayana.sh#L480.

.github/workflows/lra-test.yml
```diff
@@ -0,0 +1,38 @@
+name: LRA test
+
+on:
+  push:
+    branches: [ master ]
+    paths:
+      - "rts/lra/**"
+  pull_request:
```
tomjenkinson, line 8: @ochaloup I think we discussed to not have this on PR and just have it run on merged code for now. Is that your understanding?

Otherwise LGTM - thanks @xstefank

ochaloup (approved):

mmusgrov:
So we are potentially going to be running 3 CI systems in parallel. Is this experimental and if so what is the end goal?

LRA failures are easy to diagnose. More difficult is the AS_TESTS axis, can that be moved? Why are we only doing this for LRA?
￼
When does the GH action run?

Who is going to be monitoring the status of the GH Action?

Is there a risk of choosing the CI system that makes our code look best?

mmusgrov:
I have removed the `Hold` label because it would be useful to compare running the tests on github versus running them on the internal CI test infrastructure. To answer my own question posed in my last comment:

Q1 When does the GH action run?
A1 on pull requests and on demand (@xstefank can we run GH Action without a PR?)

Q2 Who is going to be monitoring the status of the GH Action?
A2 I will monitor it and discuss it during our weekly meetings if neccessary

Q3 Is there a risk of choosing the CI system that makes our code look best?
A3 Possible, but initially we should run them in parallel to get a feel for the false positive/negative frequency

mmusgrov (approved):
I also added a comment above

xstefank:
> Q1 When does the GH action run?
> A1 on pull requests and on demand (@xstefank can we run GH Action without a PR?)

@mmusgrov yes, it is possible, but not configured in the PR - https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/. Should I add this option?

> Q2 Who is going to be monitoring the status of the GH Action?
> A2 I will monitor it and discuss it during our weekly meetings if necessary

Also, there will be a status check pushed to each commit for which the action is running so you'll see X or check mark next to each LRA PR.

> Q3 Is there a risk of choosing the CI system that makes our code look best?
> A3 Possible, but initially we should run them in parallel to get a feel for the false positive/negative frequency
Agreed.

.github/workflows/lra-test.yml
```diff
@@ -0,0 +1,40 @@
...
+      uses: actions/cache@v2
+      with:
+        path: ~/.m2/repository
+        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+        restore-keys: |
+          ${{ runner.os }}-maven-
+    - name: LRA test
+      run: |
+        git config --global user.name "jbosstm"
+        git config --global user.email "jbosstm@users.noreply.github.com"
```
tomjenkinson, line 36: @xstefank - how did you decide on this value (and the user.name above)? Is it (and the line above) required to set this for the action to work

xstefank: I believe this was required for the script. I'll try to remove it.

xstefank: yes, correct - https://github.com/xstefank/narayana/runs/2031052160

tomjenkinson: Thanks @xstefank - I am erring on thinking that as the details are fictional, how about going with what it looks like the git output comes back with:
`git config --global user.email "you@example.com"
  git config --global user.name "Your Name"`

/cc @mmusgrov do you have any preference here?

xstefank: It's irrelevant as we are not pushing anything. This is used in the script for pulling only I believe.

tomjenkinson: Sure, but by it not being the default (can a default really be established but the output from the command seems possible to consider as such) could it be perceived as not irrelevant and somehow meaningful?

I don't think this should block the PR as it could be changed later I suppose, but it's a question.

xstefank: Personally, I think jbosstm is better than the example the git outputs by default :) by that are my 2c. Surely, I can change it, it's not a big issue for me. I will just run it in my branch to verify it works properly.

mmusgrov: I tend to agree with tom (jbosstm@users.noreply.github.com implies a real email but you@example.com does not). But this job is for internal use so the change is only a "nice to have" item and need not block merging the PR.

It looks like we are all in agreement to merge the PR so provided we don't need to do anything for this last item I think we should merge and begin acquiring some data about whether the PR improves upon the rate of false positive we are seeing with our internal CI job.

xstefank: @mmusgrov please go ahead, I will open another PR when the job passes later.

ochaloup:
I think we can merge this now. If there are some requirements for further changes we can adjust in the follow-up PR>

PR #1693, opened 2020-10-13
[SCRIPTS_ONLY] Add LRA_TCK_TIMEOUT_FACTOR env property to LRA tests

xstefank:
!MAIN !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF !LRA NO_WIN DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/scripts/hudson/narayana.sh
+++ b/scripts/hudson/narayana.sh
@@ -480,7 +480,7 @@ function lra_tests {
   # we can't use 'mvn -f' option beacuse of Thorntail plugin issue THORN-2049
   cd ./rts/lra/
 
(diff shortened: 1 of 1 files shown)


mmusgrov (changes requested):
I need either a JBTM or the PR needs to describe the motivation for the change.

ochaloup:
@mmusgrov this is only a CI script. This change makes possible to configure job from outside of the script itself. It does not break anything and makes the things just more configurable. We usually do not create a issue for such types of changes, don't we?

ochaloup:
@mmusgrov I'm sorry for misreading. There is no direct request for JBTM but for elaboration on the reason for this PR. Agree, that's a good point. And I assume @xstefank added it.

mmusgrov (approved):

PR #1696, opened 2020-10-14
[SCRIPTS ONLY] using GitHub token instead of user/password on commenting PRs 4.17

ochaloup:
MAIN WIN
!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF !RTS !AS_TESTS !TOMCAT !JACOCO !LRA


ochaloup:
It seems it works fine.

PR #1702, opened 2020-10-22
[JBTM-3376] clean CMRIntegrationTest to be more transparent and get more info when it fails

ochaloup:
https://issues.redhat.com/browse/JBTM-3376

AS_TESTS

### DIFF
--- a/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/integration/CMRIntegrationTest.java
+++ b/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/integration/CMRIntegrationTest.java
@@ -40,9 +40,9 @@ import javax.transaction.SystemException;
 import javax.transaction.TransactionManager;
 import javax.transaction.UserTransaction;
 
-import org.h2.jdbcx.JdbcDataSource;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.logging.Logger;
 import org.jboss.shrinkwrap.api.ArchivePaths;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.asset.EmptyAsset;
@@ -51,7 +51,6 @@ import org.jboss.shrinkwrap.api.spec.JavaArchive;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
-import com.arjuna.ats.arjuna.exceptions.ObjectStoreException;
 import com.arjuna.ats.arjuna.recovery.RecoveryManager;
 import com.arjuna.ats.arjuna.recovery.RecoveryModule;
 import com.arjuna.ats.internal.jta.recovery.arjunacore.CommitMarkableResourceRecordRecoveryModule;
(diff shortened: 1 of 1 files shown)


ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/integration/CMRIntegrationTest.java
```diff
@@ -169,60 +167,40 @@ public void run() {
...
+							log.errorf(e, "Error while invoking insertion to a database table with CMR resource");
 							SQLException nextException = e.getNextException();
 							while (nextException != null) {
-								nextException.printStackTrace();
-								nextException = nextException
-										.getNextException();
-							}
-							Throwable[] suppressed = e.getSuppressed();
-							for (int j = 0; j < suppressed.length; j++) {
-								suppressed[j].printStackTrace();
```
mmusgrov, line 201: Why aren't we printing the stack trace for suppressed exceptions?

ochaloup: That's done by logger automatically. It prints nicely cause exceptions and the suppressed exceptions without this manual handling. The logger is placed clearly with the level of severity and indentation under the `standalone.log` (easier to read, easier to understand the cause chain with the suppressed exception printed in the hierarchy).
That's all the reasons for these changes.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/integration/CMRIntegrationTest.java
```diff
@@ -169,60 +167,40 @@ public void run() {
...
+								nextException = nextException.getNextException();
 							}
 							try {
 								userTransaction.rollback();
 							} catch (IllegalStateException | SecurityException
 									| SystemException e1) {
-								e1.printStackTrace();
-								fail("Problem with transaction");
+								log.error("Problem with transaction", e1);
+								throw new IllegalStateException("Problem with transaction", e1);
```
mmusgrov, line 190: I would prefer that we explicitly fail here (also there are other places where we throw instead of fail). If a test throws an exception I interpret that as meaning that the test needs fixing unless the test is marked as such, eg @Test (expected = IllegalStateException.class).

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/integration/CMRIntegrationTest.java
```diff
@@ -169,60 +167,40 @@ public void run() {
...
-							// System.out
-							// .printf("committed txn iteration %d%n", i);
 							success++;
 						} catch (SQLException e) {
-							System.err.println("boom");
-							e.printStackTrace();
-							if (e.getCause() != null) {
-								e.getCause().printStackTrace();
-							}
+							log.errorf(e, "Error while invoking insertion to a database table with CMR resource");
```
mmusgrov, line 179: This is much more friendly than boom :-)

mmusgrov (changes requested):

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/integration/CMRIntegrationTest.java
ochaloup: Sure. Fixed.

mmusgrov (approved):

PR #1706, opened 2020-10-28
JBTM-3361 include JTS format id check

mmusgrov:
JBTM-3361

We need this mod to fix the [JTS recovery quickstart](https://github.com/jbosstm/quickstart/tree/master/ArjunaJTS/recovery/src/main/java).

MAIN !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF !LRA !NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/RecoveryXids.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/RecoveryXids.java
@@ -42,6 +42,7 @@ import javax.transaction.xa.XAResource;
 import javax.transaction.xa.Xid;
 
 import com.arjuna.ats.arjuna.logging.tsLogger;
+import com.arjuna.ats.internal.arjuna.FormatConstants;
 import com.arjuna.ats.jta.common.jtaPropertyManager;
 import com.arjuna.ats.jta.xa.XATxConverter;
 import com.arjuna.ats.jta.xa.XidImple;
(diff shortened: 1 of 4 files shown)


jhalliday:
Hi Mike

Sorry, but I think you're barking up the wrong tree here, probably because I did a lousy job of documenting why updateIfEquivalentRM now works the way it does...

We can replace (consider equivalent) one resource with another in two cases:

- It's got the same [non null] name (because we're optimistic enough to trust nobody is going to configure two different RMs with the same name.... this one may come back to bite us) OR the driver says isSameRM==true (because we trust the driver vendors too. hahaha.)  This is a strong i.e. trustworthy match, at least relatively speaking. Neither of these holds for the JTS recovery quickstart, as the resources there have not been retrofitted with a wrapper to supply a name. So...

- If strong matching fails, we apply a heuristic based on checking if the same xid appears in the recover scans of both resources.  BUT... this works only if bqual values are unique. If they are, then an xid is unique, so it can appear in only one RM. If it appears in scans from two XAResources, then they are attached to the same RM, i.e. equivalent. However...

JCA transaction inflow can't always create new bqual values. Infact, the ONLY case where it can is for (formatId == XATxConverter.FORMAT_ID)  There is only one version of the subordinate TransactionImple, shared by JTA and JTS mode. As a result, this rule applies to both JTA and JTS, it doesn't change depending on the configured mode, since XATxConverter.FORMAT_ID ALWAYS is FormatConstants.JTA_FORMAT_ID regardless.

Therefore, JTA and ONLY JTA xids can be guaranteed unique. All others, including our own other types (JTS, bridge, etc) can potentially be non-unique and hence appear in more than one RM. This is the root of the ongoing 02751825 problem. Because of this, it's incorrect to use the JTS format id when determining RM equivalence, even if you're running in JTS mode.

So why is the quickstart failing after my change to those rules?

It's a timing issue. The recovery isn't quite as top-down as it first appears, because in JTS the resources upcall over the orb back to the (recovery) coordinator to request replayPhase2, which then downcalls back into them to invoke commit/rollback.  They have to do that before the coordinator gives up and retires the parent transaction, or there will be nothing to call up to. Which means... you need a shorter orphan safety interval, because it's that check in RecoveryXids.toRecover that's causing the problem, not the resource matching.

I suggest you patch the quickstart with jtaPropertyManager.getJTAEnvironmentBean().setOrphanSafetyInterval(anythingShorterThanThePeriodicRecoveryPeriod) and rerun. If it passes, ditch this pr as redundant.

Jonathan

mmusgrov:
This PR is superseded by the quickstart PR: https://github.com/jbosstm/quickstart/pull/299
Closing due PR since it's redundant as per Jonathan's analysis.

PR #1707, opened 2020-10-29
[JBTM-3381] Enhance LRA Client for Quarkus

ruromero:
Hey,

I wanted to do a small contribution to the client after starting to use it. I hope you find my changes useful and relevant.

* Clean up unnecessary dependencies and reformat pom.
* Set PORT and PATH_KEY as constants. I couldn't import them properly in my code
* Fix the clientID param that was sending an object string representation.
* joinLRA with a null timelimit was throwing a NPE during the unboxing.
* Other SonarLint suggestions.
* The service provided has been replaced by the Microprofile implementation.

I didn't consider meaningful creating a Jira for this but if you feel it's a requirement even for such a small PR I'll do it.
Let me know if I didn't take something into account, specially regarding the pom.xml.

Thank you in advance.

Signed-off-by: ruromero <rromerom@redhat.com>

Fix https://issues.redhat.com/browse/JBTM-3381

LRA

### DIFF
--- a/rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
+++ b/rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
@@ -106,13 +106,13 @@ public class NarayanaLRAClient implements Closeable {
      * Key for looking up the config property that specifies which port a
      * coordinator is running on
      */
-    public static String LRA_COORDINATOR_PORT_KEY = "lra.http.port";
+    public static final String LRA_COORDINATOR_PORT_KEY = "lra.http.port";
 
     /**
      * Key for looking up the config property that specifies which JAX-RS path a
      * coordinator is running on
      */
-    public static String LRA_COORDINATOR_PATH_KEY = "lra.coordinator.path";
+    public static final String LRA_COORDINATOR_PATH_KEY = "lra.coordinator.path";
 
     public static final long DEFAULT_TIMEOUT_MILLIS = 0L;
 
@@ -179,7 +179,7 @@ public class NarayanaLRAClient implements Closeable {
      *
      * @param coordinatorUri  uri of the lra coordinator
      */
-    public NarayanaLRAClient(URI coordinatorUri) throws MalformedURLException, URISyntaxException {
+    public NarayanaLRAClient(URI coordinatorUri) {
         init(coordinatorUri);
     }
 
@@ -305,7 +305,7 @@ public class NarayanaLRAClient implements Closeable {
 
             response = client.target(base)
                 .path(START_PATH)
-                .queryParam(CLIENT_ID_PARAM_NAME, client)
+                .queryParam(CLIENT_ID_PARAM_NAME, clientID)
                 .queryParam(TIMELIMIT_PARAM_NAME, Duration.of(timeout, unit).toMillis())
                 .queryParam(PARENT_LRA_PARAM_NAME, encodedParentLRA)
                 .request()
@@ -320,13 +320,6 @@ public class NarayanaLRAClient implements Closeable {
             }
 
             lra = URI.create(response.getHeaderString(HttpHeaders.LOCATION));
-
-            if (lra == null) {
-                LRALogger.i18NLogger.error_nullLraOnCreation(response.toString());
-                throwGenericLRAException(null, INTERNAL_SERVER_ERROR.getStatusCode(), "LRA creation is null");
-                return null;
-            }
-
             lraTrace(lra, "startLRA returned");
 
             Current.push(lra);
@@ -367,7 +360,7 @@ public class NarayanaLRAClient implements Closeable {
 
     /**
      * @param lraId the URI of the LRA to join
-     * @param timelimit how long the participant is prepared to wait for LRA completion
+     * @param timeLimit how long the participant is prepared to wait for LRA completion
      * @param compensateUri URI for compensation notifications
      * @param completeUri URI for completion notifications
      * @param forgetUri URI for forget callback
@@ -377,10 +370,10 @@ public class NarayanaLRAClient implements Closeable {
      * @return a recovery URL for this enlistment
      * @throws WebApplicationException if the LRA coordinator failed to enlist the participant
      */
-    public URI joinLRA(URI lraId, Long timelimit,
+    public URI joinLRA(URI lraId, Long timeLimit,
                        URI compensateUri, URI completeUri, URI forgetUri, URI leaveUri, URI afterUri, URI statusUri,
                        String compensatorData) throws WebApplicationException {
-        return enlistCompensator(lraId, timelimit, "",
+        return enlistCompensator(lraId, timeLimit, "",
                 compensateUri, completeUri,
                 forgetUri, leaveUri, afterUri, statusUri,
                 compensatorData);
@@ -388,7 +381,7 @@ public class NarayanaLRAClient implements Closeable {
 
     public void leaveLRA(URI lraId, String body) throws WebApplicationException {
         Client client = null;
-        Response response = null;
+        Response response;
 
         try {
             client = getClient();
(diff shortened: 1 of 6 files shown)


ochaloup:
TESTIT

mmusgrov:
@ruromero Thanks, the changes look good except for some procedural tasks:

- if you haven't already signed [our CLA](url) please will you do that;
- changes must be relevant to the issue being fixed so we don't normally accept code formatting changes;
- we also prefer that any PR which updates the code has a corresponding JIRA issue (this helps with tracking changes and for generating release notes);

rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
```diff
@@ -305,7 +306,7 @@ public URI startLRA(URI parentLRA, String clientID, Long timeout, ChronoUnit uni
 
             response = client.target(base)
                 .path(START_PATH)
-                .queryParam(CLIENT_ID_PARAM_NAME, client)
+                .queryParam(CLIENT_ID_PARAM_NAME, clientID)
```
ochaloup, line 309: This is a nice spot. I think the `clientID` would be better encoded (`URLEncoder.encode`).

rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
```diff
@@ -320,13 +321,6 @@ public URI startLRA(URI parentLRA, String clientID, Long timeout, ChronoUnit uni
             }
 
             lra = URI.create(response.getHeaderString(HttpHeaders.LOCATION));
-
-            if (lra == null) {
-                LRALogger.i18NLogger.error_nullLraOnCreation(response.toString());
```
ochaloup, line 325: Please with this remove the method `error_nullLraOnCreation` from the `i18NLogger` as well.

rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
```diff
@@ -641,7 +635,7 @@ private static StringBuilder makeLink(StringBuilder b, String uriPrefix, String
         return b.append(link);
     }
 
-    private URI enlistCompensator(URI lraUri, long timelimit, String uriPrefix,
+    private URI enlistCompensator(URI lraUri, Long timelimit, String uriPrefix,
```
ochaloup, line 638: An idea: what about rather change the `joinLRA` method to be using `long` instead of `Long`?
(as we are not in 1.0 we can dare to change the public api if agreed it's desirable)

rts/lra/client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
```diff
@@ -1 +1 @@
-org.jboss.resteasy.client.jaxrs.ResteasyClientBuilder
+org.jboss.resteasy.microprofile.client.impl.MpClientBuilderImpl
```
ochaloup, line 1: With this change we should be consistent in whole LRA impl. The same change should be done in https://github.com/jbosstm/narayana/blob/master/rts/lra/jaxrs/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder.
I think especially this particular change would be nice to be tracked as a jira.
FYI @xstefank

ochaloup (changes requested):
I have few minor points about the code changes. Otherwise it's nice. Thanks.

rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
ruromero: Well, I tend to prefer the `Long` as it allows the user to send `null` which is usually the _I don't care_ equivalent other than having to know beforehand that `0L` means wait indefinitely.

rts/lra/client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
ruromero: If you agree on the change I will create the JIRA for this and move it to a specific PR.

ruromero:
@mmusgrov

* The link you provided to the CLA doesn't seem to work. I have also tried to follow cla.jboss.org but I'm being redirected to a RH site.
* I will remove code formatting specific changes.
* Finally, as I was submitting changes so different that I found during the LRA client integration into my project, I didn't know what JIRA to create because first I wanted to know if they made sense at all. Please, tell me what JIRA or JIRAs should I create for each specific change.

Thanks.

ruromero:
@ochaloup @mmusgrov I have created the Jira and updated the PR.
I have also removed a dependency that was preventing me from running the client in a Quarkus native app.

rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
mmusgrov: If you're going to encode it then use UTF-8:

> URLEncoder.encode(clientID, "UTF-8");

rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
mmusgrov: I don't care could be interpreted to mean use a random timeout ;-)
BTW this is a private method so do we really care, the public API is already using the Long datatype:

> public URI joinLRA(URI lraId, Long timelimit, ...

ochaloup (approved):
Seems fine to me. Thanks.

mmusgrov:
> @mmusgrov
> 
> * The link you provided to the CLA doesn't seem to work. I have also tried to follow cla.jboss.org but I'm being redirected to a RH site.

The revised process for contributions is to include "Signed-off-by:" followed by the contributor’s name and email address in the commit message:   

    Signed-off-by: Your Name <yourname@example.com>

With git, this step can be automated using git commit -s or git commit --signoff.

I will update the automated text generated in new PRs to include this information.  

Thanks for bringing this policy change to our attention.

rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
ruromero: it's done

rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
```diff
@@ -305,7 +305,7 @@ public URI startLRA(URI parentLRA, String clientID, Long timeout, ChronoUnit uni
 
             response = client.target(base)
                 .path(START_PATH)
-                .queryParam(CLIENT_ID_PARAM_NAME, client)
+                .queryParam(CLIENT_ID_PARAM_NAME, URLEncoder.encode(clientID, "UTF-8"))
```
mmusgrov, line 308: Is it wise to encode the clientID? A couple of points:
- it is not necessarily a URL since client ids are arbitrary strings and it may be confusing to the client to receive back something other that what he set on the start call;
- the json data returned by the API call for [returning LRAs](https://github.com/jbosstm/narayana/blob/master/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java#L115 ) and the [format of the returned data](https://github.com/jbosstm/narayana/blob/master/rts/lra/service-base/src/main/java/io/narayana/lra/LRAData.java#L42) doesn't say anything about us encoding the client id.

mmusgrov (changes requested):
There is just my issue about encoding the clientID remaining.

rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
ochaloup: Encoding of the `clientID` was my idea. I was afraid that the URL escaped characters won't be correctly transfered and interpreted. I was wrong as I can see from testing the functionality.
Agree that encoding is not necessary, sorry for misguidance @ruromero

ruromero: No worries, I have reverted the change.

ruromero:
Done @mmusgrov

mmusgrov (approved):

PR #1710, opened 2020-10-29
[JBTM-3379] make participant timeout configurable

ochaloup:
https://issues.redhat.com/browse/JBTM-3379

LRA

### DIFF
--- a/rts/lra/service-base/src/main/java/io/narayana/lra/LRAConstants.java
+++ b/rts/lra/service-base/src/main/java/io/narayana/lra/LRAConstants.java
@@ -43,7 +43,13 @@ public final class LRAConstants {
     public static final String RECOVERY_PARAM = "recoveryCount";
     public static final String HTTP_METHOD_NAME = "method"; // the name of the HTTP method used to invoke participants
 
-    public static final long PARTICIPANT_TIMEOUT = 2; // number of seconds to wait for requests
+    /**
+     * System property name which defines the connection read timeout which is used
+     * by LRA Coordinator and its friends when connecting remote participants and other coordinators.
+     * The value is in seconds. The default value is {@code 2}.
+     */
+    public static final String PARTICIPANT_TIMEOUT_PARAM = "narayana.lra.participant.timeout";
+    public static final long PARTICIPANT_TIMEOUT = Long.getLong(PARTICIPANT_TIMEOUT_PARAM, 2);
 
     private static final Pattern UID_REGEXP_EXTRACT_MATCHER = Pattern.compile(".*/([^/?]+).*");
 


rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
```diff
@@ -424,7 +424,9 @@ private boolean afterLRARequest(URI target, String payload) {
         Client client = null;
 
         try {
-            client = ClientBuilder.newClient();
+            client = ClientBuilder.newBuilder()
+                    .readTimeout(PARTICIPANT_TIMEOUT, TimeUnit.SECONDS)
```
mmusgrov, line 428: This method was added in a later revision of the JAX-RS spec (and was the reason I chose async). Will you research whether or not we need to support any of the earlier versions?

rts/lra/service-base/src/main/java/io/narayana/lra/LRAConstants.java
```diff
@@ -42,7 +42,14 @@
     public static final String RECOVERY_PARAM = "recoveryCount";
     public static final String HTTP_METHOD_NAME = "method"; // the name of the HTTP method used to invoke participants
 
-    public static final long PARTICIPANT_TIMEOUT = 2; // number of seconds to wait for requests
+    /**
+     * System property name which defines the connection read timeout which is used
+     * by LRA Coordinator and its friends when connecting remote participants and other coordinators.
+     * The value is in seconds. The default value is {@value LRAConstants#PARTICIPANT_DEFAULT_TIMEOUT_S}.
+     */
+    public static final String PARTICIPANT_TIMEOUT_PARAM = "lra.participation.timeout";
```
mmusgrov, line 50: Should we prefix these properties with `narayana` to avoid any potential conflicts with future LRA spec versions.

Also participants/compensators are referred to as participants or compensators so it would be more consistent to call this property `lra.participant.timeout`

rts/lra/service-base/src/main/java/io/narayana/lra/LRAConstants.java
```diff
@@ -42,7 +42,14 @@
...
+    private static final long PARTICIPANT_DEFAULT_TIMEOUT_S = 2; // number of seconds to wait for requests
```
mmusgrov, line 51: It seems odd to declare a private constant that is only used once and then only on the next line. The less code in an application the easier it is to understand and maintain.

mmusgrov (changes requested):

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
ochaloup: Hm... I thought that we are fine to use the JAX-RS 2.1 when we defined it in dependencies.
Regarding the need of  support of the earlier versions. I can't see any need for the support or earlier versions. The Quarkus uses the 2.1 from start, for WildFly there will be for sure no support with older versions (at least for the reason that for integration there will be created an extension and WildFly supports extensions only in new versions).

ochaloup (commented):
I changed the code as suggested.

xstefank (approved):
To me, this is a good change. Thanks.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRARecord.java
mmusgrov: Why would we want to restrict our user base without any significant discernible benefit?

mmusgrov: So if you can indicate why the benefit of making this change outweighs the drawback of restricting our user base then I'll mark it as approved.

mmusgrov: Also bear in mind that I want to move to a reactive solution in a later revision so your changes would ultimately be rolled back.

mmusgrov: Also since the config property and the implementation technique of using async calls are independent of each I would recommend that you split up the issue to avoid one blocking the other.

ochaloup:
@mmusgrov I removed the part under discussion about use of the `async` call or `readTimeout` from the JAX-RS 2.1. Now it's a simple change to parametrize the client timeout.

xstefank (approved):
Can we consider moving all configuration to MP config in the future?

mmusgrov (approved):

ochaloup:
@xstefank it's something what I was thinking about but I'm not sure if we want to force suer to include the MP config when he uses the LRA client. From easy to use perspective it would be nice.

I'm going to merge this as an enhancement of the current state for now.

mmusgrov:
@ochaloup the extra changes you included in your original PR (to use JAX-RS 2.1) threw me off this one.
The participant timeout should not be configurable, I only added it to avoid unavailable participants from reducing availability - the protocol is robust enough to retry later. May I back out your change?

ochaloup:
@mmusgrov ok, I see, good. I haven't realized that absolutely. Give me some time (the PR will be ready today) to revert it. The only addition which I would be happy about would be a short comment to the constant if you don't mind.

PR #1712, opened 2020-10-30
[SCRIPTS ONLY] adding a shell script to download WFLY

ochaloup:
and provide a way to force not building JBoss AS when WFLY is downloaded

This is no change for the currently running CI matrix but it provides a helper functions and a way to skip the jboss as build when the CI job defines it.
Not building the AS may make pretty faster the axes like RTS, XTS, LRA.

AS_TESTS

### DIFF
--- a/scripts/hudson/download-wildfly.sh
+++ b/scripts/hudson/download-wildfly.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+#  functions to download the WildFly and to setup the JBOSS_HOME directory
+
+function download_wildfly_nightly_build() {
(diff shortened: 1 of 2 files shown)


mmusgrov (commented):
Who calls the new functions?

ochaloup:
@mmusgrov it may be called from the job which wants to use them. They are not directly called from the `narayana.sh`. When the job wants to use a dist WFLY it can call it before the `narayana.sh` is executed.
(it's a similar situation as for scripts like `hw.sh` or `download-gnuwin32` or `release`)

mmusgrov (approved):

PR #1716, opened 2020-11-08
JBTM-3223 Validate LRA behaviour when multiple LRA's are active on th…

mmusgrov:
https://issues.redhat.com/browse/JBTM-3223

LRA

The issue says: Add tests that check that when multiple LRA's are active on the same thread that as one LRA finishes that:
- the next one is handled correctly
- creating another one still works fine

### DIFF
--- a/rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/client/LRAIT.java
+++ b/rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/client/LRAIT.java
@@ -1,0 +1,219 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2020, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package io.narayana.lra.arquillian.client;
+
+import io.narayana.lra.arquillian.Deployer;
+import io.narayana.lra.arquillian.resource.LRAParticipant;
+import io.narayana.lra.client.NarayanaLRAClient;
+import org.eclipse.microprofile.lra.annotation.LRAStatus;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.arquillian.test.api.ArquillianResource;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.ClientBuilder;
+import javax.ws.rs.client.Invocation;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.UriBuilder;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_CONTEXT_HEADER;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+@RunWith(Arquillian.class)
+public class LRAIT {
+    private static final String SHOULD_NOT_BE_ASSOCIATED =
+            "The narayana implementation (of the MP-LRA specification) still thinks that there is "
+                + "an active LRA associated with the current thread even though all LRAs should now be finished";
+
+    @ArquillianResource
+    private URL baseURL;
+
+    private Client client;
+
+    @Deployment
+    public static WebArchive deploy() {
+        return Deployer.deploy(LRAIT.class.getSimpleName());
+    }
+
+    @Before
+    public void before() {
+        client = ClientBuilder.newClient();
+    }
+
+    @After
+    public void after() {
+        if (client != null) {
+            client.close();
+        }
+    }
+
+    /**
+     * Invoke a resource method which in turn invokes other resources.
+     * The various resource invocations are called in new or existing LRAs.
+     * Various tests are performed to verify that the correct LRAs are used
+     * and the LRAs have the expected status (see the resource method for
+     * the detail).
+     */
+    @Test
+    public void testChainOfInvocations() {
+        NarayanaLRAClient lraClient = getLRAClient();
+
+        assertNotNull(lraClient);
+
+        // Invoke a method which starts a transaction
+        // (note that the method LRAParticipant.CREATE_OR_CONTINUE_LRA also invokes other resource methods)
+        URI lra1 = invokeInTransaction(null, LRAParticipant.CREATE_OR_CONTINUE_LRA);
+        assertEquals("LRA should still be active. The identifier of the LRA was " + lra1,
+                LRAStatus.Active, lraClient.getStatus(lra1));
+
+        // end the LRA
+        invokeInTransaction(lra1, LRAParticipant.END_EXISTING_LRA);
+
+        assertNull(SHOULD_NOT_BE_ASSOCIATED, lraClient.getCurrent());
+    }
+
+    /**
+     * test behaviour when multiple LRAs are active
+     */
+    @Test
+    public void testNoCurrent() {
+        NarayanaLRAClient lraClient = getLRAClient();
+
+        assertNotNull(lraClient);
+
+        URI lra1 = invokeInTransaction(null, LRAParticipant.START_NEW_LRA);
+        assertNull(SHOULD_NOT_BE_ASSOCIATED, lraClient.getCurrent());
+        URI lra2 = invokeInTransaction(null, LRAParticipant.START_NEW_LRA);
+        assertNull(SHOULD_NOT_BE_ASSOCIATED, lraClient.getCurrent());
+
+        invokeInTransaction(lra1, LRAParticipant.END_EXISTING_LRA);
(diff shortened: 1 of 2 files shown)


rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/client/LRAIT.java
```diff
@@ -0,0 +1,192 @@
...
+    @Test
+    public void testNoCurrent() throws URISyntaxException {
+        NarayanaLRAClient lraClient = new NarayanaLRAClient();
+        URI lra1 = invokeInTransaction(null, LRAParticipant.RESOURCE_PATH, LRAParticipant.START_NEW_LRA, 200);
+        assertNull("remote invocations should not update current (see MP-LRA issue behaviour)", lraClient.getCurrent());
+        URI lra2 = invokeInTransaction(null, LRAParticipant.RESOURCE_PATH, LRAParticipant.START_NEW_LRA, 200);
+        assertNull("remote invocations should not update current (see MP-LRA issue behaviour)", lraClient.getCurrent());
+
+        invokeInTransaction(lra1, LRAParticipant.RESOURCE_PATH, LRAParticipant.END_EXISTING_LRA, 200);
+        assertNull("remote invocations should not update current (see MP-LRA issue behaviour)", lraClient.getCurrent());
```
ochaloup, line 112: similar to my note above, maybe adding the more context and the LRA instance which should not be making the progress with changing current?

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/client/LRAIT.java
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    /**
+     * test behaviour when multiple LRAs are active
+     */
+    @Test
+    public void testNoCurrent() throws URISyntaxException {
+        NarayanaLRAClient lraClient = new NarayanaLRAClient();
+        URI lra1 = invokeInTransaction(null, LRAParticipant.RESOURCE_PATH, LRAParticipant.START_NEW_LRA, 200);
+        assertNull("remote invocations should not update current (see MP-LRA issue behaviour)", lraClient.getCurrent());
```
ochaloup, line 107: does make sense to add here some more context about the failure? What does the `current` mean - is it currently active LRA instance? I  think it could be hard to search expected behaviour in MP-LRA with this description.

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/client/LRAIT.java
```diff
@@ -0,0 +1,192 @@
...
+
+        // check that closing the last LRA will leave none active
+        lraClient.closeLRA(lra1);
+        assertNull("all LRAs are closed so none should be associated with the calling thread",
+                lraClient.getCurrent());
+
+        lraClient.close();
+    }
+
+    private URI invokeInTransaction(URI lra, String resourcePrefix, String resourcePath, int expectedStatus) {
```
ochaloup, line 167: could be the parameter `expectedStatus` removed when all calls expects `200`?

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/client/LRAIT.java
```diff
@@ -0,0 +1,192 @@
...
+                    .request();
+
+            if (lra != null) {
+                builder.header(LRA_HTTP_CONTEXT_HEADER, lra.toASCIIString());
+            }
+
+            response = builder.get();
+
+            assertEquals(expectedStatus, response.getStatus());
+            assertTrue(response.hasEntity());
```
ochaloup, line 183: it could be good to add some more context about the failure to the assertion message

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/resource/LRAParticipant.java
```diff
@@ -0,0 +1,168 @@
...
+
+    @GET
+    @Path(CREATE_OR_CONTINUE_LRA)
+    @LRA(value = LRA.Type.REQUIRED, end = false)
+    public Response beginLRAWithRemoteCalls(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lra1) {
+        assertEquals("lra1 should be active", LRAStatus.Active, lraClient.getStatus(lra1));
+
+        // start a new LRA
+        URI lra2 = remoteInvocation(lra1, START_NEW_LRA);
+        assertEquals("lra1 should still be active", LRAStatus.Active, lraClient.getStatus(lra1));
```
ochaloup, line 76: It would be good to print the lra id in the assertion message. It's the same for other calls down here.

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/resource/LRAParticipant.java
```diff
@@ -0,0 +1,168 @@
...
+
+        // lra1 should be the current context for remote invocations even though lra2 is active (see spec)
+        assertEquals("lra1 should be the current", lra1, lraClient.getCurrent());
+
+        URI lra3 = remoteInvocation(lra2, CONTINUE_LRA); // lra2 is still active, use it for the next invocation
+        assertEquals("lra2 should still be active", LRAStatus.Active, lraClient.getStatus(lra2));
+        assertEquals("lra2 should equal lra3", lra2, lra3); // lra3 was a continuation of lra2
+
+        lraClient.closeLRA(lra2); // use the (proprietary) client API to close the LRA started above (START_NEW_LRA)
+//        remoteInvocation(lra2, END_EXISTING_LRA); // or do it via an end annotation attribute
```
ochaloup, line 87: This line should be removed probably.

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/resource/LRAParticipant.java
```diff
@@ -0,0 +1,168 @@
...
+                    .path(resourcePath).build()) // the sub-path of the resource method
+                    .request();
+
+            if (lra != null) {
+                builder.header(LRA_HTTP_CONTEXT_HEADER, lra.toASCIIString());
+            }
+
+            response = builder.get();
+
+            return URI.create(response.readEntity(String.class));
```
ochaloup, line 159: What about to add here a check on the status and entity existence?

ochaloup (changes requested):
I have few minor comments for this PR. All are about providing more context in case of failure. Maybe some of them could be arguable.

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/client/LRAIT.java
```diff
@@ -0,0 +1,192 @@
...
+
+        // Invoke a method which starts a transaction
+        // (note that the method LRAParticipant.CREATE_OR_CONTINUE_LRA also invokes other resource methods)
+        URI lra1 = invokeInTransaction(null, LRAParticipant.RESOURCE_PATH, LRAParticipant.CREATE_OR_CONTINUE_LRA, 200);
+        assertEquals("lra1 should be active", LRAStatus.Active, lraClient.getStatus(lra1));
+
+        // end the LRA
+        invokeInTransaction(lra1, LRAParticipant.RESOURCE_PATH, LRAParticipant.END_EXISTING_LRA, 200);
+
+        assertNull("remote invocations should not update current (see MP-LRA issue behaviour)", lraClient.getCurrent());
```
xstefank, line 97: we should either link the issue or just generalize to see MP specification.

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/resource/LRAParticipant.java
```diff
@@ -0,0 +1,168 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+
+@ApplicationScoped
+@Path(LRAParticipant.RESOURCE_PATH)
+public class LRAParticipant {
+    public static final String RESOURCE_PATH = "participant2";
+
+    public static final String CREATE_OR_CONTINUE_LRA = "start-lra";
+    public static final String CREATE_OR_CONTINUE_LRA2 = "start-end-lra";
```
xstefank, line 57: little strange to read that path contains 'end', not critical

xstefank (approved):
small comments, looks good

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/resource/LRAParticipant.java
mmusgrov: Well the interesting point of the assertion is that it failed because the the LRA is in the wrong state. The point of unit tests is to detect regressions so that an engineer can be assigned to fix it (the actual value is not of interest). If you want this kind of information then we need to add trace logging into the test suite.

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/client/LRAIT.java
```diff
@@ -0,0 +1,192 @@
...
+     * @throws URISyntaxException
+     */
+    @Test
+    public void testChainOfInvocations() throws URISyntaxException {
+        NarayanaLRAClient lraClient = new NarayanaLRAClient();
+
+        // Invoke a method which starts a transaction
+        // (note that the method LRAParticipant.CREATE_OR_CONTINUE_LRA also invokes other resource methods)
+        URI lra1 = invokeInTransaction(null, LRAParticipant.RESOURCE_PATH, LRAParticipant.CREATE_OR_CONTINUE_LRA, 200);
+        assertEquals("lra1 should be active", LRAStatus.Active, lraClient.getStatus(lra1));
```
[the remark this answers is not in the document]
ochaloup: This is pretty useful to be printed. When investigating (either locally or on CI) it's just easy to add the trace logging for the container and then the assertion printing the id is useful. When such information is available I don't consider it as no interest to anyone. If the test fails it's important to anyone searching for the cause.

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/resource/LRAParticipant.java
ochaloup: Trace logging can be switch on.

ochaloup (approved):
I'm approving the PR.
I have still opinion on reasons why the LRA id in assertion message is handy. But if you don't agree I will add it later in different PR when I will need to investigate the reasons for some failure.

mmusgrov:
> I'm approving the PR.
> I have still opinion on reasons why the LRA id in assertion message is handy. But if you don't agree I will add it later in different PR when I will need to investigate the reasons for some failure.

Sprinkling trace logging and verbose trace messages is the old style of figuring out what's going on, back in the old days it was all we had. Nowadays we have better tools, such as Byteman for example. This desire for explicit invasive tracing makes complex code that much more tricky to parse, here clarity and parsimony brings greater benefit than verbosity.

Regarding unit testing I'd suggest using mvn debug or your tool of choice for tracing behaviour.

I'd also suggest that over-engineering unit tests is not the most efficient use of engineering resource which would be better expended on, for example, producing a set of byteman scripts that gives us the more targeted and useful traces that we can use.

ochaloup:
Ok, that's not my point of view but no problem.
(We are talking about exceptional occurrence and as such it's a good practice for the exception containing all relevant information.)

PR #1717, opened 2020-11-08
JBTM-3342 Propagate the parent context header (Long-Running-Action-Pa…

mmusgrov:
https://issues.redhat.com/browse/JBTM-3342

LRA

When propagating a nested context expose the parent context header on calls to methods carrying an LRA annotation.

### DIFF
--- a/rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
+++ b/rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
@@ -48,6 +48,7 @@ import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriInfo;
 import javax.ws.rs.ext.Provider;
+import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Method;
 import java.net.URI;
 import java.net.URISyntaxException;
@@ -207,6 +208,17 @@ public class ServerLRAFilter implements ContainerRequestFilter, ContainerRespons
 
         if (endAnnotation && incommingLRA == null) {
             return;
+        }
+
+        if (incommingLRA != null) {
+            // set the parent context header
+            try {
+                headers.putSingle(LRA_HTTP_PARENT_CONTEXT_HEADER, Current.getFirstParent(incommingLRA));
+            } catch (UnsupportedEncodingException e) {
+                String msg = String.format("incoming LRA %s contains an invalid parent", incommingLRA);
+
+                throwGenericLRAException(incommingLRA, Response.Status.PRECONDITION_FAILED.getStatusCode(), msg);
+            }
         }
 
         switch (type) {
--- a/rts/lra/service-base/src/main/java/io/narayana/lra/Current.java
+++ b/rts/lra/service-base/src/main/java/io/narayana/lra/Current.java
@@ -24,13 +24,21 @@ package io.narayana.lra;
 import javax.ws.rs.client.ClientRequestContext;
 import javax.ws.rs.container.ContainerResponseContext;
 import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.core.UriBuilder;
+import java.io.UnsupportedEncodingException;
 import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Stack;
 
+import static io.narayana.lra.LRAConstants.PARENT_LRA_PARAM_NAME;
+import static io.narayana.lra.LRAConstants.QUERY_FIELD_SEPARATOR;
+import static io.narayana.lra.LRAConstants.QUERY_PAIR_SEPARATOR;
 import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_CONTEXT_HEADER;
 
 public class Current {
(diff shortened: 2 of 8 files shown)


rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -271,6 +271,28 @@ public void filter(ContainerRequestContext containerRequestContext) {
...
+        if (query != null) {
+            for (String param : query.split("&")) { // or apply a regex matching parent=...
+                if (param.startsWith("parent=")) {
+                    String parent = param.split("=", 2)[1];
+
+                    try {
+                        // verify that parent corresponds to a valid URI
+                        new URI(parent);
+                        // the parent param was generated using URI.toASCIIString
+                        headers.putSingle(LRA_HTTP_PARENT_CONTEXT_HEADER, parent);
```
xstefank, line 285: this will override user-provided header. Is it intended behavior?

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/NestedParticipantIT.java
```diff
@@ -99,34 +106,71 @@ public void nestedParticipantAfterLRACallTest() {
...
+        URI child = lraClient.startLRA("child");
+
+        Response response = null;
+
+        try {
+            response = client.target(UriBuilder.fromUri(baseURL.toExternalForm())
+                    .path(NestedParticipant.ROOT_PATH)
+                    .path(NestedParticipant.PATH))
+                    .request()
+//                .header(LRA.LRA_HTTP_CONTEXT_HEADER, parentLRA)
```
xstefank, line 154: Remove this please and add a comment explaining that the header is populated by the filter automatically?

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/resource/NestedParticipant.java
```diff
@@ -48,16 +47,33 @@
...
     LRAMetricService lraMetricService;
 
+    @LRA(end = false)
+    @GET
+    @Path(PATH)
+    @Produces({MediaType.TEXT_PLAIN})
+    public Response runWithNestedContext(@HeaderParam(LRA.LRA_HTTP_CONTEXT_HEADER) URI lraId,
+                           @HeaderParam(LRA.LRA_HTTP_PARENT_CONTEXT_HEADER) URI parentId) {
+        if (parentId == null || lraId == null) {
+            return Response.serverError().build();
```
xstefank, line 62: we should add a message that says which one was null. The error message in test says it is only parentId.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: Come on, read the spec, aren't we allowed to implement it. Does user behaviour override the spec now (ie does the users' opinion of what constitutes the transaction hierarchy trump actuality).

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/NestedParticipantIT.java
mmusgrov: Sure. I see that the fix has broken one of the tests so I will remove it at the same time I resolve that failure. I'll also add a test for deeper transaction hierarchies.

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/resource/NestedParticipant.java
mmusgrov: I don't think so, this is a test suite in which we already have tests that check the presence of LRA_HTTP_CONTEXT_HEADER.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
xstefank: https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java#L82

-- I think we should fail if the user propagates the wrong parent header, or at least log warning. This seems like it can hide potential issues with user code.

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/resource/NestedParticipant.java
xstefank: true, this was more about this particular test only. It is unlikely it wouldn't be caught elsewhere.

mmusgrov: This is a test for nested contexts so LRA_HTTP_CONTEXT_HEADER is not relevant.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: That part of the spec is referring to LRA_HTTP_CONTEXT_HEADER.
You can't randomly change the parent context since it forms a strict hierarchy.

mmusgrov:
@xstefank This is ready for another review now. I've fixed the broken test and added a test for deeper LRA hierarchies.
Please ignore the previous commits since I was using them as temporary backups and I forget that I had an outstanding PR. I'll delete the associated builds.

rts/lra/service-base/src/main/java/io/narayana/lra/LRAConstants.java
```diff
@@ -40,6 +40,8 @@
     public static final String CLIENT_ID_PARAM_NAME = "ClientID";
     public static final String TIMELIMIT_PARAM_NAME = "TimeLimit";
     public static final String PARENT_LRA_PARAM_NAME = "ParentLRA";
+    public static final String QUERY_PAIR_SEPARATOR = "&"; //
```
xstefank, line 43:
```suggestion
    public static final String QUERY_PAIR_SEPARATOR = "&";
```

rts/lra/service-base/src/main/java/io/narayana/lra/Current.java
```diff
@@ -64,6 +72,67 @@ public static Object getState(String key) {
...
+                    String parents = param.split(QUERY_FIELD_SEPARATOR, 2)[1];
+
+                    // parents is a comma separated list of parents (the first one is the direct parent)
+                    if (parents != null) {
+                        String[] pa = parents.split(",");
+
+                        if (pa.length > 0) {
+                            if (decode) {
+                                return URLDecoder.decode(pa[0], "UTF-8");
+                            }
```
xstefank, line 124: is it correct to return null if `decode` is false? In that case, this can be refactored to not contain the parameter.

xstefank (approved):

PR #1718, opened 2020-11-09
JBTM-3386 performance tune DiskSlots store backend

jhalliday:
https://issues.redhat.com/browse/JBTM-3386


mmusgrov (approved):
Thanks @jhalliday for this update (I assume it's also a good change to combat the recent UseBiasedLocking perf issue on JDK 15).

mmusgrov:
The PR Jenkins axes that failed look unrelated. The performance regressions (see tests JTAStoreTests.jtaTest and Performance1.onePhaseTest) exercised unrelated code, but they were unexpected so we need to keep monitoring and ideally come up with a more reliable performance test bed or figure out which factors skew the comparisons (we still think running on bare metal rather than docker is the way to go).

PR #1719, opened 2020-11-11
JBTM-3385 Add maven profile for the LRA Quarkus coordinator native bu…

xstefank:
…ild in Docker

https://issues.redhat.com/browse/JBTM-3385

This is preliminary work also for LRA coordinator docker image to switch to quarkus native.

### DIFF
--- a/narayana-full/src/main/assembly/bin.xml
+++ b/narayana-full/src/main/assembly/bin.xml
@@ -203,11 +203,6 @@
 			<destName>xts_third_party_licenses.txt</destName>
 		</file>
 		<file>
-			<source>../rts/lra/coordinator-quarkus/target/lra-coordinator-runner.jar</source>
-			<fileMode>0755</fileMode>
-			<outputDirectory>rts/lra/</outputDirectory>
-		</file>
-		<file>
 			<source>../rts/lra/coordinator/target/lra-coordinator.jar</source>
 			<fileMode>0644</fileMode>
 			<outputDirectory>rts/lra/</outputDirectory>
@@ -327,8 +322,12 @@
 			<outputDirectory>quickstarts/rest-tx</outputDirectory>
 		</fileSet>
 		<fileSet>
-			<directory>../rts/lra/coordinator-quarkus/target/lib</directory>
-			<outputDirectory>rts/lra/lib</outputDirectory>
+			<directory>../rts/lra/coordinator-quarkus/target/</directory>
+			<outputDirectory>rts/lra/</outputDirectory>
+			<includes>
+				<include>lra-coordinator-runner*</include>
+				<include>lib/</include>
+			</includes>
 		</fileSet>
 	</fileSets>
 </assembly> 
--- a/narayana-release-process.sh
+++ b/narayana-release-process.sh
@@ -62,6 +62,11 @@ then
   exit
 fi
 
+# check if docker is actually podman
+if [[ -z $LRA_CONTAINER_BUILD_RUNTIME ]]; then
+  command -v podman >/dev/null 2>&1 && LRA_CONTAINER_BUILD_RUNTIME=-Dcoordinator.native.container-runtime=podman
+fi
+
 # Do this early to prevent later interactive need
 docker login docker.io
 [ $? -ne 0 ] && echo "Login to docker.io was not succesful" && exit
@@ -175,7 +180,7 @@ else
 fi
 
 rm -rf $PWD/localm2repo
-./build.sh clean install -Dmaven.repo.local=${PWD}/localm2repo -DskipTests -gs ~/.m2/settings.xml -Dorson.jar.location=$ORSON_PATH -Pcommunity
+./build.sh clean install -Dmaven.repo.local=${PWD}/localm2repo -DskipTests -gs ~/.m2/settings.xml -Dorson.jar.location=$ORSON_PATH -Pcommunity -Pnative-docker $LRA_CONTAINER_BUILD_RUNTIME
 if [[ $? != 0 ]]
 then
   echo 1>&2 Could not install narayana
--- a/rts/lra/coordinator-quarkus/pom.xml
+++ b/rts/lra/coordinator-quarkus/pom.xml
@@ -18,6 +18,8 @@
         <narayana.nodename.property>CoreEnvironmentBean.nodeIdentifier</narayana.nodename.property>
         <quarkus.http.port>8080</quarkus.http.port>
         <surefire-plugin.version>2.22.0</surefire-plugin.version>
+
+        <coordinator.native.container-runtime>docker</coordinator.native.container-runtime>
     </properties>
 
     <dependencyManagement>
@@ -118,6 +120,34 @@
                                     <systemProperties>
                                         <native.image.path>${project.build.directory}/${project.build.finalName}-runner</native.image.path>
                                     </systemProperties>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+        <profile>
+            <id>native-docker</id>
+            <activation>
+                <property>
+                    <name>native-docker</name>
+                </property>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>io.quarkus</groupId>
+                        <artifactId>quarkus-maven-plugin</artifactId>
+                        <version>${version.quarkus}</version>
+                        <executions>
+                            <execution>
+                                <goals>
+                                    <goal>native-image</goal>
+                                </goals>
+                                <configuration>
+                                    <enableHttpUrlHandler>true</enableHttpUrlHandler>
+                                    <containerRuntime>${coordinator.native.container-runtime}</containerRuntime>
                                 </configuration>
                             </execution>
                         </executions>
--- a/README.md
+++ b/README.md
@@ -87,6 +87,11 @@ STM: `./build.[sh|bat] -am -pl :stm`
 
 LRA: `./build.[sh|bat] -am -f rts/lra`
 
+LRA binary: `./build.[sh|bat] -Pnative-docker -am -f rts/lra`
+If podman is installed on the system then this native-docker maven profile will produce a native executable for a coordinator which you can then start using the command:
+
+    ./rts/lra/coordinator-quarkus/target/lra-coordinator-runner
+
 Testing Narayana
 ---------------
 
--- a/rts/lra/coordinator-quarkus/src/main/resources/application.properties
+++ b/rts/lra/coordinator-quarkus/src/main/resources/application.properties
@@ -1,5 +1,8 @@
 # native
-quarkus.native.additional-build-args=-H:ReflectionConfigurationFiles=reflection-config.json
+quarkus.native.additional-build-args=\
+  --initialize-at-run-time=com.arjuna.ats.arjuna.coordinator.TxControl, \
+  --initialize-at-run-time=com.arjuna.ats.internal.arjuna.objectstore.FileSystemStore, \
+  -H:ReflectionConfigurationFiles=reflection-config.json
 quarkus.ssl.native=true
 
 # allow Swagger in prod mode
--- a/rts/lra/coordinator-quarkus/src/main/resources/reflection-config.json
+++ b/rts/lra/coordinator-quarkus/src/main/resources/reflection-config.json
@@ -18,6 +18,45 @@
     "allPublicMethods" : true,
     "allDeclaredFields" : true,
     "allPublicFields" : true
-  }
+  },
+  {
 
+    "name" : "com.arjuna.ats.internal.arjuna.objectstore.ShadowNoFileLockStore",
+    "allDeclaredConstructors" : true,
+    "allPublicConstructors" : true,
+    "allDeclaredMethods" : true,
+    "allPublicMethods" : true,
+    "allDeclaredFields" : true,
+    "allPublicFields" : true
+  },
+  {
+
+    "name" : "com.arjuna.ats.arjuna.recovery.ActionStatusService",
+    "allDeclaredConstructors" : true,
+    "allPublicConstructors" : true,
+    "allDeclaredMethods" : true,
+    "allPublicMethods" : true,
+    "allDeclaredFields" : true,
+    "allPublicFields" : true
+  },
+  {
+
+    "name" : "com.arjuna.ats.arjuna.common.ObjectStoreEnvironmentBean",
+    "allDeclaredConstructors" : true,
+    "allPublicConstructors" : true,
+    "allDeclaredMethods" : true,
+    "allPublicMethods" : true,
+    "allDeclaredFields" : true,
+    "allPublicFields" : true
+  },
+  {
+
+    "name" : "com.arjuna.ats.arjuna.common.MetaObjectStoreEnvironmentBean",
+    "allDeclaredConstructors" : true,
+    "allPublicConstructors" : true,
+    "allDeclaredMethods" : true,
+    "allPublicMethods" : true,
+    "allDeclaredFields" : true,
+    "allPublicFields" : true
+  }
 ]


ochaloup (approved):
Seems nice to me.

narayana-release-process.sh
```diff
@@ -62,6 +62,12 @@ then
   exit
 fi
 
+# check if docker is actually podman
+podman --version
```
tomjenkinson, line 66: Please can you add something like https://github.com/jbosstm/narayana/blob/f3a0bbb5784b21d1d097311d1e15b4028b64679d/narayana-release-process.sh#L16 to check the command is there

rts/lra/coordinator-quarkus/pom.xml
```diff
@@ -13,11 +13,16 @@
 
     <properties>
         <narayana.version>${project.version}</narayana.version>
+        <maven.compiler.source>1.8</maven.compiler.source>
```
tomjenkinson, line 16: @mmusgrov would it be possible/useful to add this into the root pom.xml?

rts/lra/coordinator-quarkus/pom.xml
```diff
@@ -13,11 +13,16 @@
...
+        <maven.compiler.target>1.8</maven.compiler.target>
```
tomjenkinson, line 17: @mmusgrov FYI I note we have this in a few places but different expectations: https://github.com/jbosstm/narayana/search?p=2&q=%22maven.compiler.target%22 - perhaps we would like to unify this?

rts/lra/coordinator-quarkus/pom.xml
```diff
@@ -125,6 +130,35 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <id>native-docker</id>
```
tomjenkinson, line 134: How about calling this "create-quarkus-native-image" or "create-quarkus-native-image-with-docker"

tomjenkinson (changes requested):

mmusgrov:
When I test the PR I get the following error:
```bash
...
[INFO] LRA Coordinator .................................... FAILURE [09:28 min]                                                                                                                                                               
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
...
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  42:49 min
[INFO] Finished at: 2020-11-17T12:23:46Z
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal io.quarkus:quarkus-maven-plugin:1.8.3.Final:native-image (default) on project lra-coordinator: Failed to generate native image: io.quarkus.builder.BuildException: Build failure: Build failed due to errors
[ERROR]         [error]: Build step io.quarkus.deployment.pkg.steps.NativeImageBuildStep#build threw an exception: java.lang.RuntimeException: Failed to pull builder image quay.io/quarkus/ubi-quarkus-native-image:20.2.0-java11
[ERROR]         at io.quarkus.deployment.pkg.steps.NativeImageBuildStep.setupContainerBuild(NativeImageBuildStep.java:388)
[ERROR]         at io.quarkus.deployment.pkg.steps.NativeImageBuildStep.build(NativeImageBuildStep.java:109)
[ERROR]         at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[ERROR]         at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[ERROR]         at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[ERROR]         at java.lang.reflect.Method.invoke(Method.java:498)
[ERROR]         at io.quarkus.deployment.ExtensionLoader$2.execute(ExtensionLoader.java:936)
[ERROR]         at io.quarkus.builder.BuildContext.run(BuildContext.java:277)
[ERROR]         at org.jboss.threads.ContextClassLoaderSavingRunnable.run(ContextClassLoaderSavingRunnable.java:35)
[ERROR]         at org.jboss.threads.EnhancedQueueExecutor.safeRun(EnhancedQueueExecutor.java:2046)
[ERROR]         at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.doRunTask(EnhancedQueueExecutor.java:1578)
[ERROR]         at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1452)
[ERROR]         at java.lang.Thread.run(Thread.java:748)
[ERROR]         at org.jboss.threads.JBossThread.run(JBossThread.java:479)
[ERROR] Caused by: java.io.IOException: Cannot run program "podman": error=2, No such file or directory
[ERROR]         at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
[ERROR]         at io.quarkus.deployment.util.ProcessUtil.launchProcess(ProcessUtil.java:37)
[ERROR]         at io.quarkus.deployment.pkg.steps.NativeImageBuildStep.setupContainerBuild(NativeImageBuildStep.java:385)
[ERROR]         ... 13 more
[ERROR] Caused by: java.io.IOException: error=2, No such file or directory
[ERROR]         at java.lang.UNIXProcess.forkAndExec(Native Method)
[ERROR]         at java.lang.UNIXProcess.<init>(UNIXProcess.java:247)
[ERROR]         at java.lang.ProcessImpl.start(ProcessImpl.java:134)
[ERROR]         at java.lang.ProcessBuilder.start(ProcessBuilder.java:1029)
[ERROR]         ... 15 more
```

mmusgrov:
Ah wait I thought I had installed podman but dnf install had failed:

```bash
 * Downloading packages... Failed to install packages: Cannot download Packages/s/slirp4netns-1.1.4-1.fc31.x86_64.rpm: All mirrors were tried; Last error: Status code: 404 for https://www.mirrorservice.org/sites/dl.fedoraproject.org/pub/fedora/linux/updates/31/Everything/x86_64/Packages/s/slirp4netns-1.1.4-1.fc31.x86_64.rpm (IP: 2001:630:341:12::184
```

What commands did you use to install podman?

mmusgrov:
I take a look a the quarkus manual ...

mmusgrov:
Maybe you could update the readme instructions for the release explaining how to create the native image during the release process.

mmusgrov:
> @mmusgrov would it be possible/useful to add this into the root pom.xml?

@tomjenkinson we import the -source and -target of the Java Compiler from the WildFly parent pom so i prefer that we continue to use that. Martin why do you need override these settings?

mmusgrov:
+1 to Tom's comment, command is  the `posix` compatible way of test whether an executable exists. After running `command -v podman` test the exit status (running a program to see if it exists can produce side effects and is bad practice).

mmusgrov:
@tomjenkinson we should be using 1.8 for the default profile. The first few hits of the search https://github.com/jbosstm/narayana/search?p=2&q=%22maven.compiler.target%22 seem to be matching profiles that target jdk 1.9 and above (eg <jdk>[1.9, 11)</jdk>). I will go them all to check that none of our poms are overriding it in the default profile.

mmusgrov:
@xstefank the test run is showing an issue building "Narayana: narayana-full".

mmusgrov:
> I think it coming from the parent would be best, please do be aware we override the target quite a bit so perhaps those could be cleaned up https://github.com/jbosstm/narayana/search?p=1&q=%22maven.compiler.target%22? also please note that in some cases we are setting JDK 9 - is that expected?

Agreed, the profiles that target [1.9, 11) should be choosing 11.

mmusgrov:
1. I managed to install podman by updating my yum repo: sudo curl -L -o /etc/yum.repos.d/devel:kubic:libcontainers:stable.repo https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/CentOS_7/devel:kubic:libcontainers:stable.repo
after which dnf install podman worked.

But won't the resulting executable only run on an architecture compatible with whatever machine we do the release on? In which case there would be little benefit in having this feature?

2. The quarkus site says we can also select the container runtime to use with:

./mvnw package -Pnative -Dquarkus.native.container-runtime=docker

Do we have to use podman or is docker an alternative (if so I'd prefer to avoid podman)?

3. When I run the resulting native executable. It shows the coordinator starting which is promising.
But when I try to start an LRA the coordinator fails with:

```bash
2020-11-17 14:51:37,696 INFO  [io.quarkus] (main) lra-coordinator 5.10.7.Final-SNAPSHOT native (powered by Quarkus 1.8.3.Final) started in 0.022s. Listening on: http://0.0.0.0:8080
2020-11-17 14:51:37,696 INFO  [io.quarkus] (main) Profile prod activated.
2020-11-17 14:51:37,696 INFO  [io.quarkus] (main) Installed features: [cdi, rest-client, resteasy, resteasy-jsonb, smallrye-openapi, swagger-ui]
2020-11-17 14:58:09,726 WARN  [com.arj.ats.arjuna] (executor-thread-1) com.arjuna.ats.arjuna.exceptions.ObjectStoreException: FileSystemStore.openAndLock failed to create hierarchy /project/ObjectStore/ShadowNoFileLockStore/defaultStore/StateManager/BasicAction/TwoPhaseCoordinator/LRA/0_ffff0a000264_9ee5_5fb3dbc1_5
```

What is interesting is that a call to start an LRA returned what looks like a valid `Uid`, that should not happen if the log creation failed so we may have an issue there - I will check the code to see what that particular problem could be.

Did you manage to start and join with an LRA in your testing?
If we go with this feature would you be able to update the PR to include a test against the resultant exe?

mmusgrov:
The PR already has one approval but I don't think we can risk merging yet until the issues have been worked through.
To avoid accidentally merging it I am adding the Hold label while we work through those issues.

mmusgrov:
It might be better moving this feature out of the release scripts and into a profile instead, then users can choose to target it to their particular platform.

narayana-release-process.sh
xstefank: @tomjenkinson we don't require podman. We only need a different flag when its being used. Or is this only about `command -v`?

rts/lra/coordinator-quarkus/pom.xml
xstefank: actually I would like to move Quarkus coordinator to JDK 11 as JDK 8 support is already deprecated in Quarkus and will be removed soon - https://quarkus.io/blog/quarkus-1-4-final-released/#java-8-deprecated. This mentions dropping JDK 8 in Quarkus 1.6 (which still didn't happen up to my knowledge but we should be ready that this can come up).

rts/lra/coordinator-quarkus/pom.xml
xstefank: I can rename both profiles if you prefer.

xstefank:
Hi @mmusgrov and @tomjenkinson. Sorry, there was a holiday yesterday in Czechia.

> Maybe you could update the readme instructions for the release explaining how to create the native image during the release process.

Will do.

> Martin the parent pom already sets the version to 1.8 so why do you need override these settings?

I will remove this. However, please see my comment above about Quarkus deprecating and removing support for JDK 8 possible anytime now.

> After running command -v podman test the exit status (running a program to see if it exists can produce side effects and is bad practice)

OK. I'll change this but I think I just followed some other places in the script so maybe worth checking.

> But won't the resulting executable only run on an architecture compatible with whatever machine we do the release on? In which case there would be little benefit in having this feature?

The main intention is to put quarkus native runnable into the docker image. Yes, it is true that this will be based on the architecture on which it is built but this is how Graal works, unfortunately. This is really only something additional. Users can still run `java -jar lra-coordinator-runner.jar` if running on a different architecture or build their own native as Tom mentioned. But this will significantly help with the image size and startup times of the lra-coordinator docker image.

> Do we have to use podman or is docker an alternative (if so I'd prefer to avoid podman)?

Docker is the default. But if you are on Fedora 32+ like me you are kind of forced to have only podman if you don't want total rewrite of everything docker. Fedora also recommends to use aliases for docker-podman so you don't rewrite your system scripts which breaks with native profile here because it still default somewhere to docker which is not installed on my machine. So if you run without the property set to podman the docker will be used. But if you have podman installed, it's better to default to podman. I will make this optional. So if you set the environment variable manually it will be used otherwise if podman is installed use podman otherwise use docker.

> But when I try to start an LRA the coordinator fails with

This is strange. I was running full TCK and basic tests with built natives. I'll check again.

> I think it is already in a profile in the PR, but it is called from the release scripts and added to the zip and that might be what is not necessary

It is in profile. We need to include it in the build so it can be consumed in the docker image build in https://github.com/jboss-dockerfiles/narayana.

narayana-release-process.sh
tomjenkinson: Ah, I think I understand better now. If you are trying to verify if podman exists then I think using `command -v podman` would be good instead of running the command. Thanks

rts/lra/coordinator-quarkus/pom.xml
tomjenkinson: I was just commenting to get this generally applicable, but I think it should be in the jboss-parent so actually my review would be to remove this line (and https://github.com/jbosstm/narayana/blob/master/rts/pom.xml#L46) entirely - won't it come from: https://github.com/jboss/jboss-parent-pom/blob/jboss-parent-35/pom.xml#L151 - but that second remove would be cleanup for existing code so not necessary to do in this PR as I guess you can remove this line from your PR without needing to remove it from rts/pom.xml

rts/lra/coordinator-quarkus/pom.xml
tomjenkinson: I see, there is already a "native" - in that case I think the new name is fine for this PR. I will resolve this conversation.

narayana-full/src/main/assembly/bin.xml
```diff
@@ -202,11 +202,6 @@
 			<outputDirectory>lib/ext</outputDirectory>
 			<destName>xts_third_party_licenses.txt</destName>
 		</file>
-		<file>
```
tomjenkinson, line 205: Is this section just cleanup or did it add something related to the `native-docker` profile?

xstefank:
@mmusgrov about

> Maybe you could update the readme instructions for the release explaining how to create the native image during the release process.

there is no mention about the native profile neither. Should this go into README or not?

rts/lra/coordinator-quarkus/pom.xml
mmusgrov: We follow WildFly's lead so we should only move to JDK 11 when they do.
The LRA implementation is a shim over ArjunaCore so we should use whatever that uses.
Quarkus is only one of our target platforms so that should not dictate what we do.

mmusgrov: I would add that JDK 8 code will run on JDK 11 so there is no urgency.
And we already test LRA in our CI against both JDKs so again there is no urgency.
What features of JDK 11 do you feel we need for LRA?

mmusgrov:
> @mmusgrov about
> 
> > Maybe you could update the readme instructions for the release explaining how to create the native image during the release process.
> 
> there is no mention about the native profile neither. Should this go into README or not?

I didn't understand this response but there's a section in the README (https://github.com/jbosstm/narayana/blob/master/README.md#building-specific-components) that describes how to build specific narayana components. So you add a line after the one for LRA: 

>
> LRA binary: ./build.[sh|bat] -Pnative-docker -am -f rts/lra
>     If podman is installed on the system then this `native-docker` maven profile will produce a native executable for a coordinator which you can then start using the command:
>
>         ./rts/lra/coordinator-quarkus/target/lra-coordinator-runner

@tomjenkinson normally when a release produces binaries it does so for all the common platforms (amd, linux variants, windows, etc) and our (narayana) build system (BlackTie aside) isn't geared up for that yet.

@xstefank I am still having the same issues running the resultant binary: start takes about 10 seconds, error messages on the console about writing transaction logs and cancel/close hangs. I did see the participant getting notified during cancel/close (so it's kind of doing the right thing) but the client that issues the cancel/close call hangs waiting for the coordinator.
My laptop is a bit of twilight zone (as Tom can attest) so it could be me. Will you indicate how you tested it from initial build of narayana to running the exe.

mmusgrov:
@tomjenkinson If we release a binary for one specific platform (which could change from release to release) then it's very likely we will get users asking us to provide binaries for other platforms. For BlackTie we explicitly state which platforms we support and produce binaries for those but for the LRA coordinator we can support whatever GraalVM supports but the users should build it themselves.

mmusgrov:
Remark: We also have mac and windows users so we should test the native image on those platforms too (not in this PR though).

narayana-full/src/main/assembly/bin.xml
```diff
@@ -327,8 +322,12 @@
 			<outputDirectory>quickstarts/rest-tx</outputDirectory>
 		</fileSet>
 		<fileSet>
-			<directory>../rts/lra/coordinator-quarkus/target/lib</directory>
-			<outputDirectory>rts/lra/lib</outputDirectory>
+			<directory>../rts/lra/coordinator-quarkus/target/</directory>
+			<outputDirectory>rts/lra/</outputDirectory>
+			<includes>
+				<include>lra-coordinator-runner*</include>
```
tomjenkinson, line 328: @xstefank I am thinking that this is not just cleanup and some files will appear in the zip. Is that a correct line of thinking? If it is then we need to consider that generally and if the decision was that this file were to be included in the zip then it would be important to change the name of the zip to reflect the compatibility - like we do in BlackTie. I think this might where the assembly name is set but it's a long time ago - if we proceed in adding files to the zip please let me know if you need help to work out how to make sure the name is reflected: https://github.com/jbosstm/narayana/blob/master/blacktie/blacktie/src/main/assembly/bin.xml#L19

narayana-release-process.sh
```diff
@@ -175,7 +181,7 @@ else
 fi
 
 rm -rf $PWD/localm2repo
-./build.sh clean install -Dmaven.repo.local=${PWD}/localm2repo -DskipTests -gs ~/.m2/settings.xml -Dorson.jar.location=$ORSON_PATH -Pcommunity
+./build.sh clean install -Dmaven.repo.local=${PWD}/localm2repo -DskipTests -gs ~/.m2/settings.xml -Dorson.jar.location=$ORSON_PATH -Pcommunity -Pnative-docker $CONTAINER_BUILD_RUNTIME
```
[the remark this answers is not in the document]
tomjenkinson: If there is something also expected to be added to the zip then this review comment is not correct.

tomjenkinson (changes requested):

mmusgrov:
Also note that the LRA quarkus extension produces a binary coordinator, would that not be sufficient?

narayana-release-process.sh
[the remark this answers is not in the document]
mmusgrov: Tom these comments contradicts my [earlier comment](https://github.com/jbosstm/narayana/pull/1719#issuecomment-731128864).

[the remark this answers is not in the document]
tomjenkinson: I am not sure if they do or don't sorry but I will resolve this conversation. We can revisit the topic as necessary once I understand the scope of the change in narayana-full/src/main/assembly/bin.xml

rts/lra/coordinator-quarkus/pom.xml
```diff
@@ -125,6 +130,35 @@
...
+                        <version>${version.quarkus}</version>
+                        <executions>
+                            <execution>
+                                <goals>
+                                    <goal>native-image</goal>
+                                </goals>
+                                <configuration>
+                                    <enableHttpUrlHandler>true</enableHttpUrlHandler>
+                                    <dockerBuild>true</dockerBuild>
+                                    <containerRuntime>${quarkus.native.container-runtime}</containerRuntime>
```
tomjenkinson, line 154: I will add the request for change directly here, please can you add a quickstart PR to use this image?

tomjenkinson (changes requested):

narayana-release-process.sh
tomjenkinson, line 184: I was chatting with @ochaloup I think that if the image that is built also contains a platform dependent native image that will need to be reflected in the docker image name that we push: https://github.com/jbosstm/narayana/blob/f3a0bbb5784b21d1d097311d1e15b4028b64679d/narayana-release-process.sh#L228

tomjenkinson (changes requested):

ochaloup (changes requested):
I terribly missed the point about the architecture being built to the release of Narayana zip and the docker image. I agree with Tom and Mike that the changes are needed around this.
@tomjenkinson 's point about testing is pretty good as well but this could be managed with https://issues.redhat.com/browse/JBTM-3343

rts/lra/coordinator-quarkus/pom.xml
mmusgrov: @tomjenkinson Your comment implies that we are going ahead with this. I mentioned above about us not wanting to provide two native coordinators in two different projects. Quarkus is the place where deliver native support - and we have a native test for it already [1] and I have a quickstart somewhere.

Having two different places where users can obtain a native coordinator both dilutes our message and increases the support workload on the narayana team. I would prefer that the native is supported via the quarkus project where the expertise resides. Can you clarify your intention here.

[1] https://github.com/quarkusio/quarkus/pull/6763/files#diff-1c7e7b28df509e2d2fea89c3ba1bf65d3442637def31c479462fba48526adfa8R5

tomjenkinson: I see, my intention would be if we were going ahead then we would need some way to verify the coordinator we are producing is functional. But if Narayana project isn't going ahead then I guess this PR and JBTM-3385 would be closed.

narayana-full/src/main/assembly/bin.xml
xstefank: sorry I must have misunderstood your previous comment. Yes, this will add lra-coordinator files, jar and optionally binary, to the zip.

xstefank:
sorry for the late reply. I was on PTO and didn't get to this sooner.

@mmusgrov 

We are already shipping native profile, this is only adding and option to run it without GrallVM installed so what you are proposing is not directly tied in this PR. However, the idea here is to have an option to pull this in during release to include it in docker image. I am not sure that we would want to do this by generating a quarkus app with lra-coordinator extension but it's possible for sure. But you will be publishing quarkus tied stuff under jbosstm umbrella which seems a little off to me.

I also run into the same issues with running the native built with docker and it seems the issue is with initialization in the container which defaults user dir to `/project/` and OS is then placed on `/project/ObjectStore/` by default which is accessible in the container but usually not on the target machine. I have a fix for this that I'll push shortly. If you can please, try it again.

@tomjenkinson and @ochaloup 

We've discussed the docker architectures with Ondra and basically what I propose by changing `java -jar` which we have currently in the container image to native binary would not change anything for narayana users as the image is always tied to the platform where it was built. E.g. in https://hub.docker.com/layers/jbosstm/lra-coordinator/5.10.6.Final/images/sha256-1f6f7e001affe5e1f5ef02222f499d226a3520780f0ea91dab276740a821f508?context=explore you can see `OS/ARCH linux/amd64` which I believe won't run on 32b archs but windows or mac (or linux) 64b should NOT have any issues with this image just like they don't have any issues now.

Quickstart:

It is definitely possible to do but why? This is just a different way to run the coordinator. All current QSs must run with it.

narayana-full/src/main/assembly/bin.xml
tomjenkinson: The zip would end up on our website and the script that is being edited is the one that helps product and upload that. You can find our binaries here: https://narayana.io/downloads/index.html

xstefank: If the only reason to include the Linux 64 binary of lra-coordinator is for the docker image than I would vote to build it when the docker image is created. This way we can avoid adding additional zip artifacts.

xstefank:
@tomjenkinson agreed about the QS. I will make the QS changes in a different JBTM. If it's agreed to not include the binary in the final zip then I will update the PR to just include the maven profile a we can build the binary with the docker image.

mmusgrov:
How would we configure jbossts properties (such as the object store type and location). Normally we supply a jbossts-properties.xml file or we integrate with the configuration mechanism supplied by the container we are integrating with (eg with WildFly we would introduce a subsystem for it or with quarkus we would provide an extension). It is possible to deploy LRA in quarkus but it requires some effort from the developer (see for example the [embedded quickstart](https://github.com/jbosstm/quickstart/tree/master/rts/lra-examples/cdi-embedded/src/main/resources).

Can we close this one as will not fix and can we also mark the artefact `[org.jboss.narayana.rts:lra-coordinator-quarkus](https://github.com/jbosstm/narayana/tree/master/rts/lra/coordinator-quarkus)` as test only and move it under the test directory since it is not usable outside of a narayana test environment.

@ochaloup how do you feel about us doing the above two tasks?

xstefank:
> How would we configure jbossts properties (such as the object store type and location). Normally we supply a jbossts-properties.xml file or we integrate with the configuration mechanism supplied by the container we are integrating with (eg with WildFly we would introduce a subsystem for it or with quarkus we would provide an extension). It is possible to deploy LRA in quarkus but it requires some effort from the developer - see for example the [embedded quickstart](https://github.com/jbosstm/quickstart/tree/master/rts/lra-examples/cdi-embedded/src/main/resources).

I will need to rerun this as it's sitting idle for quite some time but last command I found in my history is:

`target/lra-coordinator-runner -Dcom.arjuna.ats.arjuna.common.ObjectStoreEnvironmentBean.objectStoreDir=$OBJECT_STORE_PATH -Dcom.arjuna.ats.arjuna.common.ObjectStoreEnvironmentBean.communicationStore.objectStoreDir=$OBJECT_STORE_PATH -Dcom.arjuna.ats.arjuna.common.ObjectStoreEnvironmentBean.stateStore.objectStoreDir=$OBJECT_STORE_PATH`

> Can we close this one as will not fix and can we also mark the artefact [org.jboss.narayana.rts:lra-coordinator-quarkus](https://github.com/jbosstm/narayana/tree/master/rts/lra/coordinator-quarkus) as test only and move it under the test directory since it is not usable outside of a narayana test environment.

This is not a good idea as we are still shipping quarkus coordinator as a docker image - https://github.com/jboss-dockerfiles/narayana/tree/master/lra/lra-coordinator

mmusgrov:
This artefact is currently only usable for testing. We normally include quickstarts for components that are intended for community use. If you think it is configurable then can you create a quickstart for it. But if you agree that it is not configurable then I will raise a PR which adds configurability to it.

xstefank:
The question is whether this is still valid as there will be a Quarkus extension that should be buildable as native executable. If we will have this extension in the following weeks (hopefully) then I see no value in merging this. Dockerfile or the docker repository can build the native lra-coordinator executable for the docker image to be published to the community and to be used in the quickstarts.

mmusgrov:
@xstefank
1. I will update one of the existing quickstarts to document how how to use the component.
2. I will also update the README in the main repo with the same information.
ie you don't need to do anything

xstefank:
OK, I am closing this PR, we will continue with the LRA extension in Quarkus.

PR #1720, opened 2020-11-12
JBTM-3387 NarayanaLRAClientIT#testGetAllLRAs doesn't close started LRA

xstefank:
https://issues.redhat.com/browse/JBTM-3387

### DIFF
--- a/rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/client/NarayanaLRAClientIT.java
+++ b/rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/client/NarayanaLRAClientIT.java
@@ -52,8 +52,12 @@ public class NarayanaLRAClientIT {
         URI lra = lraClient.startLRA("test-lra");
 
         List<LRAData> allLRAs = lraClient.getAllLRAs();
-
         Assert.assertTrue(allLRAs.stream().anyMatch(lraData -> lraData.getLraId().equals(lra.toASCIIString())));
+
+        lraClient.closeLRA(lra);
+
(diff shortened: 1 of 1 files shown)


rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/client/NarayanaLRAClientIT.java
```diff
@@ -52,8 +52,12 @@ public void testGetAllLRAs() {
         URI lra = lraClient.startLRA("test-lra");
 
         List<LRAData> allLRAs = lraClient.getAllLRAs();
-
         Assert.assertTrue(allLRAs.stream().anyMatch(lraData -> lraData.getLraId().equals(lra.toASCIIString())));
+
+        lraClient.closeLRA(lra);
```
ochaloup, line 57: Would not be good to put this under `finally` block?

ochaloup (approved):
Approving. Just asking about close in finally.

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/client/NarayanaLRAClientIT.java
xstefank: I don't think there is a need. We don't do that in other tests AFAICT. We expect that the LRA coordinator is restarted (thus LRAs are cleared) between runs. Maybe we can go through all tests and fix this, but personally, I don't think that it is necessary.

mmusgrov (approved):

PR #1725, opened 2020-11-13
[JBTM-3393] narayana LRA client: cleaning and making it more flexible 

ochaloup:
https://issues.redhat.com/browse/JBTM-3393

LRA

(note the other commits are added here to get test running but will be removed when the other PRs will be committed)

### DIFF
--- a/rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
+++ b/rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
@@ -51,6 +51,7 @@ import javax.ws.rs.core.Link;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriInfo;
+import javax.ws.rs.core.UriBuilder;
 import java.io.Closeable;
 import java.io.UnsupportedEncodingException;
 import java.lang.annotation.Annotation;
@@ -62,6 +63,7 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.time.temporal.ChronoUnit;
 import java.util.Arrays;
@@ -73,7 +75,6 @@ import static io.narayana.lra.LRAConstants.AFTER;
 import static io.narayana.lra.LRAConstants.CLIENT_ID_PARAM_NAME;
 import static io.narayana.lra.LRAConstants.COMPENSATE;
 import static io.narayana.lra.LRAConstants.COMPLETE;
-import static io.narayana.lra.LRAConstants.COORDINATOR_PATH_NAME;
 import static io.narayana.lra.LRAConstants.FORGET;
 import static io.narayana.lra.LRAConstants.LEAVE;
 import static io.narayana.lra.LRAConstants.PARENT_LRA_PARAM_NAME;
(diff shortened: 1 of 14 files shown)


mmusgrov (approved):
The LRA client (NarayanaLRAClient) is for use by the implementation. It is in a client package because the MP-LRA spec used to include a client API but that was removed. At some point we need to move it into an internal package to deter people from using it (and reintroduce it in a later release if that becomes a requirement).

Some of the grammar is difficult to parse as an English speaker, at some time in the future I'd like to go through the javadoc and put my slant on it.

When we write the docs we must remember include the properties to configure coordinator endpoints.

Will you be checking that the quickstarts run okay with these changes?

ochaloup:
@mmusgrov +1 about the quickstart. I haven't realize this as a need. I will do it shortly.

rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
```diff
@@ -530,15 +539,15 @@ public LRAStatus getStatus(URI uri) throws WebApplicationException {
...
                 .request()
                 .get();
 
             if (response.getStatus() == NOT_FOUND.getStatusCode()) {
-                String responseContent = response.readEntity(String.class);
-                String errorMsg = "The requested LRA it '" + lraId + "' was not found and the status can't be obtained"
-                    + (responseContent != null ? ": " + responseContent : "");
+                String responseEntity = response.hasEntity() ? response.readEntity(String.class) : "";
+                String errorMsg = "The requested LRA it '" + lraId + "' was not found and the status can't be obtained, "
+                        + ", response content: " + responseEntity;
```
xstefank, line 550: this will be `...obtained, , response content:...`

rts/lra/coordinator/pom.xml
```diff
@@ -18,13 +18,9 @@
             -Djboss.bind.address=[::1] -Djboss.bind.address.management=[::1] -Djboss.bind.address.unsecure=[::1]
             -Djava.net.preferIPv4Stack=false -Djava.net.preferIPv6Addresses=true
         </ipv6.server.jvm.args>
-        <modular.jdk.args/>
-        <modular.jdk.props/>
-        <!-- the resource path of the coordinator - must match the deployment URL (see arquillian.xml) -->
+        <lra.coordinator.host>localhost</lra.coordinator.host>
+        <lra.coordinator.port>8080</lra.coordinator.port>
```
xstefank, line 22: is this still used since you've switched to URL?

rts/lra/coordinator/pom.xml
```diff
@@ -142,6 +138,10 @@
             <activation>
                 <activeByDefault>false</activeByDefault>
             </activation>
+            <properties>
+                <lra.coordinator.host>localhost</lra.coordinator.host>
```
xstefank, line 142: is this needed when there is url?

ochaloup: Yes. It's needed as arquillian.xml defines `managementAddress` which is used by arq extension to find where the management can be contacted. This differentiation is needed on ipv4 vs ipv6 here (as it's already defined here in `pom.xml`)

rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
ochaloup: +1

rts/lra/coordinator/pom.xml
ochaloup: you are right, the `port` won't be needed, that's overridden with use of `lra.coordinator.url`

ochaloup:
note about quickstarts: quickstarts work fine now as they use the default host/port configuration for coordinator. Still when this will be merged I will create a PR for changing poms and readmes there.

xstefank (approved):

PR #1729, opened 2020-11-25
JBTM-3397 Ensure that LRA uses the features provided by ArjunaCore

mmusgrov:
https://issues.redhat.com/browse/JBTM-3397

LRA

This PR address the issues 1-5 described in the JIRA. Note that JIRA says we will follow up with two more to address:

6. Don't use RecordType.USER_DEF_FIRST0 for LRA participants
7. Use the support in ArjunaCore for managing nested LRAs.

### DIFF
--- a/rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/TestBase.java
+++ b/rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/TestBase.java
@@ -22,6 +22,7 @@
 package io.narayana.lra.coordinator;
 
 import io.narayana.lra.client.NarayanaLRAClient;
+import io.narayana.lra.coordinator.domain.model.LongRunningAction;
 import io.narayana.lra.logging.LRALogger;
 import org.eclipse.microprofile.lra.annotation.LRAStatus;
 import org.jboss.arquillian.container.test.api.Config;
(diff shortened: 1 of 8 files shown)


rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/TestBase.java
```diff
@@ -168,7 +169,7 @@ private void clearRecoveryLog() {
     }
 
     String getFirstLRA() {
-        Path lraDir = Paths.get(storeDir.toString(), "ShadowNoFileLockStore", "defaultStore", "StateManager", "BasicAction", "TwoPhaseCoordinator", "LRA");
+        Path lraDir = Paths.get(storeDir.toString(), "ShadowNoFileLockStore", "defaultStore", LongRunningAction.getType());
```
xstefank, line 172: just guessing but this was written separately to not include delimiters ('/') in the path I guess so this won't work on windows.

xstefank (approved):
+1 much more readable and easier to understand!

rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/TestBase.java
mmusgrov: That is a very good point. In fact the whole method is assuming the ObjectStore is on a file system (so the test would fail, for example, on JDBC and in-memory stores) . I will raise a separate issue to rewrite the test to use the ObjectStore API to get the log (and to also check we aren't doing something similar elsewhere).

mmusgrov: I will make sure the JIRA says that we need the ability to run the tests on at lease one other store type.

ochaloup (approved):
+1
Just it would be good to see CI passing before merging (if possible in current times).

rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/TestBase.java
mmusgrov: The tracking issue is https://issues.redhat.com/browse/JBTM-3398

PR #1730, opened 2020-11-26
Make 4.17 testable on JDK6

ochaloup:
This PR contains fixes for 4.17 branch to be testable on JDK6.

* It downgrades the maven version to 3.2.5 which is capable to run on JDK6
* It makes the release script to not create a release to maven central but only a tag on github
* Chages the CI script to make a way to define different JDK version for building and different JDK version for running the tests
* Chages in test sources as the tests (in contrast to Narayana runtime) were not ready to be run on JDK6


mmusgrov:
It's a big one but needed, I will wait for the tests to run before going through it in detail.

pom.xml
```diff
@@ -412,7 +412,7 @@
     <version.org.slf4j>1.6.1</version.org.slf4j>
     <version.dom4j>1.6.1</version.dom4j>
     <version.jfree>1.0.9</version.jfree>
-    <version.com.h2database>1.4.195</version.com.h2database>
+    <version.com.h2database>1.4.191</version.com.h2database>
```
tomjenkinson, line 415: Is the H2 downgrade definitely required?

narayana-release-process.sh
```diff
@@ -52,7 +52,7 @@ then
   git clean -f -d
   # Add -x (this will delete all files (e.g. IDE, new features) not under source control)
   rm -rf $PWD/localm2repo
-  ./build.sh clean deploy -Dmaven.repo.local=${PWD}/localm2repo -Prelease,all -Dmaven.javadoc.skip=true -DskipTests -pl :jbossjts-jacorb,:jbossxts,:jbossjts-integration,:byteman_support,:jbosstxbridge -am
+  ./build.sh clean install -Dmaven.repo.local=${PWD}/localm2repo -Prelease,all -Dmaven.javadoc.skip=true -DskipTests -pl :jbossjts-jacorb,:jbossxts,:jbossjts-integration,:byteman_support,:jbosstxbridge -am
```
tomjenkinson, line 55: Noted, thanks

pom.xml
ochaloup: Yes, it's needed. This H2 version is last running with JDK6.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/FailAfterPrepareBase.java
```diff
@@ -50,6 +52,12 @@
     private SimpleXAResource xaResource;
     private CommitMarkableResourceRecordRecoveryModule recoveryModule;
 
+    @Before
+    public void setUp() {
+        // not running with JDK6
+        Assume.assumeTrue(!System.getProperty("java.version").startsWith("1.6"));
```
mmusgrov, line 58: Why do we only run the tests on JDK 1.6?

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/TestCommitMarkableResourceBase.java
```diff
@@ -49,7 +50,7 @@ public final void setup() throws Exception {
 
 		File file = new File(System.getProperty("user.dir") + "/ObjectStore");
 		if (file.exists()) {
-			Utils.removeRecursive(file.toPath());
+			FileUtils.deleteRecursive(file.getAbsolutePath(), false);
```
mmusgrov, line 53: What was the problem with `Utils.removeRecursive`? Is it using features that aren't in 1.6?

scripts/hudson/narayana.sh
```diff
@@ -123,11 +123,38 @@ function check_if_pull_closed
 function build_narayana {
   echo "Building Narayana"
   cd $WORKSPACE
-  [ $NARAYANA_TESTS = 1 ] && NARAYANA_ARGS= || NARAYANA_ARGS="-DskipTests"
 
-  ./build.sh -Prelease,community,all$OBJECT_STORE_PROFILE -Didlj-enabled=true "$@" $NARAYANA_ARGS $IPV6_OPTS -B clean install
+  if [ -n "$JAVA_HOME_FOR_BUILD" ]; then
```
mmusgrov, line 127: I cannot find any documentation that describes what this `env` variable does. I can see it skipping some things but shouldn't the build instructions describe what it's for?

scripts/hudson/narayana.sh
```diff
@@ -385,6 +412,7 @@ export ANT_OPTS="$ANT_OPTS $IPV6_OPTS"
 
 # run the job
 [ $NARAYANA_BUILD = 1 ] && build_narayana "$@"
+[ $NARAYANA_TESTS = 1 ] && test_narayana "$@"
```
mmusgrov, line 415: Phew, I am so pleased that this is the last change that needs reviewing.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/FailAfterPrepareBase.java
ochaloup: It's the test for not being run on JDK 1.6 as there is the trouble as test checks for suppressed exceptions. Skipping these few tests on 1.6 was the shortcut to proceed with testing on JDK6.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/TestCommitMarkableResourceBase.java
ochaloup: Yes. It's from 1.7 in Java.

scripts/hudson/narayana.sh
ochaloup: That's a good point. I will add a comment in the `sh` script if it's fine.

pom.xml
tomjenkinson: Thanks

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/FailAfterPrepareBase.java
tomjenkinson: Will we run these tests for JDK7+ still?

mmusgrov: Is that because we aren't allowed to do introspection to see if suppressed exceptions are available (I think that is what the later code did)? And is there no check for 1.5 because we don't support 1.5?

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/TestCommitMarkableResourceBase.java
mmusgrov: So do you mean that the code inside FileUtils.deleteRecursive() is using features that are not available 1.6

scripts/hudson/narayana.sh
```diff
@@ -123,11 +123,41 @@ function check_if_pull_closed
 function build_narayana {
   echo "Building Narayana"
   cd $WORKSPACE
-  [ $NARAYANA_TESTS = 1 ] && NARAYANA_ARGS= || NARAYANA_ARGS="-DskipTests"
 
-  ./build.sh -Prelease,community,all$OBJECT_STORE_PROFILE -Didlj-enabled=true "$@" $NARAYANA_ARGS $IPV6_OPTS -B clean install
+  # JAVA_HOME_FOR_BUILD can be used in environment if JDK for building Narayana needs to be different from one which runs e.g. tests.
+  # This is an option used for testing on JDK6. The Narayana 4.17 requires JDK7 for building but target version is JDK6.
```
mmusgrov, line 128: The pom says 1.6 is used for the source and target which doesn't match this statement:
```
         <artifactId>maven-compiler-plugin</artifactId>
          <version>2.3.2</version>
          <configuration>
            <source>1.6</source>
            <target>1.6</target>
            <compilerArgument>-AtranslationFilesPath=${project.basedir}/classes/</compilerArgument>
          </configuration>
```

Do we need to change the <source>1.6</source> parameter?

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/TestCommitMarkableResourceBase.java
ochaloup: Sorry, I probably explained wrongly.
The `FileUtils` uses the code which is available in 1.6. That's the reason why I took that class it into the changeset.
The `Utils` class uses the file nio utils methods which are not available on JDK6. That's the reason for replacing it.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/FailAfterPrepareBase.java
ochaloup: @tomjenkinson  Yes. It will be still running on JDK7+.
@mmusgrov It's because the tests assertion work with the suppressed exceptions. I would need to change the logic of the tests which I don't want to do.
Yes, there is not check for 1.5 as it's not possible to start Narayana on JDK1.5 (aka. we do not support it).

scripts/hudson/narayana.sh
ochaloup: That's a good point. The code can't be build on JDK6 as the code uses methods which are not available in JDK6 (e.g. suppressed exceptions). On the other side the code does not use any JDK7 specific language syntax. That way the source is 1.6 compatible but cannot be built on JDK6. It's a question how we want to approach it. Maybe changing to 1.7 would be better for guiding the user. I'm going to change it.

ochaloup: Ok, my wrong. The `maven-compiler-plugin` requires the `source` being the same (or lower probably) than the `target`. As we need to build for JDK6 then the `source` parameter *must* be 1.6.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/FailAfterPrepareBase.java
mmusgrov: I meant why aren't we using refection to see if the method exists like we do in the main code (
https://github.com/jbosstm/narayana/blob/4.17.43.Final/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/TransactionImple.java#L1684) rather than checking for specific JDK versions.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/TestCommitMarkableResourceBase.java
mmusgrov: Ah right, got you thanks. I'll mark this conversation as resolved then.

scripts/hudson/narayana.sh
mmusgrov: My point is that the new comment contradicts what the pom says. The pom says source is 1.6 whereas your new comment says that "Narayana 4.17 requires JDK7 for building".

mmusgrov: So the pom says we can only use JDK 1.6 features in the source.
Do we require JDK 1.7 for compiling because our version of maven requires it or that we have some pom dependencies that are incompatible with JDK1.6?

mmusgrov: For clarification, the new comment in the code means:
```
We target 1.6 byte code.
And the java source is limited to 1.6 language features only.

But we must build on 1.7 because we are using a JDK API that only exists in JDK 1.7.

That is what we mean by "Narayana 4.17 requires JDK7 for building"?
```

ochaloup: +1, I'm going to adjust the comment to be more understandable.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/commitmarkable/FailAfterPrepareBase.java
ochaloup: By my opinion we can use both approaches here. The reflection targets the difference more precisely as we do use from JDK API only this method.

ochaloup: And the reason why I chose to use the check for the java version is that it sounds me more reasonable for the test approach.

ochaloup: I will use the similar approach as for TransactionImple to be done in the same way

mmusgrov (approved):

scripts/hudson/narayana.sh
```diff
@@ -123,11 +123,38 @@ function check_if_pull_closed
...
 
+  if [ -n "$pathOriginal" ]; then
+     export PATH="$pathOriginal"
+     export JAVA_HOME="$javaHomeOriginal"
+  fi
+
+  return 0
+}
+
+function test_narayana {
```
tomjenkinson, line 152: I wonder if we should put a commit like this on master too in case we need to do something similar in future? (ability to run tests with a different JDK)

scripts/hudson/narayana.sh
```diff
@@ -124,13 +124,17 @@ function build_narayana {
...
+  # Narayana targets 1.6 byte code. And the java source is limited to 1.6 language features only.
+  # We must build on 1.7 because we are using a JDK API that only exists in JDK 1.7.
   if [ -n "$JAVA_HOME_FOR_BUILD" ]; then
     local pathOriginal="$PATH"
     local javaHomeOriginal="$JAVA_HOME"
     export JAVA_HOME="$JAVA_HOME_FOR_BUILD"
     export PATH="$JAVA_HOME/bin:$PATH"
   fi
-
+FailAfterPrepareBase.java
```
tomjenkinson, line 137: This looks wrong to me, is it expected?

scripts/hudson/narayana.sh
tomjenkinson: I note there is a further comment in https://github.com/jbosstm/narayana/pull/1730/commits/0323f19dec710bfd134f5bdd7536dca2d0b5bb33#diff-234fb7d762fdc80a3f29e8d8f4e0669ad47101d550eb3724420e91f64ee044de that could be useful to adapt too. But I expect some of the references to specific JDK versions like JDK6, 1.7 and 1.6 in the script and commit message would be revised perhaps to be more future looking.

ochaloup: Currently there is not such need (as far as I know). If it sounds useful to you I can prepare it.
WDYT?

scripts/hudson/narayana.sh
ochaloup: +1, thanks for spotting this. This is a dumb copy&paste error.

ochaloup:
I created a new PR (as this seems to be merged already) with a fix for the error in narayana.sh script - it's here https://github.com/jbosstm/narayana/pull/1733

scripts/hudson/narayana.sh
ochaloup: https://github.com/jbosstm/narayana/pull/1733

PR #1734, opened 2020-12-01
[JBTM-3307] enabling LRACoordinatorRecovery2TestCase with timeout adjustments

ochaloup:
https://issues.redhat.com/browse/JBTM-3307

LRA

### DIFF
--- a/rts/lra/coordinator/pom.xml
+++ b/rts/lra/coordinator/pom.xml
@@ -163,6 +163,7 @@
                                 <lra.coordinator.host>${lra.coordinator.host}</lra.coordinator.host>
                                 <lra.coordinator.url>${lra.coordinator.url}</lra.coordinator.url>
                                 <server.jvm.args>${server.jvm.args} -Dlra.coordinator.url=${lra.coordinator.url}</server.jvm.args>
+                                <lra.test.timeout.factor>${lra.test.timeout.factor}</lra.test.timeout.factor>
                             </systemPropertyVariables>
                             <systemProperties>
                                 <arquillian.xml>arquillian.xml</arquillian.xml>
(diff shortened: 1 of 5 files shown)


rts/lra/coordinator/pom.xml
```diff
@@ -163,6 +163,7 @@
                                 <lra.coordinator.host>${lra.coordinator.host}</lra.coordinator.host>
                                 <lra.coordinator.url>${lra.coordinator.url}</lra.coordinator.url>
                                 <server.jvm.args>${server.jvm.args} -Dlra.coordinator.url=${lra.coordinator.url}</server.jvm.args>
+                                <lra.tck.timeout.factor>${lra.tck.timeout.factor}</lra.tck.timeout.factor>
```
xstefank, line 166: maybe it shouldn't be named with `tck` in the name.

xstefank (approved):

ochaloup:
There is a failure which I haven't hit during my investigation (I run LRA several times with AMS CI with my changes).
It's  with `LRACoordinatorRecovery1TestCase.testRecovery` and I haven't seen it yet. I need to investigate.

rts/lra/coordinator/pom.xml
ochaloup: That's a good point. I was thinking if it's good to use it with the name `tck` in it but then I leave it as it is as it's how it's used.
I feel it's not a great approach. I will try to come with a general name.

mmusgrov (approved):

rts/lra/coordinator/pom.xml
ochaloup: @xstefank I changed the naming. If you have a while, please, take a look.

ochaloup:
The trouble with `LRACoordinatorRecovery1TestCase.testRecovery` was the fact the both tests are running as client side and the first test does not clean the `LRAClient` context from the started LRAs and the following test was trying to join the cancelled LRA. Hopefully fixed now.

xstefank (approved):
LGTM. Can be merged.

PR #1735, opened 2020-12-02
JDK6 changes to run on CI testing (4.17)

ochaloup:
!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB

I need to provide few more changes in 4.17 branch as not all axes work fine on JDK6 and AMS CI.
I ran jobs with this additional set of changes and with them it all seems should be working finally now.

### DIFF
--- a/scripts/hudson/narayana.sh
+++ b/scripts/hudson/narayana.sh
@@ -138,6 +138,7 @@ function build_narayana {
   echo "Using Java version, Maven version with JAVA_HOME='$JAVA_HOME'"
   ./build.sh -version
 
+  # building with -Dtest + -DfailIfNoTests to get downloaded test dependencies to local maven repo on the build time
   ./build.sh -Prelease,community,all$OBJECT_STORE_PROFILE -Didlj-enabled=true "$@" $NARAYANA_ARGS -Dtest=NonExistentTest -DfailIfNoTests=false -P${ARQ_PROF} $IPV6_OPTS -B clean install
   [ $? = 0 ] || fatal "narayana build failed"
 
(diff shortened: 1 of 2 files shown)


scripts/hudson/narayana.sh
```diff
@@ -208,11 +208,25 @@ function build_as {
...
+    local pathOriginal="$PATH"
+    local javaHomeOriginal="$JAVA_HOME"
+    export JAVA_HOME="$JAVA_HOME_FOR_BUILD"
+    export PATH="$JAVA_HOME/bin:$PATH"
+  fi
+
   export MAVEN_OPTS="$MAVEN_OPTS -Xms2048m -Xmx2048m -XX:MaxPermSize=1024m"
   export JAVA_OPTS="$JAVA_OPTS -Xms1303m -Xmx1303m -XX:MaxPermSize=512m"
-  (cd .. && ./build.sh -f jboss-as/pom.xml -B clean install -DskipTests -Dts.smoke=false $IPV6_OPTS -Drelease=true -Dversion.org.jboss.jboss-transaction-spi=7.1.0.SP2 -Dversion.org.jboss.jbossts.jbossjts=4.17.44.Final-SNAPSHOT -Dversion.org.jboss.jbossts.jbossjts-integration=4.17.44.Final-SNAPSHOT -Dversion.org.jboss.jbossts.jbossxts=4.17.44.Final-SNAPSHOT -Dversion.org.jboss.jbossts=4.17.44.Final-SNAPSHOT $AS_XARGS)
+  (cd .. && ./build.sh -f jboss-as/pom.xml -B clean install -Dtest=NonExistentTest -DfailIfNoTests=false -Dts.smoke=false $IPV6_OPTS -Drelease=true -Dversion.org.jboss.jboss-transaction-spi=7.1.0.SP2 -Dversion.org.jboss.jbossts.jbossjts=4.17.44.Final-SNAPSHOT -Dversion.org.jboss.jbossts.jbossjts-integration=4.17.44.Final-SNAPSHOT -Dversion.org.jboss.jbossts.jbossxts=4.17.44.Final-SNAPSHOT -Dversion.org.jboss.jbossts=4.17.44.Final-SNAPSHOT $AS_XARGS)
```
mmusgrov, line 222: Yes I agree with Tom. This command will fail every time since `NonExistentTests` don't exist.

ochaloup: The build is run with `-DfailIfNoTests=false` which ensures there won't be a failure.
The reason for this "hack" is that the NonExistentTest is used to not skip the maven surefire plugin dependency resolution and to download even the test dependencies to maven local repo.
I will add a comment to clarify why this cryptic stuff is used here.
Thanks for objecting this.

tomjenkinson: Thanks!

mmusgrov (approved):

tomjenkinson (approved):

PR #1738, opened 2020-12-04
JBTM-3245 report request and response filter processing failures back…

mmusgrov:
https://issues.redhat.com/browse/JBTM-3245

This PR adds error handling and reporting (to the caller). In particular it verifies the behaviour when a coordinator cannot be contacted and produces a response back to the caller indicating that the LRA is potentially in doubt.

I tested the code by break-pointing the coordinator at the appropriate places to simulate coordinator unavailability. I'm not sure if we need to simulate that using byteman in the future but for now I think doing it manually is sufficient.

LRA

### DIFF
--- a/rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
+++ b/rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
@@ -56,7 +56,6 @@ import java.io.Closeable;
 import java.io.UnsupportedEncodingException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
-import java.net.ConnectException;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
@@ -70,6 +69,9 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import static io.narayana.lra.LRAConstants.AFTER;
 import static io.narayana.lra.LRAConstants.CLIENT_ID_PARAM_NAME;
@@ -85,7 +87,6 @@ import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
 import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
 import static javax.ws.rs.core.Response.Status.NOT_ACCEPTABLE;
 import static javax.ws.rs.core.Response.Status.PRECONDITION_FAILED;
-import static javax.ws.rs.core.Response.Status.SERVICE_UNAVAILABLE;
 import static javax.ws.rs.core.Response.Status.GONE;
 import static javax.ws.rs.core.Response.Status.NOT_FOUND;
 import static javax.ws.rs.core.Response.Status.NO_CONTENT;
@@ -94,6 +95,9 @@ import static io.narayana.lra.LRAConstants.RECOVERY_COORDINATOR_PATH_NAME;
 import static io.narayana.lra.LRAConstants.COORDINATOR_PATH_NAME;
 
 /**
+ * WARNING: NarayanaLRAClient is an internal utility class and is subject to change
+ * (however it will be made available with improved guarantees in a subsequent release).
+ *
  * A utility class for controlling the lifecycle of Long Running Actions (LRAs) but the preferred mechanism is to use
  * the annotation in the {@link org.eclipse.microprofile.lra.annotation} package
  */
@@ -113,6 +117,14 @@ public class NarayanaLRAClient implements Closeable {
     private static final String CANCEL_PATH = "/%s/cancel";
 
     private static final String LINK_TEXT = "Link";
+
+    /**
+     * constrain how long client operations take before giving up
+     *
+     * WARNING: NarayanaLRAClient is an internal utility class and is subject to change
+     * (but it will be made available with improved guarantees in a subsequent release).
+     */
+    private static final long CLIENT_TIMEOUT = Long.getLong("lra.internal.client.timeout", 10);
 
     private URI coordinatorUrl;
 
@@ -191,8 +203,10 @@ public class NarayanaLRAClient implements Closeable {
         try {
             client = getClient();
             Response response = client.target(coordinatorUrl)
-                .request()
-                .get();
+                        .request()
+                        .async()
+                        .get()
+                        .get(CLIENT_TIMEOUT, TimeUnit.SECONDS);
 
             if (response.getStatus() != OK.getStatusCode()) {
                 LRALogger.logger.debugf("Error getting all LRAs from the coordinator, response status: %d", response.getStatus());
@@ -200,6 +214,9 @@ public class NarayanaLRAClient implements Closeable {
             }
 
             return response.readEntity(new GenericType<List<LRAData>>() {});
+        } catch (InterruptedException | ExecutionException | TimeoutException e) {
+            throw new WebApplicationException("getAllLRAs client request timed out, try again later",
+                    Response.Status.SERVICE_UNAVAILABLE.getStatusCode());
         } finally {
             if (client != null) {
                 client.close();
@@ -235,7 +252,12 @@ public class NarayanaLRAClient implements Closeable {
         return startLRA(getCurrent(), clientID, timeout, unit);
     }
 
+
     public URI startLRA(URI parentLRA, String clientID, Long timeout, ChronoUnit unit) throws WebApplicationException {
+        return startLRA(parentLRA, clientID, timeout, unit, true);
+    }
+
+    public URI startLRA(URI parentLRA, String clientID, Long timeout, ChronoUnit unit, boolean verbose) throws WebApplicationException {
         Client client = null;
         Response response = null;
         URI lra;
@@ -265,13 +287,17 @@ public class NarayanaLRAClient implements Closeable {
                 .queryParam(TIMELIMIT_PARAM_NAME, Duration.of(timeout, unit).toMillis())
                 .queryParam(PARENT_LRA_PARAM_NAME, encodedParentLRA)
                 .request()
-                .post(null);
+                .async()
+                .post(null)
+                .get(CLIENT_TIMEOUT, TimeUnit.SECONDS);
 
             // validate the HTTP status code says an LRA resource was created
             if (isUnexpectedResponseStatus(response, Response.Status.CREATED)) {
                 String responseEntity = response.hasEntity() ? response.readEntity(String.class) : "";
-                LRALogger.i18NLogger.error_lraCreationUnexpectedStatus(response.getStatus(), responseEntity);
-                throwGenericLRAException(null, INTERNAL_SERVER_ERROR.getStatusCode(),
+                if (verbose) {
+                    LRALogger.i18NLogger.error_lraCreationUnexpectedStatus(response.getStatus(), responseEntity);
+                }
+                throwGenericLRAException(null, response.getStatus(),
                         "LRA start returned an unexpected status code: " + response.getStatus() + ", response '" + responseEntity + "'");
                 return null;
             }
@@ -281,32 +307,23 @@ public class NarayanaLRAClient implements Closeable {
 
             Current.push(lra);
 
+            return lra;
         } catch (UnsupportedEncodingException e) {
-            LRALogger.i18NLogger.error_invalidFormatToEncodeParentUri(parentLRA, e);
+            if (verbose) {
+                LRALogger.i18NLogger.error_invalidFormatToEncodeParentUri(parentLRA, e);
+            }
             throwGenericLRAException(null, INTERNAL_SERVER_ERROR.getStatusCode(),
                     "Cannot connect to the LRA coordinator: " + coordinatorUrl + " as provided parent LRA URL '" + parentLRA +
                             "' is not in URI format (" + e.getCause().getMessage() + ")");
             return null;
-        } catch (Exception e) {
-            LRALogger.i18NLogger.error_cannotContactLRACoordinator(coordinatorUrl, e);
-
-            if (e.getCause() != null && ConnectException.class.equals(e.getCause().getClass())) {
-                throwGenericLRAException(null, SERVICE_UNAVAILABLE.getStatusCode(),
-                        "Cannot connect to the LRA coordinator: " + coordinatorUrl + " (" + e.getCause().getMessage() + ")");
-            } else {
-                throwGenericLRAException(null, Response.Status.SERVICE_UNAVAILABLE.getStatusCode(), e.getMessage());
-            }
-
-            return null;
+        } catch (InterruptedException | ExecutionException | TimeoutException e) {
+            throw new WebApplicationException("start LRA client request timed out, try again later",
+                    Response.Status.SERVICE_UNAVAILABLE.getStatusCode());
         } finally {
             if (client != null) {
                 client.close();
             }
         }
-
-        // check that the lra is active
-        // isActiveLRA(lra);
-        return lra;
     }
 
     public void cancelLRA(URI lraId) throws WebApplicationException {
@@ -370,7 +387,9 @@ public class NarayanaLRAClient implements Closeable {
             response = client.target(coordinatorUrl)
                 .path(String.format(LEAVE_PATH, LRAConstants.getLRAUid(lraId)))
                 .request()
-                .put(Entity.text(body));
+                .async()
+                .put(Entity.text(body))
+            .get(CLIENT_TIMEOUT, TimeUnit.SECONDS);
 
             if (OK.getStatusCode() != response.getStatus()) {
                 LRALogger.i18NLogger.error_lraLeaveUnexpectedStatus(response.getStatus(),
@@ -378,6 +397,9 @@ public class NarayanaLRAClient implements Closeable {
                 throwGenericLRAException(null, response.getStatus(), "Leaving LRA " + lraId + " from coordinator " + coordinatorUrl
                     + " finished with unexpected response code: " + response.getStatusInfo());
             }
+        } catch (InterruptedException | ExecutionException | TimeoutException e) {
+            throw new WebApplicationException("leave LRA client request timed out, try again later",
+                    Response.Status.SERVICE_UNAVAILABLE.getStatusCode());
         } finally {
             if (client != null) {
                 client.close();
@@ -542,7 +564,9 @@ public class NarayanaLRAClient implements Closeable {
             response = client.target(coordinatorUrl)
                 .path(String.format(STATUS_PATH, LRAConstants.getLRAUid(uri)))
                 .request()
-                .get();
+                    .async()
+                .get()
+                    .get(CLIENT_TIMEOUT, TimeUnit.SECONDS);
 
             if (response.getStatus() == NOT_FOUND.getStatusCode()) {
                 String responseEntity = response.hasEntity() ? response.readEntity(String.class) : "";
@@ -579,13 +603,17 @@ public class NarayanaLRAClient implements Closeable {
                     Response.Status.INTERNAL_SERVER_ERROR.getStatusCode(),
                     "LRA coordinator returned an invalid status"
                 );
-                return null;
             }
+        } catch (InterruptedException | ExecutionException | TimeoutException e) {
+            throw new WebApplicationException("get LRA status client request timed out, try again later",
+                    Response.Status.SERVICE_UNAVAILABLE.getStatusCode());
         } finally {
             if (client != null) {
                 client.close();
             }
         }
+
+        return null;
     }
 
     /**
@@ -673,10 +701,9 @@ public class NarayanaLRAClient implements Closeable {
                     .queryParam(TIMELIMIT_PARAM_NAME, timelimit)
                     .request()
                     .header("Link", linkHeader)
-                    .put(Entity.text(compensatorData == null ? linkHeader : compensatorData));
-            } catch (WebApplicationException webApplicationException) {
-                throw new WebApplicationException(uri.toASCIIString(), GONE);
-            }
+                        .async()
+                    .put(Entity.text(compensatorData == null ? linkHeader : compensatorData))
+                .get(CLIENT_TIMEOUT, TimeUnit.SECONDS);
 
             String responseEntity = response.hasEntity() ? response.readEntity(String.class) : "";
             if (response.getStatus() == Response.Status.PRECONDITION_FAILED.getStatusCode()) {
@@ -706,6 +733,12 @@ public class NarayanaLRAClient implements Closeable {
                         "join " + lraId + " returned an invalid recovery URI '" + recoveryUrl + "' : " + responseEntity);
                 return null;
             }
+        } catch (WebApplicationException webApplicationException) {
+            throw new WebApplicationException(uri.toASCIIString(), GONE); // not sure why we think it's gone TODO
+        } catch (InterruptedException | ExecutionException | TimeoutException e) {
+            throw new WebApplicationException("join LRA client request timed out, try again later",
+                    Response.Status.SERVICE_UNAVAILABLE.getStatusCode());
+        }
         } finally {
             if (client != null) {
                 client.close();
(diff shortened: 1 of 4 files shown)


rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
```diff
@@ -265,13 +287,17 @@ public URI startLRA(URI parentLRA, String clientID, Long timeout, ChronoUnit uni
...
+                .async()
+                .post(null)
+                .get(CLIENT_TIMEOUT, TimeUnit.SECONDS);
 
             // validate the HTTP status code says an LRA resource was created
             if (isUnexpectedResponseStatus(response, Response.Status.CREATED)) {
                 String responseEntity = response.hasEntity() ? response.readEntity(String.class) : "";
-                LRALogger.i18NLogger.error_lraCreationUnexpectedStatus(response.getStatus(), responseEntity);
-                throwGenericLRAException(null, INTERNAL_SERVER_ERROR.getStatusCode(),
+                if (verbose) {
```
ochaloup, line 297: I wonder what is the benefit for having this `verbose` flag here. My opinion is that when the LRA is not created it's generally usable to understand the reason. The `throwGenericLRAException` was changed to contain this information in case of error. Then the `verbose` flag won't provide more information, will it?

rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
```diff
@@ -370,14 +387,19 @@ public void leaveLRA(URI lraId, String body) throws WebApplicationException {
             response = client.target(coordinatorUrl)
                 .path(String.format(LEAVE_PATH, LRAConstants.getLRAUid(lraId)))
                 .request()
-                .put(Entity.text(body));
+                    .async()
```
ochaloup, line 390:
```suggestion
                .async()
```

ochaloup: For me there is a strange formatting for these `async` calls and `get` calls. This formatting is used overall of the whole class. Is it intentional?

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -78,10 +83,13 @@
...
     private static final String SUSPENDED_LRA_PROP = "suspendLRA";
     private static final String NEW_LRA_PROP = "newLRA";
+    private static final String ABORT_WITH_PROP = "abortWith";
 
     private static final Pattern START_END_QUOTES_PATTERN = Pattern.compile("^\"|\"$");
     private static final long DEFAULT_TIMEOUT_MILLIS = 0L;
-
+    private static final String JOIN = "join";
+    // i18nMessageCode corresponding to lraI18NLogger#warn_LRAStatusInDoubt
+    private static final int i18nMessageCode = 25145;
```
ochaloup, line 92: What is the reason to explicitly provide the message code when it's printed by logger when logged?

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -78,10 +83,13 @@
     private static final String TERMINAL_LRA_PROP = "terminateLRA";
     private static final String SUSPENDED_LRA_PROP = "suspendLRA";
     private static final String NEW_LRA_PROP = "newLRA";
+    private static final String ABORT_WITH_PROP = "abortWith";
 
     private static final Pattern START_END_QUOTES_PATTERN = Pattern.compile("^\"|\"$");
     private static final long DEFAULT_TIMEOUT_MILLIS = 0L;
-
+    private static final String JOIN = "join";
```
ochaloup, line 90: Is this constant used somewhere?

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -460,46 +575,149 @@ private boolean isJaxRsCancel(ContainerRequestContext requestContext, ContainerR
...
+        }
     }
 
-    private String getCompensatorId(URI lraId, UriInfo uriInfo, Long timeout) {
-        Map<String, String> terminateURIs = NarayanaLRAClient.getTerminationUris(resourceInfo.getResourceClass(), uriInfo, timeout);
+    // convert the list of steps carried out by the filters into a warning message
+    private String processLRAOperationFailures(ArrayList<Progress> progress) {
+        StringJoiner badOps = new StringJoiner(", ");
+        StringJoiner goodOps = new StringJoiner(", ");
+        StringBuffer code = new StringBuffer("-");
```
ochaloup, line 631: Please, use ` StringBuilder`

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -460,46 +575,149 @@ private boolean isJaxRsCancel(ContainerRequestContext requestContext, ContainerR
...
+
+        /*
+         * return a string which encodes the result:
+         * <major code>-<failed op codes>-<successful op codes>: <details of failed ops> (<details of successful ops>)
+         *
+         * where
+         *
+         * <major code>: corresponds to the id of the message in the logs
+         * <failed op codes>: each digit corresponds to the enum ordinal calue of the ProgressStep enum value that was successful
+         * <successful op codes>: each digit corresponds to the enum ordinal calue of the ProgressStep enum value that failed
```
ochaloup, line 651:
```suggestion
         * <successful op codes>: each digit corresponds to the enum ordinal value of the ProgressStep enum value that failed
```

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -460,46 +575,149 @@ private boolean isJaxRsCancel(ContainerRequestContext requestContext, ContainerR
...
 
-    private void lraTrace(URI lraId, String reason) {
-        if (LRALogger.logger.isTraceEnabled()) {
-            Method method = resourceInfo.getResourceMethod();
-            LRALogger.logger.tracef("%s: container request for method %s: lra: %s%n",
-                    reason, method.getDeclaringClass().getName() + "#" + method.getName(),
-                    lraId == null ? "context" : lraId);
+    // add another step to the list of steps performed so far
+    private ArrayList<Progress> updateProgress(ArrayList<Progress> progress, ProgressStep step, String reason) {
+        if (progress == null) {
```
ochaloup, line 665: Could this be called from multiple threads?

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -385,10 +460,24 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
         try {
             if (current != null && isCancel) {
                 try {
-                    lraClient.cancelLRA(current);
+                    // do not attempt to cancel if the request filter tried but failed to start a new LRA
+                    if (progress == null || progressDoesNotContain(progress, ProgressStep.StartFailed)) {
+                        lraClient.cancelLRA(current);
+                        progress = updateProgress(progress, ProgressStep.Ended, null);
```
ochaloup, line 466: a minor thing, the `ClientLRAFilter` uses the call as `updateProgress(progres, ...)` while this class uses the format `progress = updateProgress(progress, ...`. Would not be better to use the same format of the call (probably just `updateProgress(...)`?

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -460,46 +575,149 @@ private boolean isJaxRsCancel(ContainerRequestContext requestContext, ContainerR
...
+        StringJoiner goodOps = new StringJoiner(", ");
+        StringBuffer code = new StringBuffer("-");
 
-        if (!terminateURIs.containsKey("Link")) {
-            throwGenericLRAException(lraId, Response.Status.BAD_REQUEST.getStatusCode(),
-                    "Missing complete or compensate annotations");
-        }
+        progress.forEach(p -> {
+            if (p.ok()) {
+                code.insert(0, p.progress.ordinal());
```
ochaloup, line 635: Using the enum `ordinal` for identity is not a good practice in java programming (https://wiki.sei.cmu.edu/confluence/display/java/DCL56-J.+Do+not+attach+significance+to+the+ordinal+associated+with+an+enum). There should be used a different approach for codes.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -460,46 +575,149 @@ private boolean isJaxRsCancel(ContainerRequestContext requestContext, ContainerR
...
+    // the following structure is used to track progress so that such failures can be reported in the response
+    // filter processing
+    private enum ProgressStep {
+        Left ("leave succeeded"),
+        LeaveFailed("leave failed"),
+        Started("start ok"),
+        StartFailed("start failed"),
+        Joined("join succeeded"),
+        JoinFailed("join failed"),
+        Ended("end ok"),
```
ochaloup, line 588: Should not be `"end suceeded"`?

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -460,46 +575,149 @@ private boolean isJaxRsCancel(ContainerRequestContext requestContext, ContainerR
...
+         * <details of failed ops>: comma separated list of failed operation details "<op name> (<exception message>)"
+         * <details of successful ops>: comma separated list of successful operation details "<op name> (<op description>)"
+         */
+
+        return badOps.length() != 0 ? String.format("%d-%s: %s (%s)", i18nMessageCode, code, badOps, goodOps) : null;
+    }
 
-        return terminateURIs.get("Link");
+    private boolean progressDoesNotContain(ArrayList<Progress> progress, ProgressStep step) {
+        return progress.stream().noneMatch(p -> p.progress == step);
```
ochaloup, line 660: When the progress data structure is used for seeking on existence would not be better to use the map structure?

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -460,46 +575,149 @@ private boolean isJaxRsCancel(ContainerRequestContext requestContext, ContainerR
...
+            return status;
+        }
     }
 
-    private void resumeTransaction(URI lraId) {
-        // nothing to do
+    // list of steps (both successful and unsuccesful) performed so far by the request and response filter
+    // and is used for error reporting
+    private static class Progress {
+        static EnumSet<ProgressStep> failures = EnumSet.of(
```
ochaloup, line 607: Would no this `EnumSet` be more appropriate to be part (a member) of the `ProgressStep` Enum itself?

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -460,46 +575,149 @@ private boolean isJaxRsCancel(ContainerRequestContext requestContext, ContainerR
...
+
+        ProgressStep progress;
+        String reason;
+
+        public Progress(ProgressStep progress, String reason) {
+            this.progress = progress;
+            this.reason = reason;
+        }
+
+        public boolean ok() {
```
ochaloup, line 622: Maybe this would be better to be name with `success`? These this term seems to be used more in the sources here than `ok` to say that's all fine.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -460,46 +575,149 @@ private boolean isJaxRsCancel(ContainerRequestContext requestContext, ContainerR
...
+        StringBuffer code = new StringBuffer("-");
 
-        if (!terminateURIs.containsKey("Link")) {
-            throwGenericLRAException(lraId, Response.Status.BAD_REQUEST.getStatusCode(),
-                    "Missing complete or compensate annotations");
-        }
+        progress.forEach(p -> {
+            if (p.ok()) {
+                code.insert(0, p.progress.ordinal());
+                goodOps.add(String.format("%s (%s)", p.progress.name(), p.progress.status));
```
ochaloup, line 636: Why the `ProgressStep#status` is reported for success and not for the failure?

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -431,6 +529,23 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
...
                         Response.Status.OK.getStatusCode());
             }
+
+            /*
+             * report any failed steps (ie if progress contains any failures) to the caller.
+             * If either filter encountered a failure they may have completed partial actions and
+             * we need tell the caller which steps failed and which ones succeeded. We use a
+             * different warning code for each scenario:
+             */
+            if (progress != null) {
```
ochaloup, line 539: Would not be clearer and faster not to concatenate the progress ops in case of succesful run (ie. the most probably much more often case than the failure)? 
Checking here for empy ops failure or similar, instead of checking the content of string of `failureMesssage`?

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -460,46 +575,149 @@ private boolean isJaxRsCancel(ContainerRequestContext requestContext, ContainerR
...
+                goodOps.add(String.format("%s (%s)", p.progress.name(), p.progress.status));
+            } else {
+                code.append(p.progress.ordinal());
+                badOps.add(String.format("%s (%s)", p.progress.name(), p.reason));
+            }
+        });
+
+        /*
+         * return a string which encodes the result:
+         * <major code>-<failed op codes>-<successful op codes>: <details of failed ops> (<details of successful ops>)
```
ochaloup, line 645: After a while of decoding I understand the idea here.
I would use probably a format but it's a matter of taste as discussed elsewhere.
I would rather ask here what happens when there are multiple failed (or successful) op codes. They seem to be appended without any delimiter, or not?

In fact I wonder what is the purpose of having these opcodes as part of the user readable failure message?

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -240,41 +286,65 @@ public void filter(ContainerRequestContext containerRequestContext) {
 
                         // if there is an LRA present nest a new LRA under it
                         suspendedLRA = incommingLRA;
-                        lraTrace(suspendedLRA, "ServerLRAFilter before: REQUIRED start new LRA");
-                        newLRA = lraId = startLRA(incommingLRA, method, timeout);
+
+                        if (progress == null) {
```
ochaloup, line 290: Why this check for null is needed when the `updateProgress` checks on `null`?

ochaloup (changes requested):
I have few points. 
I consider some of them deserve an adjustment. The most of them are questions to discuss the approach or trying to understand the processing.

rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
mmusgrov: The code is generating a stacktrace and that is not useful. The user wants a new LRA but the system responsible for fulfilling that request is not available. I don't think that is a good reason to start generating stack traces in the logs (this is a distributed system and Availability cannot be guaranteed - ie it is an expected condition).

LRA operates in an environment where failure/unavailability conditions are the norm and dumping stacktraces is not the way to go, it gives the sense that the s/w is buggy. We only need to generate a decent response back to the client and a warning in the logs for further analysis and correlation. And this is what this PR is trying to achieve.

rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
mmusgrov: Yes the whole file probably needs reformatting. I am not doing that in the PR so it can be the subject of another issue.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: Take a look at where the constant gets used.
It provides a way for correlating the response with what's in the logs.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: No. I will remove it.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: okay

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: It is a private method and there are lots of other words that mean the same thing.
If I change it to something you want then what about the next person who prefers something different.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: Because the stuff in the brackets is only there to give more information about the step.
The failure branch includes the exception message whereas in the successful branch it adds nothing other than a more verbose description of name().

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: Let's deal with performance enhancements when we have the primary functionality in place. There are many places throughout the code that need to be optimised.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: Not. The code is using `java.util.StringJoiner` which includes the delimiter in the constructor.
The purpose is for knowing which operations are potentially in doubt.
We need some response to tell the caller that something is in doubt and this gives him a fighting chance of figuring out what went wrong i.e. it's not meant for users to read, rather it is for application logic to reason about the status of the system.

You/we can provide something more comprehensive in MP release 1.x

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: Because it might be null and `updateProgress` might not have been called.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: I missed these comments, github had collapse them all. Anyway,

This code is part of a request response processing filter chain. The context for such a request is passed into the filter method and applies to the particular request context. If you check the variable `progress` you will see that it is a method local variable and therefore not subject to your implied concurrency concern. In addition the contents of `progress` are passed to the response filter via a context property (of the request context)n and again there are no concurrency concerns.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: This is a server side request filter and it is only on the server that we need to check for coordinator availability. The ClientLRAFilter is for client side processing and is of no concern to the goals of this PR.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: The advice you quote (DCL56-J) "Do not attach significance to the ordinal associated with an enum" does not apply.
I am not attaching any significance to the ordinal value and I am using it as an identifier, identifiers do not carry significance, other than the fact that they imply uniqueness.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: Maybe, or "join ok" would work too from a consistency point of view.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: I don't know. Let's get the functionality out there rather than getting the code absolutely perfect. There is plenty of scope for improving the code elsewhere.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: Not necessarily. Progress knows what constitutes a failure. Again, if we want this code to be perfect then different engineers have different concepts of perfection and such a strategy just serves to prolong the process.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
ochaloup: I don't agree. The `ordinal` is used here for definition of the op code - they are reported to user as a code with significance of particular state.
The main reason is maitanance - enum constant reordering means troubles, op codes are not unique as any other constant enum will need to be used the values will be same... Check e.g. Effective Java from Joshua Bloch, Item 35.

ochaloup (changes requested):
I have few points in the started discussions.

rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
mmusgrov: Exactly right, we should not be throwing exceptions and my code changes log the warning but avoid the exception.
The verbosity flag was added because I don't want the server side filters throwing exceptions. Existing code using this client API is unaffected (I still think it's wrong even for them but it isn't the subject of my PR).

rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
mmusgrov: Perhaps, but the formatting in this file has already diverged.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: I am not following your logic, all I've done is to include the log message id in the response to the client. I am not advocating anything complex here, we just need a story to tell the user.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: This is a trivial PR comment and is subjective. But if you're going to dig your feat in then I'll change it. I could be flippant and ask you for a coding style guide which passes your coding criteria, but I'll change it.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: We need to make progress and we don't have time for long protracted coding style debates.
After we have finally got the code in front of users we can take our foot off the pedal and you can rework it all to satisfy your sensibilities.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: I could rewrite the whole thing in many different ways. Code is an art as well as a science.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: What other method (that link just takes me to the top of the PR diffs). Is this one another complaint about my coding style not matching up with what you prefer.

ochaloup (changes requested):
After the discussion on different places at this PR I've resolved most of them.

There are remaining only few minor requests from my side (one formatting, some typos) for approving the PR. I have no more remarks now.

ochaloup (approved):
Fine, thanks.
Just there is a strange failure in TCK tests (`TckParticipantTests#testNonJaxRsCompletionStageVoid`). I've never seen that.

mmusgrov:
The failure is due to a request (from the server side LRA filter) timing out so it may be another transient. I have triggered CI again to retest the PR.

mmusgrov:
I increased the timeout on Future.get(). When we tackle JBTM-2997 (Ensure that any LRA coordinator instance can control any LRA) in a later release we can provide a better user experience.

PR #1741, opened 2020-12-10
[JBTM-3406] InboundBridge should recover only local tx

jhalliday:
https://issues.redhat.com/browse/JBTM-3406

MAIN XTS AS_TESTS 
!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !PERF NO_WIN !RTS !TOMCAT !JACOCO !LRA


mmusgrov (approved):
Thanks for the fix.
I will get it back ported to the other branches.

PR #1746, opened 2020-12-15
JBTM-3408 Include instuctions for new contributors to the Narayana pr…

mmusgrov:
https://issues.redhat.com/browse/JBTM-3408

!MAIN !CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF !LRA !NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

I took the quarkus CONTRIBUTING.md file and tweaked it to match our needs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,98 @@
+# Contributing guide
+
+**Want to contribute? Great!** 
+We try to make it easy, and all contributions, even the smaller ones, are more than welcome.
+This includes bug reports, fixes, documentation, examples... 
+But first, read this page.
+
+## Legal
+
+All original contributions to Narayana are licensed under the
+[GNU LESSER GENERAL PUBLIC LICENSE Version 2.1](https://github.com/jbosstm/narayana/blob/master/LICENSE).
+
+All contributions are subject to the [Developer Certificate of Origin (DCO)](https://developercertificate.org/).
+The DCO text is also included verbatim in the [dco.txt](dco.txt) file in the root directory of the repository.
+
+New files may contain a license header in the following format:
+
+> Copyright The Narayana Authors
+> 
+> SPDX-License-Identifier: LGPL-2.1-only
+
+## Reporting an issue
+
+This project uses the [JBTM issue tracker](https://issues.redhat.com/projects/JBTM)
+to manage the issues. Open an issue directly in this issue tracker.
+
+If you believe you found a bug, and it's likely possible, please indicate a way to reproduce it, what you are seeing and what you would expect to see.
+Don't forget to indicate your Narayana, Java and Maven version. 
+
+### Building master
+
+See the [README](README.md) file in the root directory of the repository.
+
+## Before you contribute
+
+To contribute, use GitHub Pull Requests (PRs), from your **own** fork.
+
+When you create a PR, the description field of the PR will include brief instructions on what you need to include.
+But the following guidelines provide a more detailed set of requirements that we have found useful:
+
+1. The Pull Request title is properly formatted: `JBTM-XYZ Subject`
+2. The Pull Request *should* contain a link to the JIRA issue(s) at the start of the PR description (only minor changes to script/text files are exempt from this rule). If the engineer wishes to address multiple issues and they are closely related then they can be addressed in a single PR. The JIRA must contain sufficient information to enable the reader to understand what the issue is, so at a minimum the description field of the JIRA must be present and legible/clear.
+3. Engineers are not allowed to submit PRs which only contain formatting changes. The guidance on formatting code are covered in the [Coding Guidelines](#coding-guidelines) section below.
+4. New PRs are tested against multiple Jenkins CI axes. If you know that a change only affects particular axes then you can disable the ones that aren't required. When you first create a PR, the description field will contain a basic instructions about how to disable a CI test axis.
+
+Also, make sure you have set up your Git authorship correctly:
+
+```
+git config --global user.name "Your Full Name"
+git config --global user.email your.email@example.com
+```
+
+If you use different computers to contribute, please make sure the name is the same on all your computers.
+
+### Code reviews
+
+All submissions, including submissions by project members, should be reviewed before being merged.
+A possible exception is that if a change only effects build scripts or non-source files and is "small" then self review is permitted. But use this option with care to avoid breaking build scripts or the readabilty of text files.
+
+Before asking for a review it's best to wait for [Continuous Integration tests](#continuous-integration) to finish successfully (unless early feedback is being sought).
+
+Once a review has started both parties should attempt to respond to feedback in a timely manner.
+Do not approve the PR until you have either seen a successful CI test of the PR, or you can reasonably explain why a failure is unrelated to the code changes made in the PR (and documented in a PR comment).
+
+### Guidelines
+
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
...
+We try to make it easy, and all contributions, even the smaller ones, are more than welcome.
+This includes bug reports, fixes, documentation, examples... 
+But first, read this page.
+
+## Legal
+
+All original contributions to Narayana are licensed under the
+[GNU LESSER GENERAL PUBLIC LICENSE Version 2.1](https://github.com/jbosstm/narayana/blob/master/LICENSE).
+
+All contributions are subject to the [Developer Certificate of Origin (DCO)](https://developercertificate.org/).
```
tomjenkinson, line 13: I think this could be socialised on the Narayana forum for some further feedback first but I would be fine with it

CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
...
+```
+git config --global user.name "Your Full Name"
+git config --global user.email your.email@example.com
+```
+
+If you use different computers to contribute, please make sure the name is the same on all your computers.
+
+### Code reviews
+
+All submissions, including submissions by project members, need to be reviewed before being merged.
```
tomjenkinson, line 43: I think this is "should be reviewed" I have pushed to master before. If we really want to say need to be reviewed then I suppose self review is fine but I expect our release process at least expects to push unreviewed commits up to master

CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
...
+
+Enable the checkstyle plugin via the pom file, or
+Reference the rule file from your IDE settings (for Intelij that would be Settings -> Other Settings -> Checkstyle).
+
+ * We do not use `@author` tags in the Javadoc in new code: they are hard to maintain, especially in a very active project, and we use the Git history to track authorship. GitHub also has [this nice page with your contributions](https://github.com/jbosstm/narayana/graphs/contributors).
+ * Commits should be atomic and semantic. Please properly squash your pull requests before submitting them. Fixup commits can be used temporarily during the review process but things should be squashed at the end to have meaningful commits.
+
+### Continuous Integration
+
+To ensure Narayana is stable for everyone, all changes must go through Narayana continuous integration: when you raise a pull request one of the members of the team will schedule a CI run to test your PR.
```
tomjenkinson, line 64: again, I think we should use "should" here so this is "all changes should go through Narayana continuous integration" (even when CI is done we don't do all axis always)

CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
...
+We follow [Open source participation guidelines](https://www.redhat.com/en/resources/open-source-participation-guidelines-overview)
+
+### Coding Guidelines
+
+We use a checkstyle plugin which is skipped by default because we have a lot of old files, that we don't want to change, which would fail the compilation.
+But on new code we do encourage/advise that it should follow our checkstyle rules.
+
+To see if your code follows our style guide you have two options:
+
+Enable the checkstyle plugin via the pom file, or
```
ochaloup, line 56: I think this could be hard to do for a developer (but I don't know how to better write the sentence, just saying)
My context is: 
* the checkstyle plugin is configured in root pom.xml
* it could be switched on by setting property on compilation `-Dcheckstyle.skip=false`
* when set up during compilation then it just completely fails as most of the code will not comply, ie. it's difficult to create a new file and try to check with checkstyle if it's ok
* setting up the property probably makes sense only in IDE
* the new modules - only the LRA does so now - enable the checkstyle and the checkstyle is verified during compilation automatically

ochaloup (approved):
It's nice. I've added just a short comment. It's not necessary to be addressed thus approving.

jmfinelli (commented):
I think that the "Contributing guide" looks good. Only one comment, in the root of the Narayana repository, the general README.md asks contributors to sign a CLA. Should that sentence be removed from that file or added to CONTRIBUTING.md?

mayankkunwar (approved):

CONTRIBUTING.md
mmusgrov: Great suggestion.

mmusgrov:
There was also a community request to say something about PR's without an assigned reviewer. Our response was to include something like:

```
Anyone can review a PR, as is normal for opensource projects. If the person that raised a PR just wanted to get the benefit of CI testing for his PR and is not ready for review then he just needs to add a note to the PR description to let people know his intent. 
If an engineer does choose to assign a reviewer then its best to wait for CI tests to finish successfully before doing so (unless early feedback is being sought).
```

CONTRIBUTING.md
mmusgrov: I raised the question (https://groups.google.com/g/narayana-users/c/rLR6w2RudIk). If there is still no input from the community by the end of the week then I will assume that's tacit agreement and we can go ahead with the mandate [of contributions being subject to the DCO].

mmusgrov:
I've put this on hold waiting for me to integrate all the PR comments.

mmusgrov:
@jmfinelli and @ochaloup I updated the PR and it's ready for another review

ochaloup (approved):
Seems fine to me.

PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,16 +1,9 @@
...
 
-Our style rules for submitting code are as follows:
-
-* If you add a new file it MUST adhere to our checkstyle ruleset.
-  1. If you change a file (in a non trivial way) you are allowed (MAY) to reformat the code to conform to our checkstyle ruleset (`org.jboss.narayana:checkstyle`).
-  2. If you choose not to reformat it then you SHOULD, where possible, try to follow the same style that's already in use for that file.
-
-
-The build axis can be controlled by prefixing a ! on the following as appropriate.
+The build axis can be controlled by prefixing a ! on the following as appropriate. Note that the default is not to test on windows (!NO_WIN).
```
jmfinelli, line 7: > [...] Note that the default is not to test on windows (!NO_WIN)

IIRC, NO_WIN means that the PR will not be tested on Windows, hence, !NO_WIN means that the PR will be tested on Windows (double negatives can be confusing). If this is the case, the above sentence should be changed with

[...] Note that the default behaviour is to test on Windows (!NO_WIN).

Maybe, to make it clearer, we could add an extra sentence saying:

[...] Note that the default behaviour is to test on Windows (!NO_WIN), therefore, if you want to disable testing the PR on Windows, use NO_WIN

mmusgrov: I left the text as it was before the PR.

But if you want to raise a separate one for this then go ahead, but bear in mind that I tried to keep the text minimal because most developers don't read the instructions anyway and if we add too many words there is even less chance that they will read it. In fact you would be better off changing the PROFILE to `WIN`.

jmfinelli: Agree :-)
Plus, (now that I think about it) the testing axis`NO_WIN` will be removed as soon as `BLACKTIE` gets deprecated.

jmfinelli (approved):

PR #1748, opened 2020-12-16
JBTM-3213 Allow request timeouts to be configurable. Enable throttlin…

mmusgrov:
https://issues.redhat.com/browse/JBTM-3213

LRA

This is a partial fix for JBTM-3213 and the PR increases the resilience of LRA to overload. The PR:

- allows throttling of LRA start requests to be configurable;
- allows clients to control how long coordinator requests can take before giving up. The PR adds separate properties for things like starting, joining, ending LRAs etc. We need separate properties since some request failures are more serious than others (eg failure to abort an LRA is significantly more problematic than are failures to start or join with LRAs).

There is a lot we can do in this area [fault resilience] and we have a few JIRAs for tackling the various requirements.
But this PR is a basic minimum to reduce the number of errors caused by overloading the system in which LRA operates.

### DIFF
--- a/rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
+++ b/rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
@@ -125,6 +125,11 @@ public class NarayanaLRAClient implements Closeable {
      * (but it will be made available with improved guarantees in a subsequent release).
      */
     private static final long CLIENT_TIMEOUT = Long.getLong("lra.internal.client.timeout", 10);
+    private static final long START_TIMEOUT = Long.getLong("lra.internal.client.timeout.start", CLIENT_TIMEOUT);
+    private static final long JOIN_TIMEOUT = Long.getLong("lra.internal.client.timeout.join", CLIENT_TIMEOUT);
+    private static final long END_TIMEOUT = Long.getLong("lra.internal.client.end.timeout", CLIENT_TIMEOUT);
+    private static final long LEAVE_TIMEOUT = Long.getLong("lra.internal.client.leave.timeout", CLIENT_TIMEOUT);
+    private static final long QUERY_TIMEOUT = Long.getLong("lra.internal.client.query.timeout", CLIENT_TIMEOUT);
 
     private URI coordinatorUrl;
 
@@ -206,7 +211,7 @@ public class NarayanaLRAClient implements Closeable {
                         .request()
                         .async()
                         .get()
-                        .get(CLIENT_TIMEOUT, TimeUnit.SECONDS);
+                        .get(QUERY_TIMEOUT, TimeUnit.SECONDS);
 
             if (response.getStatus() != OK.getStatusCode()) {
                 LRALogger.logger.debugf("Error getting all LRAs from the coordinator, response status: %d", response.getStatus());
@@ -289,7 +294,7 @@ public class NarayanaLRAClient implements Closeable {
                 .request()
                 .async()
                 .post(null)
-                .get(CLIENT_TIMEOUT, TimeUnit.SECONDS);
+                .get(START_TIMEOUT, TimeUnit.SECONDS);
 
             // validate the HTTP status code says an LRA resource was created
             if (isUnexpectedResponseStatus(response, Response.Status.CREATED)) {
@@ -389,7 +394,7 @@ public class NarayanaLRAClient implements Closeable {
                 .request()
                 .async()
                 .put(Entity.text(body))
-            .get(CLIENT_TIMEOUT, TimeUnit.SECONDS);
+            .get(LEAVE_TIMEOUT, TimeUnit.SECONDS);
 
             if (OK.getStatusCode() != response.getStatus()) {
                 LRALogger.i18NLogger.error_lraLeaveUnexpectedStatus(response.getStatus(),
@@ -566,7 +571,7 @@ public class NarayanaLRAClient implements Closeable {
                 .request()
                     .async()
                 .get()
-                    .get(CLIENT_TIMEOUT, TimeUnit.SECONDS);
+                    .get(QUERY_TIMEOUT, TimeUnit.SECONDS);
 
             if (response.getStatus() == NOT_FOUND.getStatusCode()) {
                 String responseEntity = response.hasEntity() ? response.readEntity(String.class) : "";
@@ -703,7 +708,7 @@ public class NarayanaLRAClient implements Closeable {
                     .header("Link", linkHeader)
                         .async()
                     .put(Entity.text(compensatorData == null ? linkHeader : compensatorData))
-                .get(CLIENT_TIMEOUT, TimeUnit.SECONDS);
+                .get(JOIN_TIMEOUT, TimeUnit.SECONDS);
 
             String responseEntity = response.hasEntity() ? response.readEntity(String.class) : "";
             if (response.getStatus() == Response.Status.PRECONDITION_FAILED.getStatusCode()) {
@@ -761,7 +766,7 @@ public class NarayanaLRAClient implements Closeable {
                         .request()
                         .async()
                         .put(null)
-                        .get(CLIENT_TIMEOUT, TimeUnit.SECONDS);
+                        .get(END_TIMEOUT, TimeUnit.SECONDS);
             } catch (InterruptedException | ExecutionException | TimeoutException e) {
                 throw new WebApplicationException("end LRA client request timed out, try again later",
                         Response.Status.SERVICE_UNAVAILABLE.getStatusCode());
--- a/rts/lra/coordinator/pom.xml
+++ b/rts/lra/coordinator/pom.xml
@@ -29,6 +29,12 @@
             <version>${version.microprofile.lra.api}</version>
         </dependency>
         <dependency>
+            <groupId>org.eclipse.microprofile.fault-tolerance</groupId>
+            <artifactId>microprofile-fault-tolerance-api</artifactId>
+            <version>${version.microprofile.fault-tolerance}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
             <groupId>org.jboss.narayana.rts</groupId>
             <artifactId>lra-service-base</artifactId>
             <version>${project.version}</version>
--- a/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
+++ b/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
@@ -63,6 +63,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
+import org.eclipse.microprofile.faulttolerance.Bulkhead;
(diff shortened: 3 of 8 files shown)


rts/lra/coordinator/pom.xml
```diff
@@ -28,6 +28,11 @@
             <artifactId>microprofile-lra-api</artifactId>
             <version>${version.microprofile.lra.api}</version>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.microprofile.fault-tolerance</groupId>
+            <artifactId>microprofile-fault-tolerance-api</artifactId>
+            <version>2.1.1</version>
```
ochaloup, line 34: The versions would be better to be configurable with properties.

rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
```diff
@@ -124,7 +124,11 @@
      * WARNING: NarayanaLRAClient is an internal utility class and is subject to change
      * (but it will be made available with improved guarantees in a subsequent release).
      */
-    private static final long CLIENT_TIMEOUT = Long.getLong("lra.internal.client.timeout", 10);
```
ochaloup, line 127: Would not be good to have one system property to define all client timeout values at once and on top of that to provide ability to granularly change the particular value?

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
```diff
@@ -189,6 +191,7 @@ public LRAData getLRAInfo(
     @POST
     @Path("start")
     @Produces(MediaType.TEXT_PLAIN)
+    @Bulkhead
```
ochaloup, line 194: I'm interested what is the reason for taking the faultolerance `@Bulkhead` on start method.
Having the fault tolerance capability is probably fine but why not other options (e.g. `@Retry`)? Why we want to work with only 10 parallel requests - what is the reason? Are we fine in adding one more dependency to the mix for the LRA processing?

rts/lra/coordinator/pom.xml
```diff
@@ -28,6 +28,11 @@
             <artifactId>microprofile-lra-api</artifactId>
             <version>${version.microprofile.lra.api}</version>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.microprofile.fault-tolerance</groupId>
+            <artifactId>microprofile-fault-tolerance-api</artifactId>
```
ochaloup, line 33: Is the compile scope considered as the right one for this?

ochaloup (commented):
I have few questions but in general it's probably fine.

rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
mmusgrov: Well that would be more code than I wanted to provide at this stage and we'd have document clearly which properties take precedence over others. I will rename the properties so that we can do what you request in a future revision, something like:

```java
    private static final long CLIENT_TIMEOUT = Long.getLong("lra.internal.client.timeout", 10);
    private static final long START_TIMEOUT = Long.getLong("lra.internal.client.timeouts.tart", 10);
    private static final long JOIN_TIMEOUT = Long.getLong("lra.internal.client.timeout.join", 10);
etc
```

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
mmusgrov: Because we want to protect the coordinator against overload. This PR is to address a particular issue I found during my stress testing of the coordinator (the issue really needs to be a blocker). Throttling start requests will reduce the number of active LRAs that the coordinator manages. So applying a bulkhead to start will reduce the total number of parallel LRAs to a manageable number and the other coordinator requests will all apply to this reduced number of LRAs.

I chose the default because, well it is the default. The user may then configure a value that is appropriate to his installation (using the usual MP FT config mechanism).

I didn't need any of the other FT annotations because the intent of the PR is to protect the coordinator. In the PR description I did explain why we don't use most of the FT features:

> There is a lot we can do in this area [fault resilience] and we have a few JIRAs for tackling the various requirements. But this PR is a basic minimum to reduce the number of errors caused by overloading the system in which LRA operates.

rts/lra/coordinator/pom.xml
mmusgrov: That's a good point, I cut and pasted the dependency from maven central. But you are right in querying this, I would have expected we'd need to be pulling in the smallrye implementation of FT. I will investigate further and update my PR depending upon what I find.

I did find that after I made the changes in this PR that I resolved the errors produced by the coordinator under heavy load (something like 5000 out of 5000 successful LRAs against two remote participants, it did take a couple of recovery passes for them all to finish okay). So I assumed my fix was good and did not drill down any further into the bulkhead.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
xstefank: Generally, this PR looks good. But I just want to point out that `@Bulkhead` is only reducing concurrent executions of the method thus at max 10 thread can run `start` at the same time but that doesn't correspond to the reduction of the number of active LRAs that the coordinator manages. The errornous state can still be reached but it would take a few retries on the /start calls.

xstefank (approved):

mmusgrov:
@xstefank yes agreed and this will still take the load off the coordinator.
I think it would be useful to still cap the total number of active LRAs, can you recommend any other FT annotations that would help us in this respect?

rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
ochaloup: ok

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
ochaloup: @mmusgrov sure, I was not aware about the problem from the stress testing. Then this makes sense. Thanks.

rts/lra/coordinator/pom.xml
ochaloup: I think the API will be bundled to the final jar file and the quarkus/thorntail is then taking the smallrye implementation into consideration and start it.
The trouble could be with WFLY maybe. In fact when I was looking at the `pom.xml` I'm not sure, from this perspective, is good to have the compile scope for `microprofile-lra-api` either. The deployment should be using what's provided by app runtime. I hope we can consider this as a nuance that could be investigated when working on WFLY extension.

ochaloup:
@mmusgrov I wonder about the idea of capping the number of active LRAs. Narayana does not do any similar thing for JTA or JTS, I think.

ochaloup (approved):
Thanks for the elaboration on my comments. There could be some further improvements but nothing big for not approving.

mmusgrov:
> @mmusgrov I wonder about the idea of capping the number of active LRAs. Narayana does not do any similar thing for JTA or JTS, I think.

The cap is a config property for users to apply only if their environment is known be constrained, AMS for example, (and we can make the default as large or as small as we like).

We don't do this with JTA or JTS because they run in app server environments whose resource needs are generally well known. To date we haven't provided any transaction products that by design are meant to run in a loosely coupled distributed system (XTS is the closest but that still runs with the app server as a primary architectural piece).

And MP-LRA is a MP component and use of other MP components is encouraged in order to achieve the use cases for which MicroProfile was created. MP-FT is designed for those environments where we expect users to be running LRA.

xstefank:
@mmusgrov FT isn't capable of such a thing up to my knowledge. The only thing that comes to mind is to cut the size of activeLRAs map in LRAService. Not sure if that is the correct solution.

Also off-topic, can we then move the configuration to MP config? I would like to work on that.

mmusgrov:
> @mmusgrov FT isn't capable of such a thing up to my knowledge. The only thing that comes to mind is to cut the size of activeLRAs map in LRAService. Not sure if that is the correct solution.

Okay, then let's stay with just `@Bulkhead` for now since that seemed to resolve the immediate problem I found in my stress testing.

> 
> Also off-topic, can we then move the configuration to MP config? I would like to work on that.

Would doing this in the narayana code base interfere with what we are doing in the Quarkus and WildFly integrations. If not then, yes, I agree that leveraging MP config would be good.

xstefank:
OK. Config shouldn't interfere. SR projects often come with their own config properties. quarkus prefix should be restricted to quarkus specifics I believe. In every case, please go ahead and merge this PR, I'll create a new JBTM to track config.

rts/lra/coordinator/pom.xml
mmusgrov: Done

rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
mmusgrov: Actually the change was rather trivial so I incorporated it into the PR.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
mmusgrov: I ran a local test to ensure that the Bulkhead was being respected.

rts/lra/coordinator/pom.xml
mmusgrov: Well I made some changes to how we consume the dependency although I didn't quite follow your last comment so you might want to check my latest push.

mmusgrov:
@xstefank and @ochaloup I made a number of changes since you both approved it so you may want to rescan the PR to verify that I addressed your comments correctly.

rts/lra/coordinator/pom.xml
ochaloup: My point is the best described in way that I think the dependency should be in scope `provided`. The app runtime then loads the api library. It won't be the application that defines it.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -87,8 +87,7 @@
 
     private static final Pattern START_END_QUOTES_PATTERN = Pattern.compile("^\"|\"$");
     private static final long DEFAULT_TIMEOUT_MILLIS = 0L;
-    // i18nMessageCode corresponding to lraI18NLogger#warn_LRAStatusInDoubt
-    private static final int i18nMessageCode = 25145;
```
ochaloup, line 91: I like this change ;-)

ochaloup (approved):
I think I'm fine with the current state of the PR.

xstefank (approved):

rts/lra/coordinator/pom.xml
mmusgrov: +1

PR #1749, opened 2020-12-16
Indicate that the artemis journal is a provided dependency

tomjenkinson:
I noticed this was missing from narayana-jta while importing the module into an IDE. I added it to some other poms and running CI to test it. It may be required in other shaded jars?

### DIFF
--- a/ArjunaJTA/narayana-jta/pom.xml
+++ b/ArjunaJTA/narayana-jta/pom.xml
@@ -168,6 +168,18 @@
         <scope>provided</scope>
     </dependency>
     <dependency>
+      <groupId>org.apache.activemq</groupId>
+      <artifactId>artemis-journal</artifactId>
+      <version>${version.org.apache.activemq}</version>
+      <scope>provided</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>io.netty</groupId>
+          <artifactId>netty</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
       <groupId>org.jboss.narayana</groupId>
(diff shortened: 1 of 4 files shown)


mmusgrov (approved):

ochaloup (approved):

mayankkunwar:
@tomjenkinson The main reason for zero % change is explained below, please correct me if my understanding to your question is not correct.

Suppose we are comparing two results that are `X` and `Y` and the deviation of the corresponding results are `dx` and `dy` respectively i.e. range of `X` is `X-dx <= X <= X+dx` and range of `Y` is `Y-dy <= Y <= Y+dy`

So when we compare these two ranges with eachother, if the range overlaps, we return the result as `zero%` change.

`Example:`
 range of X is (100 to 150), 
range of Y is (120 to 180),

in this case we say that there is no change in the performance.

mayankkunwar:
I agree @tomjenkinson rephrasing the comment would help, but only in `zero %` change case.
But if you want me to rephrase it for all the cases then I will do it.

mayankkunwar:
Okay @tomjenkinson I will raise a PR with related changes.
Thanks.

mayankkunwar:
> I guess it would be something along the lines of "considering the deviation of the numbers we do not determine this test performance to have altered significantly". Presenting zero % doesn't seem right to me considering there seems there could be some kind of change but overlapping. If you would be able to provide a PR though I would be happy to take a look and provide more targeted feedback. Thanks

@tomjenkinson I have raised a PR https://github.com/jbosstm/performance/pull/87

PR #1751, opened 2020-12-18
[JBTM-3411] enhance exception reporting for NarayanaLRAClient

ochaloup:
https://issues.redhat.com/browse/JBTM-3411

This is a change that I benefit when working with `NaraynaLRAClient` and when I'm creating new tests.
It's about adding cause for the thrown exception that could be processed in the code or assertion.
Plus, `URI` is more convenient format than String parsing.

LRA


mmusgrov (approved):
Note that this is something we do not want to support yet (parts of it are still untested).
What was your motivation for adding the extra javadoc?
Note that the PR is approved but we want to avoid giving the impression that it's supported, it is provided as is and at the users own risk. Hopefully the javadoc WARNING at the top of the file will let such users know this.

ochaloup:
My perspective is that any public method should be documented. It could be removed, deprecated or changed later but as it's present it should be documented. That was the motivation.
And we use the `LRAClient` pretty extensively in our code thus the documentation is a plus.

And as you mentioned the warning is present and indicates there will be changes in future.

ochaloup:
@mmusgrov as this PR was approved I'm merging it. If there are some further adjustment on top of it, let me know and I will provide a new PR.

PR #1747, opened 2020-12-15
[JBTM-3383] Improve ARJUNA016009: Caught::java.lang.NullPointerException log message

ivassile:
Issue: https://issues.redhat.com/browse/JBTM-3383

MAIN AS_TESTS QA_JTA
!QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS !TOMCAT !JACOCO !LRA

PR #1759, opened 2021-01-08
JBTM-3414 Update to add null check with trans on Xids trace logging d…

zhfeng:
…uring xaRecoveryFirstPass

https://issues.redhat.com/browse/JBTM-3414

MAIN CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF !LRA !NO_WIN DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
@@ -736,7 +736,7 @@ public class XARecoveryModule implements ExtendedRecoveryModule
                             + ((trans != null) ? trans.length : 0)
                             + " xids in doubt");
                 }
-				if (jtaLogger.logger.isTraceEnabled()) {
+				if (jtaLogger.logger.isTraceEnabled() && trans != null) {
 					for (Xid xid : trans) {
 						byte[] globalTransactionId = xid.getGlobalTransactionId();
 						byte[] branchQualifier = xid.getBranchQualifier();


ochaloup (approved):
Seems fine to me.

zhfeng:
Thanks @ochaloup !

mmusgrov:
@ochaloup how does this pr relate to [JBTM-3383] Improve ARJUNA016009: Caught::java.lang.NullPointerException log message

mmusgrov:
@ochaloup in particular, does it obsolete pr #1747 [JBTM-3383]

ochaloup:
@mmusgrov I would say the pr #1747 elaborates on the NPE issue and fixes it in the broader scope. This PR and the pr #1747 touches the same part of the code.

mmusgrov:
@ochaloup I thought the issue was that the logging code is throwing an NPE which, although it does not cause any functional issues, it makes EAP look bad because it suggests a problem in EAP. Why do we have two issues and two PRs for the same (non functional) problem?

zhfeng:
@mmusgrov JBTM-3414 came across the Fuse case and I have no idea about  JBTM-3383. Maybe we can mark JBTM-3413 as duplicated ?

mmusgrov:
@zhfeng thanks for the fix, I am investigating why we ended up with two JBTM's for this issue.

mmusgrov:
But ideally we should have caught the duplication before merging the fixes.

ochaloup:
@mmusgrov trace logging caused the functional issue as well. But it was caused by fact the trace logging is switched on. The trouble with NPE would probably not be considered in full scope if @zhfeng does not raise this one.

PR #1761, opened 2021-01-13
[JBTM-3343] LRA TCK Tests running on all runtimes (Thorntail/Quarkus + WildFly)

ochaloup:
https://issues.redhat.com/browse/JBTM-3343

This is a set of changes in pom.xml and Arquillian setup for `rts/lra/test/` which makes possible to run the TCK and basic tests with Quarkus/Thorntail and WildFly. The changes clean the way how the test executions are configured in general.

For the WildFly to pass the TCK there is needed to have a new release of spec thus I'm postponing to configure and verify the results until it happens.

The unfinished task, regarding the testing, could be the tests inside of `container` module. Those ones are currently running only with WildFly and it would be good to have them executed with Quarkus/Thorntail. I assume to look at it as a next step after this PR is merged and when I finish the JBTM-3294.

/CC @xstefank if you may take a look it would be nice

LRA


rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/Deployer.java
```diff
@@ -26,15 +26,21 @@
...
+        // manifest for WildFly deployment which requires access to transaction jboss module
+        final String ManifestMF = "Manifest-Version: 1.0\n"
+            + "Dependencies: org.jboss.jts, org.jboss.logging\n";
+
         return ShrinkWrap.create(WebArchive.class, appName + ".war")
                 .addPackages(true,
                     LRAMetricService.class.getPackage(),
                     org.codehaus.jettison.JSONSequenceTooLargeException.class.getPackage())
-                .addClasses(LRAClientOps.class, WrongHeaderException.class);
+                .addClasses(LRAClientOps.class, WrongHeaderException.class)
```
mmusgrov, line 43: Can we limit the use of MP-LRA tck test classes to the tck module rather than allowing them to leak out into other LRA modules. I also recall seeing the use of org.eclipse.microprofile.lra.tck.service.LRAMetricService somewhere - tying us to internal TCK test classes is going to bind us to something that we do not have full control over and there is almost zero benefit in using these classes.

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/FailedLRAIT.java
```diff
@@ -94,15 +96,19 @@ public void after() {
...
+        return LRAConstants.getLRACoordinatorUrl(lraId) + "/" + RECOVERY_COORDINATOR_PATH_NAME;
+    }
+
     /**
      * test that a failure record is created when a participant (with status reporting) fails to compensate
      */
     @Test
     public void testWithStatusCompensateFailed() throws Exception {
-        URI lra = invokeInTransaction(LRAParticipantWithStatusURI.LRA_PARTICIPANT_PATH,
+        LRAURIHolder lra = invokeInTransaction(LRAParticipantWithStatusURI.LRA_PARTICIPANT_PATH,
```
mmusgrov, line 108: Can we stay with using URIs since LRAURIHolder only adds overhead on top of that.

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/NonRootLRAParticipantIT.java
```diff
@@ -76,8 +79,9 @@ public void testNonRootLRAParticipantEnlist() {
      * @throws URISyntaxException if arquillian URL is invalid
      */
     @Test
-    public void testFinishLRA() throws URISyntaxException {
-        NarayanaLRAClient client = new NarayanaLRAClient(); // the narayana client API for using LRAs
+    public void testFinishLRA() throws URISyntaxException, MalformedURLException {
```
mmusgrov, line 82: Remark: adding an extra exception is fine from my perspective (so I am not asking for a change) but, in the past, you have told me not to do that in my own PRs.

rts/lra/test/basic/src/test/resources/logging.properties
```diff
@@ -15,9 +15,10 @@
 #
 
 handlers = java.util.logging.ConsoleHandler
-java.util.logging.ConsoleHandler.level = INFO
+java.util.logging.ConsoleHandler.level = FINEST
```
mmusgrov, line 18: Why finest by default - can't we add that in the jobs or on demand if an engineer is debugging a particular issue. This adds a lot of noise, slows down running local builds and makes it impossible to quickly scan the output looking for failures. I see some profiles in the top level pom for enabling trace level logging - how does that compare to this properties file.

rts/lra/test/pom.xml
```diff
@@ -137,8 +287,8 @@
                                         <argument>${lra.coordinator.log.file.module.narayana}</argument>
                                         <argument>-DRecoveryEnvironmentBean.periodicRecoveryPeriod=10</argument>
                                         <argument>-DRecoveryEnvironmentBean.recoveryBackoffPeriod=5</argument>
-                                        <argument>-DObjectStoreEnvironmentBean.objectStoreDir=${project.build.directory}/lra_txn_logs</argument>
-                                        <argument>-DObjectStoreEnvironmentBean.communicationStore.objectStoreDir=${project.build.directory}/lra_txn_logs</argument>
+                                        <argument>-DObjectStoreEnvironmentBean.objectStoreDir=${project.build.directory}/quarkus_lra_objectstore</argument>
```
mmusgrov, line 290: `lra_txn_logs` is more generic. Did you make this change so that a module can be tested against multiple containers concurrently. If they only ever run against one container in a single execution then we should use a generic name since it slows down developers and inhibits scripting. Besides the contents of filesystem based stores should be the same regardless of which container we run against.

rts/lra/test/tck/src/test/resources/logging.properties
```diff
@@ -15,9 +15,9 @@
 #
 
 handlers = java.util.logging.ConsoleHandler
-java.util.logging.ConsoleHandler.level = INFO
+java.util.logging.ConsoleHandler.level=INFO
```
mmusgrov, line 18: In another place in this PR you explicitly added spaces around the = yet here you have removed them. Any particular reason or is it just you noticing something that doesn't look quite right?

mmusgrov (changes requested):
I had two inline change requests to the code and a few comments.

In addition I saw some failures of the time bound tests on my laptop which I have not seen before. I did a second run which passed. But it is something to look out for - I will try soak testing the TCK for a while to verify it was a spurious failure (unless it is related to the adjustTimeoutByDefaultFactor issue I mention below).

Also would it be possible to indicate how to use the profiles for debugging. For example, to debug the tests in the basic module I have to either edit the `javaVmArguments` property in `basic/src/test/resources/arquillian-thorntail.xml` or override the entire line via a system property. Previously there had been other ways to debug that `mvn verify -Ddebug.coordinator -Parq` or `mvn verify -Parq -Ddebug.coordinator -Dmaven.failsafe.debug etc but those stopped working a while back.

And also profiles more generally (for example `-Pdo-not-start-lra-coordinator` that one obvious but I don't know of its existence unless I look) etc, and what's the difference between, say, `-Parq.wildfly` and `-Parq`. Maybe also document somewhere how to configure the debugging/ logging/port properties etc (which you do in test/pom.xml), could even be as simple as linking the README to the pom where they need to be configured.

So having something somewhere (such as a readme in one of the parent modules, for example) that tells engineers how to debug and run tests means they don't have to deconstruct pom files to figure it out. Then I could have tried testing the various profiles while doing my review (our CI probably does not exercise all the profiles).

There is also a missing initial value for `adjustTimeoutByDefaultFactor` since TckRecoveryTests fail parsing an empty string to a number. Is that the MP-LRA change you have mentioned previously. Or do I just need to pass `-Dlra.tck.timeout.factor=num` on the mvn verify command line?

rts/lra/test/pom.xml
```diff
@@ -52,14 +67,154 @@
...
+                </plugins>
+            </build>
+        </profile>
+
+        <profile>
+            <id>arq.wildfly</id>
+            <properties>
+                <!-- the AppServerCoordinatorDeploymentObserver creates the deployment named 'lra-coordinator' and deploys it to the app server,
+                     the last part of the path with value 'lra-coordinator' is defined by LRA coordinator JAX-RS endpoint -->
+                <lra.coordinator.url>http://${lra.coordinator.host}:${lra.coordinator.port}/lra-coordinator/lra-coordinator</lra.coordinator.url>
```
xstefank, line 148: maybe you can change the name of the deployment so it's different than lra-coordinator? not sure if this can be confusing for anyone.

rts/lra/test/pom.xml
```diff
@@ -52,14 +67,154 @@
...
+                            <execution>
+                                <id>enforce-env-variable-jboss-home</id>
+                                <goals>
+                                    <goal>enforce</goal>
+                                </goals>
+                                <configuration>
+                                    <rules>
+                                        <requireProperty>
+                                            <property>env.JBOSS_HOME</property>
+                                            <message>You need to define JBOSS_HOME environmental variable!</message>
```
xstefank, line 181: +1! I got questions about this before :)

rts/lra/test/pom.xml
```diff
@@ -52,14 +67,154 @@
...
+                            </execution>
+                        </executions>
+                    </plugin>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-failsafe-plugin</artifactId>
+                        <configuration>
+                            <skipITs>false</skipITs>
+                            <systemProperties>
+                                <arquillian.xml>arquillian.xml</arquillian.xml>
```
xstefank, line 197: can you rename this arquillian file to arquillian-wildfly.xml since it's explicitly listed?

rts/lra/test/pom.xml
xstefank: +1

rts/lra/test/tck/pom.xml
```diff
@@ -38,14 +32,7 @@
...
                     <systemPropertyVariables combine.children="append">
-                        <java.util.logging.config.file>${project.build.testOutputDirectory}/logging.properties</java.util.logging.config.file>
-                        <lra.tck.suite.host>${lra.tck.suite.host}</lra.tck.suite.host>
-                        <lra.tck.suite.port>${lra.tck.suite.port}</lra.tck.suite.port>
-                        <thorntail.test.logging.category.arjuna>${thorntail.test.logging.category.arjuna}</thorntail.test.logging.category.arjuna>
-                        <thorntail.test.logging.category.narayana>${thorntail.test.logging.category.narayana}</thorntail.test.logging.category.narayana>
-                        <lra.coordinator.port>${lra.coordinator.port}</lra.coordinator.port>
-                        <lra.tck.suite.debug.params>${lra.tck.suite.debug.params}</lra.tck.suite.debug.params>
-                        <lra.tck.timeout.factor>${lra.test.timeout.factor}</lra.tck.timeout.factor>
+                        <lra.tck.timeout.factor>${lra.tck.timeout.factor}</lra.tck.timeout.factor>
```
xstefank, line 35: This is where the TCK factor is not being set by default making the timelimit tests fail.

rts/lra/test/tck/pom.xml
```diff
@@ -72,36 +59,4 @@
...
-            <id>arq</id>
-            <properties>
-                <lra.coordinator.exec.plugin.phase>pre-integration-test</lra.coordinator.exec.plugin.phase>
-            </properties>
-        </profile>
-        <profile>
-            <id>debug.lra.tck.suite</id>
-            <activation>
-                <property>
-                    <name>debug.tck</name>
```
xstefank, line 87: I also used this a lot in the past. I liked that I can easy just add `-Ddebug.tck` or `-Ddebug.coordinator`.

ochaloup (commented):
Fixing the remarks from the reviews.

ochaloup:
@mmusgrov I tried to address the inline comments in my PR changeset. To the points in your comment.

I created the `README` in `lra/test` module which talks about the test setup and points to the maven profiles which can be used.

I tried the debugging options and `-Ddebug.coordinator -Dmaven.failsafe.debug` works fine for me. The only option which was changed is the `-Ddebug.tck`  which is now just `-Ddebug`.

The `-Dlra.tck.timeout.factor=num` can be defined manually but otherwise the default value of `1.0` is used. The TCK uses that default value as well (not needed to pass anything in case it's not wanted).

I have no idea about the test failures. This PR does not touch directly any test functionality. I only unified the way how the `pom.xml` was working with properties and added Arquillian extension to run the tests with WildFly as well.
When I was running the tests on my local machine I was not able to reproduce the failure, at least so far. The timeout issues for the TCK is still a bit trouble for any LRA testing unfortunately :-(

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/Deployer.java
mmusgrov: +1

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/NonRootLRAParticipantIT.java
mmusgrov: No, you complained about the exact same thing you have done here.

rts/lra/test/pom.xml
mmusgrov: To each of the four points:

1. So you are saying that the build breaks if you use a generic name (such as lra-txn-logs).
2. I am talking about developers wanting to write their own scripts to manager their own work. Plus the lack of consistency incurs greater mental effort and when our time has multiple demands upon it just adds overhead and makes working with the code base a less pleasant experience.
3. I discussed and addressed this point in my review comment.
4. Yes make the name consistent (lra-logs or whatever) regardless of which container is in use (unless doing so breaks the build as per my response to point 1).

rts/lra/test/pom.xml
mmusgrov: I have the same comment as above in point 2, let's keep it consistent. If http://localhost:8080/lra-coordinator works then use that. If the endpoint keeps changing depending upon which test is being ran or debugged then it increases the number of surprises that the developer has to deal with and, again, it makes working with the code base a less pleasant experience.

rts/lra/test/tck/pom.xml
mmusgrov: See my PR review summary - the build was failing unless I pass `-Dlra.tck.timeout.factor=num` to the build.

rts/lra/test/tck/pom.xml
mmusgrov: Thanks

mmusgrov:
> I created the `README` in `lra/test` module which talks about the test setup and points to the maven profiles which can be used.

It is spot on, thanks.

> I tried the debugging options and `-Ddebug.coordinator -Dmaven.failsafe.debug` works fine for me. The only option which was changed is the `-Ddebug.tck` which is now just `-Ddebug`.

I'm not convinced but I will use the instructions you added to the README next time.

> The `-Dlra.tck.timeout.factor=num` can be defined manually but otherwise the default value of `1.0` is used. The TCK uses that default value as well (not needed to pass anything in case it's not wanted).

I pulled the PR again, the TCK still fails in the same way. Without setting the timeout factor it fails consistently with Failures: 2, Errors: 1 but with the timeout factor set to 1 it passes consistently.


Typing `mvn verify -Parq` in the lra directory the TCK fails consistently with NumberFormatException and there are two test failures (whereas `mvn verify -Parq -Dlra.tck.timeout.factor=1` passes fine):

[ERROR] testCancelWhenParticipantIsUnavailable(org.eclipse.microprofile.lra.tck.TckRecoveryTests)  Time elapsed: 14.066 s  <<< ERROR!
java.lang.NumberFormatException: empty String
        at org.eclipse.microprofile.lra.tck.TckRecoveryTests.adjustTimeoutByDefaultFactor(TckRecoveryTests.java:234)
        at org.eclipse.microprofile.lra.tck.TckRecoveryTests.testCancelWhenParticipantIsUnavailable(TckRecoveryTests.java:192)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TckTests.timeLimit:320->TckTestBase.checkStatusReadAndCloseResponse:109 Response status on call to 'http://localhost:8180/lraresource/timeLimit' failed to match. expected:<200> but was:<500>
[ERROR]   TckTests.timeLimitWithPreConditionFailed:361 Expected 412 or 410 response
Expected: (<412> or <410>)
     but: was <500>
[ERROR] Errors: 
[ERROR]   TckRecoveryTests.testCancelWhenParticipantIsUnavailable:192->adjustTimeoutByDefaultFactor:234 » NumberFormat
[INFO] 
[ERROR] Tests run: 133, Failures: 2, Errors: 1, Skipped: 0

mmusgrov:
Why don't you just set the default in the test/tck pom otherwise `lra.tck.timeout.factor` has no value:

```java
<lra.tck.timeout.factor>${lra.tck.timeout.factor}</lra.tck.timeout.factor>
```

rts/lra/test/tck/pom.xml
ochaloup: +1, thanks for pointing that out. The scripts were wrong and I took the wrong property. I hope it will be ok now.

ochaloup:
@mmusgrov thanks, I used the wrong property. It should be working now.

rts/lra/test/pom.xml
ochaloup: Ok, I changed this back to `lra_txn_logs`.

ochaloup: I'm resolving this discussion. If I missed something to be done, please, reopen it.

rts/lra/test/pom.xml
ochaloup: Ok, let's go with what is it defined now.

I'm resolving this discussion. If there is something that should be adjusted and I missed that, please reopen it.

mmusgrov (approved):
Thanks, a nice improvement.

ochaloup:
@mmusgrov thanks. I'm going to merge the PR. I'm happy to fix/adjust what will be needed in some next PRs.

What I'm still unsure how to process with the tests in `coordinator` package. That would be good to be run with Quarkus too but my quick experimentation with byteman, was all a bit complicated.

PR #1762, opened 2021-01-13
RELEASE SCRIPT ONLY: Do not access the jobs via a Release view

tomjenkinson:
NO_TEST


mmusgrov (approved):
@tomjenkinson I don't understand the need for the change but am approving anyway. Will add a comment indicating the motivation for the change.

PR #1763, opened 2021-01-13
[JBTM-3383] Improve ARJUNA016009: Caught::

ivassile:
Upstream issue: https://issues.redhat.com/browse/JBTM-3383
Issue: https://issues.redhat.com/browse/JBEAP-20431

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
@@ -716,33 +716,27 @@ public class XARecoveryModule implements ExtendedRecoveryModule
                             + ((trans != null) ? trans.length : 0)
                             + " xids in doubt");
                 }
-				if (jtaLogger.logger.isTraceEnabled()) {
-					for (Xid xid : trans) {
-						byte[] globalTransactionId = xid.getGlobalTransactionId();
-						byte[] branchQualifier = xid.getBranchQualifier();
-	
-						StringBuilder stringBuilder = new StringBuilder();
-						stringBuilder.append("< ");
-						stringBuilder.append(xid.getFormatId());
-						stringBuilder.append(", ");
-						stringBuilder.append(globalTransactionId.length);
-						stringBuilder.append(", ");
-						stringBuilder.append(branchQualifier.length);
-						stringBuilder.append(", ");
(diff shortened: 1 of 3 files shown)


ochaloup (approved):
Seems fine to me. The AS_TEST failure (`ApplicationContextLifecycleTestCase`) should be only the CI intermittent error.

mmusgrov (commented):
The PR for this behaviour on the master branch had asked for other stuff (example https://github.com/jbosstm/narayana/pull/1747#issuecomment-756057299) but the PR was merged before the discussion had reached a conclusion. Will you indicate why that comment was not relevant to the PR?

I was also expecting that the other PR would address the reasons why this fix is improves on the one that that already addressed the issue (ie. #1759) - the only reference to the duplication was when @ochaloup said " #1759 touches (but only party) the same issue" but with no explanation as to why Amos's fix was insufficient.

ochaloup:
@mmusgrov ok. I'm sorry for my ignorance but I was thinking that merging the original PR was fine from your side (ie. that merging #1747 is fine).
I agree that the comment you refer to was not fully addressed but I didn't realize that at that time. Would you be fine if I create a new jira issue which will be asking for creating the tests for that? I can assign that work to me if you agree.

Regarding the difference on Amos's fix #1759 and the fix at #1747 : the @ivassile 's fix was adding more robustness on top of the Amos's one. Amos was fixing the trouble of `null` being assigned to `trans` variable. This fix is go beyond and fixes the NPE would not be thrown when `trans` is not null and trace logging is enabled. That was part which was not addressed by Amos. That's why I consider this PR (ie. #1747) good and convenient. Would be this explanation fine from your perspective?

mmusgrov:
The explanation is okay (although it does imply that Amos's fix did not correctly address the problem and its lack of "robustness" still left us exposed), the main issue I have is that PR comments need to be responded to before allowing a merge.
Regarding the tests, if the view is that the tests are overkill then that is fine but then such a view would need to be documented in the PR.

mmusgrov:
> I agree that the comment you refer to was not fully addressed but I didn't realize that at that time. Would you be fine if I create a new jira issue which will be asking for creating the tests for that? I can assign that work to me if you agree.

I am not in favour of adding more issues and PRs just to respond to my comment on the original PR regarding testing. But I already said that on the original PR.

mmusgrov (approved):

TomasHofman:
Upstream PR: https://github.com/jbosstm/narayana/pull/1747

PR #1767, opened 2021-01-25
[SCRIPTS ONLY] correcting the name of the timeout factor attribute for WFLY integration tests

ochaloup:
I checked the failing tests on our CI to understand what are the current troubles. By me the most of them are caused by timing issues (e.g. quite usual is error message like `Failed to establish connection in 16009ms`). From that I found that the script uses wrong timeout factor property which should be `ts.timeout.factor` (ie. https://github.com/wildfly/wildfly-core/blob/14.0.0.Final/testsuite/shared/src/main/java/org/jboss/as/test/shared/TimeoutUtil.java#L40).

This adjustment is an attempt to enhance the stability of the WFLY testsuite.

!MAIN !CORE !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS  !TOMCAT !JACOCO !LRA
AS_TESTS

/cc @tomjenkinson

### DIFF
--- a/scripts/hudson/narayana.sh
+++ b/scripts/hudson/narayana.sh
@@ -407,7 +407,7 @@ function build_as {
 
   # running WildFly testsuite if configured to be run by axis AS_TESTS
   if [ $AS_TESTS = 1 ]; then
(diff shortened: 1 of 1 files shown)


tomjenkinson (changes requested):
Given https://github.com/wildfly/wildfly/blob/master/testsuite/pom.xml#L137-L138 I think the setting is already taking affect

ochaloup:
@tomjenkinson +1, you are right. I was too hasty. Closing this PR.

PR #1773, opened 2021-01-29
JBTM-3297: Move LRA failure records to another part of the store

mayankkunwar:
https://issues.redhat.com/browse/JBTM-3297

!MAIN !CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF LRA !NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
+++ b/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
@@ -528,15 +528,15 @@ public class LongRunningAction extends BasicAction {
             updateState(LRAStatus.Closing);
         } // otherwise status is (cancel ? LRAStatus.Cancelled : LRAStatus.Closed)
 
+        finishTime = LocalDateTime.now(ZoneOffset.UTC);
+
+        updateState(); // ensure the record is removed if it finished otherwise persisted the state
+
         if (!isRecovering()) {
             if (lraService != null) {
                 lraService.finished(this, false);
             }
         }
-
-        finishTime = LocalDateTime.now(ZoneOffset.UTC);
-
-        updateState(); // ensure the record is removed if it finished otherwise persisted the state
 
         return res;
     }
--- a/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
+++ b/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
@@ -32,7 +32,9 @@ import com.arjuna.ats.arjuna.objectstore.StoreManager;
 import com.arjuna.ats.arjuna.recovery.RecoveryModule;
 import com.arjuna.ats.arjuna.recovery.TransactionStatusConnectionManager;
 import com.arjuna.ats.arjuna.state.InputObjectState;
+import com.arjuna.ats.arjuna.state.OutputObjectState;
 import io.narayana.lra.coordinator.domain.model.LongRunningAction;
+import io.narayana.lra.coordinator.domain.model.FailedLongRunningAction;
 import io.narayana.lra.coordinator.domain.service.LRAService;
 import org.eclipse.microprofile.lra.annotation.LRAStatus;
 
@@ -111,7 +113,7 @@ public class LRARecoveryModule implements RecoveryModule {
          // uids per transaction type
         InputObjectState aa_uids = new InputObjectState();
 
-        if (getUids(aa_uids)) {
+        if (getUids(_transactionType, aa_uids)) {
             processTransactionsStatus(processTransactions(aa_uids));
         }
     }
@@ -126,6 +128,7 @@ public class LRARecoveryModule implements RecoveryModule {
 
             LRAStatus lraStatus = lra.getLRAStatus();
             if (LRAStatus.FailedToCancel.equals(lraStatus) || LRAStatus.FailedToClose.equals(lraStatus)) {
+                moveEntryToFailedLRAPath(recoverUid);
                 return;
             }
 
@@ -157,6 +160,37 @@ public class LRARecoveryModule implements RecoveryModule {
         }
     }
 
+    public boolean moveEntryToFailedLRAPath (final Uid failedUid) {
+        String failedLRAType = FailedLongRunningAction.FAILED_LRA_TYPE;
+        boolean moved = false;
+        try {
+            InputObjectState inputState = _recoveryStore.read_committed(failedUid, _transactionType);
+            InputObjectState failedLRAUidState = _recoveryStore.read_committed(failedUid, failedLRAType);
+            if (inputState != null) {
+                if (failedLRAUidState != null) {
+                    // Record already exists in failedLRARecord location, hence removing it from the LRARecord location
+                    moved = true;
+                    if (!_recoveryStore.remove_committed(failedUid, _transactionType)) {
+                        LRALogger.i18NLogger.warn_UnableToRemoveDuplicateFailedLRARecord(
+                                failedUid.toString(), failedLRAType, _transactionType);
+                        moved = false;
+                    }
+                    return moved;
+                }
+
+                if (_recoveryStore.write_committed(failedUid, failedLRAType, new OutputObjectState(inputState))) {
+                    moved = _recoveryStore.remove_committed(failedUid, _transactionType);
+                    if (moved) {
+                        LRALogger.logger.infof("Failed lra record (Uid: %s) moved to new location type: %s", failedUid, failedLRAType);
+                    }
+                }
+            }
+        } catch (ObjectStoreException e) {
+            LRALogger.i18NLogger.warn_move_lra_record(failedUid.toString(), e.getMessage());
+        }
+        return moved;
+    }
+
     private Collection<Uid> processTransactions(InputObjectState uids) {
         Collection<Uid> uidCollection = new ArrayList<>();
 
@@ -166,7 +200,7 @@ public class LRARecoveryModule implements RecoveryModule {
 
         Consumer<Uid> uidUnpacker = uidCollection::add;
 
-        forEach(uids, uidUnpacker);
+        forEach(uids, uidUnpacker, _transactionType);
 
         return uidCollection;
     }
@@ -197,7 +231,7 @@ public class LRARecoveryModule implements RecoveryModule {
         InputObjectState aa_uids = new InputObjectState();
 
         // for LRA call reactivate and add it to the input list
-        if (getUids(aa_uids)) {
+        if (getUids(_transactionType, aa_uids)) {
             Collection<Uid> uids = processTransactions(aa_uids);
             uids.forEach(uid -> {
                 int status = _transactionStatusConnectionMgr.getTransactionStatus(_transactionType, uid);
@@ -244,7 +278,7 @@ public class LRARecoveryModule implements RecoveryModule {
     public void getFailedLRAs(Map<URI, LongRunningAction> lras) {
         InputObjectState aa_uids = new InputObjectState();
         Consumer<Uid> failedLRACreator = uid -> {
-            LongRunningAction lra = new LongRunningAction(service, new Uid(uid));
+            FailedLongRunningAction lra = new FailedLongRunningAction(service, new Uid(uid));
             lra.activate();
 
             LRAStatus status = lra.getLRAStatus();
@@ -253,24 +287,24 @@ public class LRARecoveryModule implements RecoveryModule {
             }
         };
 
-        if (getUids(aa_uids)) {
-            forEach(aa_uids, failedLRACreator);
+        if (getUids(FailedLongRunningAction.FAILED_LRA_TYPE, aa_uids)) {
+            forEach(aa_uids, failedLRACreator, FailedLongRunningAction.FAILED_LRA_TYPE);
         }
     }
 
-    private boolean getUids(InputObjectState aa_uids) {
+    private boolean getUids(final String type, InputObjectState aa_uids) {
         synchronized (this) {
             try {
-                return _recoveryStore.allObjUids(_transactionType, aa_uids);
+                return _recoveryStore.allObjUids(type, aa_uids);
             } catch (ObjectStoreException e) {
                 if (LRALogger.logger.isTraceEnabled()) {
                     LRALogger.logger.tracef(e,
                             "LRARecoverModule: Object store exception %s while unpacking records of type %s",
-                            e.getMessage(), _transactionType);
+                            e.getMessage(), type);
                 } else if (LRALogger.logger.isInfoEnabled()) {
                     LRALogger.logger.infof(
                             "LRARecoverModule: Object store exception %s while unpacking records of type %s",
-                            e.getMessage(), _transactionType);
+                            e.getMessage(), type);
                 }
 
                 return false;
@@ -284,11 +318,13 @@ public class LRARecoveryModule implements RecoveryModule {
      *
      * @param uids the uids to iterate over
      * @param consumer the consumer that should be called for each Uid
+     * @param transactionType within the Object Store for LRAs
      */
     // This method could be moved to an ArjunaCore class (such as UidHelper) if its useful,
     // in which case make the method returns a boolean:
     //     * @return false if there was an error processing the collection of uids
-    private void forEach(InputObjectState uids, Consumer<Uid> consumer) {
+    // The @param transactionType is only used for logging purpose
+    private void forEach(InputObjectState uids, Consumer<Uid> consumer, final String transactionType) {
         do {
             try {
                 Uid uid = new Uid(uids.unpackBytes());
@@ -302,11 +338,11 @@ public class LRARecoveryModule implements RecoveryModule {
                 if (LRALogger.logger.isTraceEnabled()) {
                     LRALogger.logger.tracef(e,
                             "LRARecoverModule: Object store exception %s while unpacking a record of type %s",
-                            e.getMessage(), _transactionType);
+                            e.getMessage(), transactionType);
                 } else if (LRALogger.logger.isInfoEnabled()) {
                     LRALogger.logger.infof(
                             "LRARecoverModule: Object store exception %s while unpacking a record of type: %s",
-                            e.getMessage());
+                            e.getMessage(), transactionType);
                 }
 
                 return;
--- a/rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/FailedLRAIT.java
+++ b/rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/FailedLRAIT.java
@@ -26,6 +26,7 @@ import io.narayana.lra.LRAConstants;
 import io.narayana.lra.arquillian.resource.LRAParticipantWithStatusURI;
 import io.narayana.lra.arquillian.resource.LRAParticipantWithoutStatusURI;
 import io.narayana.lra.arquillian.spi.NarayanaLRARecovery;
+import io.narayana.lra.coordinator.domain.model.FailedLongRunningAction;
 import org.codehaus.jettison.json.JSONArray;
 import org.codehaus.jettison.json.JSONObject;
 import org.eclipse.microprofile.lra.annotation.LRAStatus;
@@ -196,6 +197,81 @@ public class FailedLRAIT {
         // the participant should now be in a failed state so the record can be deleted
         int statusDeleted = removeFailedLRA(lraId);
         assertEquals("deleting a failed LRA", 204, statusDeleted);
+    }
+
+    /**
+     * test that a created failure record is moved to failedLRAType location
+     */
+    @Test
+    public void testToMoveFailedLRARecords() throws Exception {
+        URI lraId = invokeInTransaction(LRAParticipantWithStatusURI.LRA_PARTICIPANT_PATH,
+                LRAParticipantWithStatusURI.TRANSACTIONAL_CANCEL_PATH, 500);
+        // Checks if the record exists in the LRARecords location
+        if (!validateFailedRecordMoved(lraId)) {
+            fail("lra not in failed list location : " + FailedLongRunningAction.FAILED_LRA_TYPE);
+        }
+        // Removing Failed LRA record.
+        removeFailedLRA(lraId);
+    }
+
+    /**
+     * Validate if Failed LRA is moved to failed lraRecord's Location.
+     *
+     * @param lra the LRA whose state is to be validated
+     * @return true if the Failed LRA record is moved to another location
+     * @throws Exception if the state cannot be determined
+     */
+    private boolean validateFailedRecordMoved(URI lra) throws Exception {
+        boolean isMoved = false;
+        String failedLRAId = null;
+        List<JSONObject> failedRecords = getFailedRecords(lra);
+        for (JSONObject failedLRA : failedRecords) {
+            String lraId = (String) failedLRA.get("lraId");
+            lraId = lraId.replaceAll("\\\\", "");
+            if (lraId.contains(lra.toASCIIString())) {
+                failedLRAId = lraId;
+            }
+        }
+        List<JSONObject> allRecords = getAllRecords(lra);
+        if (allRecords.isEmpty() && failedLRAId != null) {
+            isMoved = true;
+        } else {
+            isMoved = true;
+            for (JSONObject lraRec : allRecords) {
+                String lraId = (String) lraRec.get("lraId");
+                lraId = lraId.replaceAll("\\\\", "");
+                if (lraId.contains(failedLRAId)) {
+                    return false;
+                }
+            }
+        }
+
+        return isMoved;
+    }
+
+    private List<JSONObject> getAllRecords(URI lra) throws Exception {
+        Response response = null;
+        String coordinatorUrl = LRAConstants.getLRACoordinatorUrl(lra) + "/";
+
+        try {
+            response = client.target(coordinatorUrl).path("").request().get();
+
+            Assert.assertTrue("Missing response body when querying for all LRAs", response.hasEntity());
+            String allLRAs = response.readEntity(String.class);
+
+            JSONArray jsonArray = new JSONArray(allLRAs);
+            List<JSONObject> allLRAList = new ArrayList<>();
+
+            for (int i = 0; i < jsonArray.length(); i++) {
+                allLRAList.add(new JSONObject(jsonArray.getString(i)));
+            }
+
+            return allLRAList;
+        } finally {
+            if (response != null) {
+                response.close();
+            }
+        }
     }
 
     private URI invokeInTransaction(String resourcePrefix, String resourcePath, int expectedStatus) {
--- a/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/FailedLongRunningAction.java
+++ b/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/FailedLongRunningAction.java
@@ -1,0 +1,40 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2021, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package io.narayana.lra.coordinator.domain.model;
+
+import com.arjuna.ats.arjuna.common.Uid;
+import io.narayana.lra.coordinator.domain.service.LRAService;
+
+public class FailedLongRunningAction extends LongRunningAction {
+
+    public static final String FAILED_LRA_TYPE = "/StateManager/BasicAction/LongRunningAction/Failed";
+
+
+    public FailedLongRunningAction(LRAService lraService, Uid rcvUid) {
+        super(lraService, rcvUid);
+    }
+
+    @Override
+    public String type() {
+        return FAILED_LRA_TYPE;
+    }
+}
--- a/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
+++ b/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
@@ -168,6 +168,9 @@ public class LRAService {
     }
 
     public void finished(LongRunningAction transaction, boolean fromHierarchy) {
+        if (transaction.isFailed()) {
+            getRM().moveEntryToFailedLRAPath(transaction.get_uid());
+        }
         if (transaction.isRecovering()) {
             recoveringLRAs.put(transaction.getId(), transaction);
         } else if (fromHierarchy || transaction.isTopLevel()) {
--- a/rts/lra/service-base/src/main/java/io/narayana/lra/logging/lraI18NLogger.java
+++ b/rts/lra/service-base/src/main/java/io/narayana/lra/logging/lraI18NLogger.java
@@ -220,6 +220,16 @@ public interface lraI18NLogger {
     @Message(id = 25145, value = "Unable to process LRA annotations: %s'")
     String warn_LRAStatusInDoubt(String reason);
 
+    @LogMessage(level = WARN)
+    @Message(id = 25146, value = "Unable to remove the failed duclicate failed LRA record (Uid: '%s') " +
+            "(which is already present in the failedLRA record location type: '%s'.) from LRA Record location: '%s'")
+    void warn_UnableToRemoveDuplicateFailedLRARecord(String failedUid, String failedLRAType, String lraType);
+
+    @LogMessage(level = WARN)
+    @Message(id = 25147, value = "An exception was thrown while moving failed LRA record (Uid: '%s'). " +
+            "Reason: '%s'")
+    void warn_move_lra_record(String failedUid, String exceptionMessage);
+
     /*
         Allocate new messages directly above this notice.
           - id: use the next id number in numeric sequence. Don't reuse ids.


ochaloup (changes requested):
It seems nice in general. Just I have few comments in-line.

On top of that, please add some new tests for the new behaviour.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
```diff
@@ -76,6 +76,7 @@
     private boolean inFlight;
     private LRAService lraService;
     private String uid;
+    private boolean isFailedLRAType = false;
```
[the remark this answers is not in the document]
mayankkunwar: The value of `isFailedLRAType` will always be `false` until a constructor is called from https://github.com/jbosstm/narayana/pull/1773/files/a7cdb4f993598108c9c0a974ca70a3a5e98f80d4#diff-62917e7ece8109d31366b0cd7f3a4c0b80cc996e85374f99ede7563c453db778R259
This is just to return the correct type for LRA record, from the `type()` method https://github.com/jbosstm/narayana/pull/1773/files/a7cdb4f993598108c9c0a974ca70a3a5e98f80d4#diff-24f9482f07d3654cd40bbcb491ccb2dcf5250dd05044dd478741c0798b9dd4fdR349

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
```diff
@@ -335,7 +341,14 @@ public static String getType() {
         return LRA_TYPE;
     }
 
+    public static String getFailedLraType() {
+        return LRA_TYPE + "/Failed";
```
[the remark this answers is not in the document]
mayankkunwar: okay

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -130,6 +135,35 @@ private void doRecoverTransaction(Uid recoverUid) {
...
+    private void moveEntryToFailedLRAPath (final Uid failedUid) throws ObjectStoreException {
+        String failedLRAType = LongRunningAction.getFailedLraType();
+        InputObjectState inputState = _recoveryStore.read_committed(failedUid, _transactionType);
+        InputObjectState failedLRAUidState = _recoveryStore.read_committed(failedUid, failedLRAType);
+        boolean result = false;
+
+        if (inputState != null) {
+            if (failedLRAUidState != null) {
+                // Record already exists in failedLRARecord_PATH, hence removing it from the LRA_PATH
+                _recoveryStore.remove_committed(failedUid, _transactionType);
```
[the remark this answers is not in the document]
mayankkunwar: sure

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -130,6 +135,35 @@ private void doRecoverTransaction(Uid recoverUid) {
...
+                return;
+            }
+
+            boolean moved = _recoveryStore.write_committed(failedUid, failedLRAType, new OutputObjectState(inputState));
+
+            if (!moved) {
+                if (LRALogger.logger.isDebugEnabled()) {
+                    LRALogger.logger.debugf("Unable to move failed lra record (Uid: %s) to new location type: %s", failedUid, failedLRAType);
```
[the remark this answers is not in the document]
mayankkunwar: sure

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -130,6 +135,35 @@ private void doRecoverTransaction(Uid recoverUid) {
...
+                }
+            }
+            else {
+                result = _recoveryStore.remove_committed(failedUid, _transactionType);
+                if (result) {
+                    LRALogger.logger.infof("Failed lra record (Uid: %s) moved to new location type: %s", failedUid, failedLRAType);
```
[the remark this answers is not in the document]
mayankkunwar: The `info` logging is just to inform that the Failed LRA record is moved to separate location.
For the case in which the failedLRA record is failed to get deleted from LRA_PATH, this would be retried in the next attempt when the recovery manager will rerun for the next iteration, and this condition ( https://github.com/jbosstm/narayana/pull/1773/files/a7cdb4f993598108c9c0a974ca70a3a5e98f80d4#diff-62917e7ece8109d31366b0cd7f3a4c0b80cc996e85374f99ede7563c453db778R145 ) will remove it and will also inform the same by the logs.

`Failed` records can only be fetched from the new path i.e. `/StateManager/BasicAction/LongRunningAction/failed`, so there will not be any duplicate record.

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/FailedLRAIT.java
```diff
@@ -106,7 +106,9 @@ static String getRecoveryUrl(URI lraId) {
     public void testWithStatusCompensateFailed() throws Exception {
         URI lraId = invokeInTransaction(LRAParticipantWithStatusURI.LRA_PARTICIPANT_PATH,
                 LRAParticipantWithStatusURI.TRANSACTIONAL_CANCEL_PATH, 500);
-
+        // Call the recoveryManager(recovery), after which we can remove "Thread.sleep"
+        // defaultPeriodicRecoveryInterval is 120 seconds
+        Thread.sleep(120000);
```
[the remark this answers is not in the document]
mayankkunwar: In these tests, it is expected to get the deleted record immediately as LRA fails, but after my changes, the deleted records cannot be found until the recoveryManager runs. I am not sure how to proceed on this one. It would be great if you could point me to the right direction?

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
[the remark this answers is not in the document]
ochaloup: I was thinking rather the case when the `_recoveryStore.remove_committed` failed. Then it would be good to inform about the trouble during recovery processing.

When the removal was successful then a trace log could be fine as well (but that should not be `warnf` as it's now).

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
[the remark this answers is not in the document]
ochaloup: `The info logging is just to inform that the Failed LRA record is moved to separate location.`) I see, I got it. Thanks.

`For the case in which the failedLRA record`) Good with the recovery manager and retries. For the second part I just wonder if there could be some race condition in fetching the records and using them. E.g. by LRAService#getFailedLRAs returning the failed ones and with the recovery manager using the duplicate. Or in case of learning the status of the LRA. If that's covered by you then ok, it's fine.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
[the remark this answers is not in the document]
ochaloup: Ok, then this should be `final`.

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/FailedLRAIT.java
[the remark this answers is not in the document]
ochaloup: It could be that the test expects the LRA being marked as failed. That's good and expected.
I think you should verify how the status call is invoked and why you are getting the different value than it was returned previously. I think maybe you need to merge the approaches - which may correspond with the race condition I mentioned above. The failed LRA should be considered based on the location in the object store (based on the `type`) but as well based on the LRA status when it has not been moved yet.

ochaloup (commented):
One more point I'm thinking about. Have you consider what to do when the number of the `/Failed` starts growing? I assume we need some mechanism that we can offer to clean the space. Maybe an `ExpiryScanner` (https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/ExpiryScanner.java)?

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/FailedLRAIT.java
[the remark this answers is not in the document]
mayankkunwar: ok

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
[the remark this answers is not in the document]
mayankkunwar: I will synchronize the loose ends.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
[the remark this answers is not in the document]
mayankkunwar: ok

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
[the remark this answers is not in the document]
mayankkunwar: ok

mayankkunwar:
> One more point I'm thinking about. Have you consider what to do when the number of the `/Failed` starts growing? I assume we need some mechanism that we can offer to clean the space. Maybe an `ExpiryScanner` (https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/ExpiryScanner.java)?

@ochaloup what should be the expiry time?

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
```diff
@@ -109,6 +112,22 @@ public LongRunningAction(LRAService lraService, Uid rcvUid) {
...
+        this.uid = rcvUid.fileStringForm();
+        this.lraService = lraService;
+        this.inFlight = false;
+        this.id = null;
+        this.parentId = null;
+        this.clientId = null;
+        this.finishTime = null;
+        this.status = LRAStatus.Active;
+        this.scheduler = Executors.newScheduledThreadPool(1);
+        this.isFailedLRAType = isFailed;
```
ochaloup, line 130: When the new constructor is needed, could be reused the initialization code from the existing constructor where `this(...)` would be used for the fields already handled by the existing one?

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -130,6 +137,41 @@ private void doRecoverTransaction(Uid recoverUid) {
...
+        InputObjectState failedLRAUidState = _recoveryStore.read_committed(failedUid, failedLRAType);
+        boolean result = false;
+
+        if (inputState != null) {
+            if (failedLRAUidState != null) {
+                // Record already exists in failedLRARecord_PATH, hence removing it from the LRA_PATH
+                boolean rmRec = _recoveryStore.remove_committed(failedUid, _transactionType);
+                if (!rmRec) {
+                    LRALogger.logger.warnf("Unable to remove the failed duclicate failed LRA record (Uid: %s) " +
+                            "(which is already present in the failed LRA record location type: %s.) from LRA_PATH: %s",
```
ochaloup, line 152: I would not use the word `LRA_PATH` as the type does not mean `PATH`. It's just location in the hierarchy, e.g. if you use JDBC object store then you do not work with a `PATH`.

ochaloup: I didn't realize one thing (sorry for that). The convention is for the `error` and `warning` messages to use the `Logger` - ie. add the log under `lraI18NLogger`.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -130,6 +137,41 @@ private void doRecoverTransaction(Uid recoverUid) {
...
+                }
+                LRALogger.logger.infof("The failed lra record (Uid: %s) is already present in the failed LRA record location type: %s." +
+                        "Hence removed the failed duclicate record from LRA_PATH: %s", failedUid, failedLRAType, _transactionType);
+                return;
+            }
+
+            boolean moved = _recoveryStore.write_committed(failedUid, failedLRAType, new OutputObjectState(inputState));
+
+            if (!moved) {
+                LRALogger.logger.warnf("Unable to move failed lra record (Uid: %s) to new location type: %s", failedUid, failedLRAType);
```
ochaloup, line 164: The same what I mentioned above - the `warn` messages should go to `lraI18NLogger`.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -278,7 +337,7 @@ private void forEach(InputObjectState uids, Consumer<Uid> consumer) {
                 if (LRALogger.logger.isTraceEnabled()) {
                     LRALogger.logger.tracef(e,
                             "LRARecoverModule: Object store exception %s while unpacking a record of type %s",
-                            e.getMessage(), _transactionType);
+                            e.getMessage(), transactionType);
                 } else if (LRALogger.logger.isInfoEnabled()) {
                     LRALogger.logger.infof(
                             "LRARecoverModule: Object store exception %s while unpacking a record of type: %s",
```
ochaloup, line 343: The `infof` misses the `transactionType` as the second parameter (there is only one  which is `e.getMessage()`).

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -260,11 +318,12 @@ private boolean getUids(InputObjectState aa_uids) {
      *
      * @param uids the uids to iterate over
      * @param consumer the consumer that should be called for each Uid
+     * @param transactionType within the Object Store for LRAs.
```
ochaloup, line 321: A minor, I'm not sure about the `dot` a the end when the sentence does not start with upper case.
Btw. maybe it could be good to add in the comment that the parameter is used only for logging purposes?

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -200,53 +242,69 @@ public void getRecoveringLRAs(Map<URI, LongRunningAction> lras) {
...
         InputObjectState aa_uids = new InputObjectState();
-        Consumer<Uid> failedLRACreator = uid -> {
-            LongRunningAction lra = new LongRunningAction(lraService, new Uid(uid));
-            lra.activate();
+        String failedTransactionType = LongRunningAction.getFailedLraType();
 
-            LRAStatus status = lra.getLRAStatus();
-            if (LRAStatus.FailedToCancel.equals(status) || LRAStatus.FailedToClose.equals(status)) {
-                lras.put(lra.getId(), lra);
+        synchronized (this) {
```
ochaloup, line 269: Is the `synchronized` necessary here? What's the danger of not synchronized access as it was used before?

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -200,53 +242,69 @@ public void getRecoveringLRAs(Map<URI, LongRunningAction> lras) {
...
-            forEach(aa_uids, failedLRACreator);
+            failedLRACreator = getLRACreater(lras, false);
+            if (getUids(aa_uids, _transactionType)) {
+                forEach(aa_uids, failedLRACreator, _transactionType);
+            }
         }
     }
 
-    private boolean getUids(InputObjectState aa_uids) {
+    private Consumer<Uid> getLRACreater(Map<URI, LongRunningAction> lras, final boolean isFailedLRAType) {
```
ochaloup, line 282:
```suggestion
    private Consumer<Uid> getFailedLRACreator(Map<URI, LongRunningAction> lras, final boolean isFailedLRAType) {
```

ochaloup (commented):
I have few comments but rather minor ones. I really appreciate the fix that omits the recovery to be forced for showing failed LRAs. I have concern about necessity of the `synchronized` but otherwise I think it's fine.

ochaloup:
@mayankkunwar the expiry scanner would not be configured by default, I think. By default it will be disabled.
But when provided it could be configured and the default expiration should be for long time, like a day or two. The expiration time still should be configurable.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
mayankkunwar: Since `isFailedLRAType` is final now, cannot use `this(...)` as it tries to set it's value twice, which would be a compilation error.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mayankkunwar: ok

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mayankkunwar: ok

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mayankkunwar: Here, since the records are now fetched from two different locations, but earlier we were fetching the records from single location. So, `synchronized` could be useful here in the race condition.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mayankkunwar: ok

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mayankkunwar: I will add it thanks.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
ochaloup: Try this
```
    public LongRunningAction(LRAService lraService, Uid rcvUid) {
        this(lraService, rcvUid, false);
    }
```
but to be honest maybe asking for the field to be final is not relevant when all other fields are non-final. But I assume they should be all final. Never mind. Rather to simplify the constructor and not having the field final. Finality of the all fields would be good to be reevaluated later probably.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
ochaloup: We should not be adding `synchronized` at place where we think "it could be useful". When I'm looking at this I believe it could not be necessary as the recovery will eventually finalize the state to be consistent. The only danger which I can think about (please correct me) is a possible duplicity in the output?

mayankkunwar: Yes, it could print duplicate records, if not marked `synchronized`

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
mayankkunwar: ok

mmusgrov:
I have put this PR on hold until the test is ready (so that I don't need to keep checking it).
@mayankkunwar will you remove the Hold label when you push the test, thanks.

mayankkunwar:
Okay @mmusgrov

mayankkunwar:
I have added tests, the code is ready for review

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
ochaloup: I'm really not sure if it's good idea to add `synchronize` here. Why to slow down processing when the consistency issue that are considered here is only possible duplicity - which I can't observe either how that could be for the achieved (?!) with the used data structure of {{Map}}. Use a data structure that eliminates the duplicity and that's it, or am I wrong in my assumption?

ochaloup (commented):
Seems fine to me. Just adding few generic points.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mayankkunwar: Correct, but there could also be a scenario in which thread calls get failed logs, and at the same time anoher thread deletes the failed logs, at that point of time, we are returning some logs to the user, but actually there is no record present in the system.

So if the mentioned point is not of much concern, I can remove the `synchronized` keyword.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/ExpiredFailedLRATransactionManagerScanner.java
```diff
@@ -0,0 +1,124 @@
...
+
+                            recoveryStore.remove_committed( newUid, transactionType ) ;
+                        }
+                    }
+                }
+            }
+        }
+        catch ( Exception e )
+        {
+            // end of uids!
```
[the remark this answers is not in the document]
mayankkunwar: I also don't know this, I thought there would be some reason.
Logging the exception make sense, thanks.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/ExpiredFailedLRATransactionManagerScanner.java
```diff
@@ -0,0 +1,124 @@
...
+        this.recoveryStore = StoreManager.getRecoveryStore();
+        this.transactionType = LongRunningAction.getFailedLraType();
+    }
+
+    /**
+     * This is called periodically by the RecoveryManager
+     */
+    public void scan () {
+        // calculate the time before which items will be removed
+        Date oldestSurviving = new Date( new Date().getTime() - expiryTime * 1000 ) ;
```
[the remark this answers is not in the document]
mayankkunwar: Okay, using `LocalDateTime` in LRA module would be better, then. I will change it.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/ExpiredFailedLRATransactionManagerScanner.java
```diff
@@ -0,0 +1,124 @@
...
+
+public class ExpiredFailedLRATransactionManagerScanner implements ExpiryScanner {
+
+
+    private RecoveryStore recoveryStore;
+
+    // 'type' within the Object Store for LRAs
+    private final String transactionType;
+
+    protected int expiryTime = recoveryPropertyManager.getRecoveryEnvironmentBean()
```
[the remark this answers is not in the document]
mayankkunwar: I was thinking it from the perspective of writing a test case.

For integration testing, if we want to test it, I was thinking, I would create a extended class and override the `expiryTime` time as here whatever value we are getting from `recoveryPropertyManager` we are converting it to `hours`.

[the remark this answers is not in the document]
ochaloup: In general writing a code to be testable is a great approach. But opening encapsulation to only make test easier is not  great. For me this is not a good practice.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
ochaloup: Yes, you are right. Those are good points you say. Let me scratch another two examples to continue the discussion.

Imagine that a thread gets the failed logs and it gets out of the synchronized block back to the REST endpoint but before the HTTP I/O response is sent the thread is suspended. The thread scheduler assigns the time to the recovery thread which deletes the failed logs. Does the `synchronized` helped here? Do you have some chance to defend against such "inconsistency"?

Another question. The remote API asks for failed LRAs, the coordinator returns it and the program saves that to internal map. It starts to iterate over it and starts printing it to the console. Meanwhile the coordinator recovery deletes the failed logs. Does the `synchronized` helped here? Do you have some chance to defend against such "inconsistency"?

ochaloup (commented):
Mostly fine for me. Thank you!
My last big concern is if the `synchronized` is needed. I don't like opening the encapsulation but it can be re-evaluated on creating a test.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/ExpiredFailedLRATransactionManagerScanner.java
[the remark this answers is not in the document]
mayankkunwar: Okay, I will make it `private`.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mayankkunwar: Okay, in your scenarios, `synchronized` keyword would not help.
I got your point, that the scenario of inconsistency could be possible by other means. So just `synchronizing` at the particular point would not help much, but instead make the runs slow.
I will remove this `synchronized` block then.

ochaloup (approved):
It's fine by me. Thank you.
Will you @mmusgrov take a look as well?

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -98,8 +99,14 @@ private void doRecoverTransaction(Uid recoverUid) {
             boolean inFlight = (lra.getLRAStatus() == LRAStatus.Active);
 
             LRAStatus lraStatus = lra.getLRAStatus();
-            if (LRAStatus.FailedToCancel.equals(lraStatus) || LRAStatus.FailedToClose.equals(lraStatus)) {
-                return;
+
+            synchronized (this) {
```
ochaloup, line 103: synchronized here, is necessary?

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -200,53 +241,66 @@ public void getRecoveringLRAs(Map<URI, LongRunningAction> lras) {
...
-        } catch (ObjectStoreException e) {
-            if (LRALogger.logger.isTraceEnabled()) {
-                LRALogger.logger.tracef(e,
-                        "LRARecoveryModule: Object store exception '%s' while removing LRA record %s",
-                        e.getMessage(), lraUid.fileStringForm());
-            } else if (LRALogger.logger.isInfoEnabled()) {
-                LRALogger.logger.infof(
-                        "LRARecoveryModule: Object store exception '%s' while removing LRA record %s",
-                        e.getMessage(), lraUid.fileStringForm());
+        synchronized (this) {
```
ochaloup, line 244: synchronized here, is necessary?

ochaloup (changes requested):
sorry @mayankkunwar I found there are two more usages of `synchronized` in your code. I would like to understand if they are necessary.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mayankkunwar: I will remove it for same reason as discussed earlier.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mayankkunwar: I remove it from here as well. I thought the same scenario could happen here as well, but this would not help much.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/ExpiredFailedLRATransactionManagerScanner.java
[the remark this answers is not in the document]
mmusgrov: Uid.nullUid() marks the end of the collection.
Normally one would catch ObjectStoreException when calling into ArjunaCore.
The unpacking code can throw IOException (UidHelper.unpackFrom).

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -130,6 +135,40 @@ private void doRecoverTransaction(Uid recoverUid) {
...
+                // Record already exists in failedLRARecord location, hence removing it from the LRARecord location
+                boolean rmRec = _recoveryStore.remove_committed(failedUid, _transactionType);
+                if (!rmRec) {
+                    LRALogger.i18NLogger.warn_UnableToRemoveDuplicateFailedLRARecord(
+                            failedUid.toString(), failedLRAType, _transactionType);
+                    return;
+                }
+                LRALogger.logger.infof("The failed lra record (Uid: %s) is already present in the failed LRA record location type: %s." +
+                        "Hence removed the failed duclicate record from LRARecord location: %s", failedUid, failedLRAType, _transactionType);
+                return;
```
mmusgrov, line 155: why two return statements if (!rmRec). Easy to read would be:

```java
                if (!_recoveryStore.remove_committed(failedUid, _transactionType)) {
                    LRALogger.logger.infof("The failed lra record (Uid: %s) is already present in the failed LRA record location type: %s." +
                            "Hence removed the failed duclicate record from LRARecord location: %s", failedUid, failedLRAType, _transactionType);
                } else {
                    LRALogger.i18NLogger.warn_UnableToRemoveDuplicateFailedLRARecord(
                            failedUid.toString(), failedLRAType, _transactionType);
                }
                return;
```
Under what circumstances will the record exist in two parts of the store (should only happen if the server has previously crashed or are there other scenarios).

If the delete failed then the warning will be repeated forever after.

Why is this logged at warning level, what are the consequences of being unable to delete the record. Surely it is more informational but if you want to log at warn level then only do it once.

mmusgrov: Also there is too much logging going on here, moving a record can now produce four different log statements. Why is the existing ArjunaCore code insufficient:

```java
	    InputObjectState state = _recoveryStore.read_committed(newUid, _typeName);
	    boolean res = false;
	    
	    if (state != null) // just in case recovery
	        // kicked-in
	    {
	        boolean moved = _recoveryStore.write_committed(newUid, _movedTypeName, new OutputObjectState(state));

	        if (!moved) {
                tsLogger.logger.debugf("Removing old transaction status manager item %s", newUid);
                }
	        else {
	            res = _recoveryStore.remove_committed(newUid, _typeName);
                    tsLogger.i18NLogger.warn_recovery_ExpiredTransactionStatusManagerScanner_6(newUid);
                }

	    }
```

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -201,7 +240,8 @@ public void getRecoveringLRAs(Map<URI, LongRunningAction> lras) {
      */
     public boolean removeCommitted(Uid lraUid) {
         try {
-            return _recoveryStore.remove_committed(lraUid, _transactionType);
+            return _recoveryStore.remove_committed(lraUid, _transactionType) ||
+                    _recoveryStore.remove_committed(lraUid, LongRunningAction.getFailedLraType());
```
mmusgrov, line 244: Why are we removing failed records?

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -219,34 +259,44 @@ public boolean removeCommitted(Uid lraUid) {
...
-            LongRunningAction lra = new LongRunningAction(lraService, new Uid(uid));
-            lra.activate();
+        String failedTransactionType = LongRunningAction.getFailedLraType();
 
-            LRAStatus status = lra.getLRAStatus();
-            if (LRAStatus.FailedToCancel.equals(status) || LRAStatus.FailedToClose.equals(status)) {
-                lras.put(lra.getId(), lra);
-            }
-        };
+        Consumer<Uid> failedLRACreator = getFailedLRACreator(lras, true);
```
mmusgrov, line 264: We just need a list of the records, ie we don't need to activate the record. The following should be sufficient:

``` java
_recoveryStore.allObjUids("/StateManager/BasicAction/LongRunningAction/Failed", <input buffer);
```

rts/lra/service-base/src/main/java/io/narayana/lra/LRAData.java
```diff
@@ -41,11 +41,12 @@
     private long startTime;
     private long finishTime;
     private int httpStatus;
+    private String lraType;
 
     public LRAData() {}
 
     public LRAData(URI lraId, String clientId, LRAStatus status, boolean isTopLevel, boolean isRecovering,
-                    long startTime, long finishTime, int httpStatus) {
+                    long startTime, long finishTime, int httpStatus, String lraType) {
```
mmusgrov, line 49: the status indicates the terminal state so the lraType data is not required.

mmusgrov (changes requested):

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
```diff
@@ -64,6 +64,7 @@
 
 public class LongRunningAction extends BasicAction {
     private static final String LRA_TYPE = "/StateManager/BasicAction/LongRunningAction";
+    private static final String FAILED_LRA_TYPE = LRA_TYPE + "/Failed";
```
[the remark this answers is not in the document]
mayankkunwar: Okay, I will move it to RecoveringLRA

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
```diff
@@ -93,11 +95,16 @@ public LongRunningAction(LRAService lraService, String baseUrl, URI parentId, St
...
+        this.isFailedLRAType = false;
 
         this.scheduler = Executors.newScheduledThreadPool(1);
     }
 
     public LongRunningAction(LRAService lraService, Uid rcvUid) {
+        this(lraService, rcvUid, false);
+    }
+
+    public LongRunningAction(LRAService lraService, Uid rcvUid, final boolean isFailed) {
```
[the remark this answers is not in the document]
mayankkunwar: Okay

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
```diff
@@ -121,7 +129,7 @@ public LRAData getLRAData() {
         return new LRAData(id, clientId, status, isTopLevel(), isRecovering(),
                 startTime.toInstant(ZoneOffset.UTC).toEpochMilli(),
                 finishTime == null ? 0L : finishTime.toInstant(ZoneOffset.UTC).toEpochMilli(),
-                getHttpStatus());
+                getHttpStatus(), type());
```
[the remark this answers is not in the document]
mayankkunwar: But failed status will not verify that the failed records have moved to another location. How do I verify this?

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
```diff
@@ -335,7 +343,19 @@ public static String getType() {
...
+        return FAILED_LRA_TYPE;
+    }
+
+    public LocalDateTime getFinishTime() {
+        return this.finishTime;
+    }
+
+
     public String type() {
+        if (this.isFailedLRAType)
```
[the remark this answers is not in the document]
mayankkunwar: okay, I will move it to RecoveringLRA

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/ExpiredFailedLRATransactionManagerScanner.java
```diff
@@ -0,0 +1,126 @@
...
+import com.arjuna.ats.arjuna.objectstore.StoreManager;
+import com.arjuna.ats.arjuna.recovery.ExpiryScanner;
+import com.arjuna.ats.arjuna.state.InputObjectState;
+import com.arjuna.ats.internal.arjuna.common.UidHelper;
+
+import io.narayana.lra.logging.LRALogger;
+import io.narayana.lra.coordinator.domain.model.LongRunningAction;
+import io.narayana.lra.coordinator.domain.service.LRAService;
+
+public class ExpiredFailedLRATransactionManagerScanner implements ExpiryScanner {
```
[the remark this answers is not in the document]
mayankkunwar: In `ExpiredTransactionStatusManagerScanner` there was no way I could pass the `transactionType` and I thought modifying that class was not good idea.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -75,7 +76,7 @@ public void periodicWorkFirstPass() {
         // uids per transaction type
         InputObjectState aa_uids = new InputObjectState();
 
-        if (getUids(aa_uids)) {
+        if (getUids(aa_uids, _transactionType)) {
```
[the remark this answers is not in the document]
mayankkunwar: Actually, here the problem was, if there is a failed record present then, it would not be moved to failed location until recoveryManager executes its next run, so before that we need to check both the locations.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -98,7 +99,11 @@ private void doRecoverTransaction(Uid recoverUid) {
             boolean inFlight = (lra.getLRAStatus() == LRAStatus.Active);
 
             LRAStatus lraStatus = lra.getLRAStatus();
-            if (LRAStatus.FailedToCancel.equals(lraStatus) || LRAStatus.FailedToClose.equals(lraStatus)) {
+
+            // This condition is to check if there in any failed LRA record
+            if (lra.isFailed()) {
+                // Moving this failed record to another part of store
+                moveEntryToFailedLRAPath(recoverUid);
```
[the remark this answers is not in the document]
mayankkunwar: okay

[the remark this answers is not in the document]
mayankkunwar: Here also it is the same scenario as I mentioned above. If there is a failed record present then, it would not be moved to failed location until recoveryManager executes its next run, so before that we need to check both the locations.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
```diff
@@ -130,6 +135,40 @@ private void doRecoverTransaction(Uid recoverUid) {
...
+            if (failedLRAUidState != null) {
+                // Record already exists in failedLRARecord location, hence removing it from the LRARecord location
+                boolean rmRec = _recoveryStore.remove_committed(failedUid, _transactionType);
+                if (!rmRec) {
+                    LRALogger.i18NLogger.warn_UnableToRemoveDuplicateFailedLRARecord(
+                            failedUid.toString(), failedLRAType, _transactionType);
+                    return;
+                }
+                LRALogger.logger.infof("The failed lra record (Uid: %s) is already present in the failed LRA record location type: %s." +
+                        "Hence removed the failed duclicate record from LRARecord location: %s", failedUid, failedLRAType, _transactionType);
```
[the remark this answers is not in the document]
mayankkunwar: okay

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mayankkunwar: Okay, I will log warn level only once.

Here is the scenario in which the record will be available in both the location:
1. Failed records are in `/StateManager/BasicAction/LongRunningAction/` -> RecoveryManager Kicked in -> reading from `/StateManager/BasicAction/LongRunningAction/` for first time -> writing the failed record at `/StateManager/BasicAction/LongRunningAction/Failed` -> now deleting/removing the failed record from `/StateManager/BasicAction/LongRunningAction/` -> failed to `delete/remove` due to system crash or maybe for some other reason.

Now at this point of time we will have the failed record in both the locations.

Now recoveryManager kicks in for second time -> from here we can check if the record already exists in `/StateManager/BasicAction/LongRunningAction/Failed`?

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mayankkunwar: Because `removeCommitted` is called from LRAService#removeLog (https://github.com/jbosstm/narayana/blob/master/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java#L188)
and `LRAService#removeLog` is called from RecoveryCoordinator#deleteFailedLRA (https://github.com/jbosstm/narayana/blob/master/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/RecoveryCoordinator.java#L205)

Here we are checking both the location as explained in previous comments.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mayankkunwar: Now above comments will change the implementation here.

rts/lra/service-base/src/main/java/io/narayana/lra/LRAData.java
mayankkunwar: Status would not verify that the failed records are moved to another location. Is there any other way to do this?

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/ExpiredFailedLRATransactionManagerScanner.java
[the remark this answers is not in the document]
mmusgrov: Don't delete failure records. Transactional software provides a higher level of guarantee than many other systems and the failure records provide key information for post failure analysis of the consequences of those failures.

At most, if you are worried about storage/denial of service, then stop creating new LRAs until the situation has been resolved. But that is the subject of another issue/design proposal.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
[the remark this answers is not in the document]
mmusgrov: I would advise against modifying the implementation just to test behaviour (use a test to do that instead).
You can verify it by attempting to read the record from the object store.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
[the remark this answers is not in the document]
mmusgrov: My point was that the requirement for this new functionality is twofold: to move the failed record and to query for failed transactions by looking in a specific part of the store.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mmusgrov: Okay

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mmusgrov: You should only be removing the original record. Do not delete failure records from the new location.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mmusgrov: Okay I will review that change when you next push an update to this PR, thanks. If we can keep the changes as minimal as possible then we can reduce opportunities for bugs to creep in (the ratio of bugs to new lines of code is still quite high in our industry).

rts/lra/service-base/src/main/java/io/narayana/lra/LRAData.java
mmusgrov: I addressed this in one of my earlier comments.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/ExpiredFailedLRATransactionManagerScanner.java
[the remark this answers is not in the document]
mayankkunwar: Okay, I have removed this class. We can discuss this as part of another issue.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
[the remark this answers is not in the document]
mayankkunwar: okay

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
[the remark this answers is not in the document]
mayankkunwar: okay

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mayankkunwar: okay

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/LRARecoveryModule.java
mayankkunwar: I have pushed new changes. Here almost the code is same a before, just I did some code refactoring, and checking the failed record in both the location.

rts/lra/service-base/src/main/java/io/narayana/lra/LRAData.java
mayankkunwar: okay, thanks

mmusgrov (approved):
Thanks Mayank, it looks good.
When you have resolved the conflicts will you merge the PR.

mayankkunwar:
Okay, Mike, I will resolve the conflict and merge.
Thanks

mmusgrov:
> One more point I'm thinking about. Have you consider what to do when the number of the `/Failed` starts growing? I assume we need some mechanism that we can offer to clean the space. Maybe an `ExpiryScanner` (https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/ExpiryScanner.java)?

I addressed that comment earlier, can you respond to my response instead.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/internal/ExpiredFailedLRATransactionManagerScanner.java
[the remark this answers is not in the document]
mmusgrov: @ochaloup can you respond directly to this comment rather than requesting changes to Mayank's PR?

mmusgrov:
> > One more point I'm thinking about. Have you consider what to do when the number of the `/Failed` starts growing? I assume we need some mechanism that we can offer to clean the space. Maybe an `ExpiryScanner` (https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/ExpiryScanner.java)?
> 
> I addressed that comment earlier, can you respond to my response instead.

Ah this is an old comment from the original review, apologies, I thought you had re-reviewed it and withheld your approval, sorry.
Anyway, @ochaloup it looks like Mayank has resolved all of the issues.

ochaloup:
@mmusgrov yes, I think all issues were resolved. By me, after the rebase, it's fine to be merged.

ochaloup (approved):

PR #1774, opened 2021-02-02
JBTM-3421 remove provided scope for artefact mashona:mashona-logwriting

mmusgrov:
https://issues.redhat.com/browse/JBTM-3421

This is a follow up to JBTM-3417 (Add Persistent Memory SlotStore backend) to workaround a javadoc issue compiling on JDK 11. I'm not sure if it is the best fix but it will do as a stop gap (an example of the compilation failure is http://narayanaci1.eng.hst.ams2.redhat.com/job/narayana/PROFILE=CORE,jdk=jdk11.latest,label=linux/21/consoleFull).

We are using OpenJDK 64-Bit Server VM 18.9 (build 11.0.6+10-LTS, mixed mode, sharing)

MAIN CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF !LRA !NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/ArjunaCore/arjuna/pom.xml
+++ b/ArjunaCore/arjuna/pom.xml
@@ -93,7 +93,7 @@
     <dependency>
       <groupId>io.mashona</groupId>
       <artifactId>mashona-logwriting</artifactId>
-      <version>1.0.0.Beta1</version>
+      <version>${version.io.mashona}</version>
       <scope>provided</scope>
     </dependency>
     <dependency>
--- a/narayana-full/pom.xml
+++ b/narayana-full/pom.xml
@@ -417,6 +417,14 @@
           </plugin>
         </plugins>
       </build>
+      <dependencies>
+        <dependency>
+          <groupId>io.mashona</groupId>
+          <artifactId>mashona-logwriting</artifactId>
+          <version>${version.io.mashona}</version>
+          <scope>compile</scope>
+        </dependency>
+      </dependencies>
     </profile>
     <profile>
       <id>community</id>
--- a/pom.xml
+++ b/pom.xml
@@ -496,6 +496,7 @@
     <version.com.github.docker-java>3.0.14</version.com.github.docker-java>
     <version.org.eclipse.microprofile.openapi>1.1.2</version.org.eclipse.microprofile.openapi>
     <version.smallrye-converter-api>1.0.10</version.smallrye-converter-api>
+    <version.io.mashona>1.0.0.Beta1</version.io.mashona>
 
     <!-- Maven plugin versions -->
     <version.org.codehaus.mojo.jboss-maven-plugin>1.5.0</version.org.codehaus.mojo.jboss-maven-plugin>


ArjunaCore/arjuna/pom.xml
```diff
@@ -94,7 +94,7 @@
       <groupId>io.mashona</groupId>
       <artifactId>mashona-logwriting</artifactId>
       <version>1.0.0.Beta1</version>
-      <scope>provided</scope>
+      <scope>compile</scope>
```
tomjenkinson, line 97: Is it intentional to also add this to the runtime classpath (which is what would happen by making compile)? https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html

narayana-full/pom.xml
```diff
@@ -356,6 +356,7 @@
                     -exclude com.arjuna.orbportability.internal -exclude com.arjuna.ats.jbossatx
                     -exclude com.arjuna.mw.wstx -exclude org.jboss.jbossts.txbridge
                     -exclude org.jboss.jbossts.xts -exclude org.jboss.jbossts.xts11
+                    -exclude com.arjuna.ats.internal.arjuna.objectstore.slot
```
tomjenkinson, line 359: Do you know why the exclude of com.arjuna.ats.internal on line 355 is not enough? perhaps we need to somehow tell Javadoc exclude recursively. If that is not available as an option this works for me

ArjunaCore/arjuna/pom.xml
mmusgrov: Jonathan originally had it as provided which I understood to mean that it was provided by the JDK so it was confusing why it only failed on JDK 11 (presumably it is a modules issue). But as a minimum it is needed for compilation. Let's wait for Jonathan to return.

narayana-full/pom.xml
mmusgrov: I don't think this line actually had any effect. The main change was to remove the provided scope from the dependency.

tomjenkinson: To test this, how about a commit reverting that line to see if it still passes?

ArjunaCore/arjuna/pom.xml
tomjenkinson: In maven terms provided is when the dependency is available present within the runtime environment already, so yeah if the JDK is providing it it seems that it should not be added in by Narayana too. But if the JDK is providing it then also I would think that it wouldn't have to be added for compilation (e.g. we don't need to for java.lang). as you say, it could be a modules thing we need to add. That said, would the JDK have io.mashona:mashona-logwriting things in it?

tomjenkinson: I think I was on a bit the wrong steer as provided is also present at compile too. So this is about exposing it to the javadoc compiler which might help explain the other change

narayana-full/pom.xml
tomjenkinson: I think try a commit without the other change rather than this, we don't want class in the javadocs so probably the exclusion here is what we need and if so then although I don't think we should need to add each package independently perhaps we do for now? We should try to get this excluded recursively for all internal sub-packages though.

jhalliday (approved):

ochaloup (approved):

PR #1780, opened 2021-02-04
bumping version of org.apache.ant:ant to 1.10.9

ochaloup:
MAIN QA_JTA
!QA_JTS_OPENJDKORB !QA_JTS_JACORB !XTS !PERF NO_WIN !RTS !AS_TESTS !TOMCAT !JACOCO !LRA ! BLACKTIE !QA_JTS_JDKORB

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -466,7 +466,7 @@
     <version.jfree>1.0.9</version.jfree>
     <version.com.h2database>1.4.195</version.com.h2database>
     <version.org.apache.tomcat>9.0.7</version.org.apache.tomcat>
-    <version.org.apache.ant>1.8.2</version.org.apache.ant>
+    <version.org.apache.ant>1.10.9</version.org.apache.ant>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -466,7 +466,7 @@
     <version.jfree>1.0.9</version.jfree>
     <version.com.h2database>1.4.195</version.com.h2database>
     <version.org.apache.tomcat>9.0.7</version.org.apache.tomcat>
-    <version.org.apache.ant>1.8.2</version.org.apache.ant>
+    <version.org.apache.ant>1.10.9</version.org.apache.ant>
```
mmusgrov, line 469: We don't have this change on master, I wonder why we didn't get an alert for it?

mmusgrov (approved):
You will need to resolve the two conflicts.

pom.xml
ochaloup: Maybe because the resolution mechanism is not clever enough and is not capable to manage the version in property.

PR #1782, opened 2021-02-08
[JBTM-3398] New test to use different types of object stores with Undertow

jmfinelli:
This PR resolves the JIRA issue [JBTM-3398](https://issues.redhat.com/browse/JBTM-3398)

!MAIN !CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF !NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

LRA


mmusgrov:
@jmfinelli how does this PR relate to pr #1765 ?
We normally stipulate 1 PR per issue. I will update the project guidelines to make this clear.

jmfinelli:
As we have split JBTM-3398 in two different tickets, I thought it might be beneficial to close #1765 and leave only this PR to fix JBTM-3398

mmusgrov:
But #1765 already captures some discussion about the issue would be lost if you withdrew it. What would be the benefit of closing the other one and using this PR instead?

jmfinelli:
> But #1765 already captures some discussion about the issue would be lost if you withdrew it. What would be the benefit of closing the other one and using this PR instead?

As far as I understand, #1765 was created to modify two recovery tests that ran as integration tests using a Wildfly container. As we decided to [move](https://issues.redhat.com/browse/JBTM-3419) those integration tests to `lra-test/basic` and convert those two tests in `coordinator` to a single (not-recovery) unit test to verify just that the coordinator works with different types of Object Stores (using Undertow instead of Wildfly), I thought that the conversation in #1765 arrived at an EOL and it would have been clearer to start a new PR to kick off a new conversation.

If you think that this is unnecessary, please, let me know and I will close this PR. Thanks

mmusgrov:
#1765 was not created to modify two recovery tests that ran as integration tests using a Wildfly container, I think you are mixing it up with https://issues.redhat.com/browse/JBTM-3419

JBTM-3398 (Test LRA behaviour on different Object Stores) was the subject of your other PR and it is also the subject of this PR too so we should not have both. The other one already contains some review history so that is the one to keep.

jmfinelli:
OK, no problem. Thanks for the clarification @mmusgrov

PR #1786, opened 2021-02-17
JBTM-3428 jbossts-properties.xml not read when deploying LRA coodinat…

mmusgrov:
https://issues.redhat.com/browse/JBTM-3428

LRA


mmusgrov:
The test failure is a know transient error on our CI infrastructure and is unrelated to this PR.

ochaloup (approved):
It's a bit pity we need to hardcode the recovery module definition but per discussion I think it's a good solution for now.
It should be re-considered later.

mmusgrov:
> It's a bit pity we need to hardcode the recovery module definition but per discussion I think it's a good solution for now.
> It should be re-considered later.

I agree, but it will need to stay until we have the LRA subsystem in place.

PR #1797, opened 2021-03-10
subresource

mmusgrov:
https://issues.redhat.com/browse/JBTM-3441

!MAIN !CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF !LRA NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
+++ b/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
@@ -112,6 +112,7 @@ public class Coordinator extends Application {
     private UriInfo context;
 
     private final LRAService lraService;
+    private final RecoveryCoordinator recoveryCoordinator;
 
     @Override
     public Set<Class<?>> getClasses() {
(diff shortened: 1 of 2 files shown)


mmusgrov:
@xstefank it doesn't let me assign you as the reviewer. Can you take a look please.

xstefank (approved):

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
```diff
@@ -122,6 +123,12 @@
 
     public Coordinator() {
         lraService = LRARecoveryModule.getService();
+        recoveryCoordinator = new RecoveryCoordinator();
+    }
+
+    @Path(RECOVERY_COORDINATOR_PATH_NAME)
+    public RecoveryCoordinator getRecoveryCoordinator() {
+        return recoveryCoordinator;
```
xstefank, line 131: @ochaloup correctly pointed out that this will use the same instance of RecoveryCoordinator for each call. I believe that quarkus defaults to application scoped beans for JAX-RS resources so I don't think that this should be an issue, just something to keep in mind.

mmusgrov: It's irrelevant because RecoveryCoordinator is not a singleton, ie you can create however many instances you like, ie the front end API is stateless.

PR #1799, opened 2021-03-15
Added more notes on handling Oracle

tomjenkinson:
NO_TEST


mmusgrov (approved):
Thanks.

I will try again, bearing these notes in mind, to see if I can get the db working on my laptop.

PR #1801, opened 2021-03-17
[JBTM-3443] added the orson dependency to coordinator-thorntail

jmfinelli:
This PR fixes the JIRA [JBTM-3443](https://issues.redhat.com/browse/JBTM-3443)

MAIN WIN
!TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF LRA !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/rts/pom.xml
+++ b/rts/pom.xml
@@ -90,28 +90,37 @@ Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
     </profile>
     <profile>
 	  <id>jdk11</id>
-          <activation>
-             <jdk>[11,)</jdk>
-          </activation>
-          <properties>
-             <jvm.args.modular>--add-modules=java.se</jvm.args.modular>
-	     <version.org.jboss.spec.javax.xml.bind.jboss-jaxb-api_2.3_spec>1.0.1.Final</version.org.jboss.spec.javax.xml.bind.jboss-jaxb-api_2.3_spec>
-	  </properties>
-	  <dependencies>
-             <dependency>
-                <groupId>org.jboss.spec.javax.xml.bind</groupId>
-                <artifactId>jboss-jaxb-api_2.3_spec</artifactId>
-                <version>${version.org.jboss.spec.javax.xml.bind.jboss-jaxb-api_2.3_spec}</version>
-		<scope>provided</scope>
-             </dependency>
-	  </dependencies>
-      </profile>
+      <activation>
+        <jdk>[11,)</jdk>
+      </activation>
+        <properties>
+          <jvm.args.modular>--add-modules=java.se</jvm.args.modular>
+          <version.org.jboss.spec.javax.xml.bind.jboss-jaxb-api_2.3_spec>1.0.1.Final</version.org.jboss.spec.javax.xml.bind.jboss-jaxb-api_2.3_spec>
+        </properties>
+      <dependencies>
+        <dependency>
+        <groupId>org.jboss.spec.javax.xml.bind</groupId>
+        <artifactId>jboss-jaxb-api_2.3_spec</artifactId>
+        <version>${version.org.jboss.spec.javax.xml.bind.jboss-jaxb-api_2.3_spec}</version>
+        <scope>provided</scope>
+        </dependency>
+      </dependencies>
+    </profile>
     <profile>
     <id>community</id>
       <modules>
         <module>lra</module>
         <module>sra</module>
       </modules>
+      <dependencies>
+        <dependency>
+          <groupId>orson</groupId>
+          <artifactId>orson</artifactId>
+          <version>${version.orson}</version>
+          <scope>system</scope>
+          <systemPath>${orson.jar.location}/orson-0.5.0.jar</systemPath>
+        </dependency>
+      </dependencies>
     </profile>
   </profiles>
   <modules>


mmusgrov (changes requested):
Can you add the dependency in a community profile instead like we do in the other modules (ArjunaCore/arjunacore/pom.xml for example).

mmusgrov:
@jmfinelli I think you should assign yourself to the JIRA (it's currently set to Ondra).

jmfinelli:
> Can you add the dependency in a community profile instead like we do in the other modules (ArjunaCore/arjunacore/pom.xml for example).

@ochaloup and I discussed this option. @ochaloup suggested that the entire `rts/lra` module is activated under the `community` profile (@ochaloup, @mmusgrov please, correct me if I am wrong on this)

jmfinelli:
Thanks @ochaloup for your review and approval.

Let's continue this conversation tomorrow as I think we won't get any result from the CI before then

mmusgrov:
My comment was that we should only add the dependency when the community profile is active.

jmfinelli:
> My comment was that we should only add the dependency when the community profile is active.

Please, read my edited comment above.
I think it makes sense to give it a go :+1: I'll update the PR

mmusgrov:
orson has nothing to do with the lra module and the failure is a quirk of our build system.
So add the dependency to the community profile in the rts pom (org.jboss.narayana.rts:rest-tx).

ochaloup (commented):
I'm abstaining from approval because I'm unsure.
Should not be the profile added only to `coordiantor-thorntail` where is the only place where the dependency is needed? I have a similiar thinking in way the `orson` is not specified in root `pom.xml` but only in a child module `narayana-jta`.

mmusgrov:
> I'm abstaining from approval because I'm unsure.
> Should not be the profile added only to `coordiantor-thorntail` where is the only place where the dependency is needed? I have a similiar thinking in way the `orson` is not specified in root `pom.xml` but only in a child module `narayana-jta`.

 `coordiantor-thorntail` does not need the orson dependency

mmusgrov:
In fact the correct place to fix the problem is in the module that requires it (namely ArjunaCore/arjuna - see TxPerfGraph.java).
How would you (@ochaloup ) feel about providing a review if the dependency was constrained to only the module that requires it?

rts/pom.xml
```diff
@@ -90,28 +90,37 @@ Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
     </profile>
     <profile>
 	  <id>jdk11</id>
-          <activation>
```
mmusgrov, line 93: Remark: we normally only include changes related to the issue. Also notice that even if your goal was to clean up the white space there is a tab on the previous line (line 92).

I don't block PRs for spurious white space clean up because it is very difficult to stop the IDE from correcting white space (and I often make the same error myself and would not like my PRs being blocked because of it).

jmfinelli: Thanks for spotting that tab; I'll correct it and push another commit.

Thanks for clarifying this. So, should I open a new issue/create a new PR if I think a code would benefit from a bit of cleaning up?

ochaloup:
@mmusgrov @jmfinelli 
The point of the issue was to get rid of the error during the build of the thorntail coordinator. If we know that the `lra-coordinator` does not need the orson we can use the fix of exclusion of the dependency - https://github.com/ochaloup/narayana/commit/77300e1fdd999fc2168e95e79e24e99f2494346a

The issue is with the Thorntail resolution process that the orson is transitively defined in the chain of dependencies and Thorntail wants to download all  of them. Otherwise there could be CNFE when fat jar is started. Thorntail does not know what is needed for the runtime.
@jmfinelli When we know the coordinator does not need it then probably my "fix" is better :-)

@mmusgrov `Do you know why we need the dependency in so many modules when it is only used by ArjunaCore/arjuna`) Not sure. I was thinking it's because of the the dependency chain. But with the `system` scope it does not much sense. I make a short check and removed it from the modules like this: https://github.com/ochaloup/narayana/commit/91398744053e2c2e08daf7a51ce759d4f3ccdf54
and the build works.

In fact, I've never used the orson and jchart in Narayana before. Do you know @mmusgrov where or how that is used?

rts/pom.xml
mmusgrov: We don't clean up historical code because it makes analysing diffs problematic.
We have a policy somewhere about it so I will dig that out and pass it on.

mmusgrov: But yes, one PR per issue (unless multiple issues are tightly interrelated).

mmusgrov:
Okay then let's go with an exclusion in the thorntail pom.

I believe we use orson for charting performance. It is historical and probably not used by anyone but we can't remove it without deprecating. But can say it's annoying ;-)

jmfinelli:
> @jmfinelli When we know the coordinator does not need it then probably my "fix" is better :-)

Yep, I think so :smile: and also @mmusgrov agrees with you. @ochaloup I'll close this PR then and let you open a new one from your branch :+1:

rts/pom.xml
jmfinelli: Thanks for clarifying this :+1:

ochaloup:
@jmfinelli @mmusgrov about the orson dependencies - do you think that I should create a PR from changes like these?
https://github.com/ochaloup/narayana/commit/91398744053e2c2e08daf7a51ce759d4f3ccdf54

jmfinelli:
> I believe we use orson for charting performance. It is historical and probably not used by anyone but we can't remove it without deprecating. But can say it's annoying ;-)

@ochaloup I agree with @mmusgrov  that we cannot remove this library before deprecating it

ochaloup:
@jmfinelli @mmusgrov I was not thinking about removing library, I was thinking to remove the dependency declaration at other places except in core. Just and idea. Let's leave it as it is.

PR #1803, opened 2021-03-22
[RELEASE] on release deploy to deploy lra-coordinator-quarkus-runner jar as well

ochaloup:
MAIN

When we do the release we use the `-Prelease` which activates the release profile. The maven behaviour for `<activeByDefault>true</activeByDefault>` is that `activeByDefault` profile is not activated when another profile is activated explicitly with `mvn -P...`.
As we use `-Prelease` for release then the `uber-jar` profile is not activated and the `-runnner.jar` is not deployed to nexus. The uber jar profile has to be activated regardless if `-P` is or isn't used.

### DIFF
--- a/rts/lra/coordinator-quarkus/pom.xml
+++ b/rts/lra/coordinator-quarkus/pom.xml
@@ -155,7 +155,9 @@
       <profile>
         <id>quarkus-uber</id>
         <activation>
-          <activeByDefault>true</activeByDefault>
+          <property>
+            <name>!no-quarkus-uber</name>
+          </property>
         </activation>
         <properties>
           <quarkus.package.type>uber-jar</quarkus.package.type>


mmusgrov (approved):

rts/lra/coordinator-quarkus/pom.xml
```diff
@@ -155,7 +155,9 @@
       <profile>
         <id>quarkus-uber</id>
         <activation>
-          <activeByDefault>true</activeByDefault>
+          <property>
+            <name>!quarkus-uber</name>
```
xstefank, line 159: I don't understand this. The profile will run when you specify -Pquarkus-uber but when you do -Dquarkus-uber it will not run. What is the intent of this?

mmusgrov: I think it is the name of the property that's confusing. Perhaps it would be more intuitive to call it something like `no-quarkus-uber` instead.

ochaloup: That's right. Thanks @xstefank and @mmusgrov I changed it to `no-quarkus-uber`.

xstefank (approved):

PR #1804, opened 2021-03-22
Adding version badge to readme file

mayankkunwar:
This is just a minor enhancement in *readme* file to add a project version which would point to maven repository page.

These badges would be a quick info to the Narayana users/contributors or new comer to the repository. As whosever visit to any repository in git, tries to have a quick look into the readme of that project.

The eye catchy badges would give quick and important piece of information about our project without digging into files for the info.

Since the change is only in readme file, therefore triggering a CI build would not be necessary.
!MAIN !CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF !LRA !NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,8 @@
+[![Narayana](https://design.jboss.org/narayana/logo/final/narayana_logo_600px.png)](https://narayana.io/)
+
+[![Version](https://img.shields.io/maven-central/v/org.jboss.narayana/narayana-all?logo=apache-maven&style=for-the-badge)](https://search.maven.org/artifact/org.jboss.narayana/narayana-all)
+[![License](https://img.shields.io/badge/License-LGPL%20v2.1-green.svg?style=for-the-badge)](http://www.gnu.org/licenses/lgpl-2.1)
+
 Narayana
 ========
 


ochaloup (approved):

mmusgrov (changes requested):
There is no explanation or justification for why we need this change.
We already have a LICENSE file.
What is the relationship between the LICENSE and the README.md?
Now we are referring to the license from multiple places (at least three references), do we want this amount of duplication.

ochaloup:
I like the idea and I think the duplication is good for the license.

mmusgrov:
Yes but there is zero justification on the PR. It might look nice but why do we want it?

mayankkunwar:
These badges would be a quick info to the Narayana users/contributors or new comer to the repository. As whosever visit to any repository in git tries to have a quick look into the readme of that project. 

The eye catchy badges would be give quick and important piece of information about our project without digging into files for the info. 

If you say this badge is not required, I would remove it or if you want me to add more badges for displaying other information, will add them as part of this PR.

mmusgrov:
I don't like the badges so I will remove my review.

mayankkunwar:
Okay I have added more description to the PR and removed the LICENSE badge

jmfinelli:
TBH, I like badges and I think that having something similar to [quarkus](https://github.com/quarkusio/quarkus) might help newcomers (but also old committers) to immediately get important info about Narayana (e.g. current version and compatible JDKs)

mmusgrov:
> Okay I have added more description to the PR and removed the LICENSE badge

Don't remove them just because I said I didn't like the badges. If people want them then, since this is a community effort, no individual should be able to overrule a majority preference without good reason.

mayankkunwar:
Okay Mike, I thought this discussion should be unanimous. But if the case of majority then I will push these changes. Thanks

PR #1805, opened 2021-03-23
JBTM-3448 update dependencies for jdk-16 build.

jhalliday:
https://issues.redhat.com/browse/JBTM-3448

### DIFF
--- a/XTS/WS-C/pom.xml
+++ b/XTS/WS-C/pom.xml
@@ -38,7 +38,6 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-war-plugin</artifactId>
-        <version>3.0.0</version>
         <executions>
           <execution>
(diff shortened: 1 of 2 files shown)


mmusgrov:
Looks good but why a full CI test run of everything given that we don't test with jdk-16?

jhalliday:
> Looks good but why a full CI test run of everything given that we don't test with jdk-16?

Because I don't test on anything but 16, so I've no idea about possible regressions.

mmusgrov:
My point is that we only have JDK 8 and 11 on AMS so the full CI run will not prove anything.

mmusgrov (approved):
The main narayana job passed.
@jhalliday Do you want to wait for the rest to pass too ;-) or should I merge now?

jhalliday:
Merge it if you're happy, the other runs probably didn't survive the CI server rebuild anyhow.

PR #1806, opened 2021-03-25
JBTM-3447 Improve diagnostics for slow transactions

jhalliday:
https://issues.redhat.com/browse/JBTM-3447
Transaction reaper modifications to support periodically capturing thread stack trace information as the tx executes and logging them if it timesout.

MAIN CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF !LRA !NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/CoordinatorEnvironmentBean.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/CoordinatorEnvironmentBean.java
@@ -70,6 +70,8 @@ public class CoordinatorEnvironmentBean implements CoordinatorEnvironmentBeanMBe
     private volatile long txReaperCancelWaitPeriod = TransactionReaper.defaultCancelWaitPeriod;
     private volatile long txReaperCancelFailWaitPeriod = TransactionReaper.defaultCancelFailWaitPeriod;
     private volatile int txReaperZombieMax = TransactionReaper.defaultZombieMax;
+    private volatile long txReaperTraceGracePeriod = TransactionReaper.defaultUntracedPeriod;
+    private volatile long txReaperTraceInterval = TransactionReaper.defaultTracePeriod;
 
     private volatile int defaultTimeout = 60; // seconds
     private volatile boolean transactionStatusManagerEnable = true;
@@ -517,6 +519,47 @@ public class CoordinatorEnvironmentBean implements CoordinatorEnvironmentBeanMBe
     public void setTxReaperZombieMax(int txReaperZombieMax)
     {
         this.txReaperZombieMax = txReaperZombieMax;
+    }
+
+    /**
+     * Returns the number of milliseconds delay after a transaction is started,
+     * before the reaper will start taking periodic stack traces from it.
+     *
+     * Default: 180000 (3 minutes)
+     *
+     * @return the reaper tracing grace period, in milliseconds.
+     */
+    public long getTxReaperTraceGracePeriod() {
+        return txReaperTraceGracePeriod;
+    }
+
+    /**
+     * Sets the delay after a transaction is started, before it is eligible for periodic tracing.
+     *
+     * @param txReaperTraceGracePeriod in milliseconds.
+     */
+    public void setTxReaperTraceGracePeriod(long txReaperTraceGracePeriod) {
+        this.txReaperTraceGracePeriod = txReaperTraceGracePeriod;
+    }
+
+    /**
+     * Returns the number of milliseconds interval between transaction stack trace snapshots.
+     *
+     * Default: 30000 (30 seconds)
+     *
+     * @return the reaper tracing interval, in milliseconds.
+     */
+    public long getTxReaperTraceInterval() {
+        return txReaperTraceInterval;
+    }
+
+    /**
+     * Sets the interval between stack trace snapshots.
+     *
+     * @param txReaperTraceInterval in milliseconds.
+     */
+    public void setTxReaperTraceInterval(long txReaperTraceInterval) {
+        this.txReaperTraceInterval = txReaperTraceInterval;
     }
 
     /**
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
@@ -32,11 +32,16 @@
 package com.arjuna.ats.arjuna.coordinator;
 
 import java.io.IOException;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Date;
 import java.util.Enumeration;
 import java.util.Hashtable;
 import java.util.List;
+import java.util.Map;
+import java.util.HashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 
@@ -3377,6 +3382,37 @@ public class BasicAction extends StateManager
         return isCurrent;
     }
 
+    // called by the reaper system via TwoPhaseCoordinator.recordStackTraces
+    protected synchronized void recordStackTraces() {
+        Map<String,String> currentCapture = createStackTraces();
+        StackTraceCapture stackTraceCapture = new StackTraceCapture(System.currentTimeMillis(), currentCapture);
+        threadStackTraceHistoryList.add(stackTraceCapture);
+    }
+
+    // called by the reaper system via TwoPhaseCoordinator.outputCapturedStackTraces
+    protected synchronized void outputCapturedStackTraces() {
+        DateFormat dateFormat = new SimpleDateFormat("HH:mm:ss.SSS");
+        for(StackTraceCapture stackTraceCapture : threadStackTraceHistoryList) {
+            for(Map.Entry<String,String> entry : createStackTraces().entrySet()) {
+                tsLogger.i18NLogger.info_historic_stack_trace(objectUid, entry.getKey(), dateFormat.format(new Date(stackTraceCapture.getTime())), entry.getValue());
+            }
+        }
+    }
+
+    protected synchronized Map<String,String> createStackTraces() {
+        Map<String,String> results = new HashMap<>();
+        for (Thread entry : _childThreads.values()) {
+            StackTraceElement[] stackTrace = entry.getStackTrace();
+            StringBuilder sb = new StringBuilder();
+            for (StackTraceElement element : stackTrace) {
+                sb.append(element.toString());
+                sb.append("\n");
+            }
+            results.put(entry.getName(), sb.toString());
+        }
+        return results;
+    }
+
     private final boolean checkChildren (boolean isCommit)
     {
         boolean problem = false;
@@ -3403,14 +3439,8 @@ public class BasicAction extends StateManager
                     // This was added for JBTM-2476 - it could be used during commit
                     // but as the commit path needs to be optimum and getStackTrace is expensive
                     // I did not include it
-                    for (Thread entry : _childThreads.values()) {
-                        StackTraceElement[] stackTrace = entry.getStackTrace();
-                        StringBuilder sb = new StringBuilder();
-                        for (StackTraceElement element : stackTrace) {
-                            sb.append(element.toString());
-                            sb.append("\n");
-                        }
-                        tsLogger.i18NLogger.warn_multiple_threads(objectUid, entry.getName(), sb.toString());
+                    for(Map.Entry<String,String> entry : createStackTraces().entrySet()) {
+                        tsLogger.i18NLogger.warn_multiple_threads(objectUid, entry.getKey(), entry.getValue());
                     }
                 }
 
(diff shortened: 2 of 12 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
```diff
@@ -3377,6 +3382,37 @@ private final boolean checkIsCurrent ()
...
+        Map<String,String> currentCapture = createStackTraces();
+        StackTraceCapture stackTraceCapture = new StackTraceCapture(System.currentTimeMillis(), currentCapture);
+        threadStackTraceHistoryList.add(stackTraceCapture);
+    }
+
+    // called by the reaper system via TwoPhaseCoordinator.outputCapturedStackTraces
+    protected synchronized void outputCapturedStackTraces() {
+        DateFormat dateFormat = new SimpleDateFormat("HH:mm:ss.SSS");
+        for(StackTraceCapture stackTraceCapture : threadStackTraceHistoryList) {
+            for(Map.Entry<String,String> entry : createStackTraces().entrySet()) {
```
mayankkunwar, line 3396: Please correct me if I am wrong here, I think in the `for` loop we should use the recorded value i.e using `stackTraceCapture.getStackTraces()` instead of `createStackTraces()`?

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
```diff
@@ -3377,6 +3382,37 @@ private final boolean checkIsCurrent ()
...
+                tsLogger.i18NLogger.info_historic_stack_trace(objectUid, entry.getKey(), dateFormat.format(new Date(stackTraceCapture.getTime())), entry.getValue());
```
mayankkunwar, line 3397: Is it fine to use it as logger level INFO?

mayankkunwar (changes requested):
Now we have this improved diagnostics for transaction, won't it affect the performance. Suppose there is a scenario, the reaper is collecting the stack trace and just after it passes the status check to collect stack trace, at the same moment `commit` is performed. Or there could be some other scenarios that can get impacted by this. If so then, we should add a note somewhere, informing user the downside of this feature as well.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/CoordinatorEnvironmentBean.java
```diff
@@ -519,6 +521,47 @@ public void setTxReaperZombieMax(int txReaperZombieMax)
...
 
+    /**
+     * Returns the number of milliseconds delay after a transaction is started,
+     * before the reaper will start taking periodic stack traces from it.
+     *
+     * Default: 180000 (3 minutes)
+     *
+     * @return the reaper tracing grace period, in milliseconds.
+     */
+    public long getTxReaperTraceGracePeriod() {
```
mmusgrov, line 532: A remark, the grace period will avoid generating StackTraceCapture object instances (which could be thousands for high throughput) when the transactions are short running. I wonder if there are other reasons for the grace period.

mmusgrov (commented):
The feature looks useful. I had a few gripes (changing an I18N WARN message to INFO, a query about the tests for TwoPhaseCoordinator and child threads, and Mayank's comment about the method BasicAction.outputCapturedStackTraces), but otherwise the PR looks good.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
jhalliday: D'oh, yes, that's a bug for sure. Good catch, thanks!

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
jhalliday: Support (and me) don't want it at DEBUG, because many users are reluctant to enable debug logging in production due to (perceived or actual) slowness. One option I did toy with is having a separate logger just for this feature, which means it can be independently set to the desired level and also that it can write only the stacktraces to a different file from other logging, which would potentially allow stack captures to be logged immediately instead of held in memory until the end of the tx (which is done so they appear grouped in the log for easer reading), so then you would get info even if the tx did not timeout but just committed more slowly than intended.  However, for the way it's currently designed, INFO is intended.

mayankkunwar: I see, thanks for the clarification.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java
```diff
@@ -931,6 +975,8 @@ public static synchronized long transactionLifetime()
     public static final long defaultCheckPeriod = 120000; // in milliseconds
     public static final long defaultCancelWaitPeriod = 500; // in milliseconds
     public static final long defaultCancelFailWaitPeriod = 500; // in milliseconds
+    public static final long defaultUntracedPeriod = 180000; // in milliseconds
+    public static final long defaultTracePeriod = 30000; // in milliseconds
```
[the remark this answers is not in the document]
jhalliday: It is intended for use in prod, so it does need to be relatively cheap, but I think we're there. For ordinary use you'd set the gracePeriod greater than the default tx timeout, which means it would take no snapshots and have no overhead. When you want diagnostics, you temporarily lower it enough to catch the tx you are worried about. It's similar to support asking user to enable debug logging, but a much lighter weight, focussed alternative.

Two quibbles: Like most arj config properties, you have to restart to change it, which may not be desirable in prod environments. It could be read directly from the bean instead of cached in a reaper field, and thus be changeable through the mbean interface, but that may make it slower. Second, to turn it off entirely you have to set it not only higher than the default timeout, but higher than any timeout a tx may set explicitly, which takes a bit of thought unless you just go for e.g. max int.  However, that probably only matters for e.g. the wildfly integration code's default, not for the one we hardcode in core.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java
```diff
@@ -137,15 +138,15 @@ public final void check()
...
+                    final long nextCheck = reaperElement.getNextCheckAbsoluteMillis();
+                    if(nextCheck > now) {
+                        nextDynamicCheckTime.set(nextCheck);
                         return; // nothing to do yet.
                     }
                 }
             }
 
-            tsLogger.i18NLogger.warn_coordinator_TransactionReaper_18(reaperElement._control.get_uid(), reaperElement.statusName());
+            tsLogger.i18NLogger.info_coordinator_TransactionReaper_18(reaperElement._control.get_uid(), reaperElement.statusName());
```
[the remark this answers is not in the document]
jhalliday: Yes, INFO was a compromise between dropping it all the way down to DEBUG or leaving it at WARN. When stacktracing is enabled it's the stacktraces themselves in the log that will really be noisy, a single extra INFO won't make much odds. When stacktracing is not enabled, it's printed only in timeout cases. To be fair those will also log WARN from the reapable (i.e. BasicAction) itself, so we're also a bit noisy there, but there shouldn't bee many timeouts anyhow and it's still better than it historically has been.
I'm not fussed though - let me know if you want it at trace/debug, I'll have to update the PR for the iterator fix anyhow so it's easy to bundle in.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/reaper/ReaperStacktracingTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Exercises TransactionReaper periodic stack tracing functionality.
+ *
+ * @author jonathan.halliday@redhat.com, 2021-03
+ */
+public class ReaperStacktracingTest {
```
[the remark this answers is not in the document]
jhalliday: Yes I did, but it's a pain to automate because you need to either mock the logging library or read back the file output it generates.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/logging/arjunaI18NLogger.java
```diff
@@ -510,9 +511,9 @@
 	@LogMessage(level = WARN)
 	public void warn_coordinator_TransactionReaper_16(String arg0, Uid arg1, @Cause() Throwable arg2);
 
-	@Message(id = 12117, value = "TransactionReaper::check timeout for TX {0} in state  {1}", format = MESSAGE_FORMAT)
-	@LogMessage(level = WARN)
-	public void warn_coordinator_TransactionReaper_18(Uid arg0, String arg1);
+	@Message(id = 12117, value = "TransactionReaper::check processing TX {0} in state  {1}", format = MESSAGE_FORMAT)
```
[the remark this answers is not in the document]
jhalliday: Certainly the text change should not be a problem as scripts have to cope with the i18n variations anyhow. The level change is more interesting - does changing the level perhaps make it a new message requiring a new id number, and is that more disruptive to tooling than modifying the existing one? No obvious answer I think, just put your project lead hat on and make an executive decision...

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/CoordinatorEnvironmentBean.java
jhalliday: Nope, that is indeed pretty much what it's intended to do. Credit to Stephen Fikes for that idea.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java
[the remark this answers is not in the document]
mmusgrov: Okay we will capture this logic in the docs.
With regards to reading the property directly from the bean; let's see what the support team say (which I'd think would be Sephen).

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/reaper/ReaperStacktracingTest.java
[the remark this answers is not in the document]
mmusgrov: okay

mayankkunwar (approved):
Looks good now. Thanks

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/logging/arjunaI18NLogger.java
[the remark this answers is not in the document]
mmusgrov: Ideally the PR would log at warn if the reaper state is not TRACE (to maintain backwards compatibility). I don't think an if (repear mode != TRACE) then log using WARN) is adding much overhead.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java
```diff
@@ -137,15 +138,19 @@ public final void check()
...
+                    final long nextCheck = reaperElement.getNextCheckAbsoluteMillis();
+                    if(nextCheck > now) {
+                        nextDynamicCheckTime.set(nextCheck);
                         return; // nothing to do yet.
                     }
                 }
             }
 
-            tsLogger.i18NLogger.warn_coordinator_TransactionReaper_18(reaperElement._control.get_uid(), reaperElement.statusName());
+            if(reaperElement._status != ReaperElement.RUN ||
```
mmusgrov, line 149: I cannot follow the logic path that stops the warning from being printed when reaperElement._status is ReaperElement.TRACE. @jhalliday will you clarify this for me?

jhalliday: It's never TRACE at that point. It's RUN, until the reaper sets it to TRACE below, based on the same comparison of the timestamps. If TRACE was used in the way you think, there would need to be a mechanism for the worker thread to wake the reaper thread so that e.g. TRACE_DONE could be reset to RUN/TRACE. That doesn't exist as present - the wake is always time based, never completion based. We could do it be having an interval within which the worker is expected to complete the tracing, much like we have an interval in which it's expected to complete the cancel, but that seems like overkill.

mmusgrov: OK, thanks for the explanation. Let's hope I don't have to change this code very often when you're not around.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java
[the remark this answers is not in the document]
mmusgrov: I understand that you have already discussed the feature with Stephen so I marked this one as resolved.

mmusgrov (approved):

PR #1809, opened 2021-03-30
[SCRIPTS ONLY] do not fail on XTS when skipping tests is activated

ochaloup:
XTS
!MAIN !CORE !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !PERF NO_WIN !RTS !AS_TESTS !TOMCAT !JACOCO !LRA


tomjenkinson (approved):
Thanks!

PR #1817, opened 2021-04-16
Changes for JBTM-3461 using quarkus instead of thorntail

swetapatra:
Task:
https://issues.redhat.com/browse/JBTM-3461

RTS
!MAIN !CORE !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !AS_TESTS !TOMCAT !JACOCO !LRA

### DIFF
--- a/rts/sra/pom.xml
+++ b/rts/sra/pom.xml
@@ -11,52 +11,76 @@
     <artifactId>sra-parent</artifactId>
     <name>SRA PoC</name>
     <description>SRA PoC</description>
-    <packaging>war</packaging>
+    <packaging>jar</packaging>
 
     <properties>
         <failOnMissingWebXml>false</failOnMissingWebXml>
     </properties>
+     <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-bom</artifactId>
+                <version>${version.quarkus}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
 
     <build>
-        <finalName>lra-test</finalName>
-        <plugins>
+        <finalName>sra-test</finalName>
+            <plugins>
             <plugin>
-                <groupId>io.thorntail</groupId>
-                <artifactId>thorntail-maven-plugin</artifactId>
-                <version>${version.thorntail}</version>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-maven-plugin</artifactId>
+                <version>${version.quarkus}</version>
                 <executions>
                     <execution>
                         <goals>
-                            <goal>package</goal>
+                            <goal>build</goal>
                         </goals>
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                  <groupId>org.apache.maven.plugins</groupId>
+                  <artifactId>maven-jar-plugin</artifactId>
+                  <executions>
+                    <execution>
+                      <goals>
+                      <goal>test-jar</goal>
+                     </goals>
+                    </execution>
+                  </executions>
+                  <inherited>false</inherited>
+            </plugin>
         </plugins>
     </build>
-
     <dependencies>
-        <!-- WildFly Thorntail Fractions -->
+        <!-- Quarkus dependencies -->
         <dependency>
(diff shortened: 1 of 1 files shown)


rts/sra/pom.xml
```diff
@@ -11,52 +11,76 @@
...
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
 
     <build>
-        <finalName>lra-test</finalName>
-        <plugins>
+        <finalName>sra-test</finalName>
+         <plugins>
```
mmusgrov, line 33: When we use indentation try to be consistent. Normally we indent by 4 spaces but this file is using a random mix of 1, 2 and 4 spaces.

rts/sra/pom.xml
```diff
@@ -11,52 +11,76 @@
...
+                    </execution>
+                  </executions>
+                  <inherited>false</inherited>
+            </plugin>
         </plugins>
     </build>
-
     <dependencies>
-        <!-- WildFly Thorntail Fractions -->
+        <!-- WildFly Quarkus Fractions -->
```
mmusgrov, line 61: There is no such thing as a "WildFly Quarkus Fraction". Just use "quarkus dependencies" which is preferable since we my want to include quarkus dependencies that are not extensions.

mmusgrov (approved):
Looks good apart from a minor observation.

Note that when we ignore failed tests (AS_TESTS, XTS, QA_JTS_JDKORB, QA_JTS_JDKORB and DB_TESTS) we should include a note, on the PR, about why it is safe to ignore them.

ochaloup (commented):
I would put down here my observations that I've already discussed with @swetapatra .
@mmusgrov  I'm not sure what is the goal so I'm in doubt.

The trouble is that I'm not sure what should be the point of the https://issues.redhat.com/browse/JBTM-3461
I don't say that I don't understand the reason for the existence of the issue and I think I understand the motivation but the trouble is that I'm not sure how is adressed in this PR.

Until now the SRA was a project which generates a war file to be deployed to app container (i.e. to WildFly) and a jar runner for thorntail. This whole project is a demo (as far as I can say) showing the capabilities of the annotations.
This PR creates only a jar file and a jar runner. The jar file won't be possible to  be deployed as a web application to WildFly. Thus it changes a bit what was here before.
Now, I'm not sure if SRA should not be somehow split to a client library and a demo. If so then I think some change similar to LRA would be needed. I mean having separate maven modules for different usage - having a separate jar library for client, having a demo war to be deployed on WFLY, having a demo runnable jar for quarkus. If not then at least the `war` deployment should be probably remained.

mmusgrov:
This PR is for JBTM-3461 which is a subtask of [JBTM-3457](https://issues.redhat.com/browse/JBTM-3457).
The goal of the overall task is to ensure that REST-AT is in a suitable state for resolving a [quarkus issue](https://github.com/quarkusio/quarkus/issues/15364) requesting distributed JTA style transactions. The SRA annotations are meant to be integrated with the REST-AT module (JBTM-3461).

SRA was originally a PoC meant for a MicroProfile specification but that is no longer the primary goal. The new goal is to ensure that it is in state suitable for use in Quarkus. Once it has had some usage there we can look at getting it back into MicroProfile.

Regarding jar versus war, I'm not sure about that change but once all of sub-tasks are done we will be using whatever packaging is suitable for the REST-AT module while, at the same time, ensuring that we can build a quarkus extension for it and still support the WildFly RTS extension (so I'd anticipate that we will produce both packaging types like we do with the LRA module).

Perhaps the confusion is that the issues were missing the task <-> sub-task relationship - I've fixed that now.

ochaloup (approved):
I do understand the points said here and as this PR is part of the bigger task it's good to be merged. The only think which I'm not sure is the trouble that the result artifact is `jar` and not `war`. But until the demo is split to a separate project in quickstart it probably does not make any issue.
Mike approved, I'm going to merge.

ochaloup:
@mmusgrov just a note, it was me who changed the testing only for `RTS` as other axes has not meaning and was only taking time in CI. Clarification for why I did so is that the `sra` module is not integrated in any other testcase and there are no tests in `sra` module. The `RTS` is fine to see that compilation passes.

PR #1818, opened 2021-04-20
[SCRIPTS ONLY] narayana.sh to split the AS_TESTS to multiple functions

ochaloup:
spliting makes possible to separately download, build and test the application server. Such a parametrization is useful mostly for a customized CI jobs which runs some specific aspects of the testing.

Motivation: investigating on errors of 5.11.1.Final for WildFly builds - https://issues.redhat.com/browse/WFLY-14701

MAIN RTS LRA AS_TESTS
!QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !TOMCAT !JACOCO

### DIFF
--- a/scripts/hudson/narayana.sh
+++ b/scripts/hudson/narayana.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -x
 
 function fatal {
   if [[ -z $PROFILE ]]; then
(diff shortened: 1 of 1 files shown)


ochaloup:
I'm proposing a change for CI script which makes me possible to configure CI job to not compile whole WFLY but only clone tests and run that against published WFLY server.
In some way these changes could be used within the https://issues.redhat.com/projects/JBTM/issues/JBTM-3465 but I did not add anything up to what was in script. I just split the one function into two and made small adjustment with env variables. For example there was not added a functionality to only compile Narayana and push the result jar files into the downloaded WFLY server. On the other hand running the SNAPSHOT WFLY tests has an issue that tests  requires the SNAPSHOT version of WFLY integration libraries and these are not available in nexus.

Anyway, this change helps me to run the CI AS tests in way that I was able to investigate at https://issues.redhat.com/browse/JBTM-3468 and I think it could be useful for usage in future too.

note: AS TESTS failures are only test failures not failures of the CI script

tomjenkinson (approved):
Thanks!

ochaloup:
Thanks @tomjenkinson . I talked with @mmusgrov he is mostly fine as well.
I'm going to merge this PR. If there are needed some fixes or further enhancements, we (or happily me) address it in a follow-up PR.

PR #1819, opened 2021-04-20
Additional changes for JBTM-3461 SRA Demo

swetapatra:
https://issues.redhat.com/browse/JBTM-3461

RTS
!MAIN !CORE !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN  !AS_TESTS !TOMCAT !JACOCO !LRA


mmusgrov (changes requested):
It is not clear why these changes are required.
It will be quicker if we can discuss it via gchat when you return from your bank holiday.

mmusgrov (approved):
This is good first pass at getting the PoC for the SRA annotations and demo working with Quarkus.
Next steps will be to make it more robust, for example some things are partially implemented and there are references to LRA (since LRA was used as the starting point for the design and implementation). But the code and demo serves its main purpose as proof that the module can be made to work in a quarkus environment. Thanks

PR #1821, opened 2021-04-20
[JBTM-3468] omit throwing NPE on generating stack trace in reaper worker thread

ochaloup:
https://issues.redhat.com/browse/JBTM-3468

AS_TESTS
!MAIN !CORE !QA_JTA !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !QA_JTS_JACORB !BLACKTIE !XTS !PERF NO_WIN !RTS  !TOMCAT !JACOCO !LRA

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
@@ -3401,6 +3401,7 @@ public class BasicAction extends StateManager
 
     protected synchronized Map<String,String> createStackTraces() {
         Map<String,String> results = new HashMap<>();
+        if (_childThreads == null) return results;
         for (Thread entry : _childThreads.values()) {
             StackTraceElement[] stackTrace = entry.getStackTrace();
             StringBuilder sb = new StringBuilder();


mmusgrov (approved):
Good spot.

ochaloup:
@jhalliday from what I can say the issue was about a NPE which can be thrown and then worker thread gets stopped. I'm providing a simple fix as part of this PR. Please, let me know if it's sufficient from your perspective. From perspective of AS tests (as far as I was able to check, the PR CI run was not yet started because of overwhelmed CI) it was enough to get the tests passed.

jhalliday:
Ahh, yes. The reaper system cares about the worker getting stuck (usually on db driver I/O) and would spawn a replacement in such case, but it doesn't so much consider the worker failing or quitting for other reasons. Not so sure why it would OOM in those cases, but if its fixed I don't suppose that matters.

ochaloup:
Thanks @mmusgrov and @jhalliday. I will wait for the CI tests results and then merge this.

ochaloup:
The AS testsuite is pretty stable - in comparision to standard run when we can see 5-10 intermittent failures. For this PR the affected tests passed smoothly. But the the one failure that is observed is `testEjbTimeoutOnOtherNode(org.jboss.as.test.multinode.ejb.timer.database.DatabaseTimerServiceMultiNodeTestCase` which sounds suspicious. I'm investigating if it has some connection to the changes or it's only an environment trouble.

ochaloup:
I still need to investigate on `DatabaseTimerServiceMultiNodeTestCase` failures (stil not sure if it's only env issue of our CI as I'm not able to hit it in any way on my local machine) but this PR fixes the basic AS tests failures which are needed to be addressed.

PR #1825, opened 2021-04-26
JBTM-3474 LRA filters log an invalid warning about missing CDI

xstefank:
https://issues.redhat.com/browse/JBTM-3474

!MAIN !CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF LRA NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/rts/lra/service-base/src/main/java/io/narayana/lra/logging/LRALogger.java
+++ b/rts/lra/service-base/src/main/java/io/narayana/lra/logging/LRALogger.java
@@ -29,5 +29,5 @@ public final class LRALogger {
     }
 
     public static final Logger logger = Logger.getLogger("io.narayana.lra");
-    public static final lraI18NLogger i18NLogger = Logger.getMessageLogger(lraI18NLogger.class, "io.narayana.lra");
+    public static final LraI18nLogger i18nLogger = Logger.getMessageLogger(LraI18nLogger.class, "io.narayana.lra");
 }
--- a/rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
+++ b/rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java
@@ -198,7 +198,7 @@ public class NarayanaLRAClient implements Closeable {
         try {
             this.coordinatorUrl = LRAConstants.getLRACoordinatorUrl(lraId);
         } catch (IllegalStateException e) {
-            LRALogger.i18NLogger.error_invalidLraIdFormatToConvertToCoordinatorUrl(lraId.toASCIIString(), e);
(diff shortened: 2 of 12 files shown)


rts/lra/service-base/src/main/java/io/narayana/lra/logging/LRALogger.java
```diff
@@ -29,5 +29,5 @@ private LRALogger() {
     }
 
     public static final Logger logger = Logger.getLogger("io.narayana.lra");
-    public static final lraI18NLogger i18NLogger = Logger.getMessageLogger(lraI18NLogger.class, "io.narayana.lra");
+    public static final LraI18nLogger i18nLogger = Logger.getMessageLogger(LraI18nLogger.class, "io.narayana.lra");
```
ochaloup, line 32: The naming should be (probably) left as it was, i.e. `lraI18NLogger` and `i18NLogger` - it's because of consistency. Narayana project just uses this naming convention for `I18n` classses - see https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/logging/arjunaI18NLogger.java, https://github.com/jbosstm/narayana/blob/master/ArjunaJTA/jta/classes/com/arjuna/ats/jta/logging/jtaI18NLogger.java or whatever else.

ochaloup (changes requested):
I think the naming should not be change (but feel free to start the discussion that I miss some thing).
Otherwise +1 for the change.

rts/lra/service-base/src/main/java/io/narayana/lra/logging/LRALogger.java
xstefank: ok, I didn't know that this is customary. But still why is it a Java class starting with lowercase letter and if we use Java convention of PascalCase or camelCase then only new words in the name should be capitalized, so i18*N*Logger is wrong because N is not a new word in i18n. But I understand that it is done this way, I will change it back, but this is not a correct Java form.

ochaloup: Agree, it's not a correct form from Java checkstyle guidelines. Here I was thinking that as the LRA is part of the Narayana project it would be good to follow the style that Narayana uses for all other i18n classes. I feel the project consistency will be be better in this case.

ochaloup (approved):
After discussion in the team we agreed that we should follow the Java coding checkstyle in new Narayana modules (not changing the old ones, being better for the new ones).

PR #1829, opened 2021-04-30
[JBTM-3450] Oracle JDBC drivers updated to version 18.3.0.0 (18c)

jmfinelli:
This PR fixes JIRA: https://issues.redhat.com/browse/JBTM-3450

Profiles to test:
CORE QA_JTA QA_JTS_OPENJDKORB oracle

Profiles to skip:
!MAIN !TOMCAT !QA_JTS_JACORB !QA_JTS_JDKORB !AS_TESTS !RTS !JACOCO !XTS  !BLACKTIE !PERF !LRA NO_WIN !DB_TESTS !mysql !db2 !postgres

### DIFF
--- a/qa/build.xml
+++ b/qa/build.xml
@@ -37,7 +37,7 @@
   <property name="org.jboss.jbossts.qa.server_manager_location" location="ext/jboss-server-manager-0.1.1.GA.jar"/>
   <property name="org.jboss.jbossts.qa.dist.buildroot" location="build"/>
 
-  <target name="get.drivers" depends="clean-dbdrivers">
+  <target name="get.drivers">
     <!--
         https://docspace.corp.redhat.com/clearspace/docs/DOC-16080
         http://www.jboss.com/products/platforms/application/supportedconfigurations/
@@ -45,8 +45,6 @@
         https://docspace.corp.redhat.com/clearspace/community/bu/middleware/jboss-qe/lab
         server connection params in config/jdbc_profiles/
         -->
-    <mkdir dir="${driver.home}/oracle_10_2_0_4"/>
-    <get src="${driver.url}/ojdbc14.jar" dest="${driver.home}/oracle_10_2_0_4/ojdbc14.jar"/>
     <mkdir dir="${driver.home}/mssql2005_sqljdbc_2.0/enu"/>
     <get src="${driver.url}/sqljdbc_2.0/enu/sqljdbc.jar" dest="${driver.home}/mssql2005_sqljdbc_2.0/enu/sqljdbc.jar"/>
     <get src="${driver.url}/sqljdbc_2.0/enu/sqljdbc4.jar" dest="${driver.home}/mssql2005_sqljdbc_2.0/enu/sqljdbc4.jar"/>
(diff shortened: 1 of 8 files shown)


qa/build.xml
```diff
@@ -29,11 +29,12 @@
...
         This is a redhat vpn server. for non-vpn use, mirror it locally or copy the drivers by hand.
-        wget \-\-no-parent -m http://www.qa.jboss.com/jdbc-drivers/
-        wget \-\-no-parent -m http://www.qa.jboss.com/jdbc-drivers/jConnect-6_0/classes
+        wget \-\-no-parent -m http://www-qa-jboss.hosts.mwqe.eng.bos.redhat.com/jdbc-drivers/
+        wget \-\-no-parent -m http://www-qa-jboss.hosts.mwqe.eng.bos.redhat.com/jdbc-drivers/jConnect-6_0/classes
         ant -Ddriver.url=file:///path/to/www.qa.jboss.com/jdbc-drivers get.drivers
     -->
-  <property name="driver.url" value="http://www.qa.jboss.com/jdbc-drivers/"/>
+  <property name="driver.url" value="https://www-qa-jboss.hosts.mwqe.eng.bos.redhat.com/jdbc-drivers/"/>
+  <property name="driver.oracle.url" value="https://svn.devel.redhat.com/repos/jboss-qa/hibernate-properties-and-drivers/jdbc/EAP/7.4.0/"/>
```
mmusgrov, line 37: I will review this PR after the CI runs finish but this line depends on a private repo managed by Red Hat whereas narayana is a public community project. This means that the QA tests would only be able to be ran behind a firewall. Is the driver available from any public repository?

tomjenkinson: That line should be removed. @jmfinelli if it is necessary to do this then you could consider if getting the URL to use from an environment variable in the job config is possible?

jmfinelli: My bad, I added the two servers without double-checking they are private servers. Problem is that http://www.qa.jboss.com/ looks to be down and the only alternative I found was www-qa-jboss.hosts.mwqe.eng.bos.redhat.com (which I realise now it is a private server)

tomjenkinson: @jmfinelli if it is anywhere here: https://search.maven.org/search?q=ojdbc if so you could trying putting the dependency under https://github.com/jbosstm/narayana/blob/master/qa/pom.xml#L200

qa/build.xml
```diff
@@ -29,11 +29,12 @@
   <!-- you should not need to configure anything below here -->
   <!--
         This is a redhat vpn server. for non-vpn use, mirror it locally or copy the drivers by hand.
```
jmfinelli, line 31: Just noticed that the old server www.qa.jboss.com was also behind the VPN. In fact, for non-VPN use, it is suggested to copy all needed drivers manually.

qa/build.xml
jmfinelli: I'll try to implement your suggestion @tomjenkinson. As this modules runs an ant script to execute tests, I thought it was worth to give a shot to the (already-in-place) logic to fetch drivers. As you and @mmusgrov suggested, the server hosting JDBC drivers should not be behind a VPN.

I'll try to copy the new oracle driver from Maven to the folder where the ant script looks into and, if successful, we might think to apply the same logic to all other drivers.

ochaloup (approved):

PR #1839, opened 2021-05-24
Changes for JBTM-3464

swetapatra:

  1. If you change a file (in a non trivial way) you are allowed (MAY) to reformat the code to conform to our checkstyle ruleset (`org.jboss.narayana:checkstyle`).


mmusgrov:
Note that the SRA annotation code is now duplicated. So you will need to delete the code from the original sra directory and update the demo to use the artefact where the new code resides.

mmusgrov:
I aborted the CORE profile test run since we'll need an update to fix the RTS profile tests failure. Note also that the windows run showed the new test (SRATest) passing.

PR #1840, opened 2021-05-25
[CHECKSTYLE] adding checkstyle rules for Narayana CDI module

ochaloup:
I would like to add the checkstyle rules for the JTA/cdi module. It's a module which is in development and it would be good to have that considered as being read and written quite often.

MAIN


jmfinelli (approved):

mmusgrov:
All of our modules are maintained. Plus, isn't our policy to only apply check style fixes to a file if we are making functional changes or bug fixes to the file.

ochaloup:
@mmusgrov Applying the checkstyle is beneficial for any module for any future work. The CDI module is one of the actively maintained modules where not only Narayana team but other developers participate in.

I'm proposing this change as I think it's a good thing.
When there is the policy that we can't add the checkstyle then I would put this PR on hold and bring this topic on the team meeting. I would like to discuss if a policy which does not permit applying the checkstyle rules to modules where different contributors work at is beneficial for the project. Is that considerable?

jmfinelli:
> [..] isn't our policy to only apply check style fixes to a file if we are making functional changes or bug fixes to the file

If IIRC, our check style is activated with the property `<checkstyle.skip>false</checkstyle.skip>`. This applies the check style to the whole module. As a consequence, I was wondering how we can apply check style rules to only one file and not to the whole module (maybe manually?). Plus, I do not know how the end result would look like if only some files are modified in a module and all others are left as they are.

Maybe this could be a good place to discuss about the policy you were talking about @mmusgrov?

ochaloup:
@jmfinelli I have no experience with running the checkstyle just against one file but it could be that running the maven module has some configuration for such usage.
Still I would be happy to have a chance to enable checkstyle for the whole module.

PR #1841, opened 2021-05-25
[JBTM-3442] transactional cdi InterceptionFactory

ochaloup:
https://issues.redhat.com/browse/JBTM-3442

MAIN AS_TESTS

@manovotn could you, please, take a look at this PR?

### DIFF
--- a/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
+++ b/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
@@ -59,6 +59,9 @@ import static java.security.AccessController.doPrivileged;
 public abstract class TransactionalInterceptorBase implements Serializable {
     private static final long serialVersionUID = 1L;
 
+    // to distinguish Weld implementation where WeldInvocationContext defines context data key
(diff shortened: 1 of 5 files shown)


manovotn (commented):
Yeah, looks good to me.

@ochaloup the only issue I see is that non-weld impl will still blow up with hard to read error (I think it was NPE?). It might be worth guarding for that `null` and rethrowing something more understandable. Can even be something like "not supported for interception factory with non-weld CDI impl". But it is OFC up to you; I have no idea in which environments is Narayana usually used.

mmusgrov:
@manovotn narayana is used by many products (quarkus, fuse osgi, wildfly, etc) so throwing the exception early is preferable.

ochaloup:
@manovotn thanks, that's a really good point. I added a check on `null` in the problematic place. Please, take a look if you consider it fine or to be refined.

manovotn (approved):
LGTM now

mmusgrov (approved):
I think Matej's approval is sufficient.

ochaloup:
the CI tests seem fine, the AS_TESTS fails with the intermittent failure at `DatabaseTimerServiceMultiNodeTestCase.testEjbTimeoutOnOtherNode`
it's the CI issue. it's something that I'm trying to hit with other testing job and find if some timeout adjust may help but I'm not able so far to say where the trouble is

ochaloup:
@manovotn thanks!

PR #1845, opened 2021-05-27
[JBTM-3480] Change the logic to download DB drivers in the QA module

jmfinelli:
This PR addresses the JIRA [JBTM-3480](https://issues.redhat.com/browse/JBTM-3480)

Profiles passed:
!MAIN !CORE QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !DB_TESTS !postgres !mysql !oracle !db2

Profiles to skip:
!TOMCAT !AS_TESTS !RTS !JACOCO !XTS !BLACKTIE !PERF !LRA NO_WIN


ArjunaCore/arjuna/pom.xml
```diff
@@ -222,269 +222,273 @@
...
+      </build>
+    </profile>
+    <!--
+    <profile>
+      <id>sybase-jdbc-store</id>
+      <dependencies>
+        <dependency>
+          <groupId>com.sybase.jConnect</groupId>
+          <artifactId>jconn3</artifactId>
+          <version>${version.com.sybase.jConnect}</version>
```
ochaloup, line 249: is needed to specify the version here when the scope is `system`?

ArjunaCore/arjuna/pom.xml
```diff
@@ -222,269 +222,273 @@
...
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+              <systemProperties>
+                <property>
+                  <name>com.arjuna.ats.arjuna.common.propertiesFile</name>
+                  <value>oraclejbossts-properties.xml</value>
```
ochaloup, line 291: would not be more conscise to define the property like `jbossts.propertyxml.file.name` and define it in `properties` section of the `profile` and setting up the than setting up the properties file just once in `build` parametrized with that properties?

ochaloup (approved):
To me the changes seem fine. I added just two questions in comments but it's out of my curiosity.
One thing which I think should be changed before potential merging is to squash the commits under one (or at least significantly less than ~8 as there are now).

ArjunaCore/arjuna/pom.xml
jmfinelli: Sorry, this area was highlighted as "new" because of the modifications introduced by the checkstyle.

Having said that, I agree with you that the `pom.xml` can be improved with a property. @mmusgrov WDYT?

jmfinelli:
> One thing which I think should be changed before potential merging is to squash the commits under one (or at least significantly less than ~8 as there are now).

Forgot to mention in the previous comment that I will squash all commits together. I have not done it yet as I am still working on this PR and an extensive history might help in case I need to rollback

ArjunaCore/arjuna/pom.xml
ochaloup: +1

mmusgrov:
@jmfinelli What was the issue with the DB_TESTS profile tests failures?

ArjunaJTA/jta/pom.xml
```diff
@@ -25,9 +26,9 @@
   <packaging>jar</packaging>
 
   <properties>
-      <server.jvm.args>
-          ${jvm.args.other} ${jvm.args.memory} ${jvm.args.debug} ${jvm.args.modular}
-      </server.jvm.args>
+    <server.jvm.args>
```
mmusgrov, line 29: Part of the reason for asking engineers to minimise reformatting code is that it makes it difficult to track the reasons for fixes. If you really need to reformat the code amongst other functional changes then can you do the reformatting in a separate commit.

jmfinelli: > I'm not sure we have formally documented that requirement so I will look to add a doc/text explaining the project requirements.

Yes, please. It is not very clear whether we can reformat old code when we carry out functional modifications or if we should apply style rules on a separated commit.

jmfinelli:
> @jmfinelli What was the issue with the DB_TESTS profile tests failures?

Still working on this

jmfinelli:
>> @jmfinelli What was the issue with the DB_TESTS profile tests failures?

> Still working on this

Failures of `btny-pulls-narayana-jdbcobjectstore` were due to performances of DBs on our CI. Running DB_TESTS again cleared the failures against Oracle and DB2.

This PR is ready to be reviewed :-)

qa/build.xml
```diff
@@ -21,55 +21,49 @@
...
   <property name="org.jboss.jbossts.qa.ts.home" location="dist"/>
   <property name="narayana.version" value="5.11.3.Final-SNAPSHOT"/>
-  <!-- you should not need to configure anything below here -->
+
   <!--
-        This is a redhat vpn server. for non-vpn use, mirror it locally or copy the drivers by hand.
-        wget \-\-no-parent -m http://www.qa.jboss.com/jdbc-drivers/
-        wget \-\-no-parent -m http://www.qa.jboss.com/jdbc-drivers/jConnect-6_0/classes
-        ant -Ddriver.url=file:///path/to/www.qa.jboss.com/jdbc-drivers get.drivers
+        The previous logic used to download drivers has been replaced for
```
ochaloup, line 30: Just a personal opinion. To me it's not necessary to have this "long" explanation that there was some old functionality. All information about the any prior handling is easily available from version control system.
It could be fine to just say the drivers are downloaded from maven repos.
From what I'm a little bit missing is the information to what the drivers are used to. What does mean "QA tests against ...". Here I think it could be beneficial to have information that databases are for running the object stores.

qa/build.xml
```diff
@@ -84,45 +78,51 @@
...
   <target name="help">
-    <echo>Supported targets:
-  jta - build tests for the standalone jta transaction engine
-  jts - build tests for the standalone jts transaction engine
-  narayana-jta - build tests for the integrated jta transaction engine
-  narayana-jts - build tests for the integrated jts transaction engine
-  dist - build tests for the last build of the transaction engine
-</echo>
+    <echo>
+      Supported targets: jta - build tests for the standalone jta transaction engine jts - build tests for the
```
ochaloup, line 122: This formatting seems a bit weird to me. Should not be targets printed on a separate line?

ochaloup (approved):
I consider the changes nice. Just one or two small remarks to consider for adjustment from me.

jmfinelli (commented):
Thanks Ondra

mmusgrov:
> > > @jmfinelli What was the issue with the DB_TESTS profile tests failures?
> 
> > Still working on this
> 
> Failures of `btny-pulls-narayana-jdbcobjectstore` were due to performances of DBs on our CI. Running DB_TESTS again cleared the failures against Oracle and DB2.
> 
> This PR is ready to be reviewed :-)

What does "performances of DBs on our CI" mean. For example if the transactions time out then that would call for a tweak to the timeouts used for the tests.

ArjunaCore/arjuna/pom.xml
mmusgrov: okay

> Sorry, this area was highlighted as "new" because of the modifications introduced by the checkstyle.

This is a prime example of why reformatting code causes difficulties problem

jmfinelli:
> What does "performances of DBs on our CI" mean. For example if the transactions time out then that would call for a tweak to the timeouts used for the tests.

DB2:
It is likely that the root cause of the previous failure was a deadlock between parallel running instances of Narayana (`DB2 SQL Error: SQLCODE=-911, SQLSTATE=40001`). At the time `btny-pulls-narayana-jdbcobjectstore` was running the build number 8, the CI was running the maximum number of builds (most of which against DBs) and, among them, there were 2 builds (IIRC) against DB2 (one of which was another `narayana-jdbcobjectstore`). As DB2 is running as a container, I thought that a deadlock could be an understandable result if two identical builds run at the same time while the CI is under stress (hence "performances of DBs on our CI"). Rebuilding the same job against DB2 (but this time in isolation) cleared the failure.

Oracle:
On the other side, `btny-pulls-narayana-jdbcobjectstore` failed against Oracle because of the impossibility to write in a table (`ORA-00942 table or view does not exist`). To me, it sounds a lot like the above deadlock against DB2. Again, rebuilding the same job against Oracle (but this time in isolation) cleared the failure.

It is hard to reproduce this kind of failures locally and remotely on our CI (for their random nature and their "apparent" dependency on the hardware of the CI). If you think that the above errors have been caused by something different than the performance of our CI, I can try my best to reproduce the failures and investigate on them. In this case, I will use my local laptop to easily access debug information. Please, let me know what is your take on this :-)

ArjunaCore/arjuna/pom.xml
jmfinelli: > This is a prime example of why reformatting code causes difficulties.

True. This is a good example. Fortunately, reformatting code only happens once :-) (and, according with what has been recently discussed for old files, it can take place when there is a functional modification)

qa/build.xml
ochaloup: The previous formatting was in style with end of lines after each target listed, i.e.
```
Supported targets:
  jta - build tests for the standalone jta transaction engine 
  jts - build tests for the standalone jts transaction engine
  ...
```

while your proposed change is about having it somehow mixed
```
Supported targets: jta - build tests for the standalone jta transaction engine jts - build tests for the
  standalone jts transaction engin
  ...
```
I mean every target `jta`,`jts`,`narayana-jta`,`narayana-jts`,`dist` should be placed on a separate line.

jmfinelli: Oh, I see now! Thank you very much for the explanation @ochaloup  +1

PR #1852, opened 2021-06-08
JBTM-3494 Lazily construct a NarayanaLRAClient helper

mmusgrov:
https://issues.redhat.com/browse/JBTM-3494

Lazily constructs NarayanaLRAClient helper. Doing it this way avoids it's construction if the filter is never used and it also minimises our dependency on CDI.

!MAIN !CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF LRA NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
+++ b/rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
@@ -94,11 +94,7 @@ public class ServerLRAFilter implements ContainerRequestFilter, ContainerRespons
     @Inject
     private LRAParticipantRegistry lraParticipantRegistry;
 
-    private final NarayanaLRAClient lraClient;
-
-    public ServerLRAFilter() {
-        lraClient = new NarayanaLRAClient();
-    }
+    private NarayanaLRAClient lraClient;
 
     private boolean isTxInvalid(ContainerRequestContext containerRequestContext, LRA.Type type, URI lraId,
                                 boolean shouldNotBeNull, ArrayList<Progress> progress) {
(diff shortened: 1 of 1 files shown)


xstefank (approved):

mmusgrov:
The jdk 8 github action failed with Error: Unable to access jarfile /home/runner/work/narayana/narayana/rts/lra/test/../coordinator-quarkus/target/lra-coordinator-runner.jar

I don't know what causes this but we I saw it before. I don't think it is related to the PR because it passed on our own CI and on the JDK 11 github action. So unless there are objections I propose to merge this PR.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
```diff
@@ -554,6 +550,15 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
         }
     }
 
+    private NarayanaLRAClient getLRAClient() {
+        if (lraClient == null) {
+            // no need to lock
+            lraClient = new NarayanaLRAClient();
```
ochaloup, line 556: my 2pc: while I think this is not fully safe as it could still potentially provide a partially initialized object it is probably not be a trouble in later usages

ochaloup (approved):

ochaloup:
`I propose to merge this PR.` I tried to rerun the gh actions but it fails with the usual dependency problem. By me the PR is fine for meging as well.

rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
mmusgrov: Java guarantees that assigning a constructed object to a variable will always result in the variable referencing a fully initialised Java object.

PR #1855, opened 2021-06-15
[JBTM-3479] Added resteasy-client and resteasy-jackson2-provider dependencies to the lra-client module

jmfinelli:
This PR addressed [JBTM-3479](https://issues.redhat.com/browse/JBTM-3479)

LRA
!MAIN !CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !BLACKTIE !PERF NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/rts/lra/client/pom.xml
+++ b/rts/lra/client/pom.xml
@@ -30,6 +30,16 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
+      <groupId>org.jboss.resteasy</groupId>
+      <artifactId>resteasy-client</artifactId>
+      <version>${version.org.jboss.resteasy}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jboss.resteasy</groupId>
+      <artifactId>resteasy-jackson2-provider</artifactId>
+      <version>${version.org.jboss.resteasy}</version>
+    </dependency>
+    <dependency>
       <groupId>javax.enterprise</groupId>
       <artifactId>cdi-api</artifactId>
       <version>${version.cdi-api}</version>
--- a/rts/lra/client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
+++ b/rts/lra/client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
@@ -1,1 +1,2 @@
+org.jboss.resteasy.client.jaxrs.ResteasyClientBuilder
 org.jboss.resteasy.microprofile.client.impl.MpClientBuilderImpl


rts/lra/client/pom.xml
```diff
@@ -29,6 +29,16 @@
       <artifactId>lra-service-base</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
```
mmusgrov, line 32: Why do we need these and where is the evidence that they are required. Is there an example of something failing, and if not can we have a reproducer.

rts/lra/client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
```diff
@@ -1 +1,2 @@
-org.jboss.resteasy.microprofile.client.impl.MpClientBuilderImpl
+org.jboss.resteasy.client.jaxrs.ResteasyClientBuilder
+org.jboss.resteasy.microprofile.client.impl.MpClientBuilderImpl
```
mmusgrov, line 2: Where do we use the dependency on MpClientBuilderImpl?

rts/lra/client/pom.xml
jmfinelli: Basically, when the LRA client is employed, an implementation of the JAX-RS Client API must be specified among the dependencies. As we already specify that we want to use `org.jboss.resteasy.microprofile.client.impl.MpClientBuilderImpl` in the ServiceLoader, I think it might be beneficial to also add the resteasy-client dependency.

For what concerns `resteasy-jackson2-provider`, when the client is used to fetch all LRAs (getAllLRAs()), an exception RESTEASY003145 is thrown.

Reproducer:
```
/**
 * Copyright Red Hat
 *
 * SPDX-License-Identifier: LGPL-2.1-only
 */

import io.narayana.lra.LRAData;
import io.narayana.lra.client.NarayanaLRAClient;
import org.junit.Test;

import java.util.List;

import static org.junit.Assert.assertNotNull;
import static org.junit.Assert.assertTrue;

public class CheckDeploymentTestCase {

    // I suppose you have a running lra-coordinator at this address
    private final NarayanaLRAClient lraClient =
            new NarayanaLRAClient("http://127.0.0.1:8080/lra-coordinator/lra-coordinator");

    @Test
    public void testWildflyIsRunning() {
        final List<LRAData> allLRAs = lraClient.getAllLRAs();

        assertNotNull(allLRAs);
        assertTrue(allLRAs.isEmpty());
    }
}
```

rts/lra/client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
```diff
@@ -1 +1,2 @@
-org.jboss.resteasy.microprofile.client.impl.MpClientBuilderImpl
+org.jboss.resteasy.client.jaxrs.ResteasyClientBuilder
```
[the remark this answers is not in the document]
jmfinelli: I get `java.lang.ClassNotFoundException: org.glassfish.jersey.client.JerseyClientBuilder` when I use the client straight from the code base. IIRC `org.glassfish.jersey.client.JerseyClientBuilder` is the last attempt to find a ClientBuilder implementation in the JAX-RS specification

rts/lra/client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
jmfinelli: It is the implementation of the ClientBuilder, so (for example) we use it [here](https://github.com/jbosstm/narayana/blob/30a7d8e676606d69017ea84d75b0218b1d163c9d/rts/lra/client/src/main/java/io/narayana/lra/client/NarayanaLRAClient.java#L886)

mmusgrov: The first implementation dependency `public class ResteasyClientBuilder extends javax.ws.rs.client.ClientBuilder` 
and `public class MpClientBuilderImpl extends ResteasyClientBuilderImpl` so why do we need the both? I'll ask again, how come it worked before: which use case requires MpClientBuilderImpl and which use cases require the super class ResteasyClientBuilderImpl. I assume that the TCK requires MpClientBuilderImpl (or is it the MP platform that requires it). So if that is the case then we just need to identify the ResteasyClientBuilderImpl use case.

We used to have just `ResteasyClientBuilder` but then someone on the team replaced it with just `MpClientBuilderImpl`. So if you're not sure about the use cases then can you look back through the commit history and tag the engineer who made the change and ask them if we need tell the service loader about both implementation classes.

rts/lra/client/pom.xml
mmusgrov: Can you add this as a unit test.

rts/lra/client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
jmfinelli: Of course, I can do that :-). Just to mention, it is possible to specify a list of dependencies separated with newline, therefore, specifying two dependencies for `javax.ws.rs.client.ClientBuilder` should not be a problem

rts/lra/client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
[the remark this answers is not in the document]
mmusgrov: Okay, so your reproducer in the previous comment is such an example and is a good reason to include it. I'd still like to get some clarification on my question below about whether we need to tell the service loader about both classes, in particular if we can get rid of the most derived class (MpClientBuilderImpl) then that would be beneficial since it would remove the dependency on whatever library is providing MpClientBuilderImpl.

rts/lra/client/pom.xml
jmfinelli: I have added this TestCase in the (future) PR for [JBTM-3470](https://issues.redhat.com/browse/JBTM-3470). If you also want to test lra-client on its own, I can add a bit of code to test it with Arquillian :-)

rts/lra/client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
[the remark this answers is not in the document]
jmfinelli: I'll dive deeper into this and I'll report back :+1:

rts/lra/client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
mmusgrov: My concern is that depending on MpClientBuilderImpl is coupling the `lra-client` module to whatever library it belongs to. Your change back to the generic case makes the inclusion of MpClientBuilderImpl suspicious.

rts/lra/client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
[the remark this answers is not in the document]
mmusgrov: What happens if you just remove `org.jboss.resteasy.microprofile.client.impl.MpClientBuilderImpl`. Does the TCK and other tests still pass.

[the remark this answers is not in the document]
mmusgrov: If you check the git commit history the commit comment (and associated JIRA) may indicate the rationale for replacing ResteasyClientBuilder with the more specific class.

rts/lra/client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
mmusgrov: So your analysis shows MpClientBuilderImpl was added whilst fixing https://issues.redhat.com/browse/JBTM-3381 and was required for use with quarkus (thanks for researching that). So it looks like your proposal (to add both) is the correct one. I will mark this thread as resolved.

rts/lra/client/pom.xml
mmusgrov: Your test case would not need to have any server running, I think just asserting that `java.lang.ClassNotFoundException` is not thrown when calling `lraClient.getAllLRAs()` would be enough (or asserting 404 or whatever gets returned from the call when there is no server).

rts/lra/client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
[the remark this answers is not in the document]
mmusgrov: Thanks for checking this, even though the TCK still works, then as you reported it was need by a community member when he was testing against quarkus (https://issues.redhat.com/browse/JBTM-3381).

I willl mark this conversation as resolved.

rts/lra/client/pom.xml
```diff
@@ -29,6 +29,16 @@
...
+      <groupId>org.jboss.resteasy</groupId>
+      <artifactId>resteasy-client</artifactId>
+      <version>${version.org.jboss.resteasy}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jboss.resteasy</groupId>
+      <artifactId>resteasy-jackson2-provider</artifactId>
+      <version>${version.org.jboss.resteasy}</version>
+    </dependency>
```
jmfinelli, line 41: After a chat with @ruromero, we can confirm that the `resteasy-jackson2-provider` dependency will work nicely with Quarkus based applications :+1:
Thanks Ruben

rts/lra/client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
jmfinelli, line 2: `MpClientBuilderImpl` was added because of Quarkus-based applications. We'll keep `MpClientBuilderImpl` among possible `ClientBuilder` implementations to allow Quarkus users to use lra-client out of the box.

rts/lra/client/pom.xml
jmfinelli: I think that a server (basically, an Arquillian container) should be used for testing the client as more unit tests should be written to test different aspects of lra-client (for example, the exception thrown when `resteasy-jackson2-provider` is removed)

rts/lra/client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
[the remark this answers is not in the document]
jmfinelli: I can confirm that `MpClientBuilderImpl` was added because of Quarkus-based applications. We'll keep `MpClientBuilderImpl` among possible `ClientBuilder` implementations to allow Quarkus users to use lra-client out of the box.

ochaloup (approved):
I was wondering about the resteasy dependencies for the LRA project that I think it should be implementation agnostic but after reading the discussion here, briefly looking at the behaviour and with having a chat with @xstefank the changeset seems to me reasonable to be merged.

rts/lra/client/pom.xml
jmfinelli: What about creating a new JIRA to track unit tests? I would like to keep this PR focused on dependencies. @mmusgrov If you agree with this, could you please resolve this conversation?

jmfinelli:
> I was wondering about the resteasy dependencies for the LRA project that I think it should be implementation agnostic but after reading the discussion here, briefly looking at the behaviour and with having a chat with @xstefank the changeset seems to me reasonable to be merged.

Thank you very much @ochaloup @xstefank +1

rts/lra/client/pom.xml
mmusgrov: It's not normal to add functionality and to then defer the test to a future patch.
But I guess in this instance it's not a big enough deal to argue the case so I'll approve the PR and you can merge it.

PR #1866, opened 2021-06-23
JBTM-3503 NPE when using @Singleton @Transactional(TxType.REQUIRES_NEW)

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/JBTM-3503

MAIN AS_TESTS 

### DIFF
--- a/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
+++ b/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
@@ -107,14 +107,16 @@ public abstract class TransactionalInterceptorBase implements Serializable {
      */
     private Transactional getTransactional(InvocationContext ic) {
         // when the CDI implementation is Weld then using the Weld API for accessing the annotated type
-        if (ic.getContextData().containsKey(WELD_INTERCEPTOR_BINDINGS_KEY)) {
+        // when weld doesn't provide the annotation as per JBTM-3503 sample test, we switch to more complicated processing.
+        if (interceptedBean != null && ic.getContextData().get(WELD_INTERCEPTOR_BINDINGS_KEY) != null) {
             Set<Annotation> annotationBindings = (Set<Annotation>) ic.getContextData().get(WELD_INTERCEPTOR_BINDINGS_KEY);
             for (Annotation annotation : annotationBindings) {
                 if (annotation.annotationType() == Transactional.class) {
                     return (Transactional) annotation;
                 }
             }
-        } else if (interceptedBean != null) { // not-null for CDI
+        }
+        if (interceptedBean != null) { // not-null for CDI
             // getting annotated type and method corresponding of the intercepted bean and method
             AnnotatedType<?> currentAnnotatedType = extension.getBeanToAnnotatedTypeMapping().get(interceptedBean);
             if (currentAnnotatedType == null) {
--- a/ArjunaJTA/cdi/tests/classes/com/hp/mwtests/ts/jta/cdi/transactional/stereotype/extension/NoTransactional.java
+++ b/ArjunaJTA/cdi/tests/classes/com/hp/mwtests/ts/jta/cdi/transactional/stereotype/extension/NoTransactional.java
@@ -1,0 +1,43 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2021, Red Hat Middleware LLC, and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package com.hp.mwtests.ts.jta.cdi.transactional.stereotype.extension;
+
+import javax.ejb.ConcurrencyManagement;
+import javax.ejb.Singleton;
+import javax.ejb.Startup;
+import javax.transaction.Transactional;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Startup
+@Singleton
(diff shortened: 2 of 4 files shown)


ochaloup:
TESTIT

ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorRequiresNew.java
```diff
@@ -49,7 +49,7 @@ public Object intercept(InvocationContext ic) throws Exception {
 
     @Override
     protected Object doIntercept(TransactionManager tm, Transaction tx, InvocationContext ic) throws Exception {
-        if (tx != null) {
+        if (tx != null && ic != null) {
```
ochaloup, line 52: So if I understand correctly then this fix expects that the tx could not be `null` - i.e., in case of EJB manages  transactions and the transactions was started - but the `ic` is null as it's EJB that is the manager and is not the responsibility of CDI to work here.
Do I interpret this right?

ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
```diff
@@ -109,9 +109,11 @@ private Transactional getTransactional(InvocationContext ic) {
         // when the CDI implementation is Weld then using the Weld API for accessing the annotated type
         if (ic.getContextData().containsKey(WELD_INTERCEPTOR_BINDINGS_KEY)) {
```
ochaloup, line 110: I think that "an easier" and in way of how the code is written better (feel free to correct me) fix would be like

```
// not-null for CDI && CDI is managed by Weld
if (interceptedBean != null && ic.getContextData().containsKey(WELD_INTERCEPTOR_BINDINGS_KEY)) {
```

What do you think?

ochaloup (changes requested):
I'm providing and alternative idea of the fix (not tested, though).

When we conclude that the code proposed in this PR right now is better to what I suggested then I assume it's needed to cover the check of `ic != null` not only in `TransactionalInterceptorRequiresNew` but for other inherited interceptor classes as well.

Then, how do you think about testing? It seems this is about EJB handling so the test will go to WildFly integration testsuite?

ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorRequiresNew.java
RanabirChakraborty: Yes, I got you. Whenever EJB manages the transactions the value of ic could be null. I have missed that scenario, will change it to as it was before.

ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
RanabirChakraborty: I guess checking `annotationBindings` will be an efficient one and tested it in my local not able to see any NPE.

ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
```diff
@@ -109,9 +109,11 @@ private Transactional getTransactional(InvocationContext ic) {
...
             Set<Annotation> annotationBindings = (Set<Annotation>) ic.getContextData().get(WELD_INTERCEPTOR_BINDINGS_KEY);
-            for (Annotation annotation : annotationBindings) {
-                if (annotation.annotationType() == Transactional.class) {
-                    return (Transactional) annotation;
+            if (annotationBindings != null) {
```
ochaloup, line 112: I'm still not sure about the fix. This was the code before the adjustment for handling Weld in other path to fix 
```
[JBTM-3442] using Weld API to get transactional annotation at CDI interceptor if Weld is available
```
-> https://github.com/jbosstm/narayana/blob/5.11.2.Final/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java#L106

Which works in way: `if CDI -> go the branch otherwhise -> EJB branch`
Your fix does `if Weld -> go the branch -> otherwhise -> do nothing`
I think it should be `if Weld and CDI -> go the branch otherwise check if CDI -> go the branch otherwise -> EJB branch`

I'm still considering the fix should be in way

```
// not-null for CDI && CDI is managed by Weld
if (interceptedBean != null && ic.getContextData().containsKey(WELD_INTERCEPTOR_BINDINGS_KEY)) {
 // when the CDI implementation is Weld then using the Weld API for accessing the annotated type
...
```

Could you please elaborate on your solution for me to undestand if I'm missing something?

ochaloup (changes requested):
I'm still not sure if the provided fix follows what was in the code before.

RanabirChakraborty:
Will change the PR according to the discussion with @ochaloup.

ochaloup:
Just some details form discussion with @RanabirChakraborty 

On investigation I found that the weld IC does not provide (`ic.getContextData().getKey(WELD_INTERCEPTOR_BINDINGS_KEY)` the `@Transactional` annotation despite the fact the method contains it. The original processing was capable to find the annotation via the loop over the CDI bindings.
I think it could be some trouble of Weld or it could be some design decision of Weld.

We need to be using the prior processing when there is not found annotation as backing way when Weld gives us nothing.
Then the test based on the reproducer needs to be added to testsuite (Narayana JTA CDI or WildFly, not sure where it belogs better as this error - reported by jira - happens for case when EJB and CDI is combined)

mmusgrov:
@ochaloup Are you saying that the original fix (git SHA ced7cdf74 made on 24th May) was insufficient?

RanabirChakraborty:
NPE issue is resolved, will write a test for this.

ochaloup:
@RanabirChakraborty the code seems fine to me now.
I'm just *not* approving as we need a test. When the test is provided I will be glad to approve your fix. Thank you!

ochaloup:
Thanks @RanabirChakraborty for providing the test. I'm just not sure if it's in all details what we need for the issue to be tested.
I have got few points you should focus on and enahance the code.

* when the test is executed then it fails[1]
* your changes influene other tests in the same package. They all fail with the same error. Please, study how to write the Arquillian tests and consider using separate package or this
* the test could be better named (a TransactionTest is not much descriptive) or/and slightly commented about functionality
* Please study what is the stereotype in CDI
* When the test passes with your fix you need to verify that the test fails without the fix in this PR. I'm worry that the test do not verify what was the scope of the issue.

[1]
```
"jboss.deployment.unit.\"JBTM-3503-test.war\".WeldStartService" => "Failed to start service
    Caused by: org.jboss.weld.exceptions.DeploymentException: WELD-001408: Unsatisfied dependencies for type TransactionalBean with qualifiers @Default
  at injection point [BackedAnnotatedField] @Inject com.hp.mwtests.ts.jta.cdi.transactional.stereotype.extension.TransactionalTest.TransactionalBean
  at com.hp.mwtests.ts.jta.cdi.transactional.stereotype.extension.TransactionalTest.TransactionalBean(TransactionalTest.java:0)
WELD-001474: Class com.hp.mwtests.ts.jta.cdi.transactional.stereotype.extension.TransactionalBean is on the classpath, but was ignored because a class it references was not found: com.hp.mwtests.ts.jta.cdi.transactional.stereotype.TestException from [Module \"deployment.JBTM-3503-test.war\" from Service Module Loader].
"}}}}
```

ochaloup:
Hi @RanabirChakraborty . As I created the test and decided to supersede this PR with my one at https://github.com/jbosstm/narayana/pull/1875

PR #1870, opened 2021-06-28
JBTM-3509 Small optimisations in setting Transaction Timeout

Sanne:
https://issues.redhat.com/browse/JBTM-3509

MAIN CORE TOMCAT AS_TESTS !RTS JACOCO !XTS QA_JTA QA_JTS_JACORB QA_JTS_JDKORB QA_JTS_OPENJDKORB !BLACKTIE !PERF !LRA !NO_WIN !DB_TESTS mysql db2 postgres oracle

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/BaseTransaction.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/BaseTransaction.java
@@ -86,7 +86,7 @@ public class BaseTransaction
 		}
 
 		Integer value = _timeouts.get();
-		int v = 0; // if not set then assume 0. What else can we do?
+		int v;
 
 		if (value != null)
 		{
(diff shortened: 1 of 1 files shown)


ochaloup:
TESTIT

mmusgrov (approved):
Testing this against the CORE/MAIN axis is sufficient so I will review it now (but may as well allow the other axes to finish since they have already started).

The tomcat axis failed with could not locate org.jboss.narayana:common:pom:5.9.0.Final which does exist (https://mvnrepository.com/artifact/org.jboss.narayana/common/5.9.0.Final) so it can be ignored as a transient failure.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/BaseTransaction.java
```diff
@@ -174,9 +174,13 @@ public int getStatus() throws javax.transaction.SystemException
...
 			throws javax.transaction.SystemException
 	{
-		if (seconds >= 0)
+		if (seconds > 0)
+		{
+		    _timeouts.set(Integer.valueOf(seconds));
+		}
+		else if (seconds == 0)
 		{
-		    _timeouts.set(new Integer(seconds));
```
Sanne, line 179: FWIW, doing `new Integer(int)` is deprecated in most recent JVMs and ill be removed. To be future-proof, I'd suggest consistently changing all code to use the better `Integer.valueOf(int)` variant.

ochaloup (approved):

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/BaseTransaction.java
mmusgrov: We do that quite a lot so thanks for the tip

PR #1875, opened 2021-07-07
[JBTM-3503] when Weld does not provide @Transactional annotations using prior way of searching it

ochaloup:
https://issues.redhat.com/browse/JBTM-3503

superseding https://github.com/jbosstm/narayana/pull/1866

MAIN AS_TESTS

### DIFF
--- a/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
+++ b/ArjunaJTA/cdi/classes/com/arjuna/ats/jta/cdi/transactional/TransactionalInterceptorBase.java
@@ -107,14 +107,16 @@ public abstract class TransactionalInterceptorBase implements Serializable {
      */
     private Transactional getTransactional(InvocationContext ic) {
         // when the CDI implementation is Weld then using the Weld API for accessing the annotated type
-        if (ic.getContextData().containsKey(WELD_INTERCEPTOR_BINDINGS_KEY)) {
+        // when Weld does not provide the annotation we switch to more complicated processing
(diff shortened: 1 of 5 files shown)


ochaloup:
@mmusgrov by me this could be merged when reviewed (saying as I won't be available next week)

I verified the test fails without the fix and checked with debugger what the fix does. @RanabirChakraborty verified this type of fix works with the reproducer.

mmusgrov (approved):
I don't have a lot of experience with this module but it looks correct and Ranabir has re-ran his reproducer and we have re-ran our own CDI tests. Is there evidence that the fix still works with the WildFly test suite?

RanabirChakraborty:
I have checked this fix with WildFly and it's running fine with WildFly test suite.

This fix looks good. one thing is that though we are using single instance but do we need to add `@singleton` annotation like we have in the reproducer?

ochaloup:
@RanabirChakraborty  the `@Singleton` is one of the EJB types and in the reproducer it provides a possibility to execute code just after the archive is deployed (otherwise there is needed to have some servlet that is `curl`ed or other way to execute the code).
The Weld (from what I can see from may investigation) has got troubles to combine the EJB+CDI to lookup the annotation. The testsuite works with Arquillian and the execution can be done via the injection.

In summary, we don't need a `Singleton` bean but we need the combination of EJB bean + CDI annotation.

ochaloup:
@mmusgrov + the AS test axis passes with this fix applied - http://narayanaci1.eng.hst.ams2.redhat.com/job/btny-pulls-narayana/PROFILE=AS_TESTS,jdk=jdk8.latest,label=linux/101/

mmusgrov:
Good point @ochaloup . So, if @RanabirChakraborty is happy with the fix then can this one can be merged?

RanabirChakraborty (approved):

PR #1877, opened 2021-07-12
Keep narayana-tomcat under jbosstm org version updated

tomjenkinson:
NO_TEST NO_WIN


mmusgrov (approved):
Thanks, I wasn't sure what change you had been proposing.

PR #1878, opened 2021-07-12
Replace the version number of Narayana in the narayana-tomcat checkout

tomjenkinson:
!MAIN !CORE TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !PERF !LRA NO_WIN !DB_TESTS mysql db2 postgres oracle

### DIFF
--- a/scripts/hudson/narayana.sh
+++ b/scripts/hudson/narayana.sh
@@ -628,10 +628,15 @@ EOT
     rm -rf narayana-tomcat
     git clone https://github.com/jbosstm/narayana-tomcat.git
     echo "Executing Narayana Tomcat tests"
-    ./build.sh -f narayana-tomcat/pom.xml -fae -B -P${ARQ_PROF}-tomcat ${CODE_COVERAGE_ARGS} -Dtest.db.type=h2 -Dversion.org.jboss.narayana=${NARAYANA_CURRENT_VERSION} "$@" ${IPV6_OPTS} clean install "$@"
+    # When the version of Narayana in the jbosstm/narayana-tomcat pom.xml changes, the following line needs to be updated
+    NARAYANA_TOMCAT_NARAYANA_VERSION=5.9.11.Final
+    grep "<version.org.jboss.narayana>${NARAYANA_TOMCAT_NARAYANA_VERSION}</version.org.jboss.narayana>" narayana-tomcat/pom.xml
+    [ $? -eq 0 ] || fatal "The version of Narayana in the jbosstm fork is not the expected version to replace"
+    sed -i "" "s#<version.org.jboss.narayana>${NARAYANA_TOMCAT_NARAYANA_VERSION}</version.org.jboss.narayana>#<version.org.jboss.narayana>${NARAYANA_CURRENT_VERSION}</version.org.jboss.narayana>#g" narayana-tomcat/pom.xml
+    ./build.sh -f narayana-tomcat/pom.xml -fae -B -P${ARQ_PROF}-tomcat ${CODE_COVERAGE_ARGS} -Dtest.db.type=h2 "$@" ${IPV6_OPTS} clean install "$@"
     RESULT=$?
     [ $RESULT = 0 ] || fatal "Narayana Tomcat tests failed H2"
-    ./build.sh -f narayana-tomcat/pom.xml -fae -B -P${ARQ_PROF}-tomcat ${CODE_COVERAGE_ARGS} -Dtest.db.type=pgsql -Dversion.org.jboss.narayana=${NARAYANA_CURRENT_VERSION} "$@" ${IPV6_OPTS} clean install "$@"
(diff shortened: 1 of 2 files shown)


scripts/hudson/narayana.sh
```diff
@@ -628,10 +628,13 @@ EOT
     rm -rf narayana-tomcat
     git clone https://github.com/jbosstm/narayana-tomcat.git
     echo "Executing Narayana Tomcat tests"
-    ./build.sh -f narayana-tomcat/pom.xml -fae -B -P${ARQ_PROF}-tomcat ${CODE_COVERAGE_ARGS} -Dtest.db.type=h2 -Dversion.org.jboss.narayana=${NARAYANA_CURRENT_VERSION} "$@" ${IPV6_OPTS} clean install "$@"
+    grep "<version.org.jboss.narayana>5.9.11.Final</version.org.jboss.narayana>" narayana-tomcat/pom.xml
```
mmusgrov, line 631: Seems a bit brittle hardcoding reliance on 5.9.11.Final but I guess we will quickly see the problem if the version in the narayana-tomcat changes.

scripts/hudson/narayana.sh
```diff
@@ -628,10 +628,13 @@ EOT
...
+    [ $? -eq 0 ] || fatal "The version of Narayana in the jbosstm fork is not the expected version to replace"
+    sed -i "" "s#<version.org.jboss.narayana>5.9.11.Final</version.org.jboss.narayana>#<version.org.jboss.narayana>${NARAYANA_CURRENT_VERSION}</version.org.jboss.narayana>#g" narayana-tomcat/pom.xml
```
mmusgrov, line 633: I know the CI test run passed but I don't understand your sed syntax and it doesn't work on my Fedora release 34 laptop). Something like the following is what I'd expect:

`sed -i s/\<version\.org\.jboss\.narayana\>5.9.11.Final/\<version\.org\.jboss\.narayana\>${NARAYANA_CURRENT_VERSION}/g x`

I also didn't bother matching on the closing element.

mmusgrov (commented):
I think I understood your changes.
I reviewed them on face value and had an issue with your sed syntax.

scripts/hudson/narayana.sh
tomjenkinson: Regarding the initial `""` I see this being done here: https://github.com/jbosstm/narayana/blob/master/scripts/pre-release.sh#L63

I can't really remember why it was that way but maybe it was for MacOS?

I am not sure why that line isn't working on Fedora 34 - did you run it in isolation or did you run it as part of the script? Perhaps there is something else that has been set in the script that enables this kind of syntax?

scripts/hudson/narayana.sh
tomjenkinson: +1, but I should have said that somewhere maybe - i.e. when the version in our fork of narayana-tomcat's pom.xml changes, to update this. I am not sure if hiding the comment above here helps in isolation (but maybe it would help when debugging) but I guess we might try to find somewhere else (maybe a second comment in our fork of narayana-tomcat's pom.xml?)

I will add a comment to this file though

scripts/hudson/narayana.sh
mmusgrov: I did the following:

```bash
[mmusgrov@dev2 tmp]$ cat x
line1
<version.org.jboss.narayana>5.9.11.Final</version.org.jboss.narayana>
line2

[mmusgrov@dev2 tmp]$ echo ${NARAYANA_CURRENT_VERSION}
5.CURRENT
[mmusgrov@dev2 tmp]$ sed -i "" "s#<version.org.jboss.narayana>5.9.11.Final</version.org.jboss.narayana>#<version.org.jboss.narayana>${NARAYANA_CURRENT_VERSION}</version.org.jboss.narayana>#g" x
sed: can't read s#<version.org.jboss.narayana>5.9.11.Final</version.org.jboss.narayana>#<version.org.jboss.narayana>5.CURRENT</version.org.jboss.narayana>#g: No such file or directory
[mmusgrov@dev2 tmp]$ 
```

mmusgrov: I will research (later) what the alternate syntax is about.

scripts/hudson/narayana.sh
tomjenkinson: I did a little more than the comment, hopefully the changes in this repo are clearer.

scripts/hudson/narayana.sh
mmusgrov: Even though it (appears to) break on my system, I will merge anyway after the test finishes. If does break on CI we would just push another update anyway.

mmusgrov (approved):

PR #1879, opened 2021-07-14
JBTM-3108 enable site plugin

mmusgrov:
https://issues.redhat.com/browse/JBTM-3108

!MAIN !CORE !TOMCAT !AS_TESTS !RTS JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB PERF LRA NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/rts/lra/pom.xml
+++ b/rts/lra/pom.xml
@@ -164,13 +164,6 @@
           </includes>
         </configuration>
       </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-site-plugin</artifactId>
-        <configuration>
-          <skip>true</skip>
-        </configuration>
-      </plugin>
     </plugins>
   </build>
 


mayankkunwar (approved):

mayankkunwar:
The `maven site` is enabled with or without these changes. In my local system it is creating `site` folder under the `target` directory in all the cases.
The site folder is not shown in the CI jobs, maybe we need to add the regex for the `maven site` directory in the CI artifact configuration.

mmusgrov:
We disabled the site plugin because it [caused an issue](https://issues.redhat.com/browse/JBTM-3093) with the code coverage (the JACOCO CI axis) - something to do with JACOCO not resolving microprofile-lra-api snapshot builds correctly. But now that microprofile-lra is released we no longer rely on snapshot builds and therefore we no longer require the workaround introduced by the fix for JBTM-3093.

Also note that I removed the the maven-site-plugin config from the lra parent pom since it is already configured in the main narayana pom. I will trigger another run.

mayankkunwar:
> We disabled the site plugin because it [caused an issue](https://issues.redhat.com/browse/JBTM-3093) with the code coverage (the JACOCO CI axis) - something to do with JACOCO not resolving microprofile-lra-api snapshot builds correctly. But now that microprofile-lra is released we no longer rely on snapshot builds and therefore we no longer require the workaround introduced by the fix for JBTM-3093.
> 
> Also note that I removed the the maven-site-plugin config from the lra parent pom since it is already configured in the main narayana pom. I will trigger another run.

Thanks for the clarification.
Looks like everything is working now.

mmusgrov:
@mayankkunwar the JACOCO axis passed so I will merge it now

PR #1881, opened 2021-07-16
[JBTM-3427] split arq and arq.quarkus profiles

jmfinelli:
This PR addressed this [JIRA](https://issues.redhat.com/browse/JBTM-3427).

NOTE: there are two new testing axes for Long Running Action

MAIN CORE RTS LRA

!TOMCAT !AS_TESTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !PERF NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/rts/lra/test/README.adoc
+++ b/rts/lra/test/README.adoc
@@ -1,31 +1,35 @@
 = `lra/test` module - integration tests
 
+This module has been designed to simulate an environment with three actors: the LRA coordinator, the LRA participant,
+and a client (which in this case is the test class). This is implemented using three JVMs running separately.
+This reproduces a production scenario where the LRA coordinator runs independently from the LRA participant(s)
+and from the client(s).
+All tests are executed with help of https://arquillian.org[Arquillian] platform. In some tests, Arquillian is coupled
+with maven plugin to start the runtime and deploy the LRA coordinator.
+
 The `lra/test` module contains the following submodules
 
-* link:./arquillian-extension/[`arquillian-extension`] contains set of
-  Arquillian extension classes which makes possible to execute the test modules
-  on supported runtimes
+* link:./arquillian-extension/[`arquillian-extension`] contains Arquillian extension classes
+  which make possible to execute the test modules on supported runtimes
 * link:./basic[`basic`] contains integration tests which are written for testing
   the Narayana features of the
   https://github.com/eclipse/microprofile-lra[LRA specification]
-* link:./tck[`tck`] is a module which is responsible for execution of
+* link:./tck[`tck`] is a module which is responsible for the execution of
   https://github.com/eclipse/microprofile-lra/tree/master/tck[LRA TCK tests]
   with Narayana supported runtimes
-
-All tests are executed with help of https://arquillian.org[Arquillian] platform.
-Arquillian is responsible to launch the runtime and makes possible to execute
-the tests.
+* link:./crash[`crash`] hosts tests that simulate a crash of either the LRA coordinator or
+  the LRA participant or both. So far, only Wildfly is considered in this module as Quarkus
+  does not have an Arquillian adapter to run unmanaged tests
 
 == How to run integration tests
 
-The itegration tests which should be run on the application runtime
-are configured to be started with https://maven.apache.org/surefire/maven-failsafe-plugin/[Maven Failsafe Plugin].
+Integration tests are tested with
+https://maven.apache.org/surefire/maven-failsafe-plugin/[Maven Failsafe Plugin], using the target 'verify'.
 
-To run the tests the target `verify` has to be used.
-Single test with `-Dit.test=...`.
-Debug the Maven test execution with `-Dmaven.failsafe.debug`.
+Single test can be executed with `-Dit.test=...`
+To use Failsafe in debug mode, use `-Dmaven.failsafe.debug`
 
-The Failsafe Plugin link:../pom.xml#114[is configured] to run tests with names
+The Failsafe Plugin link:../pom.xml[is configured] to run tests with names
 `*IT.java`, `*Tests.java` and `*TestCase.java`.
 
 === Application runtimes
@@ -33,14 +37,17 @@ The Failsafe Plugin link:../pom.xml#114[is configured] to run tests with names
 The module is configured to run two runtimes - https://www.wildfly.org[WildFly]
 and link:https://quarkus.io[Quarkus]/link:https://thorntail.io[Thorntail].
 
-The **WildFly** runtime deploys the coordinator and the application with the client
-library at the same application server. The coordinator is deployed as
-a separated `war` archive, and the same the applications which packs together
-the LRA client library.
+When the **WildFly** runtime is used (through the 'arq' profile), two instances of Wildfly are initiated:
+ * The first instance is started using WildFly Maven Plugin, then the LRA coordinator is deployed using the WAR produced
+ by the module `lra/lra-coordinator-war`
+ * The second instance is started using Arquillian, then an LRA participant is deployed
+Moreover, the testing class is hosted on separated JVM.
 
-**Quarkus/Thorntail** splits the deployments to two separate JVMs.
-The Quarkus is started as Narayana LRA coordinator and
-Thorntail packs the client application with the LRA client library.
+Also, the configuration using **Quarkus** (through the 'arq.thorntail') runs the LRA coordinator and the LRA participant(s)
+in two separate JVMs:
+ * Quarkus is started with its maven plugin and configured to host Narayana LRA coordinator
+ * A Thorntail (Arquillian) container hosts the LRA participant
+Again, the testing class is hosted on a separated JVM.
 
 === Running tests with Quarkus/Thorntail
 
@@ -48,10 +55,10 @@ To start integration tests:
 
 [source,sh]
 ----
-mvn clean verify -Parq
+mvn clean verify -Parq.thorntail
 
-# single test in one module
-mvn clean verify -Parq -pl basic -Dit.test=FailedLRAIT
+# single test in one module (example)
+mvn clean verify -Parq.thorntail -pl basic -Dit.test=FailedLRAIT
 ----
 
 .Configuration options of Quarkus/Thorntail
@@ -86,10 +93,10 @@ To start all integration tests:
 [source,sh]
 ----
 export JBOSS_HOME=<path-to-wildfly-distribution-directory>
-mvn clean verify -Parq.wildfly
+mvn clean verify -Parq
 
 # single test in one module
-mvn clean verify -Parq.wildfly -pl basic -Dit.test=FailedLRAIT
+mvn clean verify -Parq -pl basic -Dit.test=FailedLRAIT
 ----
 
 .Configuration options of WildFly
--- a/rts/lra/test/pom.xml
+++ b/rts/lra/test/pom.xml
@@ -107,8 +107,8 @@
 
   <profiles>
     <profile>
-      <!-- when arq profile is activated we expect to start coordinator -->
-      <id>arq</id>
+      <!-- when arq.thorntail profile is activated we expect to start coordinator -->
+      <id>arq.thorntail</id>
       <properties>
         <!-- making the Quarkus coordinator to be started -->
         <lra.coordinator.exec.plugin.phase>pre-integration-test</lra.coordinator.exec.plugin.phase>
@@ -149,7 +149,7 @@
     </profile>
 
     <profile>
-      <id>arq.wildfly</id>
+      <id>arq</id>
       <modules>
         <module>crash</module>
       </modules>
@@ -159,19 +159,25 @@
         <!-- Qualifier of the lra-coordinator of the deployment -->
         <lra.coordinator.deployment.qualifier>lra-coordinator</lra.coordinator.deployment.qualifier>
         <!-- the AppServerCoordinatorDeploymentObserver creates the deployment named 'lra-coordinator' and deploys it to the app server,
-                     the last part of the path with value 'lra-coordinator' is defined by LRA coordinator JAX-RS endpoint -->
+             the last part of the path with value 'lra-coordinator' is defined by LRA coordinator JAX-RS endpoint -->
         <lra.coordinator.url>http://${lra.coordinator.host}:${lra.coordinator.port}/lra-coordinator/lra-coordinator</lra.coordinator.url>
+        <lra.coordinator.war.name>lra-coordinator.war</lra.coordinator.war.name>
+        <!-- The following filename is used for starting the WildFly instance that hosts the LRA coordinator; this file is a copy of the
+             standalone.xml file provided in WildFly (${env.JBOSS_HOME}/standalone/configuration/standalone.xml). Please, see the plugins
+             section of this profile (specifically, maven-antrun-plugin) for further info -->
         <lra.coordinator.xml.filename>lra-coordinator.xml</lra.coordinator.xml.filename>
         <!-- Qualifier of the container hosting lra-participant -->
         <lra.participant.container.qualifier>lra-participant-server</lra.participant.container.qualifier>
         <!-- Qualifier of the lra-participant of the deployment -->
         <lra.participant.deployment.qualifier>lra-participant</lra.participant.deployment.qualifier>
+        <!-- The following filename is used for starting the WildFly instance that hosts the LRA participant; this file is a copy of the
+             standalone.xml file provided in WildFly (${env.JBOSS_HOME}/standalone/configuration/standalone.xml). Please, see the plugins
+             section of this profile (specifically, maven-antrun-plugin) for further info -->
         <lra.participant.xml.filename>lra-participant.xml</lra.participant.xml.filename>
         <skip.wildfly.plugin>true</skip.wildfly.plugin>
         <!-- where the WFLY test application will be started at (the coordinator is war deployed for the same app server at the same port) -->
         <test.application.port>8080</test.application.port>
         <version.wildfly-maven-plugin>2.1.0.Beta1</version.wildfly-maven-plugin>
-        <lra.coordinator.war.name>lra-coordinator.war</lra.coordinator.war.name>
       </properties>
 
       <dependencies>
--- a/scripts/hudson/narayana.sh
+++ b/scripts/hudson/narayana.sh
@@ -24,7 +24,7 @@ function get_pull_xargs {
   res=$(echo $res | sed 's/"/ /g')
   OLDIFS=$IFS
   IFS=', ' read -r -a array <<< "$res"
-  echo "get_pull_xargs: parsing $1" 
+  echo "get_pull_xargs: parsing $1"
 
   for element in "${array[@]}"
   do
@@ -54,6 +54,7 @@ function init_test_options {
     [ $CODE_COVERAGE ] || CODE_COVERAGE=0
     [ x"$CODE_COVERAGE_ARGS" != "x" ] || CODE_COVERAGE_ARGS=""
     [ $ARQ_PROF ] || ARQ_PROF=arq	# IPv4 arquillian profile
+    [ $ARQ_THORNTAIL ] || ARQ_THORNTAIL=arq.thorntail
     [ $IBM_ORB ] || IBM_ORB=0
     [ $ENABLE_LRA_TRACE_LOGS ] || ENABLE_LRA_TRACE_LOGS=" -Dtest.logs.to.file=true -Dtrace.thorntail.test -Dtrace.lra.coordinator"
 
@@ -210,7 +211,7 @@ function init_test_options {
     [ $PERF_TESTS ] || PERF_TESTS=0 # benchmarks
     [ $REDUCE_SPACE ] || REDUCE_SPACE=1 # Whether to reduce the space used
 
-    get_pull_xargs "$PULL_DESCRIPTION_BODY" $PROFILE # see if the PR description overrides any of the defaults 
+    get_pull_xargs "$PULL_DESCRIPTION_BODY" $PROFILE # see if the PR description overrides any of the defaults
 
     JAVA_VERSION=$(java -version 2>&1 | grep "\(java\|openjdk\) version" | cut -d\  -f3 | tr -d '"' | tr -d '[:space:]' | awk -F . '{if ($1==1) print $2; else print $1}')
 }
@@ -286,7 +287,7 @@ function kill_qa_suite_processes
       [[ $main == ${pat}* ]] && killit=1
     done
 
-    if [[ $killit == 1 ]]; then 
+    if [[ $killit == 1 ]]; then
       echo "Test suite process $pid still running - terminating it with signal 9"
       kill -9 $pid
     fi
@@ -312,7 +313,7 @@ function build_narayana {
     ./build.sh -f jboss-transaction-spi/pom.xml -B clean install
     [ $? -eq 0 ] || fatal "Build of SPI failed"
   fi
-  
+
   echo "Building Narayana"
   cd $WORKSPACE
 
@@ -324,7 +325,7 @@ function build_narayana {
     [ $? -eq 0 ] || fatal "You must use the IBM jdk to build with ibmorb"
   fi
   echo "Using MAVEN_OPTS: $MAVEN_OPTS"
-  
+
   ./build.sh -B -Prelease,community$OBJECT_STORE_PROFILE $ORBARG "$@" $NARAYANA_ARGS $IPV6_OPTS $CODE_COVERAGE_ARGS clean install
 
   [ $? -eq 0 ] || fatal "narayana build failed"
@@ -381,7 +382,7 @@ function clone_as {
   echo "Rebasing the wildfly upstream/main on top of the AS_BRANCH $AS_BRANCH"
   git pull --rebase --ff-only upstream main
   [ $? -eq 0 ] || fatal "git rebase failed"
-  
+
   if [ $REDUCE_SPACE = 1 ]; then
     echo "Deleting git dir to reduce disk usage"
     rm -rf .git
@@ -523,8 +524,15 @@ function lra_tests {
   ./build.sh install -pl ArjunaCore/arjunacore,rts -am -DskipTests -Pcommunity "$@"
   # we can't use 'mvn -f' option beacuse of Thorntail plugin issue THORN-2049
   cd ./rts/lra/
+  echo "#0. Running LRA tests using $ARQ_PROF profile"
   PRESERVE_WORKING_DIR=true ../../build.sh -fae -B -Pcommunity -P$ARQ_PROF $CODE_COVERAGE_ARGS $ENABLE_LRA_TRACE_LOGS -Dlra.test.timeout.factor="${LRA_TEST_TIMEOUT_FACTOR:-1.5}" "$@"
-  [ $? = 0 ] || fatal "LRA Test failed"
+  lra_arq=$?
+  echo "#0. Running LRA tests using $ARQ_THORNTAIL profile"
+  PRESERVE_WORKING_DIR=true ../../build.sh -fae -B -Pcommunity -P$ARQ_THORNTAIL $CODE_COVERAGE_ARGS $ENABLE_LRA_TRACE_LOGS -Dlra.test.timeout.factor="${LRA_TEST_TIMEOUT_FACTOR:-1.5}" "$@"
+  lra_thorn=$?
+  if [ $lra_arq != 0 ] && [ $lra_thorn != 0 ] ; then fatal "LRA Test failed with failures in both $ARQ_PROF and $ARQ_THORNTAIL profiles" ;  fi
+  if [ $lra_arq != 0 ] ; then fatal "LRA Test failed with failures in $ARQ_PROF profile" ; fi
+  if [ $lra_thorn != 0 ] ; then fatal "LRA tests failed with failures in $ARQ_THORNTAIL profile" ; fi
 }
 
 function jta_cdi_tests {
@@ -704,13 +712,12 @@ function add_qa_xargs {
 
 function qa_tests_once {
   echo "QA Test Suite $@"
-  
-  
+
   # Download dependencies
   cd $WORKSPACE
   ./build.sh -f qa/pom.xml -B dependency:copy-dependencies
   [ $? -eq 0 ] || fatal "Copy dependency failed"
-  
+
   cd $WORKSPACE/qa
   unset orb
   codeCoverage=false;
@@ -769,7 +776,7 @@ function qa_tests_once {
     let txtimeout=$MFACTOR*120
     sed -e "s/COMMAND_LINE_13=-DCoordinatorEnvironmentBean.defaultTimeout=[0-9]*/COMMAND_LINE_13=-DCoordinatorEnvironmentBean.defaultTimeout=${txtimeout}/" TaskImpl.properties > "TaskImpl.properties.tmp" && mv "TaskImpl.properties.tmp" "TaskImpl.properties"
   fi
-  
+
   [ -z "${IPV6_OPTS+x}" ] && ant -Dorbtype=$orbtype "$QA_BUILD_ARGS" dist ||
     ant -Dorbtype=$orbtype "$QA_BUILD_ARGS" dist
 
@@ -833,7 +840,7 @@ function qa_tests_once {
 
     if [ -f TEST-failures.txt ]; then
       echo "Test Failures:"
-      cat TEST-failures.txt 
+      cat TEST-failures.txt
     fi
 
     if [ $codeCoverage = true ]; then
@@ -918,7 +925,7 @@ function perf_tests {
   rm -rf performance
   git clone https://github.com/jbosstm/performance
   cd performance/
-  if [ -n "$PULL_NUMBER" ]; 
+  if [ -n "$PULL_NUMBER" ];
   then
     echo $PULL_DESCRIPTION | grep https://github.com/jbosstm/performance/pull/
     if [[ "$?" -eq 0 ]]; then
@@ -931,7 +938,7 @@ function perf_tests {
       [ $? -eq 0 ] || fatal "git rebase failed"
     fi
   fi
-  
+
   ./scripts/run_bm.sh
   res=$?
   cd $WORKSPACE
@@ -987,9 +994,9 @@ init_test_options
 # FOR DEBUGGING SUBSEQUENT ISSUES
 if [ -x /usr/bin/free ]; then
     /usr/bin/free
-elif [ -x /usr/bin/vm_stat ]; then 
+elif [ -x /usr/bin/vm_stat ]; then
     /usr/bin/vm_stat
-else 
+else
     echo "Skipping memory report: no free or vm_stat"
 fi
 


.github/workflows/lra-test.yml
```diff
@@ -32,7 +32,8 @@ jobs:
         run: |
           git config --global user.email "you@example.com"
           git config --global user.name "Your Name"
-          WORKSPACE=$PWD PROFILE=LRA ./scripts/hudson/narayana.sh
+          WORKSPACE=$PWD PROFILE=LRA_THORNTAIL ./scripts/hudson/narayana.sh
```
mmusgrov, line 35: We need to migrate away from thorntail, for example I removed thorntail in favour of quarkus from the quickstarts because it was misbehaving, since we now have full support for running with quarkus can we use that instead.

In the lra/test pom `arq.thorntail` profile there's a comment saying `making the Quarkus coordinator to be started` which indicates we were once testing with quarkus.

What is the difference between ARQ_LRA_PROF and ARQ_PROF, why do we need an LRA specific variable, rts, xts, tomcat etc all use ARQ_PROF, what makes ARQ_LRA_PROF special?

scripts/hudson/narayana.sh
```diff
@@ -519,12 +530,16 @@ function rts_tests {
 }
 
 function lra_tests {
-  echo "#0. LRA Test"
+  if [ $LRA_WILDFLY_TESTS -eq 1 ]; then
+    echo "#0. LRA Tests with Wildfly"
```
ochaloup, line 534: a minor
```suggestion
    echo "#0. LRA Tests with WildFly"
```

scripts/hudson/narayana.sh
```diff
@@ -519,12 +530,16 @@ function rts_tests {
...
+  else
+    echo "#0. LRA Tests with Thorntail"
+  fi
   ./build.sh install -pl ArjunaCore/arjunacore,rts -am -DskipTests -Pcommunity "$@"
   # we can't use 'mvn -f' option beacuse of Thorntail plugin issue THORN-2049
   cd ./rts/lra/
-  PRESERVE_WORKING_DIR=true ../../build.sh -fae -B -Pcommunity -P$ARQ_PROF $CODE_COVERAGE_ARGS $ENABLE_LRA_TRACE_LOGS -Dlra.test.timeout.factor="${LRA_TEST_TIMEOUT_FACTOR:-1.5}" "$@"
-  [ $? = 0 ] || fatal "LRA Test failed"
+  PRESERVE_WORKING_DIR=true ../../build.sh -fae -B -Pcommunity -P$ARQ_LRA_PROF $CODE_COVERAGE_ARGS $ENABLE_LRA_TRACE_LOGS -Dlra.test.timeout.factor="${LRA_TEST_TIMEOUT_FACTOR:-1.5}" "$@"
+  [ $? = 0 ] || fatal "LRA Tests"
```
ochaloup, line 542: it would be good to inform in `fatal` what profile was executed and failed.

PULL_REQUEST_TEMPLATE.md
```diff
@@ -6,4 +6,4 @@ Please refer to our [guidelines for making contributions](https://github.com/jbo
 
 The build axis can be controlled by prefixing a ! on the following as appropriate:
 
-MAIN CORE TOMCAT AS_TESTS RTS JACOCO XTS QA_JTA QA_JTS_JACORB QA_JTS_JDKORB QA_JTS_OPENJDKORB PERF LRA !NO_WIN DB_TESTS mysql db2 postgres oracle
+MAIN CORE TOMCAT AS_TESTS RTS JACOCO XTS QA_JTA QA_JTS_JACORB QA_JTS_JDKORB QA_JTS_OPENJDKORB PERF LRA_THORNTAIL LRA_WILDFLY !NO_WIN DB_TESTS mysql db2 postgres oracle
```
mmusgrov, line 9: Under which circumstances would we make a change to LRA that would *not necessitate* running on all supported platforms. It seems niche to force a submitter to have to understand what all these axes mean. I think anyone, apart from the team, doesn't touch these because they are all niche. Maybe we need one or two variables which are clear and which under the covers starts a group of axes, and for those in the know such as ourselves we can slip in other variables. For example:

!LRA implies don't run any LRA axes and if you say make a wildfly specific change you would use LRA_WILDFLY etc

scripts/hudson/narayana.sh
```diff
@@ -62,13 +63,13 @@ function init_test_options {
         export COMMENT_ON_PULL=""
         export AS_BUILD=0 AS_CLONE=0 AS_DOWNLOAD=0 AS_TESTS=0 NARAYANA_BUILD=0 NARAYANA_TESTS=0 XTS_AS_TESTS=0 XTS_TESTS=0 TXF_TESTS=0 txbridge=0
         export RTS_AS_TESTS=0 RTS_TESTS=0 JTA_CDI_TESTS=0 QA_TESTS=0 SUN_ORB=0 OPENJDK_ORB=0 JAC_ORB=0 JTA_AS_TESTS=0
-        export PERF_TESTS=0 OSGI_TESTS=0 TOMCAT_TESTS=0 LRA_TESTS=0
+        export PERF_TESTS=0 OSGI_TESTS=0 TOMCAT_TESTS=0 LRA_THORNTAIL_TESTS=0 LRA_WILDFLY_TESTS=0
```
mmusgrov, line 66: Similar comment as my first one, we need to migrate to quarkus since thorntail isn't getting much love anymore.

And following on from my earlier comment, having a single LRA variable which we can conditionally modify using LRA_THORNTAIL_TESTS and LRA_WILDFLY_TESTS is going help with avoiding adding complexity to this script (we're already well over thousand lines of shell).

ochaloup (changes requested):
I have two minor comments to the PR otherwise I like it and I would approve it.

Though I'm marking it as "request changes" as I think it would be good to run the AMS CI axis for this change. The narayana poller job and pull job should be adjusted to allow running with `LRA_WIDLFY` and `LRA_THORNTAIL` profiles and comment back here. Or is there some intention about this?

The second point are the GitHub actions from me. It's good that the GitHub actions are not slowed down too much (like it seems only few minutes ~5-10) when the new testing runtime is added. But I think it would be good to consider spit the runs and execute separate "github action axis" for each runtime. Not saying it's required as part of this PR but maybe it could be done.

scripts/hudson/narayana.sh
mmusgrov: The addition of these extra profiles is already causing confusion, even for the author of the PR, lol

.github/workflows/lra-test.yml
jmfinelli: I think we are still running lra-coordinator on Quarkus. Thorntail is used to create the container that hosts lra-participants. Do not you think I should open a new JIRA to track the migration from Thorntail to Quarkus? @mmusgrov

I agree with you about ARQ_PROF. What do you think about this @ochaloup?

PULL_REQUEST_TEMPLATE.md
jmfinelli: +1

I agree that a single profile is easier to understand for the community and I also think we should test always on both platforms (i.e Wildfly and Thorntail/Quarkus). I am not sure about adding extra variables behind the scene though: I would be fine to start both profiles when LRA is tested (it does not take much time to test both Wildfly and Quarkus)

scripts/hudson/narayana.sh
jmfinelli: See previous comment

scripts/hudson/narayana.sh
jmfinelli: +1 thanks Ondra

scripts/hudson/narayana.sh
jmfinelli: Answered above

jmfinelli:
> Though I'm marking it as "request changes" as I think it would be good to run the AMS CI axis for this change. The narayana poller job and pull job should be adjusted to allow running with `LRA_WIDLFY` and `LRA_THORNTAIL` profiles and comment back here. Or is there some intention about this?

Sorry, forgot to answer this in the previous "review round". Yes, if the two profiles get merged, we should change jobs in our CI

.github/workflows/lra-test.yml
ochaloup: @jmfinelli yes, the `ARQ_PROF` is fine to be jused. Agree with Mike that the LRA is not special and using the same variable name for all axes makes very good sense.

Quarkus/Thorntain: We need to move from Thorntail to Quarkus but we need a way how to do so. Quarkus needs to pass the TCKs.

At time I was experimenting with it the Quarkus was not able to work with the LRA. As far as I can say we need the Quarkus  extension. Before spending more time to have Quarkus running as the LRA client without extension I was thinking to wait until the LRA extension is developed (https://github.com/quarkusio/quarkus/pull/17903).
(plus, the part of the LRA spec are the non-JAX-RS participants that was the main trouble before - without the extension)

mmusgrov: @ochaloup I think you are overestimating the capabilities of the quarkus lra extension. Basically, all it does is pull in the correct dependencies (microprofile-lra-api and narayana-lra), adds a config property for the location of the coordinator, and adds some excludes from the application. 

Using quarkus to build LRA capable apps without the extension is pretty simple.

The blocker is that quarkus does not support CDI extensions (ie it's not the lra extension that's causing the issue).
When I wrote the initial implementation of non JAX-RS support I didn't need CDI at all (CDI was only added later because I had lost control over the implementation changes). I will investigate workarounds or I will look to removing the narayana LRA dependency on CDI extensions.

mmusgrov: It looks like the easiest (recommended) solution is to port the CDI extension code and I am looking into that. I will provide an update if/when I have more ...

ochaloup: @mmusgrov yeap, agree. I wrote it wrongly - the trouble is to transform the CDI extension to make Quarkus working with LRA CDI.

mmusgrov: I will resolve the conversation.

PULL_REQUEST_TEMPLATE.md
mmusgrov: I'll repeat my motivation; we want to aim to limit the number of options available to the person raising the PR. So the following would satisfy that goal:

LRA enables both axes. Then if one of the team wants to skip thorntail they would add !LRA_THORNTAIL. Similarly for !LRA_WILDFLY. But I'm a little unsure here as well, isn't LRA_THORNTAIL for running the TCK and LRA_WILDFLY for running integration tests and if so we should use LRA_TCK and LRA_INTEGRATION, in this way we abstract away the details of our (private) CI infrastructure and we also won't need to change the scripts when we remove thorntail.

The key point being that the person raising the PR is generally only interested in asking for LRA to be tested not the internals of what gets tested. We have too many axes being exposed to people raising PRs.

scripts/hudson/narayana.sh
mmusgrov: Let's keep it to just one axis then which has the benefit that the job configs won't need to change.

PULL_REQUEST_TEMPLATE.md
jmfinelli: LRA TCK gets tested with WFLY and QUARKUS. The two profiles LRA_THORNTAIL (maybe this should be renamed to LRA_QUARKUS) and LRA_WILDFLY just select where `lra-coordinator` will be deployed (i.e. QUARKUS or WILDFLY respectively)

IMO, testing a PR on both WFLY and QUARKUS should be mandatory as something that works on one of them could not work on the other (which is the actual scenario we have in the main branch)

ochaloup: Ok.

mmusgrov: > LRA TCK gets tested with WFLY and QUARKUS. The two profiles LRA_THORNTAIL (maybe this should be renamed to LRA_QUARKUS) and LRA_WILDFLY just select where `lra-coordinator` will be deployed (i.e. QUARKUS or WILDFLY respectively)
> 
> IMO, testing a PR on both WFLY and QUARKUS should be mandatory as something that works on one of them could not work on the other (which is the actual scenario we have in the main branch)

Ondra said he found an issue when running the TCK with quarkus?

jmfinelli: > If we aren't using thorntail can we get rid of all of the references to it?

We are still using Thorntail as an Arquillian container to host `lra-participant`. I haven't checked your PR in `jbosstm/quickstart` but if Quarkus has developed a valid Arquillian container then the answer is yes :-)

.github/workflows/lra-test.yml
xstefank: @mmusgrov @ochaloup That is what I indent to look into once Mike's PR is merged in Quarkus. The CDI extension we have is fairly simple functionality which can be rewritten in quarkus to be done during build time phase so we don't need to rely on CDI extension to do this processing. However, the CDI extension will be required in WFLY.

PULL_REQUEST_TEMPLATE.md
mmusgrov: I have no idea but I will mark it as approved once you have gone back to just one axis for LRA testing.

scripts/hudson/narayana.sh
```diff
@@ -523,8 +524,15 @@ function lra_tests {
   ./build.sh install -pl ArjunaCore/arjunacore,rts -am -DskipTests -Pcommunity "$@"
   # we can't use 'mvn -f' option beacuse of Thorntail plugin issue THORN-2049
   cd ./rts/lra/
+  echo "#0. Running LRA tests using $ARQ_PROF profile"
   PRESERVE_WORKING_DIR=true ../../build.sh -fae -B -Pcommunity -P$ARQ_PROF $CODE_COVERAGE_ARGS $ENABLE_LRA_TRACE_LOGS -Dlra.test.timeout.factor="${LRA_TEST_TIMEOUT_FACTOR:-1.5}" "$@"
-  [ $? = 0 ] || fatal "LRA Test failed"
+  LRA_ARQ=$?
```
mmusgrov, line 529: We use capitals for environment variables. Use lower case for this temporary variable otherwise it makes this complex script harder to grok, something obvious like ok or res or success or anything that's lower case and meaningful.

jmfinelli: done :-) thanks @mmusgrov

jmfinelli:
This PR is ready to be re-reviewed. Could you please take a look @ochaloup?

rts/lra/test/README.adoc
```diff
@@ -1,57 +1,63 @@
...
 
-The Failsafe Plugin link:../pom.xml#114[is configured] to run tests with names
+The Failsafe Plugin link:../pom.xml[is configured] to run tests with names
 `*IT.java`, `*Tests.java` and `*TestCase.java`.
 
 === Application runtimes
 
 The module is configured to run two runtimes - https://www.wildfly.org[WildFly]
-and link:https://quarkus.io[Quarkus]/link:https://thorntail.io[Thorntail].
+and link:https://quarkus.io[Quarkus].
```
mmusgrov, line 38: What about thorntail?

rts/lra/test/README.adoc
```diff
@@ -1,57 +1,63 @@
...
 
-The **WildFly** runtime deploys the coordinator and the application with the client
-library at the same application server. The coordinator is deployed as
-a separated `war` archive, and the same the applications which packs together
-the LRA client library.
+When the **WildFly** runtime is used (through the 'arq' profile), two instances of Wildfly are initiated:
```
mmusgrov, line 40: So there is a profile called `arq` that runs the integration tests against WildFly?
And there is another one called `arq.thorntail` which runs against what?
And what does ARQ_THORN do?
And do we use ARQ_PROF=arqIPv6 in the LRA subsystem at all?

rts/lra/test/README.adoc
```diff
@@ -1,57 +1,63 @@
...
+ * The first instance is started using WildFly Maven Plugin, then lra-coordinator is deployed using the dedicated WAR
```
mmusgrov, line 41: What is the "dedicated WAR"?

rts/lra/test/README.adoc
jmfinelli: Thorntail is not used to host lra-coordinator. At the moment, Thorntail is only used to host lra-participants. Also, in a next PR, Thorntail will be removed completely

rts/lra/test/README.adoc
```diff
@@ -1,57 +1,63 @@
...
+ * The second instance is started using Arquillian, then an lra-participant is deployed
```
mmusgrov, line 42: What is `lra-participant`, I cannot find any other reference to this?

rts/lra/test/README.adoc
```diff
@@ -1,57 +1,63 @@
...
+Moreover, the testing class is hosted on a 'local' JVM, which runs separately from the servers.
```
mmusgrov, line 43: Why have you highlighted `local` JVM, do we also test using remote JVMs?

rts/lra/test/README.adoc
jmfinelli: `arq.thorntail` could also be renamed to `arq.quarkus`.
I do not think we use ARQ_PROF=arqIPv6 ATM. Should we open a new JIRA about this?

rts/lra/test/README.adoc
```diff
@@ -1,57 +1,63 @@
...
 
-**Quarkus/Thorntail** splits the deployments to two separate JVMs.
-The Quarkus is started as Narayana LRA coordinator and
-Thorntail packs the client application with the LRA client library.
+Also, the configuration using **Quarkus** (through the 'arq.thorntail') runs lra-coordinator and lra-partipant(s)
```
mmusgrov, line 45: This line is mixing concepts, why would we use 'arq.thorntail' to configure quarkus - they are very different runtimes (one is based on a the WildFly application server whereas the other runs on a standard JVM (with bytecode augmentation).

rts/lra/test/README.adoc
jmfinelli: it is the WAR coming from `lra-coordinator-war`. Should I include this info in the README?

rts/lra/test/README.adoc
jmfinelli: `lra-participant`s are JAX-RS applications implementing the LRA specification. IIRC, `participant` is an actor defined in the MP LRA specification

rts/lra/test/README.adoc
jmfinelli: I highlighted `local` because the other two JVMs (also local) have the role of remote JVMs. I see that this is not clear thought...I will remove it. Thanks Michael

rts/lra/test/README.adoc
jmfinelli: I see your point and I agree with you. I will rename `arq.thorntail` to `arq.quarkus`.

rts/lra/test/pom.xml
```diff
@@ -161,6 +161,7 @@
         <!-- the AppServerCoordinatorDeploymentObserver creates the deployment named 'lra-coordinator' and deploys it to the app server,
                      the last part of the path with value 'lra-coordinator' is defined by LRA coordinator JAX-RS endpoint -->
         <lra.coordinator.url>http://${lra.coordinator.host}:${lra.coordinator.port}/lra-coordinator/lra-coordinator</lra.coordinator.url>
+        <lra.coordinator.war.name>lra-coordinator.war</lra.coordinator.war.name>
         <lra.coordinator.xml.filename>lra-coordinator.xml</lra.coordinator.xml.filename>
```
mmusgrov, line 165: What is this file for, it is not referenced or documented anywhere and we don't use it in any of our builds.
Note I am reviewing this line because the last three commits you made are all about the rewrite of our approach to testing LRA.

scripts/hudson/narayana.sh
```diff
@@ -523,8 +524,15 @@ function lra_tests {
   ./build.sh install -pl ArjunaCore/arjunacore,rts -am -DskipTests -Pcommunity "$@"
   # we can't use 'mvn -f' option beacuse of Thorntail plugin issue THORN-2049
   cd ./rts/lra/
+  echo "#0. Running LRA tests using $ARQ_PROF profile"
   PRESERVE_WORKING_DIR=true ../../build.sh -fae -B -Pcommunity -P$ARQ_PROF $CODE_COVERAGE_ARGS $ENABLE_LRA_TRACE_LOGS -Dlra.test.timeout.factor="${LRA_TEST_TIMEOUT_FACTOR:-1.5}" "$@"
-  [ $? = 0 ] || fatal "LRA Test failed"
+  lra_arq=$?
```
mmusgrov, line 529: Ah right, okay then.

scripts/hudson/narayana.sh
```diff
@@ -704,13 +712,13 @@ function add_qa_xargs {
 
 function qa_tests_once {
   echo "QA Test Suite $@"
-  
-  
+
+
```
mmusgrov, line 716: If you are changing white space then why not the whole of the extra line?

PULL_REQUEST_TEMPLATE.md
mmusgrov: I am marking this as resolved since Manuel's changes have addressed both of our (mine and Ondra) comments

scripts/hudson/narayana.sh
jmfinelli: This is my IDE doing the work for me. I'll remove the extra line manually. Thanks for spotting this Michael

scripts/hudson/narayana.sh
jmfinelli: +1

scripts/hudson/narayana.sh
mmusgrov: resolving

rts/lra/test/pom.xml
jmfinelli: IIRC, `lra.coordinator.war.name` is used by Maven Wildfly Plugin to name the deployed WAR

scripts/hudson/narayana.sh
mmusgrov: You just replaced upper case with lower case, but okay thanks.

rts/lra/test/README.adoc
mmusgrov: Which module is being referred to? Thorntail is still one of the `=== Application runtimes` even though it used to host a different application?

rts/lra/test/README.adoc
mmusgrov: The question is not only for my benefit, it is for the benefit of any community member that wants to run our test suite.

rts/lra/test/README.adoc
mmusgrov: The question is not only for my benefit, it is for the benefit of any community member that wants to run our test suite.

rts/lra/test/README.adoc
mmusgrov: The question is not only for my benefit, it is for the benefit of any community member that wants to run our test suite.

rts/lra/test/README.adoc
mmusgrov: The question is not only for my benefit, it is for the benefit of any community member that wants to run our test suite.

rts/lra/test/README.adoc
mmusgrov: I'm still not clear about what `lra-participant` and `lra-coordinator` mean.

mmusgrov: Are they module names or artifacts names or property names, something else?

rts/lra/test/README.adoc
jmfinelli: I get it and I agree with you; it does not make much sense to highlight `local` if I do not explain why

rts/lra/test/README.adoc
jmfinelli: > The model concerns participants (aka compensators), a logical coordinator and a client. A client starts a new LRA via an annotation which results in the creation of a new LRA. When a business service does work that may have to be later compensated it enlists a participant with the LRA which, when the client later closes or cancels the LRA, operates on behalf of the service to undo the work the service performed within the scope of an LRA or to compensate for the fact that the original work could not be completed.

from [here](https://github.com/eclipse/microprofile-lra/blob/master/spec/src/main/asciidoc/microprofile-lra-spec.adoc)

From the LRA spec, I understood that `lra-coordinator` is the logical coordinator, `lra-participant` is the compensator (aka the application implementing the LRA APIs) and the client is the service that makes the calls. Please, let me know if I totally misunderstood this

rts/lra/test/README.adoc
jmfinelli: I take it as a yes :-) I'll add this to the README. Thanks Michael

rts/lra/test/README.adoc
jmfinelli: I will rename `arq.thorntail` to `arq.quarkus`. What about the new JIRA for ARQ_PROF=arqIFv6? I think it would be good to have a discussion about this in a separate JIRA/PR, WDYT @mmusgrov?

rts/lra/test/README.adoc
jmfinelli: I will restore the reference to Thorntail then. Thanks Michael :-)

scripts/hudson/narayana.sh
```diff
@@ -521,10 +522,17 @@ function rts_tests {
 function lra_tests {
   echo "#0. LRA Test"
   ./build.sh install -pl ArjunaCore/arjunacore,rts -am -DskipTests -Pcommunity "$@"
-  # we can't use 'mvn -f' option beacuse of Thorntail plugin issue THORN-2049
+  # we can't use 'mvn -f' option beacuse of Thorntail plugin issue QUARKUS-2049
```
ochaloup, line 525: I think this change will be wrong as the QUARKUS issues are under github and the plugin would not have the same issue under the same number.

rts/lra/test/README.adoc
ochaloup: Yes, what I remember we are not using the ipv6 profile for LRA testing. To me the ipv6 question would be good to be opened as a new issue.

scripts/hudson/narayana.sh
```diff
@@ -521,10 +522,17 @@ function rts_tests {
...
+  echo "#0. Running LRA tests using $ARQ_PROF profile"
   PRESERVE_WORKING_DIR=true ../../build.sh -fae -B -Pcommunity -P$ARQ_PROF $CODE_COVERAGE_ARGS $ENABLE_LRA_TRACE_LOGS -Dlra.test.timeout.factor="${LRA_TEST_TIMEOUT_FACTOR:-1.5}" "$@"
-  [ $? = 0 ] || fatal "LRA Test failed"
+  lra_arq=$?
+  [ $lra_arq = 0 ] || echo "LRA tests using $ARQ_PROF profile failed"
+  echo "#0. Running LRA tests using $ARQ_QUARKUS profile"
+  PRESERVE_WORKING_DIR=true ../../build.sh -fae -B -Pcommunity -P$ARQ_QUARKUS $CODE_COVERAGE_ARGS $ENABLE_LRA_TRACE_LOGS -Dlra.test.timeout.factor="${LRA_TEST_TIMEOUT_FACTOR:-1.5}" "$@"
+  lra_thorn=$?
+  [ $lra_thorn = 0 ] || echo "LRA tests using $ARQ_QUARKUS profile failed"
+  if [ $lra_arq = 0 ] || [ $lra_thorn = 0 ] ; then fatal "LRA Test failed" ;  fi
```
ochaloup, line 535: Please, do you think it would be possible to add to the `fatal` message what is the "profile" which failed - i.e., if it was the wfly or quarkus?

ochaloup (commented):
I have just two minor comments. To me it seems nice.

scripts/hudson/narayana.sh
ochaloup, line 535: Oh, one more thing. I think this condition is wrong. The `0` means success and not `fatal`.

ochaloup (changes requested):
The condition should be fixed.

rts/lra/test/README.adoc
jmfinelli: +1 Thanks Ondra

scripts/hudson/narayana.sh
jmfinelli: Thanks for spotting this Ondra. A huge oversight probably due to "find and replace".

scripts/hudson/narayana.sh
mmusgrov: You can change the IDE settings to not update files without your knowledge. It also makes the reviewers job that little bit easier (since it becomes clear what the functional changes are). We have said in the past that reformatting changes should go into a separate commit (but I generally ignore them if there aren't many since a commit with just reformatting makes the git commit history noisy; there's pros and cons).

scripts/hudson/narayana.sh
jmfinelli: I echo which profile fails at line 531 and line 534, do you think we should also repeat that on 'fatal'?

rts/lra/test/pom.xml
mmusgrov: I am referring to line 164.

scripts/hudson/narayana.sh
jmfinelli: yeah, just realised that...already changed and ready to be committed. Thanks Ondra

rts/lra/test/pom.xml
jmfinelli: `<lra.coordinator.war.name>lra-coordinator.war</lra.coordinator.war.name>` is used at line [280](https://github.com/jbosstm/narayana/pull/1881/files#diff-89636f088428917c8cfb617b2ba8af009f37ddbc65281cd6d63651c861853a45R280) of the same file. I use a property just in case this name should be used somewhere else

rts/lra/test/README.adoc
mmusgrov: Yes I think you have. `lra-coordinator` and `lra-participant` look like a variable names or artifacts etc, i.e. something that a computer would parse. The specification is written using words that mostly can be looked up in a dictionary so something like LRA coordinator or LRA participant is probably what you meant.

rts/lra/test/README.adoc
mmusgrov: It's very complicated but fair enough, it's okay to resolve this conversation now.

rts/lra/test/README.adoc
jmfinelli: Thanks Michael, I am gonna clarify the wording so that the community does not get confused

rts/lra/test/pom.xml
mmusgrov: I am referring to the line on which I added my comment (and my comment said `Note I am reviewing this line because the last three commits you made are all about the rewrite of our approach to testing LRA. to clarify why I was picking on that line`), namely

``` xml
<lra.coordinator.xml.filename>lra-coordinator.xml</lra.coordinator.xml.filename>
```

rts/lra/test/README.adoc
mmusgrov: Thanks

rts/lra/test/pom.xml
jmfinelli: > ... namely

Thanks for clarifying this.

```<lra.coordinator.xml.filename>lra-coordinator.xml</lra.coordinator.xml.filename>```
basically, the filename indicated here is used by one of the instances of Wildfly, specifically, the one that hosts the LRA coordinator. Both the Maven WildFly Plugin and Arquillian use this file to load the WildFly configuration needed. Another identical file is used for LRA participants (IIRC the filename [`lra-participant.xml`] is defined in the property `lra.participant.xml.filename`)

File: scripts/hudson/narayana.sh (hunk 17 above)
Comment (ochaloup) [reply scripts/hudson/narayana.sh:535]: Yes, I think so. It's because the `fatal` goes to github comment.

Comment (jmfinelli) [reply scripts/hudson/narayana.sh]: NP, ready to be committed :-) Thanks Ondra

File: rts/lra/test/pom.xml (hunk 13 above)
Comment (mmusgrov) [reply rts/lra/test/pom.xml:165]: > What is this file for, it is not referenced or documented anywhere and we don't use it in any of our builds.
> Note I am reviewing this line because the last three commits you made are all about the rewrite of our approach to testing LRA.

I cannot find any references to or documentation about this file.

Comment (jmfinelli) [reply rts/lra/test/pom.xml:165]: This is not a file we provide. This file is copied from `${env.JBOSS_HOME}/standalone/configuration/standalone.xml` and renamed to `lra-coordinator.xml`. We need two standalone XML files to start two separated instances of WildFly (AFAIK). `lra-coordinator.xml` and `lra-participant.xml` are created at lines 239-240 of `rts/lra/test/pom.xml`.

File: rts/lra/test/README.adoc (hunk 12 above)
Comment (mmusgrov) [reply rts/lra/test/README.adoc]: But the profile starts thorntail not quarkus.

Comment (jmfinelli) [reply rts/lra/test/README.adoc]: This profile starts Quarkus (using a maven plugin) AND Thorntail (using Arquillian) but the "main" runtime is Quarkus as `lra-coordinator` runs in it.

**Comment by mmusgrov** [comment]
@jmfinelli There have been a lot of conversations in this PR so to summarise my outstanding queries:

1. Clarification of what `lra-participant.xml` is used for (and just as important, what it should contain).
2. The profile `<id>arq.quarkus</id>` starts thorntail and has nothing to do with quarkus.
3. Replace `lra-partipant(s)` with a meaningful expression (such as LRA participant). And I couldn't resist but, is it referring to the pants one should wear to a party ;-)

**Comment by jmfinelli** [comment]
> 1. Clarification of what `lra-participant.xml` is used for (and just as important, what it should contain).
> There is `lra-coordinator.xml` as well, what is that used for and what should it contain. It too is not used anywhere.

These are not files we provide. These files are copied from `${env.JBOSS_HOME}/standalone/configuration/standalone.xml` and renamed to `lra-coordinator.xml` and `lra-participant.xml`. We need two standalone XML files to start two separated instances of WildFly (AFAIK). `lra-coordinator.xml` and `lra-participant.xml` are created at lines 239-240 of `rts/lra/test/pom.xml`.

I could add a comment just to explain that we copy both files from `standalone.xml` (and point the user to lines 239-240). I have to say that it sounds to me like repeating what is already clearly documented in the WildFly community.

> 2. The profile `<id>arq.quarkus</id>` starts thorntail and has nothing to do with quarkus.

This profile starts Quarkus. Then, Arquillian starts a Thonrtail container to host a `lra-participant`. Moreover, very soon we are gonna replace Thorntail with Quarkus/Wildfly so I think that `arq.quarkus` should be the right name to go for.

> 3. Replace `lra-partipant(s)` with a meaningful expression (such as LRA participant). And I couldn't resist but, is it referring to the pants one should wear to a party ;-)

If you mean replacing `lra-participant` with `LRA participant` in the `README.md` file, I have already done so. If you mean somewhere else, could you please point me where?

..."party pants"...now I understand where you guys got the idea when working on the LRA spec XD

**Comment by mmusgrov** [comment]
> > 1. Clarification of what `lra-participant.xml` is used for (and just as important, what it should contain).
> >    There is `lra-coordinator.xml` as well, what is that used for and what should it contain. It too is not used anywhere.
>
> These are not files we provide. These files are copied from `${env.JBOSS_HOME}/standalone/configuration/standalone.xml` and renamed to `lra-coordinator.xml` and `lra-participant.xml`. We need two standalone XML files to start two separated instances of WildFly (AFAIK). `lra-coordinator.xml` and `lra-participant.xml` are created at lines 239-240 of `rts/lra/test/pom.xml`.

Well you need to document somewhere that that is what you are doing, how is a user supposed to know?

>
> I could add a comment just to explain that we copy both files from `standalone.xml` (and point the user to lines 239-240). I have to say that it sounds to me like repeating what is already clearly documented in the WildFly community.

There has never been WildFly config file for narayana lra so why would an occasional user understand that that's what you are doing. And if the standalone.xml file that you have copied is unmodified then why would you copy it?

>
> > 1. The profile `<id>arq.quarkus</id>` starts thorntail and has nothing to do with quarkus.
>
> This profile starts Quarkus. Then, Arquillian starts a Thonrtail container to host a `lra-participant`. Moreover, very soon we are gonna replace Thorntail with Quarkus/Wildfly so I think that `arq.quarkus` should be the right name to go for.

If you take a look at the profile you will observe that there are no properties referring to or dependent on quarkus.
And if there is some magic involved in setting the property `lra.coordinator.exec.plugin.phase` then you would need to explain what's going on.

>
> > 1. Replace `lra-partipant(s)` with a meaningful expression (such as LRA participant). And I couldn't resist but, is it referring to the pants one should wear to a party ;-)
>
> If you mean replacing `lra-participant` with `LRA participant` in the `README.md` file, I have already done so. If you mean somewhere else, could you please point me where?
>
> ..."party pants"...now I understand where you guys got the idea when working on the LRA spec XD

It is still there. Why do I feel as if I am having to keep repeating my comments (each recheck of your PR after you push updates is time consuming).

**Comment by jmfinelli** [comment]
> Well you need to document somewhere that that is what you are doing, how is a user supposed to know?

done

> And if the standalone.xml file that you have copied is unmodified then why would you copy it?

We need two standalone XML files to start two separated instances of WildFly (AFAIK)

> And if there is some magic involved in setting the property lra.coordinator.exec.plugin.phase then you would need to explain what's going on.

This has been introduced in previous PRs. This PR is mainly to test LRA with Wildfly

**Comment by mmusgrov** [comment]
> > Well you need to document somewhere that that is what you are doing, how is a user supposed to know?
>
> done

Thanks

>
> > And if the standalone.xml file that you have copied is unmodified then why would you copy it?
>
> We need two standalone XML files to start two separated instances of WildFly (AFAIK)

But the files you have copied are unmodified so why do you need to copythem. Does using `"${env.JBOSS_HOME}/standalone/configuration/standalone.xml"` not work?

>
> > And if there is some magic involved in setting the property lra.coordinator.exec.plugin.phase then you would need to explain what's going on.
>
> This has been introduced in previous PRs. This PR is mainly to test LRA with Wildfly

This response is taken out of context. My actual comment was:
[quote]
```
If you take a look at the profile you will observe that there are no properties referring to or dependent on quarkus.
```

jmfinelli:
> But the files you have copied are unmodified so why do you need to copythem. Does using "${env.JBOSS_HOME}/standalone/configuration/standalone.xml" not work?

Using only one `standalone.xml` file means that both servers will be configured according with that file. For example, if an application is deployed on the first server, and a second application is deployed on the second one, the latter will magically deploy also the application deployed on the first server.

> If you take a look at the profile you will observe that there are no properties referring to or dependent on quarkus.

I will rollback to `arq.thorntail`

mmusgrov (approved):
My approval is continent on addressing Ondra's review comments and fixing/explaining the rebase failure against AS_BRANCH.

ochaloup (approved):
Thanks @jmfinelli

PR #1884, opened 2021-07-22
JBTM-3517 Collection.isEmpty() should be used to test for emptiness

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/JBTM-3517

PERF !MAIN !CORE !RTS !LRA !TOMCAT !AS_TESTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
@@ -870,7 +870,7 @@ public class StateManager
             
             synchronized (modifyingActions)
             {
-                if ((modifyingActions.size() > 0)
+                if ((!modifyingActions.isEmpty())
                         && (modifyingActions.get(action.get_uid()) != null))
                 {
                     return true;
@@ -962,7 +962,7 @@ public class StateManager
         {
 	    createLists();
 
-            if (usingActions.size() > 0)
+            if (!usingActions.isEmpty())
             {
                 Enumeration e = usingActions.keys();
 
@@ -1229,7 +1229,7 @@ public class StateManager
                 {
                     usingActions.remove(action.get_uid());
 
-                    if (usingActions.size() == 0)
+                    if (usingActions.isEmpty())
                     {
                         if (committed)
                         {
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TwoPhaseCoordinator.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TwoPhaseCoordinator.java
@@ -214,7 +214,7 @@ public class TwoPhaseCoordinator extends BasicAction implements Reapable
             do {
                 synchronized (_syncLock) {
                     if (processed == runningSynchronizations.size()) {
-                        if (executingInterposedSynchs || interposedSynchs.size() == 0)
+                        if (executingInterposedSynchs || interposedSynchs.isEmpty())
                             break; // all synchronizations have been executed
 
                         // all non interposed synchronizations have been executed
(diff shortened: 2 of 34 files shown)


ochaloup (changes requested):
What about the places where `Collection.size()` is used to compare of being greater than `0` (i.e., non-empty)?
Like https://github.com/jbosstm/narayana/blob/master/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/CommitMarkableResourceRecordRecoveryModule.java#L621

RanabirChakraborty:
@ochaloup at the beginning I also thought to change `> 0` but then I don't think that will be appropriate because there we are not checking non-empty, we are checking non-empty as well as the values greater than 0. what if is some cases we get negative values? and that time if we use `!Collection.isEmpty()` it will get passed.

Let me know If I need to change those too.

mayankkunwar:
@RanabirChakraborty I agree with Ondra's comment. I can see that same is applied in many places in the PRs that were merged as part of the WFLY-14972 issue that is linked in the JBTM-3517. You can see the example here https://github.com/wildfly/wildfly/pull/14437/files#diff-52c607982ae5169a234ebdaf6a9533bb85bfcc2677316c1a57ec4d1c5b44fdd8R502
Maybe the Hash issue that you mentioned is resolved now

RanabirChakraborty:
@ochaloup @mayankkunwar Thanks for the review. Done the changes :)

ochaloup:
@RanabirChakraborty ok, I see but you know the `size()` is expected to return size. By definition the cardinality of a set cannot be negative. The code the `size() > 0` considers it. Still it could be good to check one by one the places where you are appling the refactoring and discuss here if you think the code considers that statement differently.

The fact that JDK `HashSet` `size()` call can return negative number in JDK 6 is an error in JDK not an error in statement what `size of the set` means. In fact when we change for `!isEmpty()` we could be fixing that trouble by not using `size()` explicitly (while I'm not sure how `isEmpty` is implemented in HashSet, while I think we haven't been supporting JDK 6 in last few years already).

To me this PR should be updated.

RanabirChakraborty:
@ochaloup thanks! I also feel the same even I have checked each place while using `!isEmpty()` and it looks fine to me. I have modified the PR a few days back, you can check it now.

ochaloup (approved):
Thanks @RanabirChakraborty/ To me it seems fine.
The tests passed, only the perf test objects. The perf tests are usually a bit unstable. It seems to be fine for merging by me.

For merging it would be good to wait for review from @mayankkunwar (he would know better about perf test results as well).

mmusgrov:
> Thanks @RanabirChakraborty .
> The tests passed, only the perf test objects. The perf tests are usually a bit unstable. It seems to be fine for merging by me.
> 
The perf tests are very unstable but I don't see the harm in rerunning just that axis to be sure.

mmusgrov:
testit

Except the the "build requires parameters PR trigger" does not include an option of the PERF axis.

mmusgrov:
@ochaloup do you know how to trigger a re-test of the PERF axis, I want to see if the failure is consistent?

ochaloup:
@mmusgrov unfortunately the AMS CI is behind the VPN and CI polls on checking if github repo changed. It is not capable to find a new comment with some content was added. This "issue" could be fixed when @jmfinelli process with his changes to allow webhooks being passed under VPN via proxy.

Currently retesting needs to be manually launched withing the narayana-pull job on AMS CI. I did so now.
With that it seems the pull job defined a wrong name of the axis instead what narayana script uses (https://github.com/jbosstm/narayana/blob/master/scripts/hudson/narayana.sh#L158). It should be fixed now.

mmusgrov:
This is a false positive since there appears to be a bug in the benchmark code:

```
./scripts/run_bm.sh: line 66: mvn: command not found
Error: Unable to access jarfile narayana/ArjunaCore/arjuna/target/benchmarks.jar
cp: cannot stat 'narayana/ArjunaCore/arjuna/target/jmh/com.hp.mwtests.ts.arjuna.performance.Performance1-master.csv': No such file or directory
```

RanabirChakraborty:
Sorry I couldn't understand what needed to be done for the perf test, if anything needs from my side please let me know in detail. Thank you!

mayankkunwar (approved):
I think we can merge the changes

PR #1891, opened 2021-08-23
[JBTM-3523] print warning with cause when AfterLRA call does not proceed well

ochaloup:
LRA

https://issues.redhat.com/browse/JBTM-3523

I would like to propose a change in the logging of the error during after lra call from coordinator. Currently it's hidden under `debug` without the whole stacktrace. When an error happens then it's "invisible" without details. When the call goes wrong then it leads to an unrecoverable lra records (`afterLRARequest` returns `false` and the `TwoPhaseOutcome.HEURISTIC_HAZARD` is returned, https://github.com/jbosstm/narayana/blob/b4bd01f695af5fb92bed70fe1502bb606b305579/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRAParticipantRecord.java#L512) while the reason for the error is difficult to find.

@mmusgrov @mayankkunwar what do you think about his kind of change? If it's  fine with you I will create a JBTM issue and will remove the `Hold` label from this PR.

### DIFF
--- a/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRAParticipantRecord.java
+++ b/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRAParticipantRecord.java
@@ -448,13 +448,7 @@ public class LRAParticipantRecord extends AbstractRecord implements Comparable<A
                 return true;
             }
         } catch (Exception e) {
-            Throwable cause = e.getCause();
-            if (!(e instanceof WebApplicationException) && cause instanceof WebApplicationException) {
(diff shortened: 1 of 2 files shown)


mayankkunwar:
Hi @ochaloup 
I think this change would be helpful.

mayankkunwar (approved):

mmusgrov (approved):
I added a comment to the JIRA:
https://issues.redhat.com/browse/JBTM-3523?focusedCommentId=18868069&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-18868069

PR #1896, opened 2021-08-31
JBTM-3529 Upgrade the activemq artemis dependency

mmusgrov:
https://issues.redhat.com/browse/JBTM-3529

MAIN CORE !TOMCAT AS_TESTS !RTS !JACOCO !XTS QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !PERF !LRA NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle


ochaloup:
I would think we should merge this only after WildFly 25 is released. If I understand the WildFly 25 will be released with version 2.16.0 and we are still chasing a chance to get there the LRA integration.
If we did it we need to have a new Narayana release within the WFLY 25 and this change would be incompatible with the current WildFly integration.

mmusgrov:
Agreed, I included that in the JIRA description.
I created the PR early so that it wouldn't drop off my radar (I will put it on Hold once the CI run has finished).
@jmfinelli is the PR poller still operational?

mmusgrov:
The CI tests passed so I will put this on hold.

@ochaloup Was your comment a review, if not can you review it so that we are ready to go as soon as WFLY master moves to 2.18.0

ochaloup (approved):

jmfinelli:
@mmusgrov the poller is operational in our AMS CI. Sometimes it just does not work

mmusgrov:
> @mmusgrov the poller is operational in our AMS CI. Sometimes it just does not work

You're right there is a window where it gets missed. Apologies for the ping.

PR #1897, opened 2021-09-01
[JBTM-3513] Temporary workaround to fix JBTM-3513

jmfinelli:
This PR *TEMPORARY* addresses [JBTM-3513](https://issues.redhat.com/browse/JBTM-3513)

MAIN CORE !TOMCAT !AS_TESTS RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !PERF LRA NO_WIN !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
+++ b/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
@@ -54,16 +54,16 @@ import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.Executors;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantLock;
 
 public class LongRunningAction extends BasicAction {
     private static final String LRA_TYPE = "/StateManager/BasicAction/LongRunningAction";
-    private final ScheduledExecutorService scheduler;
+    private static final ScheduledExecutorService scheduler = new ScheduledThreadPoolExecutor(10);
     private URI id;
     private URI parentId;
     private String clientId;
@@ -90,8 +90,6 @@ public class LongRunningAction extends BasicAction {
         this.clientId = clientId;
         this.finishTime = null;
         this.status = LRAStatus.Active;
-
-        this.scheduler = Executors.newScheduledThreadPool(1);
     }
 
     public LongRunningAction(LRAService lraService, Uid rcvUid) {
@@ -103,7 +101,6 @@ public class LongRunningAction extends BasicAction {
         this.clientId = null;
         this.finishTime = null;
         this.status = LRAStatus.Active;
-        this.scheduler = Executors.newScheduledThreadPool(1);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


mmusgrov (changes requested):
I don't think that the extra complexity is necessary, just run the `abortLRA` method when the existing scheduler fires:

``` bash
git diff ../../lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java

@@ -939,7 +939,7 @@ public class LongRunningAction extends BasicAction {
                     }
 
                     updateState(LRAStatus.Cancelling);
-                    CompletableFuture.supplyAsync(this::cancelLRA); // use a future to avoid hogging the ScheduledExecutorService
+                    cancelLRA();
                 }
             } finally {
                 lock.unlock();
```

@jmfinelli Note also that when scheduling the abort task the possible RejectedExecutionException should be okay to leave uncaught (apart from logging a warning) but we should both check that allowing this RuntimeException results in the LRA being cancelled: if the timeout was set on the LRA then the request to start a new LRA will just fail (which is fine since the business logic is meant to handle that) and if it is set on the participant enlistment then it should be caught by the JAX-RS request filter (ServerLRAFilter.java) and the participant enlistment will fail and the LRA will be automatically cancelled (unless the user has modified the LRA.CancelOn annotation attribute). Since the existing code does not catch RejectedExecutionException this paragraph is effectively a different issue. You can either include the fix (catch, log a warning and rethrow) in this PR or you or I can raise another PR to address the correct handling of RejectedExecutionException.

ochaloup (approved):
Apart from what is said in comment above I like the PR and way how the processing is enhanced.

xstefank (approved):

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
```diff
@@ -54,16 +54,16 @@
...
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantLock;
 
 public class LongRunningAction extends BasicAction {
     private static final String LRA_TYPE = "/StateManager/BasicAction/LongRunningAction";
-    private final ScheduledExecutorService scheduler;
+    private static final ScheduledExecutorService scheduler = new ScheduledThreadPoolExecutor(10);
```
mmusgrov, line 66: Although 10 seems low, the scheduler only runs for LRAs that time out at the same time. In a typical user application I'd i) expect most LRAs to be explicitly finished, and (ii) it's unlikely that lot's of LRAs will time out at similar deadlines.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
```diff
@@ -919,7 +916,15 @@ private int scheduleCancelation(Runnable runnable, Long timeLimit) {
             finishTime = LocalDateTime.now(ZoneOffset.UTC).plusNanos(timeLimit * 1000000);
         }
 
-        scheduledAbort = scheduler.schedule(runnable, timeLimit, TimeUnit.MILLISECONDS);
+        try {
+            scheduledAbort = scheduler.schedule(runnable, timeLimit, TimeUnit.MILLISECONDS);
+        } catch (RejectedExecutionException executionException) {
+            updateState(LRAStatus.Cancelling);
+            LRALogger.logger.warnf(
```
mmusgrov, line 923: Logging the exception message is vital for diagnosis of the underlying cause. Note that I recently fixed JBTM-3525 (include exception messages in log output) and would not want to reopen it ;-)

Note that the failure is not fatal since, when periodic recovery runs it will eventually discover that it is eligible for cancellation. It might be worthwhile adding that comment in the catch block for future maintainers of this code.

mmusgrov (changes requested):
The log message needs to include the exception message.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
jmfinelli: +1 I will update the PR. Thanks

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
mmusgrov: The number of available `hardware threads` is restricted to the number of cores times the hyperthreading.

BTW the size of the pool relates to software threads and there can be hundreds of those (they are managed by the kernel and they also increase parallelism - eg if the thread is executing I/O bound code then there will be a context switch so that other code can be executed.

mmusgrov (approved):
I am approving this one but note that I am preparing a PR to use the transaction reaper to perform cancellations.

PR #1899, opened 2021-09-02
[JBTM-3530] no Weld classloading error logs at lra/basic FailedLRAIT

ochaloup:
https://issues.redhat.com/browse/JBTM-3530

LRA

### DIFF
--- a/rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/Deployer.java
+++ b/rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/Deployer.java
@@ -44,6 +44,9 @@ public class Deployer {
                         "io.narayana.lra.provider",
                         "io.narayana.lra.client",
                         "io.narayana.lra.arquillian.spi")
+                // adds the TestBase class, the test class itself seems to be uploaded during deployment by Arquillian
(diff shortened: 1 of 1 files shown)


mmusgrov (approved):

jmfinelli:
Unfortunately, there is not a way to tell Arquillian to avoid appending the test class to the deployed archive (even though the class is annotated with `@RunAsClient`). Yeah, agree with you, appending the `TestBase` is the only solution to get rid of the WARN. Thanks Ondra for fixing this

jmfinelli (approved):

mayankkunwar (approved):

xstefank (approved):
you can try also `testable = false` in `@Deployment` annotation but I don't see big issues with this solution.

jmfinelli:
@xstefank Tried it already but no luck. It seems that Arquillian appends the test class a priori

ochaloup:
Thanks @jmfinelli for the elaboration and @xstefank for the tip.
I tried with the `testable` and looking about the `RunAsClient` and I don't understand why the testcase is loaded. Maybe because of `@Before` or `@BeforeClass` methods?
Anyway, this fix seems to me the easiest one as Manuel said.

I'm going to merge this.

PR #1902, opened 2021-09-07
JBTM-3533 use the reaper for cancellations

mmusgrov:
https://issues.redhat.com/browse/JBTM-3533

!MAIN !CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !PERF LRA !NO_WIN DB_TESTS !mysql !db2 !postgres !oracle


mmusgrov:
@jmfinelli Mayank and I have seen this test failure before and it looks unrelated to my change, although it's suspicious that it failed twice in a row. 

I ran 25 successful runs on my laptop:

`for i in {1..25}; do mvn clean verify -Dit.test=FailedLRAIT#testCompleteFailed -Parq; done`

I will continue investigating (and raise a issue for it) but personally I think my PR is in a fit state for review, especially since the last run passed and since the timeout tests passed on all CI runs.

jmfinelli (changes requested):
Only a small change. Other than that, this PR LGTM.

jmfinelli:
> @jmfinelli Mayank and I have seen this test failure before and it looks unrelated to my change, although it's suspicious that it failed twice in a row.
> 
> I ran 25 successful runs on my laptop:
> 
> `for i in {1..25}; do mvn clean verify -Dit.test=FailedLRAIT#testCompleteFailed -Parq; done`
> 
> I will continue investigating (and raise a issue for it) but personally I think my PR is in a fit state for review, especially since the last run passed and since the timeout tests passed on all CI runs.

I also ran FailedLRAIT#testCompleteFailed many times on my laptop and once on the CSB CI and both machines are happy with it.

ochaloup:
ad `FailedLRAIT#testCompleteFailed`) I could see the same failure last week when working on a different PRs. Those were unrelated to the code changes (e.g. changes in `pom.xml`). It seems this test started to intermittently fail but I don't know when that started. From that I believe this particular failure is not related to the code changes here.

rts/lra/test/arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
```diff
@@ -57,7 +57,14 @@
 
     @Override
     public void waitForCallbacks(URI lraId) {
-        // no action needed
+        // for tests that do not involve timeouts no action is needed,
+        // but there are races on time sensitive tests (depending upon how busy the test system is).
+        // To handle these cases introduce an arbitrary 1000 ms delay (note that since LRA provides
+        // no real-time guarantees, introducing the delay is a valid approach):
+        try {
+            Thread.sleep(1000);
```
ochaloup, line 65: I read the explanation https://issues.redhat.com/browse/JBTM-3513?focusedCommentId=18965208#comment-18965208 and I do agree with that. But I don't think it's good to add this arbitrary long sleep time here.
For other environment the `1000` can be not enough, on many other it means to slow down the test execution when not needed.

I was trying to address the same issue with adding the annotation extension which consider `lra.test.timeout.factor` and applies it to the `@LRA` annotation definition.
https://github.com/jbosstm/narayana/blob/master/rts/lra/test/arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/LRAAnnotationAdjuster.java#L166
-> https://github.com/jbosstm/narayana/blob/master/rts/lra/test/tck/pom.xml#L54

I hope that's the reason why we don't experience the `timeLimit` failure at CI runs.

Additional note:
I can see that a trouble could be with `-Parq` that the timeout is not passed to the WildFly runtime (https://github.com/jbosstm/narayana/blob/master/rts/lra/test/tck/src/test/resources/arquillian-wildfly.xml).

To me this adjustment could not be needed, maybe.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
```diff
@@ -1029,4 +1005,30 @@ void forget() {
...
+    @Override
+    public boolean running() {
+        return status.equals(LRAStatus.Active);
+    }
+
+    @Override
+    public int cancel() {
+        abortLRA();
+
+        if (status.equals(LRAStatus.Cancelled)) {
```
ochaloup, line 1018: Just a question, is it correct to verify the LRA status only against the `Cancelled`? It seems that other LRA statuses could not be properly converted to `ActionStatus` codes.

mmusgrov: If the LRA is not in the `Cancelled` state we return the ArjunaCore ActionStatus to the reaper (it is on the next line 2022):

```
return status();
```

rts/lra/test/arquillian-extension/src/main/java/io/narayana/lra/arquillian/spi/NarayanaLRARecovery.java
mmusgrov: @ochaloup 
> To me this adjustment could not be needed, maybe.

The comment that you referenced (the one in JBTM-3513) already stated why I didn't use `lra.test.timeout.factor`.

mmusgrov:
@ochaloup and @jmfinelli  the latest CI run passed and the previous CI runs we argued were unrelated and I have addressed each of the reviewer comments, ie I'm not seeing any persuasive arguments to linger excessively over this PR.

jmfinelli (approved):

mmusgrov:
@ochaloup I have answered all of the question on https://github.com/jbosstm/narayana/pull/1902 and I've worked with the primary reviewer and I see that you are quite busy on other things so I am going to merge this PR and if you find that you  find new or still have issues with it then can you either re-open the JIRA or create a new one.

PR #1905, opened 2021-09-13
[JBTM-3537] using CDI.current().select() to get LRAParticipantRegistry

ochaloup:
as a workaround for WildFly integration (WFLY-14869)

https://issues.redhat.com/browse/JBTM-3537

LRA

### DIFF
--- a/rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
+++ b/rts/lra/jaxrs/src/main/java/io/narayana/lra/filter/ServerLRAFilter.java
@@ -35,6 +35,7 @@ import org.eclipse.microprofile.lra.annotation.Status;
 import org.eclipse.microprofile.lra.annotation.ws.rs.LRA;
 import org.eclipse.microprofile.lra.annotation.ws.rs.Leave;
 
+import javax.enterprise.inject.spi.CDI;
 import javax.inject.Inject;
 import javax.ws.rs.NotFoundException;
 import javax.ws.rs.ProcessingException;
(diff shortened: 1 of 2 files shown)


ochaloup:
@xstefank can you please take a look?

xstefank (changes requested):
@ochaloup please create a follow-up JBTM and add comments to both places.

mmusgrov (approved):

mmusgrov:
@ochaloup What are your plans for this PR, what will it take to remove the `Hold` label?

ochaloup:
@mmusgrov I'm currently planning to decline it without merging. After a discussion with Matej from Weld/Arc I think I found the way to make `@Inject` working.
https://github.com/ochaloup/wildfly/commit/0a03438a08e5239e52d3a09c33d5c087346fcd55

Just need to verify with other TCK. The trouble is that the TCK is unstable and I'm not sure if there is some connection to CDI injection.
I'm currently investigating on the context test errors.

mmusgrov:
> @mmusgrov I'm currently planning to decline it without merging. After a discussion with Matej from Weld/Arc I think I found the way to make `@Inject` working.
> [ochaloup/wildfly@0a03438](https://github.com/ochaloup/wildfly/commit/0a03438a08e5239e52d3a09c33d5c087346fcd55)
> 

that's good news

> Just need to verify with other TCK. The trouble is that the TCK is unstable and I'm not sure if there is some connection to CDI injection.

It is more to do with the test picking up the wrong LRA id, ie it's a straight bug in how the implementation is managing the LRA context.

> I'm currently investigating on the context test errors.

If you find anything of interest do keep me informed, even if it is just a suspicion. My main focus is on how we are cleaning up Current and the HTTP context header when the LRA is terminated.

Your point about the response filter and the business logic not being guaranteed to be running in the same thread I think could well be the root cause since that implementation assumes they use the same thread.

ochaloup:
@mmusgrov +1, I will close this PR.

The point that `response filter and the business logic not being guaranteed to be running in the same thread` was investigated by @xstefank actually.

PR #1907, opened 2021-09-15
JBTM 3538 Fix for TCK timeout test failures

mmusgrov:
https://issues.redhat.com/browse/JBTM-3538

This PR is designed to fix timeout test instabilities.
It is in 4 separate, but related, commits. It also reverts the recent change to use the reaper for timeouts since that does not work with WildFly (we can revisit that in a future commit after discussion with the resteasy team).
After fixing the timeout instability I found that 3 of the nested LRA tests failed which I fixed in the last commit.

LRA

### DIFF
--- a/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRAParticipantRecord.java
+++ b/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRAParticipantRecord.java
@@ -53,6 +53,8 @@ import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URLDecoder;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
 import java.util.Arrays;
 import java.util.Objects;
 import java.util.SortedMap;
@@ -114,11 +116,13 @@ public class LRAParticipantRecord extends AbstractRecord implements Comparable<A
                 if (parseException[0] != null) {
                     String errorMsg = LRALogger.i18nLogger.error_invalidCompensator(lra.getId(), parseException[0].getMessage(), linkURI);
                     LRALogger.logger.error(errorMsg);
+                    trace_progress(errorMsg);
                     throw new WebApplicationException(errorMsg, parseException[0],
                             Response.status(BAD_REQUEST).entity(errorMsg).build());
                 } else if (compensateURI == null && afterURI == null) {
                     String errorMsg = LRALogger.i18nLogger.error_missingCompensator(lra.getId(), linkURI);
                     LRALogger.logger.error(errorMsg);
+                    trace_progress(errorMsg);
                     throw new WebApplicationException(errorMsg, Response.status(BAD_REQUEST).entity(errorMsg).build());
                 }
             } else {
@@ -138,9 +142,13 @@ public class LRAParticipantRecord extends AbstractRecord implements Comparable<A
 
             this.recoveryURI = null;
             this.compensatorData = compensatorData;
+
+            trace_progress("created");
         } catch (URISyntaxException e) {
             String logMsg = LRALogger.i18nLogger.error_invalidFormatToCreateLRARecord(lraId.toASCIIString(), linkURI, e.getMessage());
             LRALogger.logger.error(logMsg);
+            trace_progress(logMsg);
+
             throw new WebApplicationException(logMsg, e,
                     Response.status(BAD_REQUEST).entity(logMsg).build());
         }
@@ -274,6 +282,8 @@ public class LRAParticipantRecord extends AbstractRecord implements Comparable<A
         URI endPath;
         Client client = null;
 
+        trace_progress("finishing");
+
         if (isFinished()) {
             return atEnd(status == ParticipantStatus.FailedToComplete || status == ParticipantStatus.FailedToCompensate
                     ? TwoPhaseOutcome.FINISH_ERROR : TwoPhaseOutcome.FINISH_OK);
@@ -325,6 +335,8 @@ public class LRAParticipantRecord extends AbstractRecord implements Comparable<A
         if (httpStatus == -1) {
             // the local invocation was not made so fallback to using JAX-RS
 
+            trace_progress("notifying participant");
+
             try {
                 // ask the participant to complete or compensate
                 client = ClientBuilder.newClient();
@@ -360,6 +372,8 @@ public class LRAParticipantRecord extends AbstractRecord implements Comparable<A
                 LRALogger.logger.infof("LRARecord.doEnd put %s failed for LRA %s (reason: %s)",
                         endPath, lraId, e.getMessage());
             } finally {
+                trace_progress("notified participant");
+
                 if (client != null) {
                     client.close();
                 }
@@ -388,6 +402,7 @@ public class LRAParticipantRecord extends AbstractRecord implements Comparable<A
             // recovery will figure out the status via the status url
             status = compensate ? ParticipantStatus.Compensating : ParticipantStatus.Completing;
             accepted = true;
+            trace_progress("notify participant failed");
         }
 
         updateStatus(compensate);
@@ -444,11 +459,13 @@ public class LRAParticipantRecord extends AbstractRecord implements Comparable<A
             Response response = responseFuture.get(PARTICIPANT_TIMEOUT, TimeUnit.SECONDS);
 
             if (response.getStatus() == 200) {
+                trace_progress("notified participant");
                 return true;
             }
         } catch (Exception e) {
             LRALogger.i18nLogger.warn_cannotNotifyAfterLRAURI(target, e);
         } finally {
+            trace_progress("finished notifying participant");
             if (client != null) {
                 client.close();
             }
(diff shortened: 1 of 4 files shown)


rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRAParticipantRecord.java
```diff
@@ -114,11 +116,13 @@ public LRAParticipantRecord() {
                 if (parseException[0] != null) {
                     String errorMsg = LRALogger.i18nLogger.error_invalidCompensator(lra.getId(), parseException[0].getMessage(), linkURI);
                     LRALogger.logger.error(errorMsg);
+                    trace_progress(errorMsg);
```
ochaloup, line 119: First, I'm fine this being in the code as you prefer it. This is just I would like just express my opinion.
This is kind of like debugging with `System.out.println` and that way it slows down the processing in general. Hopefully the JIT will cover it.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
```diff
@@ -397,8 +410,13 @@ protected ReentrantLock tryLockTransaction() {
     public int finishLRA(boolean cancel) {
         ReentrantLock lock = null;
 
-        // remove the LRA from the timeout scheduler (no need to take the lock)
-        TransactionReaper.transactionReaper().remove(this);
+        // check whether the transaction should cancel due to a timeout:
+        if (finishTime != null && !cancel && ChronoUnit.MILLIS.between(LocalDateTime.now(ZoneOffset.UTC), finishTime) <= 0) {
```
ochaloup, line 414: This sounds good. As it makes the timeout processing stable it's a great fix.

rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
```diff
@@ -165,21 +164,6 @@ public Response actionWithLRA(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI contextI
             return getResult(cancel, contextId);
         }
 
-        @GET
-        @Path("time-limit")
```
ochaloup, line 169: Would you consider to have the test `time-limit` left in the code even there will be no reaper used?

ochaloup (commented):
The PR is fine I would say. I have got just few comments.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRAParticipantRecord.java
mmusgrov: Your statement is incorrect: trace_progress is coded as `if (LRALogger.logger.isTraceEnabled()) {...}` and I make sure I only call `trace_progress` with static strings from the heap, ie the only performance hit is an extra frame.

rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
mmusgrov: That is part of `git revert` for the reaper change.
We have timeout tests in the TCK as well.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
mmusgrov: Thanks, simple fixes are so much sweater.

rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
mmusgrov: But you are right, if I deemed it useful then, then it is still useful. Thanks for picking up on it.

Note that the reason I have included a number of unit tests with the coordinator is becuase they generally complement the TCK and they also directly test coordinator functionality so they are handy to have when making coordinator specific changes.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRAParticipantRecord.java
ochaloup: I think my statement is quite correct. The performance hit is an extra frame to the stack and an if operation to verify the trace is enabled. In the hot code path it's not a great strategy. If you stress this as an benefit for other reasons, no problem.

rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
ochaloup: +1

ochaloup (approved):

ochaloup:
The github CI actions fail with `LRATest.mixedMultiLevelNestedActivity:578->multiLevelNestedActivity:706 multiLevelNestedActivity: step 9 (called test path http://localhost:8081/base/test/multiLevelNestedActivity) expected:<1> but was:<2>`
It's a test which happens to start failing quite regurarly on the github CI. I don't know what's the reason for that. Just mentioning, it's some instability

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRAParticipantRecord.java
mmusgrov: The logical conclusion of your argument is to remove all of our debug and trace level logging and we have plenty of that.

A single frame is minimal and if you really were worried about performance then there are loads of other opportunities to improve the throughput, besides the calls are only ever made with strings from the heap. If we find sufficient uptake of LRA then we can begin looking at sprucing up the code base, until then maintainability is more important.

Is this an impasse for which we need an adjudicator?

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRAParticipantRecord.java
```diff
@@ -1000,4 +1033,16 @@ private String getLRAId(URI lraId) {
...
+        if (LRALogger.logger.isTraceEnabled()) {
+            LRALogger.logger.tracef("%s: Participant id: %s, LRA id: %s, reason: %s, state: %s, accepted: %b",
+                    LocalDateTime.now(ZoneOffset.UTC), // use the same time function as used for LRA timeouts
+                    lraId,
+                    participantPath,
+                    reason,
+                    status,
+                    accepted);
+        }
+    }
```
jmfinelli, line 1047: WDYT about moving `trace_progress` to `io.narayana.lra.logging`? For example, adding it to `LRALogger` or creating a new utility class?

jmfinelli (commented):
LGTM +1

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRAParticipantRecord.java
mmusgrov: It's just a trace statement, there is no desire to over-engineer this.

jmfinelli (approved):

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRAParticipantRecord.java
mmusgrov: Do also notice that the module is I/O bound and most of the trace logging happens in response to a state change which needs to be persisted, ie the performance hit of an extra frame is dwarfed by other costs.

mmusgrov:
> The github CI actions fail with `LRATest.mixedMultiLevelNestedActivity:578->multiLevelNestedActivity:706 multiLevelNestedActivity: step 9 (called test path http://localhost:8081/base/test/multiLevelNestedActivity) expected:<1> but was:<2>`
> It's a test which happens to start failing quite regurarly on the github CI. I don't know what's the reason for that. Just mentioning, it's some instability

The specification allows the callback to be resent (e.g. to take account of network errors) so the test needs to take account of that (i.e. it should be asserting that the compensate count is >= 1).

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRAParticipantRecord.java
ochaloup: Sorry for not being clear. This is not an impasse. I'm fine with the PR to be merged and I've approved it already.

Just I have a different perspective on coding practices. For example, I would be supporting and idea to find the places where to emit open tracing events. While I would not track the processing with log trace messages.
The most of the log are written at places of a non standard processing. That's how it should be by me. Not logging on hot code path but in strange and exceptional situations.

Never-mind, I don't want to both you with my arguments. I feel I need to express my perspective which can lead to some discussion on the topic. I don't oppose merging this PR.

ochaloup:
ad `LRATest.mixedMultiLevelNestedActivity:578->multiLevelNestedActivity`) do you plan to put the fix for the test as part of this PR or should I create a new issue for adjusting the test?

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LRAParticipantRecord.java
mmusgrov: Let's wait until we have an opentracing or a bytecode/byteman based solution before throwing out trace logging.
Until that work get's done we need to be able trace the progress of LRAs and compensators.

PR #1912, opened 2021-10-01
remove reference to MAIN

mmusgrov:
script only change. I've removed `MAIN` from the PR comment since `MAIN` does not do anything.

@tomjenkinson will you review this. I removed MAIN but there is a reference to in a test in `scripts/hudson/narayana.sh` that you added (sha 49d61042c51113e72a277c844902edad93c62e82) that I don't understand so I left it in there (but with my change the test definitely has no effect).

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -6,4 +6,4 @@ Please refer to our [guidelines for making contributions](https://github.com/jbo
 
 The build axis can be controlled by prefixing a ! on the following as appropriate:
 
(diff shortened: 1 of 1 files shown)


mmusgrov:
The PR reported that it `started testing the MAIN profile` followed by `CORE profile tests passed`, rofl!
I've raised a PR to fix it.

ochaloup:
The discrepancy should be that the CI poller script uses the`MAIN` (http://narayanaci1.eng.hst.ams2.redhat.com/view/Pulls/job/btny-pulls-narayana-poller/configure) and the `narayana.sh` uses the `CORE` (https://github.com/jbosstm/narayana/blob/master/scripts/hudson/narayana.sh#L68). Up to that the `sh` checks for profile `CORE` and it prints `MAIN` (https://github.com/jbosstm/narayana/blob/master/scripts/hudson/narayana.sh#L69).

mmusgrov:
It's quite inconsistent.

tomjenkinson (approved):

PR #1917, opened 2021-11-14
narayana-jts-integration migration to jakarta

tadamski:
I needed this transformation as a dependency of IronJacamar


mmusgrov:
I am in the process of producing migrations for all of our relevant modules ie https://issues.redhat.com/browse/JBTM-3544
I have already done narayana-jts-integration migration (subtask 5) but I was hoping to raise a pr for the full task this week - I have been delayed get the XTS one to pass the crash rec tests.

I'd prefer not to have to make a release for each one - hence the reason I want to raise a PR for the overall task (and not for each subtask). Can you wait a little longer.

tadamski:
OK. I thought that I will push it cause it seems to work, but in that case I'm closing this PR.

mmusgrov:
We have a lot of modules that need migrating so i have put the rules in a consistent place and I am putting them all under a separate jakarta directory, if I merge this PR then it means a lot more work for us dealing the different approach to the directory layouts. Can't you just use this migration in your own testing and change the wildfly pom to depend on `<version.org.jboss.narayana>5.12.2.Final-SNAPSHOT</version.org.jboss.narayana>`

mmusgrov:
The transaction subsystem is targeted for sprint 1.

tadamski:
OK. Will do.

PR #1946, opened 2022-01-31
[JBTM-3575] Introduced ServiceName when the JDBC object store is tested

jmfinelli:
This PR addresses [JBTM-3575](https://issues.redhat.com/browse/JBTM-3575)

Build axis:
DB_TESTS oracle

Skipped:
!CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !PERF !LRA !mysql !db2 !postgres

PR #1947, opened 2022-02-02
Picked up an ORB implementation for the DB_TESTS testing axis

jmfinelli:
This PR explicitly pick up an ORB implementation (i.e. OPENJDK_ORB, which is also used in WildFly) when the DB_TESTS testing axis is selected.

Axis to tests:
QA_JTA DB_TESTS mysql db2 postgres oracle

Axes to skip:
!CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !PERF !LRA

### DIFF
--- a/qa/run-tests.xml
+++ b/qa/run-tests.xml
@@ -537,11 +537,9 @@
             <then>
               <var name="jdbcstore.database.properties" value="ClassName=${jdbc.datasource.class};ServiceName=${jdbc.db.name};User=${jdbc.db.user};Password=${jdbc.db.password};ServerName=${jdbc.db.url};PortNumber=${jdbc.db.port}${jdbc.db.properties}" />
             </then>
-          <else>
-            <then>
+            <else>
               <var name="jdbcstore.database.properties" value="ClassName=${jdbc.datasource.class};DatabaseName=${jdbc.db.name};User=${jdbc.db.user};Password=${jdbc.db.password};ServerName=${jdbc.db.url};PortNumber=${jdbc.db.port}${jdbc.db.properties}" />
-            </then>
-          </else>
+            </else>
           </if>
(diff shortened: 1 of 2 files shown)


tomjenkinson (approved):
I haven't verified the patch, but it looks reasonable to me - thanks!

qa/run-tests.xml
```diff
@@ -537,12 +537,10 @@
...
             </then>
-          <else>
-            <then>
+            <else>
               <var name="jdbcstore.database.properties" value="ClassName=${jdbc.datasource.class};DatabaseName=${jdbc.db.name};User=${jdbc.db.user};Password=${jdbc.db.password};ServerName=${jdbc.db.url};PortNumber=${jdbc.db.port}${jdbc.db.properties}" />
-            </then>
-          </else>
-          </if>
+            </else>
+	  </if>
```
tomjenkinson, line 543: @jmfinelli I am not sure what is going on with these changes in here, please can you clarify?

tomjenkinson (changes requested):
I am not sure what is going on with the changes in qa/run-tests.xml, please can you clarify?

qa/run-tests.xml
jmfinelli: Oversight. `else` does not need `then`.

tomjenkinson: Ah, so IIUC then the change in https://github.com/jbosstm/narayana/pull/1946/files did not work as it was expected to and the syntax is therefore incorrect as it stands?

I think this change should be in a separate commit but can be in the same PR

jmfinelli: > Ah, so IIUC then the change in https://github.com/jbosstm/narayana/pull/1946/files did not work as it was expected to and the syntax is therefore incorrect as it stands?

Yeah, that is what happened.

> I think this change should be in a separate commit but can be in the same PR

No problem, I will update this PR

tomjenkinson (approved):
LGTM from visual inspection - let's see what CI says

PR #1955, opened 2022-02-13
JBTM-3579: Revert "JBTM-3532 Add LRA support for EE 9"

mmusgrov:
https://issues.redhat.com/browse/JBTM-3579

This reverts commit ead7de415b65c0073df7075b8549ec249bd3bb49.

The reason for backing out EE 9 support is because there are too many issues with running it on various platforms, for example:
    - it breaks the quickstarts
    - we cannot do releases until we have a working LRA implementation
    - the coordinator resource is used by various community users and we do not wish to break their applications without offering them an alternative
    - the coordinator-quarkus maven model does not work
    - the change severely restricts which JAX-RS servers are able to run the coordinator

Until there is better support for Jakarta APIs on our target run-times we will choose byte code transformation - and the alternative of using a branch for EE 9 support is too much overhead.

JDK11 !CORE !TOMCAT !AS_TESTS !RTS !XTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !PERF LRA !DB_TESTS !mysql !db2 !postgres !oracle


mmusgrov:
@jmfinelli  the JDK 11 job failure is unrelated to this PR (it will most likely be related to pr #1956 )
Since the JDK 8 job passed this PR is ready for review (note that I tested both using both Java versions on my laptop).

jmfinelli (approved):
Thank you very much @mmusgrov. This was a big revert pr. it LGTM!

mmusgrov:
@xstefank I am merging this one. We will use bytecode transformation for now (since maintaining separate code bases is less than optimal) and then we can reinstate your PR when more runtimes support jakarta. We are quite under resourced at the moment so I don't know when we'd be able to add jakarta support - do you know what impact will this have upstream to help me to prioritise the work?

xstefank:
AFAIK WFLY 27 would like to have all artifacts in native jakarta.* namespace. Quarkus 3 will move to jakarta too but I haven't heard dates yet. Maybe @bstansberry can shine more light on what is expected on the WFLY side. Note that LRA is still not properly integrated into WFLY AFAIK.

bstansberry:
For WildFly to include LRA, it would need to be based on jakarta.* namespace artifacts available in maven.

I don't think LRA is critical for WF 27 though.

mmusgrov:
@xstefank FYI I have created a branch for this (JBTM-3532) but it is going to be awkward maintaining both so we still look to an alternate solution. If we need to we will do testing of the branch on our CI, example if you need to consume the branch then do let us know and we will add it to our CI.

PR #1956, opened 2022-02-14
JBTM-3582 Check that WildFly and java versions are compatible

mmusgrov:
https://issues.redhat.com/browse/JBTM-3582

I added a check for running axes that require Java 11 or above (AS_TESTS|RTS|JACOCO|XTS). Note that this should catch the XTS profile which did indeed fail but the LRA axis should pass.

The jdk checks happen in two places:
1. function clone_as which uses our jboss-as 5_BRANCH - this axis can only run on JDK 11 or above
2. function download_as which just grabs either the latest WildFly if the jdk version is 11 or above and otherwise gets the last version that can run with Java 8. For example the LRA axis just needs a JAX-RS compliant container so WildFly 26 is fine

JDK11 !CORE !TOMCAT !AS_TESTS !RTS !JACOCO XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !PERF LRA !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/scripts/hudson/narayana.sh
+++ b/scripts/hudson/narayana.sh
@@ -12,6 +12,29 @@ function fatal {
   exit 1
 }
 
+function which_java {
+  type -p java 2>&1 > /dev/null
+  if [ $? = 0 ]; then
+    _java=java
+  elif [[ -n "$JAVA_HOME" ]] && [[ -x "$JAVA_HOME/bin/java" ]]; then
+    _java="$JAVA_HOME/bin/java"
+  else
+    unset _java
+  fi
+
+  if [[ "$_java" ]]; then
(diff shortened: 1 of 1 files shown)


jmfinelli (approved):
LGTM! Before merging, I would try to run it with the argument/profile "JDK11" in the PR's description (just to double check that your modifications pass on our CI with JDK11)

mmusgrov:
I will merge this one since it passed on the JDK 8 and 11 axes. Note that to run any of the AS_TESTS|RTS|JACOCO|XTS axes the build must use Java 11 since that is the only one supported by WildFly latest at the moment.

PR #1957, opened 2022-02-15
JBTM-3582 download the correct version of the AS depending upon which…

mmusgrov:
https://issues.redhat.com/browse/JBTM-3582

This change follows on from #1956 and downloads the correct version of WildFly depending upon which version of Java is being used for the build.

I've enabled the LRA and RTS axes to build with JDK8 and then JDK11:

- LRA should build with both Java versions. For a Java 8 build will use the last WildFly release that supported Java  and for a Java 11 build the latest WildFly release will be used;
- RTS should only build with Java 11 since it does AS testing and since WildFly latest no longer supports Java 8

!JDK8 !JDK11 !LRA !RTS
!CORE !TOMCAT !AS_TESTS !XTS !JACOCO XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !PERF !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/scripts/hudson/narayana.sh
+++ b/scripts/hudson/narayana.sh
@@ -77,7 +77,7 @@ function init_test_options {
     # WildFly 27 requires JDK 11 (see JBTM-3582 for details)
     _jdk=`which_java`
     if [[ "$_jdk" -lt 11 && "$PROFILE" =~ ^(AS_TESTS|RTS|JACOCO|XTS) ]]; then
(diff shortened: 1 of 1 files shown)


mmusgrov:
@jmfinelli I just pushed a correction to a typo in a print message and disabled the testing. If you feel that a retest is required then I will trigger it manually.

jmfinelli (approved):
LGTM! Thanks @mmusgrov

PR #1959, opened 2022-02-21
Updated byteman and replaced maven-injection-plugin by maven-resource…

mayankkunwar:
…s-plugin
https://issues.redhat.com/browse/JBTM-3482

CORE TOMCAT AS_TESTS RTS JACOCO XTS QA_JTA QA_JTS_JACORB QA_JTS_JDKORB QA_JTS_OPENJDKORB PERF LRA DB_TESTS mysql db2 postgres oracle

### DIFF
--- a/ArjunaJTS/narayana-jts-ibmorb/pom.xml
+++ b/ArjunaJTS/narayana-jts-ibmorb/pom.xml
@@ -23,7 +23,16 @@
   <name>Narayana: ArjunaJTS narayana-jts-ibmorb</name>
   <packaging>jar</packaging>
   <description>JTS uber jar with ibmorb stubs</description>
+  <properties>
+    <propertiesFileName>jbossts-ibmorb-properties.xml</propertiesFileName>
+  </properties>
   <build>
+    <resources>
+      <resource>
+        <directory>${project.basedir}/src/main/resources</directory>
+        <filtering>true</filtering>
+      </resource>
+    </resources>
     <plugins>
       <plugin>
         <inherited>false</inherited>
@@ -129,49 +138,6 @@
           <doUpdate>false</doUpdate>
           <shortRevisionLength>5</shortRevisionLength>
           <revisionOnScmFailure>${project.version}-noscm</revisionOnScmFailure>
-        </configuration>
-      </plugin>
-      <plugin>
-        <groupId>org.jboss.maven.plugins</groupId>
-        <artifactId>maven-injection-plugin</artifactId>
-        <executions>
-          <execution>
-            <phase>compile</phase>
-            <goals>
-              <goal>bytecode</goal>
-            </goals>
-          </execution>
-        </executions>
-        <configuration>
-          <bytecodeInjections>
-            <bytecodeInjection>
-              <expression>${buildNumber}</expression>
-              <targetMembers>
-                <methodBodyReturn>
-                  <className>com.arjuna.common.util.ConfigurationInfo</className>
-                  <methodName>getSourceId</methodName>
-                </methodBodyReturn>
-              </targetMembers>
-            </bytecodeInjection>
-            <bytecodeInjection>
-              <expression>jbossts-ibmorb-properties.xml</expression>
-              <targetMembers>
-                <methodBodyReturn>
-                  <className>com.arjuna.common.util.ConfigurationInfo</className>
-                  <methodName>getPropertiesFile</methodName>
-                </methodBodyReturn>
-              </targetMembers>
-            </bytecodeInjection>
-            <bytecodeInjection>
-              <expression>JBoss Inc. [${user.name}] ${os.name} ${os.version} ${buildproperty.date}</expression>
-              <targetMembers>
-                <methodBodyReturn>
-                  <className>com.arjuna.common.util.ConfigurationInfo</className>
-                  <methodName>getBuildId</methodName>
-                </methodBodyReturn>
-              </targetMembers>
-            </bytecodeInjection>
-          </bytecodeInjections>
         </configuration>
       </plugin>
     </plugins>
--- a/common/classes/com/arjuna/common/util/ConfigurationInfo.java
+++ b/common/classes/com/arjuna/common/util/ConfigurationInfo.java
@@ -21,6 +21,11 @@
 package com.arjuna.common.util;
 
 import com.arjuna.common.internal.util.propertyservice.BeanPopulator;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.util.Properties;
 
 /**
  * Utility class providing access to build time and runtime configuration reporting functions.
@@ -28,12 +33,36 @@ import com.arjuna.common.internal.util.propertyservice.BeanPopulator;
  * Replaces the old per-module Info (and in some cases Configuration and report) classes.
  *
  * The actual build information is injected during the build via the
- * org.jboss.maven.plugins:maven-injection-plugin plugin
+ * Maven Resources Plugin
  *
  * @author Jonathan Halliday (jonathan.halliday@redhat.com) 2009-10
  */
 public class ConfigurationInfo
 {
+    private static final String SOURCE_ID;
+    private static final String PROPERTIES_FILE_NAME;
+    private static final String BUILD_ID;
+
+    static {
+        Properties configInfoProps = new Properties();
+        String sourceId = "(unknown)";
+        String propertiesFileName = "(unknown)";
+        String buildId = "(unknown)";
+        try (final InputStream stream = ConfigurationInfo.class.getResourceAsStream("/ConfigurationInfo.properties")) {
+            if (stream != null)
+                try (final InputStreamReader reader = new InputStreamReader(stream, StandardCharsets.UTF_8)) {
+                    configInfoProps.load(reader);
+                    sourceId = configInfoProps.getProperty("sourceId", sourceId);
+                    propertiesFileName = configInfoProps.getProperty("propertiesFileName", propertiesFileName);
+                    buildId = configInfoProps.getProperty("buildId", buildId);
+                }
+        } catch (IOException ignored) {
+        }
+        SOURCE_ID = sourceId;
+        PROPERTIES_FILE_NAME = propertiesFileName;
+        BUILD_ID = buildId;
+    }
+
     /**
      * @see getSourceId
      * @return the version, if known.
@@ -46,21 +75,21 @@ public class ConfigurationInfo
      * @return the version control tag of the source used, or "unknown".
      */
     public static String getSourceId() {
-        return ""; // the bytecode for the method is replaced at build time
+        return SOURCE_ID; // reads the property from properties file in whilch the value of keys are replaced at build time
     }
 
     /**
      * @return the name (not path) of the properties file
      */
     public static String getPropertiesFile() {
-        return ""; // the bytecode for the method is replaced at build time
+        return PROPERTIES_FILE_NAME; // define a property as <propertiesFileName></propertiesFileName> in pom.xml
     }
 
     /**
      * @return the build identification line indicating the os name and version and build date
      */
     public static String getBuildId() {
-        return ""; // the bytecode for the method is replaced at build time
+        return BUILD_ID; // reads the property from properties file in whilch the value of keys are replaced at build time
     }
 
     /**
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -18,10 +18,17 @@
   <packaging>jar</packaging>
   <properties>
     <sortpom.skip>true</sortpom.skip>
+    <propertiesFileName>jbossts-properties.xml</propertiesFileName>
   </properties>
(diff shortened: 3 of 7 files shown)


jmfinelli:
LGTM! Thanks @mayankkunwar

I do not know if it makes sense to wait for the CI to finish though. If I understand correctly, it seems that you did not introduce any unit test to check if properties get fetched correctly. I should also mention that there was not any unit test even before your PR (where properties were injected with the plugin). If we want to introduce some unit tests for this PR, we might consider using [surefire system properties](https://maven.apache.org/surefire/maven-surefire-plugin/examples/system-properties.html), WDYT?

Other than that, I have manually tested your PR and it works correctly!

mayankkunwar:
Thanks for the review @jmfinelli 
I will add the tests for the same.

mmusgrov:
It looks great @mayankkunwar - again some good work here.
I/we need to test it by building on Java 11 and running on Java 17. Have you done that? I will test it myself but I don't have time today, and tomorrow I am on PTO and Fri is earmarked for learning something new. @jmfinelli (or @mayankkunwar ) I don't suppose you will have time to test this will you, if not I will try it out over the weekend.

mayankkunwar:
Thanks @mmusgrov 
In this PR there is one issue that I am working on, I will fix it and push the changes. The issue I have already discussed with Manuel.
Maybe if someone (you or @jmfinelli ) could verify/review the code changes after that, it would be great.
Thanks

mayankkunwar:
> Also I believe we print out the build number etc when booting on WildFly so I/we need to check that that is still getting printed.

I see, we can do this, thanks

jmfinelli:
> Thanks @mmusgrov In this PR there is one issue that I am working on, I will fix it and push the changes. The issue I have already discussed with Manuel. Maybe if someone (you or @jmfinelli ) could verify/review the code changes after that, it would be great. Thanks

No problem! Give me a shout when the PR is ready to be reviewed

mmusgrov:
> > Also I believe we print out the build number etc when booting on WildFly so I/we need to check that that is still getting printed.
> 
> I see, we can do this, thanks

I meant that this should already happen so the task is to check that after your changes the message still gets printed correctly - some customers will no doubt be relying on the INFO.

jmfinelli:
Do we need to run all the axis on this PR? AFAICT, you did not introduce tests for the axis LRA, QA_JTA, etc.

ArjunaJTS/narayana-jts-ibmorb/pom.xml
```diff
@@ -23,7 +23,22 @@
...
+        <directory>${project.basedir}/src/main/resources</directory>
+        <filtering>true</filtering>
+      </resource>
+    </resources>
+    <testResources>
+      <testResource>
+        <directory>${project.basedir}/src/test/resources</directory>
+        <filtering>true</filtering>
+      </testResource>
+    </testResources>
```
jmfinelli, line 41: This should not be here. We do not need testResources AFAIU

ArjunaJTS/narayana-jts-ibmorb/src/test/resources/ConfigurationInfo.properties
```diff
@@ -0,0 +1,3 @@
+sourceId=${buildNumber}
+propertiesFileName=${propertiesFileName}
+buildId=JBoss Inc. [${user.name}] ${os.name} ${os.version} ${buildproperty.date}
```
jmfinelli, line 3: Duplication. This file is not needed

common/classes/com/arjuna/common/util/ConfigurationInfo.java
```diff
@@ -21,19 +21,48 @@
...
+    private static final String SOURCE_ID;
+    private static final String PROPERTIES_FILE_NAME;
+    private static final String BUILD_ID;
+
+    static {
+        Properties configInfoProps = new Properties();
+        String sourceId = "(unknown)";
+        String propertiesFileName = "(unknown)";
+        String buildId = "(unknown)";
+        try (final InputStream stream = Thread.currentThread().getContextClassLoader().getResourceAsStream("ConfigurationInfo.properties")) {
```
jmfinelli, line 51: Why `Thread.currentThread().getContextClassLoader()`? Personally, I am always overcareful when it comes to Threads (in this case, `ConfigurationInfo` is used in ArjunaJTS -> JTS ... so I am even more worried (maybe too much lol)). Could you please help me understanding why you went for `Thread.currentThread().getContextClassLoader()`? Thanks

common/pom.xml
```diff
@@ -22,6 +22,18 @@
...
+        <directory>${project.basedir}/src/main/resources</directory>
+        <filtering>true</filtering>
+      </resource>
+    </resources>
+    <testResources>
+      <testResource>
+        <directory>${project.basedir}/src/test/resources</directory>
+        <filtering>true</filtering>
+      </testResource>
+    </testResources>
```
jmfinelli, line 36: Same as above

common/pom.xml
```diff
@@ -145,5 +118,16 @@
...
+        <id>setPropFileName</id>
+        <activation>
+          <property>
+            <name>!propertiesFileName</name>
+          </property>
+        </activation>
+        <properties>
+          <propertiesFileName>jbossts-properties.xml</propertiesFileName>
+        </properties>
+      </profile>
```
jmfinelli, line 131: I do not see why you need a profile. I guess you should be fine defining a property:
```
<properties>
  <propertiesFileName>jbossts-properties.xml</propertiesFileName>
</properties>
```

common/src/test/resources/ConfigurationInfo.properties
```diff
@@ -0,0 +1,3 @@
+sourceId=${buildNumber}
+propertiesFileName=${propertiesFileName}
+buildId=JBoss Inc. [${user.name}] ${os.name} ${os.version} ${buildproperty.date}
```
jmfinelli, line 3: Same as above

jmfinelli (changes requested):

mayankkunwar:
> > > Also I believe we print out the build number etc when booting on WildFly so I/we need to check that that is still getting printed.
> > 
> > 
> > I see, we can do this, thanks
> 
> I meant that this should already happen so the task is to check that after your changes the message still gets printed correctly - some customers will no doubt be relying on the INFO.

Okay

mayankkunwar:
> Do we need to run all the axis on this PR? AFAICT, you did not introduce tests for the axis LRA, QA_JTA, etc.

I by mistake ran other axis. Will comment them out

ArjunaJTS/narayana-jts-ibmorb/pom.xml
mayankkunwar: Here I was thinking if someone uses this dependency and they want to write a test, so in order to execute that test they would require `testResource` i.e. the build info file. So at that time user will have to explicitly include this resource in test target directory. To avoid that I added it here.

ArjunaJTS/narayana-jts-ibmorb/src/test/resources/ConfigurationInfo.properties
mayankkunwar: If we provide `testResource`, we would need this file in `test` directory, but I can also give the reference of file that is present in `src` directory, but that would not be a good approach I guess.

common/pom.xml
mayankkunwar: There are `tests` that are using this `filter`

common/pom.xml
mayankkunwar: Okay, I will provide separate property for this

ArjunaJTS/narayana-jts-ibmorb/pom.xml
jmfinelli: I am not sure I follow you. I thought that [Maven Resources Plugin filters](https://maven.apache.org/plugins/maven-resources-plugin/examples/filter.html) items that has been indicated as resources (e.g. `<directory>${project.basedir}/src/main/resources</directory>`). Why tests should require a `testResource`?

ArjunaJTS/narayana-jts-ibmorb/src/test/resources/ConfigurationInfo.properties
jmfinelli: I do not think you need a reference to a resource during tests

common/classes/com/arjuna/common/util/ConfigurationInfo.java
jmfinelli: I am not sure I follow you. If `A` and `B` are modules and `B` uses `A`, why should they run on different threads?

common/pom.xml
jmfinelli: Same as above

common/pom.xml
jmfinelli: Thanks

common/src/test/resources/ConfigurationInfo.properties
jmfinelli: Same as above

jmfinelli (changes requested):

common/classes/com/arjuna/common/util/ConfigurationInfo.java
mayankkunwar: Suppose module A is already built and now its module B's turn to get build and run. When B starts building at that time all the A's class and resource files package are ready to be used by module B. Now during the build `maven resource plugin filter` will be applied to B's resource file and then all the variables inside B's resource file will be replaces by their respective values and depending on the type of resource filter that is applied, will decide the location to resource file in the `/target` folder (which can be any of `/target/classes` or `/target/test-classes`)

Now its time to run the tests in B. First test in B tries to run. At this time there will be single `thread` that will be running. So the first test will try to call a class of A which has `Thread.currentThread().getContextClassLoader().getResourceAsStream("R")` in it. Here same test thread will try to get the resource file `R`. So instead of looking resource file into A's resource package, Thread will try to find the resource path of the module from where it started executing first, this means it will get the path of B's resource package.
You can verify this by checking the path value returned by `Thread.currentThread().getContextClassLoader().getResource("")`

ArjunaJTS/narayana-jts-ibmorb/pom.xml
mayankkunwar: In the process of maven builds when different classes and test-classes are packaged inside `/target` folder, there are sub-folders created. One of them is `/target/classes` and another is `/target/test-classes`. So under `/target/classes` we have all the classes with their resources all packed together and in the same way in `/target/test-classes` contains all the test classes with their resources.

Now here comes `maven resource plugin filter` into picture. The main function of filtering is to replace maven related variables (like `${variable}`) with their value. The `<resources/>` will replace all the variables inside `/target/classes` folder resource files and `<testResources/>` will replace all the variables inside `/target/test-classes` folder resource files with their respective values.

So, until these resource file variables are not replaced with their values, the tests will fail.

jmfinelli (approved):

jmfinelli:
LGTM :-) Thanks Mayank :+1:

mayankkunwar:
> LGTM :-) Thanks Mayank +1

Thanks for the review @jmfinelli

mmusgrov (approved):
I checked out the PR and it does what we need to do (nb: I only tested the ArjunaCore/arjuna module):

```
 git co pr/1959                                            
 sudo update-alternatives --config java
 export JAVA_HOME=/usr/lib/jvm/java-17-openjdk-17.0.2.0.8-2.rolling.fc34.x86_64
 export PATH=/usr/lib/jvm/java-17-openjdk-17.0.2.0.8-2.rolling.fc34.x86_64/bin:$PATH
 ./build.sh clean install -DskipTests -Pcommunity -Dmaven.compiler.source=1.8 -Dmaven.compiler.target=17
 cd ArjunaCore/arjuna                                        
 mvn test -Dmaven.compiler.source=1.8 -Dmaven.compiler.target=17
```

Is there a plan to add a CI job that tests this?

mayankkunwar:
> Is there a plan to add a CI job that tests this?

I had added few changes similar to your configuration here https://github.com/jbosstm/narayana/commit/c62ceec7439fa63f8644c03278c89fa90e72667c which will run by http://narayanaci1.eng.hst.ams2.redhat.com/view/Pulls/job/btny-pulls-narayana-jdk17/configure. For now I have commented out these changes in CI

I can uncomment my changes and try to run it in CI

pom.xml
```diff
@@ -863,5 +863,16 @@
...
     </profile>
+    <profile>
+      <id>jdk17</id>
+      <activation>
+        <jdk>[17,)</jdk>
+      </activation>
+      <properties>
+        <!-- maven-compiler-plugin -->
+        <maven.compiler.source>1.8</maven.compiler.source>
+        <maven.compiler.target>17</maven.compiler.target>
```
tomjenkinson, line 874: I don't think we are proposing to ship jdk17 versions of Narayana, so I am not clear what setting the maven.compiler.target to 17 achieves for the narayana artifacts that would be accessible on maven central?

PR #1935, opened 2021-12-16
JBTM-3482:Build Narayana using JDK 17

mayankkunwar:
https://issues.redhat.com/browse/JBTM-3482

!JDK8 JDK17

!CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !PERF !LRA !DB_TESTS !mysql !db2 !postgres !oracle

PR #1962, opened 2022-04-03
JBTM-3592 Use standard wildfly (not preview)

mmusgrov:
https://issues.redhat.com/browse/JBTM-3592

The PR corrects pr #1935 which forgot to revert the script change to use wildfly preview

JDK11 LRA !RTS !CORE !TOMCAT !AS_TESTS !XTS !JACOCO XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !PERF !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/scripts/hudson/narayana.sh
+++ b/scripts/hudson/narayana.sh
@@ -457,31 +457,32 @@ function tests_as {
 function download_as {
   echo "Downloading WildFly Build"
 
+  cd $WORKSPACE
+
   # clean up any previously downloaded zip files (this will not clean up old directories)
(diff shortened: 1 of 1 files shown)


mmusgrov:
What is wrong the build script, CI is running with with JDK 8 but not JDK 11 - it is supposed to run both axes

mmusgrov:
The first JDK failure was unrelated to my PR (I re-ran it to check and the second run passed), although it's a failure I've not seen before.

Since this is a script only change I am going to merge the PR.

PR #1966, opened 2022-05-06
[JBTM-3596] Quarkus LRA coodinator produces wrong OpenAPI and Swagger-UI

xstefank:
Upgrade Quarkus to 2.9.0.Final

https://issues.redhat.com/browse/JBTM-3596


xstefank:
@tomjenkinson yes, I totally forgot about that, I should be able to go back to some version of Quarkus that is still on JDK 8 but we are still shipping coordinator with very old version of Quarkus.

xstefank:
@tomjenkinson @mmusgrov I will downgrade the quarkus in this PR and let's discuss how to best approach possible conditional compilation of quarkus modules only on JDK 11.

mayankkunwar:
> @mayankkunwar do you think #1935 is needed to help with the JDK 17 failure on this PR?

Yes @tomjenkinson #1935 is needed for JDK 17

xstefank:
JDK 8 failure is unrelated, please restart the action or I think we can merge this as CI also runs on JDK 8 AFAIK.

xstefank:
@tomjenkinson this PR is fixing different issue so please go ahead and merge it. It fixes wrong OpenAPI document which is important problem in community.

xstefank:
Any ETA when we can get lra-coordinator image with this change please?

rts/pom.xml
```diff
@@ -31,7 +31,7 @@ Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
   <url>http://www.jboss.org/jbosstm</url>
 
   <properties>
-    <version.quarkus>1.8.3.Final</version.quarkus>
+    <version.quarkus>1.13.7.Final</version.quarkus>
```
tomjenkinson, line 34: This should be removed and depend upon the Quarkus upgrade PR from over here: https://github.com/jbosstm/narayana/pull/1967

xstefank:
@tomjenkinson Quarkus version can be updated to later together with JDK 11 work. This version is not that different to what is merged currently and I have external factors waiting for correct OpenAPI document.

rts/pom.xml
tomjenkinson: If you like, you could update it here to 2.9.0.Final too as that is the version we are targeting for the next release? But to mention that assuming https://github.com/jbosstm/narayana/pull/1972 is merged, I think there will then be a merge conflict here. Perhaps you could instead try using https://github.com/jbosstm/narayana/pull/1972/commits/6d838530a0226cfd706309bb94ec6b2c6e47dc1b in your branch? I think there could be some TCK hang with 2.9.0.Final though.

tomjenkinson: Apparent hanging is mentioned in https://github.com/jbosstm/narayana/pull/1967

rts/lra/coordinator-quarkus/src/main/resources/application.properties
```diff
@@ -4,3 +4,6 @@ quarkus.ssl.native=true
 
 # allow Swagger in prod mode
 quarkus.swagger-ui.always-include=true
+
+quarkus.index-dependency.mp-lra.group-id=org.eclipse.microprofile.lra
```
tomjenkinson, line 8: Please can you clarify what this and the next line change does?

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
```diff
@@ -103,29 +95,8 @@
 
 @ApplicationScoped
 @Path(COORDINATOR_PATH_NAME)
-
-@OpenAPIDefinition(
```
tomjenkinson, line 107: Please can you clarify why this was needed to be moved out of here and into the new JaxRsApplication class?

rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
```diff
@@ -345,7 +355,7 @@ public void before() {
 
         client = ClientBuilder.newClient();
         coordinatorPath = TestPortProvider.generateURL('/' + COORDINATOR_PATH_NAME);
-        server.deploy(Coordinator.class);
+        server.deploy(LRACoordinator.class);
```
tomjenkinson, line 358: Please can you explain why is it deploying this test LRACoordinator and not the new `JaxRsApplication`?

rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/objectstore/TestBase.java
```diff
@@ -64,7 +74,7 @@ public void before() {
         client = ClientBuilder.newClient();
 
         server = new UndertowJaxrsServer().start();
-        server.deploy(Coordinator.class);
+        server.deploy(LRACoordinator.class);
```
tomjenkinson, line 77: Same comment here as per LRATest "Please can you explain why is it deploying this test LRACoordinator and not the new JaxRsApplication?"

rts/lra/coordinator-quarkus/src/main/resources/application.properties
mmusgrov, line 8: The team is discussing the removal of the quarkus runner for the coordinator since it doesn't provide enough added value. If users want a runner for a dockerised version of the coordinator then we will provide a WildFly based container.

The other changes in this PR look reasonable.

mmusgrov (changes requested):
I like the

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
xstefank: This annotation is by the OpenAPI spec required to be on the Application class which this class no longer extends.

xstefank:
> Please can you clarify which part of the change removes the duplicate "/lra-coordinator" part of the path?

The duplication was caused by getClasses in the Application class. Quarkus just doesn't use Application class if you don't need it for OpenAPI like I am changing it also in this PR.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/JaxRsApplication.java
```diff
@@ -0,0 +1,40 @@
+package io.narayana.lra.coordinator.api;
```
tomjenkinson, line 1: Missing a copyright, but I have a question I will ask on the class itself

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/JaxRsApplication.java
```diff
@@ -0,0 +1,40 @@
...
+                description = "Narayana LRA API version", schema = @Schema(ref = "LRAApiVersionSchema"))
+        },
+        headers = {
+            @Header(name = LRAConstants.NARAYANA_LRA_API_VERSION_HEADER_NAME, description = "Narayana LRA API version",
+                schema = @Schema(ref = "LRAApiVersionSchema"))
+        }
+    )
+)
+@ApplicationPath("/")
+public class JaxRsApplication extends Application {
```
tomjenkinson, line 39: Am I right to see that a big part of the change is to move the `Coordinator` from extending `Application` to this new class? Does that somehow prevent the duplicate `/lra-coordinator` path? And if so, please can you share why that is?

xstefank: Yes, correct. But TBH I am not totally sure. It is inside of the smallrye openapi processing into which I didn't dig too much.

tomjenkinson: Is there somewhere good we can ask for some information to help clarify if this is what works? I notice you have the addition of the ApplicationPath("/") in here, I wonder if it helps somehow. I wonder what would happen if many of the changes were reverted here and instead the existing `Coordinator` class had @ApplicationPath("/") instead of the previous @Path(COORDINATOR_PATH_NAME)

xstefank: Sorry, PTO. I don't think that it would help. This is the recommended Quarkus setup which I would like to preserve. Extending Application from the resource class is not a typical ways in JAX-RS. But if you want I will try to find out why it works this way, but I just don't see the value in it.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
```diff
@@ -134,20 +105,6 @@ public class Coordinator extends Application {
...
 
-    @Override
-    public Set<Class<?>> getClasses() {
-        HashSet<Class<?>> classes = new HashSet<>();
-        classes.add(Coordinator.class);
-        return classes;
-    }
-
-    @Override
-    public Set<Object> getSingletons() {
```
tomjenkinson, line 145: I notice that this method is completely going. I don't know why it was present before, do you know?

tomjenkinson: I am thinking we might want to keep this somewhere - https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/Application.html#getSingletons-- - unless we no longer need CoordinatorContainerFilter to be a singleton?

rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
tomjenkinson: @xstefank I am sorry but I don't understand why the test would deploy a test `LRACoordinator` Application and not the `JaxRsApplication` which is also an application

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/JaxRsApplication.java
tomjenkinson: Please will you share a reference to where it's recommended to separate the Application and the resource class?

rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
```diff
@@ -326,6 +326,16 @@ public Set<Class<?>> getClasses() {
         }
     }
```
tomjenkinson, line 327: I would have preferred to add this comment to the beans.xml but I can't seem to do that (perhaps because it's empty).

Given the getClasses method is not in the `main` anymore I think, is it the addition of the empty `beans.xml` that cause something to trigger which scans the binary to find the `Coordinator` class and treat is as a resource instead?

Similarly, is that file the thing that somehow determines to treat `CoordinatorContainerFilter` as a singleton?

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
tomjenkinson: (I am asking elsewhere in this PR if the empty beans.xml is somehow related to this)

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/JaxRsApplication.java
xstefank: https://quarkus.io/guides/resteasy#whats-different-from-jakarta-ee-development and https://quarkus.io/guides/openapi-swaggerui#providing-application-level-openapi-annotations. Also I would like to note that this still passes all TCK tests so there is no functional change in this PR.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
xstefank: It is singleton by default - https://quarkus.io/guides/resteasy#lifecycle-of-resources

rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
xstefank: not sure what you are referring to. Maybe misplaced comment.

rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
xstefank: This is not a JAX-RS container so there is no scanning. I can try to play around it probably by manually adding all classes but this seems easier to me.

rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
tomjenkinson: I will move it somewhere else, I wanted to add the comment to the beans.xml file but I couldn't spot a way to do that.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
```diff
@@ -134,20 +105,6 @@ public class Coordinator extends Application {
     private final LRAService lraService;
     private final RecoveryCoordinator recoveryCoordinator;
 
-    @Override
-    public Set<Class<?>> getClasses() {
```
tomjenkinson, line 138: I would have preferred to add this comment to the beans.xml but I can't seem to do that (perhaps because it's empty).

Given this `getClasses` method is being removed from here, is it the addition of the empty beans.xml that allows for this?

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
tomjenkinson: For clarity, are you saying that https://github.com/jbosstm/narayana/blob/master/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/CoordinatorContainerFilter.java is a Singleton by default? And is it the same for other containers (like WildFly)?

rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
tomjenkinson: I added an updated comment here https://github.com/jbosstm/narayana/pull/1966/files#r882565440

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/JaxRsApplication.java
tomjenkinson: Do you know if separating these classes is appropriate in WildFly too or is the information you are referring to specific to Quarkus?

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
xstefank: It is a singleton by default as stated at the documentation I posted above. I'm sorry Tom, but to answer your second question I would need to do the research which point I don't see as it is irrelevant to this PR and Quarkus.

xstefank:
@tomjenkinson I see that you are worried that my changes will break the OpenAPI document in WFLY. So I built both Quarkus and WFLY deployed coordinators and the values are here: https://gist.github.com/xstefank/a3ce61d16d02f875fb78ef714aa280af here https://gist.github.com/xstefank/4c9c6f80932ee07898ba6c3bd63ce2f0. And the diff is here https://gist.github.com/xstefank/e81db0366b9ba60407e89065fc1f99ab which doesn't show any big difference so this PR can be merged I think.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
tomjenkinson: In particular I would like to know if the singletons thing is something in Quarkus or defined by a spec

rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
tomjenkinson: When you say it's not a JAX-RS container, what are you referring to please? Quarkus, WildFly, both?

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
mmusgrov: Yes it does seem like there are a lot more changes/deletions than the issue report would suggest. If we need to back out your OpenAPI.Swagger changes then we will end up removing a lot more than just the annotations. Ideally we would want these extra "clean up"/"whatever the rationale is" changes to go into a separate commit. In earlier versions of the JAX-RS spec this was the standard way to declare resources (and it's still supported).

Will you update the issue indicating why you need the extra changes, over and above the changes to the OpenAPI annotations.

rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
mmusgrov: > This is not a JAX-RS container so there is no scanning. I can try to play around it probably by manually adding all classes but this seems easier to me.

Will it still run in the Undertow container?

xstefank:
@tomjenkinson correct, I generated these on top of the current master (sorry) but yes, do I need to generate on this PR? I think it's clear that we are not changing behavior.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
tomjenkinson: Regarding CoordinatorContainerFilter.java (https://github.com/jbosstm/narayana/blob/master/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/CoordinatorContainerFilter.java#L53) I think it seems the default for Providers per Application in EE 10 is a single instance (https://jakarta.ee/specifications/restful-ws/3.1/jakarta-restful-ws-spec-3.1.html#lifecycle_and_environment)

tomjenkinson: Regarding the original question in this thread, it is resolved from my point of view and limited understanding of the it.

rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/objectstore/TestBase.java
```diff
@@ -50,6 +50,16 @@ public Set<Class<?>> getClasses() {
         }
     }
 
+    @ApplicationPath("/")
+    public static class LRACoordinator extends Application {
```
tomjenkinson, line 54: How about extending `JaxRsApplication` - would we get the OpenAPI deployed so it could be tested that there is no /lra-coordinator/lra-coordinator references in the yaml?

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/JaxRsApplication.java
tomjenkinson: @xstefank mainly just by adding the ApplicationPath annotation and setting it to "/" (as you added it to the new Applications below) but in the existing Application - I find that (on Quarkus) the openapi renders fine then on "/lra-coordinator": https://github.com/jbosstm/narayana/compare/jbosstm:master...tomjenkinson:JBTM-3596?expand=1

That being said, on WildFly the coordinator is currently deploying on "/lra-coordinator/lra-coordinator" so I am thinking that with this change it could be that the application will now be on "/lra-coordinator" so this will be a public API change.

xstefank:
@tomjenkinson so here are openapi docs from the PR branch - https://gist.github.com/xstefank/f9e67867bdf57675c1b02b8dc8c69736, https://gist.github.com/xstefank/def98a10aa50b20065d001b98edd9359, and diff https://gist.github.com/xstefank/41ce35174ad6ecb0cbd405dbe6647264. These are totally the same. I expect that server declaration in wildfly is the context that is correctly /lra-coordinator.

xstefank:
@tomjenkinson so indeed there is a change in wildlfy OpenAPI where the base path was changed from `/lra-coordinator/lra-coordinator` to just `/lra-coordinator` but this is indeed a different bug that we didn't notice. Because this path is hardcoded if you deploy LRA coordinator WAR underdifferent name, say test-lra-coordinator the OpenAPI looks like this:

```
---
openapi: 3.0.3
info:
  title: LRA Coordinator
  contact:
    name: Narayana
    url: https://narayana.io
  version: "1.0"
servers:
- url: /test-lra-coordinator
tags:
- name: LRA Coordinator
  description: "Operations to work with active LRAs (to start, to get a status, to\
    \ finish, etc.)"
- name: LRA Recovery
paths:
  /lra-coordinator/lra-coordinator:
    get:
      tags:
      - LRA Coordinator
      summary: Returns all LRAs
      description: Gets both active and recovering LRAs
      parameters:
...
```

As you can see the servers directive is correctly changed but the OpenAPI points users still to `/lra-coordinator/lra-coordinator` which returns 404 as the context is now different and thus the correct API to call is at `/test-lra-coordinator/lra-coordinator`.

xstefank:
@tomjenkinson yes it does fix the problem. Thanks for finding out what makes OpenAPI change the paths. But I don't think it's the correct solution because exactly as you mentioned we are combining default JAX-RS resource scanning and manual registrations. Which is unnecessary and problematic. My PR removes manual registrations which is the way I am seeing JAX-RS applications being developed today. But as always, if you want to go with minimal changes, close this PR and make a new one. Important is to have the issue fixed. We can also fix the manual registration in separate PR if you'd like. Just let's make some progress.

xstefank:
Closing as per comments.

PR #1971, opened 2022-05-17
Remove support for Java SE 8

mayankkunwar:
https://issues.redhat.com/browse/JBTM-3590

JDK11 !JDK8

!CORE !TOMCAT !AS_TESTS RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !PERF LRA !DB_TESTS !mysql !db2 !postgres !oracle


mmusgrov:
I haven't gone through the PR in much detail (mainly running grep commands on the sources) so I will certainly have missed some items, @jmfinelli can you check the PR as well.

Some items that I did notice that would need to be addressed include:

1. We track WildFly parent so update to the latest which is 39 (which has the benefit of enforcing the correct compiler version)

2. There are plenty of profiles for JDK 1.9 [2] and I think I saw one for JDK 1.8 as well

3. I see a number of poms still specifying 1.8 and 1.9 for maven.compiler.source [1]

4. Make sure we specify the maven.compiler.target and maven.compiler.source correctly, eg
```
  <maven.compiler.target>11</maven.compiler.target>
  <maven.compiler.source>11</maven.compiler.source>
```

5. I see some code in scripts/hudson/narayana.sh that does particular things depending upon on Java 8, that code needs updating.

6. Make sure the quickstart and performance repo still work with this change (in particular ensure those repos enforce the correct Java version).

7. Some of the CI axes are disabled (CORE for example). I think such a major change requires more indepth testing.

[1] find . -name pom.xml -exec grep "maven.compiler.source" {} \;
[2] find . -name pom.xml -type f -exec grep  "<jdk>" {} \;

mmusgrov:
There are also some shell scripts in the source tree that refer to older Java SE version. I'm not sure what they do and I don't know how/who uses them so but they should be checked anyway.

mayankkunwar:
Okay, thanks @tomjenkinson

PR #1972, opened 2022-05-17
JBTM-3590 Remove support for Java versions below SE 11 and upgrade quarkus dependency version

tomjenkinson:
https://issues.redhat.com/browse/JBTM-3590

JDK11 !JDK8

CORE TOMCAT AS_TESTS RTS JACOCO XTS QA_JTA QA_JTS_JACORB QA_JTS_JDKORB QA_JTS_OPENJDKORB PERF LRA DB_TESTS mysql db2 postgres oracle


mayankkunwar:
Thanks @tomjenkinson

mayankkunwar:
AS_TESTS failure is not related to the PR changes

mmusgrov:
@tomjenkinson I am investigating the hang.
I reproduced it once on my laptop so it should be diagnosable.

mmusgrov:
@tomjenkinson I'm not sure it's reproducible now after you added those extra 13 commits. I've ran the test suites successfully 6 times now without failure. Prior to those commits I got it to hang 1 out of 2 times.

mmusgrov:
TESTIT

mmusgrov:
@tomjenkinson the TCK is running against WildFly, not Quarkus so it must be some other change introduced by this PR.

I have set off a 100 runs on my laptop to see if it can be reproduced.

mmusgrov:
> I have set off a 100 runs on my laptop to see if it can be reproduced.

@tomjenkinson They all passed. I am leaving the current CI job/docker container running while I investigate what it's doing. Suffice to say that we cannot really merge this PR until we have a handle on why it keeps hanging in CI otherwise it will break our regular jobs

mmusgrov:
Remark: the following log message is also a potential bug resolution:

```[mmusgrov@dev2 narayana] (JBTM-3590-with-debug) $ git diff rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
diff --git a/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java b/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
index 4a6c58fa4..c0655b46a 100644
--- a/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
+++ b/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
@@ -170,8 +170,13 @@ public class LRAService {
     public void finished(LongRunningAction transaction, boolean fromHierarchy) {
         if (transaction.isFailed()) {
             getRM().moveEntryToFailedLRAPath(transaction.get_uid());
+            LRALogger.logger.tracef("LRAService: moved %s to the failed store", transaction.getId());
         }
+        // TODO else if (transaction.isRecovering()) {
+        // this might be the bug since it will be left in the recoveringLRAs cache but absent from the main store
+        // but don't fix it until we see the hang with extra debug on
         if (transaction.isRecovering()) {
+            LRALogger.logger.tracef("LRAService: caching recovering LRA %s", transaction.getId());
             recoveringLRAs.put(transaction.getId(), transaction);
         } else if (fromHierarchy || transaction.isTopLevel()) {
```

mmusgrov:
About the location of the store [1], I think it should be the standard WildFly data dir but that seems empty when I run the TCK, whenever I look, on my laptop. It also uses the same store location for two distinct WildFly instances which on the face of it looks wrong.

[1] https://github.com/jbosstm/narayana/blob/master/rts/lra/test/pom.xml#L201

scripts/hudson/narayana.sh
```diff
@@ -392,7 +393,7 @@ function clone_as {
     git reset --hard jbosstm/5_BRANCH
     [ $? -eq 0 ] || fatal "git reset 5_BRANCH failed"
     git clean -f -d -x
-    [ $? -gt 1 ] || fatal "git clean failed"
+    [ $? -ge 0 ] || fatal "git clean failed"
```
mmusgrov, line 396: Shouldn't this be `[ $? -ne 0 ] && fatal "git clean failed"` or `[ $? -eq 0 ] || fatal "git clean failed"`

mmusgrov (changes requested):
It looks good Tom and Mayank, I only had a couple of minor change requests.
We need to follow up with changes to the quickstart and performance repos since those are still expecting 1.8 for the target and source compiler settings, once this PR is integrated we need to follow up with those asap. Can you create a JIRA for it please.

scripts/hudson/narayana.sh
```diff
@@ -93,8 +94,8 @@ function init_test_options {
     elif [[ $PROFILE == "CORE" ]]; then
         if [[ ! $PULL_DESCRIPTION_BODY == *!MAIN* ]] && [[ ! $PULL_DESCRIPTION_BODY == *!CORE* ]]; then
           comment_on_pull "Started testing this pull request with $PROFILE profile: $BUILD_URL"
-          export AS_BUILD=0 AS_CLONE=0 AS_DOWNLOAD=0 AS_TESTS=0 NARAYANA_BUILD=1 NARAYANA_TESTS=1 XTS_AS_TESTS=0 XTS_TESTS=0 TXF_TESTS=0 txbridge=0
-          export RTS_AS_TESTS=0 RTS_TESTS=0 JTA_CDI_TESTS=0 QA_TESTS=0 SUN_ORB=0 JAC_ORB=0 JTA_AS_TESTS=0 OSGI_TESTS=1
+          export AS_BUILD=1 AS_CLONE=1 AS_DOWNLOAD=0 AS_TESTS=0 NARAYANA_BUILD=1 NARAYANA_TESTS=1 XTS_AS_TESTS=0 XTS_TESTS=0 TXF_TESTS=0 txbridge=0
```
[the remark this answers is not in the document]
tomjenkinson: I think possibly this should go in a different PR, I will pull it out of here. Briefly though these are tests that we provide in our repo, I think AS_TESTS axis should really just be for testing new versions of Narayana in the existing WildFly.

scripts/hudson/narayana.sh
tomjenkinson: I think I probably erred on caution but realised that it was a 0. Anyway, I will pull this commit out for now and put it in a different PR

pom.xml
```diff
@@ -180,7 +180,10 @@
     <jvm.args.modular></jvm.args.modular>
     <jvm.args.other>-server</jvm.args.other>
     <maven.build.timestamp.format>yyyy/MMM/dd HH:mm</maven.build.timestamp.format>
+    <maven.compiler.source>11</maven.compiler.source>
```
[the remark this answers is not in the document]
tomjenkinson: Done, I would mention that whole section should go when we merge https://github.com/jbosstm/narayana/pull/1987

mmusgrov:
@tomjenkinson would you be able to rebase against master to pick up the fix for JBTM-3590 (force participant and coordinator to use distinct stores).

scripts/hudson/narayana.sh
mmusgrov: I didn't mind it being fixed here since it was a tiny change, but if you want do in a separate PR then that's okay too even though it will consume CI resources (since you would need to run one of the AS_TESTS, XTS or RTS axes).

mmusgrov (approved):
Thanks

mmusgrov:
1) LRA run number 894 failed (https://issues.redhat.com/browse/JBTM-3617)
  - looks like a combination of having to put a timelimit on the tests (eventual consistency) and a slow machine
2) LRA run number 871 I cancelled, it was hanging
  - I assume this predated the test pom fix to force the participant and coordinator to use different stores
The JMS subsystem failures (in the AS-TESTS axis) look coincidental - do you know if the JMS tests are using a transaction manager that includes all of the changes made by this report, if not then we may safely ignore these failures as being unrelated.
3) I only briefly looked at the DBTESTS axis but I'd be surprised if they are caused by these changes and your analysis of a potential cause seems like reasonable justification for ignoring them. Besides, they passed on Friday and you haven't updated the PR since.

To summarise, @tomjenkinson , I still approve of the PR, I will leave it with you to merge.

PR #1983, opened 2022-06-15
[JBTM-3602] Compatibility issues with other Java vendors in WS-AT (on Hold)

rmartinc:
Issue: https://issues.redhat.com/browse/JBTM-3602

The PR removes the service-name and endpoint-name from any reference associated to WS-AT and WS-COOR specs. I tried to not modify anything else. Tests modified because sometimes they also configure the endpoints from the test code (this way we test this in all the possible places).

@mmusgrov I have also prepared the PR for 5.11 (it's the same, no conflicts) if you need it.

JDK11 CORE AS_TESTS XTS !RTS !JACOCO !TOMCAT !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !PERF !LRA !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/XTS/WSTX/classes/com/arjuna/mwlabs/wst11/at/ContextFactoryImple.java
+++ b/XTS/WSTX/classes/com/arjuna/mwlabs/wst11/at/ContextFactoryImple.java
@@ -350,13 +350,9 @@ public class ContextFactoryImple implements ContextFactory, LocalFactory
 
     private W3CEndpointReference getParticipant(final String id, final boolean isSecure)
     {
-        final QName serviceName = AtomicTransactionConstants.PARTICIPANT_SERVICE_QNAME;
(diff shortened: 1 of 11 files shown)


mmusgrov:
This issue is on hold pending a discussion about whether to merge it into the master branch or created a topic branch for it,

mmusgrov:
@rmartinc The use case that this PR was designed to facilitate is no longer required. Even though the contribution was very helpful during discussions about the requirement, I would now like to decline the PR but it will remain available if there is a requirement for something similar in the future. I will delay for a couple of more days to give you a chance to respond.

rmartinc:
@mmusgrov No problem! You can decline it and we can close JBTM-3602 too, as won't fix or similar.

PR #1985, opened 2022-06-16
[JBTM-3611] Add a jakarta-direct.properties for text strings that nee…

jamezp:
…d to be transformed.

https://issues.redhat.com/browse/JBTM-3611

Note that there are other modules that transform byte code. However, in WildFly these were the only ones that were causing issues. If we feel these changes should go into all the transformed modules please let me know and I'll make the changes.

### DIFF
--- a/XTS/jbossxts-jakarta/pom.xml
+++ b/XTS/jbossxts-jakarta/pom.xml
@@ -39,6 +39,8 @@
                   <arg value="-o"/>
                   <arg value="-tr"/>
                   <arg value="${project.basedir}/../../jakarta-rules/jakarta-renames.properties"/>
+                  <arg value="-td"/>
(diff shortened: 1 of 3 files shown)


zhfeng:
TESTIT

jmfinelli:
TESTIT

mmusgrov (approved):
Thanks very much @jamezp, I will arrange for a new narayana release and WFLY component upgrade PR early next week (we have two other issues I'd like to get into the release but I'd expect those to be done towards the middle of the week but if not then I will do a release anyway).

jamezp:
Okay, excellent. Thanks @mmusgrov! Just as an FYI I filed a WFLY upgrade issue already https://issues.redhat.com/browse/WFLY-16496.

PR #1988, opened 2022-06-21
JBTM-3596 Includes a test to verify current behaviour for WildFly for OpenAPI

tomjenkinson:
https://issues.redhat.com/browse/JBTM-3596

!CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !PERF LRA !DB_TESTS mysql db2 postgres oracle

Follows up on https://github.com/jbosstm/narayana/pull/1966

### DIFF
--- a/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
+++ b/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
@@ -32,6 +32,7 @@ import io.narayana.lra.logging.LRALogger;
 
 import javax.enterprise.context.ApplicationScoped;
 
+import javax.ws.rs.ApplicationPath;
 import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
 import javax.ws.rs.HeaderParam;
@@ -102,6 +103,7 @@ import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_CONTEXT
 import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_RECOVERY_HEADER;
 
 @ApplicationScoped
+@ApplicationPath("/")
 @Path(COORDINATOR_PATH_NAME)
 
 @OpenAPIDefinition(
--- a/rts/lra/test/basic/pom.xml
+++ b/rts/lra/test/basic/pom.xml
@@ -12,6 +12,8 @@
   <description>LRA basic integration tests</description>
 
   <properties>
+    <!-- This configuration allows WildFly to be ran with various extra subsystems, including an OpenAPI subsystem -->
+    <config.to.replace>standalone-microprofile.xml</config.to.replace>
     <lra.coordinator.jar.path>${project.basedir}/../../coordinator-quarkus/target/lra-coordinator-runner.jar</lra.coordinator.jar.path>
     <skip.wildfly.plugin>false</skip.wildfly.plugin>
   </properties>
--- a/rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/OpenAPIIT.java
+++ b/rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/OpenAPIIT.java
@@ -1,0 +1,63 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2022, Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package io.narayana.lra.arquillian;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.logging.Logger;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TestName;
+
+import javax.ws.rs.core.Response;
+import java.net.MalformedURLException;
+import java.net.URISyntaxException;
+import java.net.URL;
+
+public class OpenAPIIT extends TestBase {
+    private static final Logger log = Logger.getLogger(OpenAPIIT.class);
+
+    @Rule
+    public TestName testName = new TestName();
+
+    @Override
+    public void before() {
+        super.before();
+        log.info("Running test " + testName.getMethodName());
+    }
+
+    @Deployment
+    public static WebArchive deploy() {
+        return Deployer.deploy(OpenAPIIT.class.getSimpleName());
+    }
+
+    @Test
+    public void test() throws URISyntaxException, MalformedURLException {
+        URL url = new URL("http://" + System.getProperty("lra.coordinator.host", "localhost")+":"+ System.getProperty("lra.coordinator.port", "8080")+"/openapi");
+        Response response = client.target(url.toURI()).request().get();
+        String output = response.readEntity(String.class);
+        org.junit.Assert.assertTrue("WildFly OpenAPI document has paths at wrong location", !output.contains("/lra-coordinator/lra-coordinator:"));
+        org.junit.Assert.assertTrue("WildFly OpenAPI document does not have paths for expected location", output.contains("/lra-coordinator:"));
+        org.junit.Assert.assertTrue("WildFly OpenAPI document does not have server URL", output.contains("url: /lra-coordinator"));
+    }
+}
--- a/rts/lra/test/pom.xml
+++ b/rts/lra/test/pom.xml
@@ -110,6 +110,7 @@
         <module>crash</module>
       </modules>
       <properties>
+        <config.to.replace>standalone.xml</config.to.replace>
         <!-- Qualifier of the container hosting lra-coordinator -->
         <lra.coordinator.container.qualifier>lra-coordinator-server</lra.coordinator.container.qualifier>
         <!-- Qualifier of the lra-coordinator of the deployment -->
@@ -203,8 +204,9 @@
                   <target>
                     <zip basedir="../../" destfile="../../ObjectStore-${project.artifactId}.zip" includes="ObjectStore" whenempty="skip"></zip>
                     <delete dir="../../ObjectStore" failonerror="false"></delete>
-                    <copy file="${env.JBOSS_HOME}/standalone/configuration/standalone.xml" tofile="${env.JBOSS_HOME}/standalone/configuration/${lra.coordinator.xml.filename}"></copy>
-                    <copy file="${env.JBOSS_HOME}/standalone/configuration/standalone.xml" tofile="${env.JBOSS_HOME}/standalone/configuration/${lra.participant.xml.filename}"></copy>
+                    <!-- We overwrite the config files as the config.to.replace can be different in different modules and without overwrite being the case, then the file will not be written over -->
+                    <copy file="${env.JBOSS_HOME}/standalone/configuration/${config.to.replace}" overwrite="true" tofile="${env.JBOSS_HOME}/standalone/configuration/${lra.coordinator.xml.filename}"></copy>
+                    <copy file="${env.JBOSS_HOME}/standalone/configuration/${config.to.replace}" overwrite="true" tofile="${env.JBOSS_HOME}/standalone/configuration/${lra.participant.xml.filename}"></copy>
                   </target>
                 </configuration>
               </execution>


rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/OpenAPITestIT.java
```diff
@@ -0,0 +1,73 @@
...
+                SimpleLRAParticipant.class);
+    }
+
+    @Test
+    public void test() throws URISyntaxException, MalformedURLException {
+        assert(coordinatorUrl.endsWith("/lra-coordinator/lra-coordinator"));
+        URL url = new URL("http://" + System.getProperty("lra.coordinator.host", "localhost")+":"+ System.getProperty("lra.coordinator.port", "8080")+"/openapi");
+        Response response = client.target(url.toURI()).request().get();
+        String output = response.readEntity(String.class);
+        assert(output.contains("/lra-coordinator/lra-coordinator"));
```
xstefank, line 71: The test fails as this is not now part of the openAPI output. Also, why not use Junit Assert or Assertions (not sure which version we use here) with a more descriptive error message?

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/OpenAPITestIT.java
```diff
@@ -0,0 +1,73 @@
...
+    public static WebArchive deploy() {
+        return Deployer.deploy(OpenAPITestIT.class.getSimpleName(),
+                LRAParticipantWithStatusURI.class,
+                LRAParticipantWithoutStatusURI.class,
+                SimpleLRAParticipant.class);
+    }
+
+    @Test
+    public void test() throws URISyntaxException, MalformedURLException {
+        assert(coordinatorUrl.endsWith("/lra-coordinator/lra-coordinator"));
```
xstefank, line 67: not sure why this is needed? You are manually setting URL to call later, so why does it matter what is passed into the client?

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/OpenAPITestIT.java
```diff
@@ -0,0 +1,73 @@
...
+    @Override
+    public void before() {
+        super.before();
+        log.info("Running test " + testName.getMethodName());
+    }
+
+    @Deployment
+    public static WebArchive deploy() {
+        return Deployer.deploy(OpenAPITestIT.class.getSimpleName(),
+                LRAParticipantWithStatusURI.class,
```
xstefank, line 60: Does this test need participant classes if you are testing only coordinator? which is kind of side effect of the running tests but I think that's ok.

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/OpenAPITestIT.java
```diff
@@ -0,0 +1,73 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2019, Red Hat, Inc., and individual contributors
```
xstefank, line 3: 2022

pom.xml
```diff
@@ -259,6 +259,7 @@
     <version.org.sonatype.plugins.nexus-staging-maven-plugin>1.6.6</version.org.sonatype.plugins.nexus-staging-maven-plugin>
     <version.org.wildfly.arquillian>3.0.1.Final</version.org.wildfly.arquillian>
     <version.postgresql>42.2.20</version.postgresql>
+    <version.quarkus>2.9.2.Final</version.quarkus>
```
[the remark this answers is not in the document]
tomjenkinson: Quarkus upgrade some come in with #1972

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/OpenAPITestIT.java
tomjenkinson: Thank you

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/OpenAPITestIT.java
tomjenkinson: Good question - thanks for pointing it out, will try removing them

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/OpenAPITestIT.java
tomjenkinson: I think this could be coming from config. I think I had it at the start when trying to understand what to look for. I am going to remove this

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/OpenAPITestIT.java
tomjenkinson: I wonder if I can just have an empty set of participant classes, will try it

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/OpenAPITestIT.java
tomjenkinson: I don't quite understand how I missed it when preparing the test but the test is failing here on WildFly. I think this could be a problem then. The reason I say that is the endpoint will be different with this change so it's a public API change. I thought WildFly would stay on "/lra-coordinator/lra-coordinator". I am going to put a hold here.

tomjenkinson: I am finding that for WildFly the coordinator is indeed on /lra-coordinator/lra-coordinator but the openapi is now showing /lra-coordinator which I was not expecting. I do understand from the change what might cause this but it's not what I would expect we need.

tomjenkinson: I think my addition of https://github.com/jbosstm/narayana/blob/eed7548e37ed5a915ba92bdd8cc61c880605a39c/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java#L129 has changed this, at least from WildFly point of view. I need to find out what is expected when a class has an `@Path` and `@ApplicationPath`

tomjenkinson: I probably thought I was using JUnit assert - will change that - thanks for mentioning.

tomjenkinson: Done, thanks

rts/lra/test/basic/pom.xml
```diff
@@ -56,4 +57,18 @@
...
   </dependencies>
 
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-failsafe-plugin</artifactId>
+        <configuration>
+          <includes>
+            <include>**/*Test*.java</include>
```
mmusgrov, line 67: Why have you added this specification? How does it relate to your PR. There are a number of basic tests that don't have Test in their name (all of the integration tests for example), In fact the only thing that will now get tested is the new test you have added in this PR. In fact if you did want to be consistent with the naming then, since the failsafe plugin is for integration testing, `<include>**/*IT*.java</include>` would have been more natural. But why do you need explicitly tell the build which classes to use, all of the tests in the `src/test` directory need to be ran.

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/OpenAPITestIT.java
```diff
@@ -0,0 +1,66 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TestName;
+
+import javax.ws.rs.core.Response;
+import java.net.MalformedURLException;
+import java.net.URISyntaxException;
+import java.net.URL;
+
+public class OpenAPITestIT extends TestBase {
```
mmusgrov, line 38: `IT` already implies it's a Test so using the suffix `TestIT` is a tautology.
But, it is just a name so if you want to keep it then fine.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
```diff
@@ -125,6 +126,7 @@
         )
 )
 @Tag(name = "LRA Coordinator", description = "Operations to work with active LRAs (to start, to get a status, to finish, etc.)")
+//@ApplicationPath("/")
```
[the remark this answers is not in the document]
mmusgrov: I believe the new test added in this PR is expecting `lra-coordinator/lra-coordinator`. How does the fact that it's commented out here square with that?

rts/lra/test/basic/pom.xml
```diff
@@ -12,6 +12,7 @@
   <description>LRA basic integration tests</description>
 
   <properties>
+    <config.to.replace>standalone-microprofile.xml</config.to.replace>
```
mmusgrov, line 15: Are you sure this will override the value being set in the parent pom?

mmusgrov (changes requested):

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/OpenAPITestIT.java
mmusgrov: Actually using both `IT` and `Test` in the name means that it would not be possible to discriminate between unit and integration tests.

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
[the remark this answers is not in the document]
tomjenkinson: It started out as more than a test - I thought we had a fix for Quarkus too but it didn't work out so far, I will remove this commented out code from the PR

rts/lra/test/basic/pom.xml
tomjenkinson: The LRA axis was passing still I think so yes. In particular this change is to ensure that WildFly is ran with OpenAPI which is important for the test I am proposing to add. I will add a comment.

rts/lra/test/basic/pom.xml
tomjenkinson: I think it might not be needed - I can see some config here (https://github.com/jbosstm/narayana/blob/master/rts/lra/pom.xml#L134), I will try without it

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/OpenAPITestIT.java
tomjenkinson: I will rename it

rts/lra/test/pom.xml
```diff
@@ -90,6 +90,8 @@
             <lra.tck.suite.host>${test.application.host}</lra.tck.suite.host>
             <!-- not necessary for test/basic but for TCK it's necessary -->
             <lra.tck.suite.port>${test.application.port}</lra.tck.suite.port>
+            <lra.coordinator.host>${lra.coordinator.host}</lra.coordinator.host>
```
tomjenkinson, line 93: I am wondering this and the next line are added...

tomjenkinson: I don't see a reason, removing them

rts/lra/test/basic/src/test/java/io/narayana/lra/arquillian/OpenAPIIT.java
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TestName;
+
+import javax.ws.rs.core.Response;
+import java.net.MalformedURLException;
+import java.net.URISyntaxException;
+import java.net.URL;
+
+public class OpenAPIIT extends TestBase {
```
tomjenkinson, line 37: I removed fields `url` and `baseURL` (and it's related import) which were not needed

rts/lra/test/pom.xml
```diff
@@ -198,8 +199,8 @@
                 <phase>generate-test-resources</phase>
                 <configuration>
                   <target>
-                    <copy file="${env.JBOSS_HOME}/standalone/configuration/standalone.xml" tofile="${env.JBOSS_HOME}/standalone/configuration/${lra.coordinator.xml.filename}"></copy>
-                    <copy file="${env.JBOSS_HOME}/standalone/configuration/standalone.xml" tofile="${env.JBOSS_HOME}/standalone/configuration/${lra.participant.xml.filename}"></copy>
+                    <copy file="${env.JBOSS_HOME}/standalone/configuration/${config.to.replace}" overwrite="true" tofile="${env.JBOSS_HOME}/standalone/configuration/${lra.coordinator.xml.filename}"></copy>
+                    <copy file="${env.JBOSS_HOME}/standalone/configuration/${config.to.replace}" overwrite="true" tofile="${env.JBOSS_HOME}/standalone/configuration/${lra.participant.xml.filename}"></copy>
```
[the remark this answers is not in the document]
mmusgrov: setting overwrite to true is ok so I'll resolve this conversation

rts/lra/test/basic/pom.xml
mmusgrov: ok

mmusgrov (approved):
The changes look good, thanks @tomjenkinson for this one

xstefank (changes requested):
@tomjenkinson Since you removed `ApplicationPath("/")` the behavior is unchanged and thus this PR is *not* longer fixing the mentioned issue. So this is just adding now a test for the wrong behavior. So please don't merge it in this state as it's just further verifying the wrong openapi and swagger.

mmusgrov:
I have triggered build http://narayanaci1.eng.hst.ams2.redhat.com/job/btny-pulls-narayana/905/ to test this PR

xstefank:
@tomjenkinson so once again: this is a wrong behaviour. Try the following steps:

1. Start the server with `bin/standalone.sh -c standalone-microprofile.xml`
2. Deploy the coordinator but with different context (different name of war): `cp /path/to/lra-coordinator.war standalone/deployments/lra-coordinator-with-different-context-root.war`
3. :8080/openapi (trimmed)
```
---
openapi: 3.0.3
info:
  title: LRA Coordinator
  contact:
    name: Narayana
    url: https://narayana.io
  version: "1.0"
servers:
- url: /lra-coordinator-with-different-context-root
tags:
- name: LRA Coordinator
  description: "Operations to work with active LRAs (to start, to get a status, to\
    \ finish, etc.)"
- name: LRA Recovery
paths:
  /lra-coordinator/lra-coordinator:
    get:
      tags:
      - LRA Coordinator
     ...
```
4. so accessing `:8080/lra-coordinator/lra-coordinator` == 404
5. but accessing `:8080/lra-coordinator-with-different-context-root/lra-coordinator` 200 with correct response as this is the correct deployment root context

xstefank (approved):

mmusgrov (approved):

mmusgrov:
The changes look good, thanks (and I presume the path is different on Quarkus because the context root is `/` whereas on WildFly it needs to be something specific because many web apps can be deployed on WildFly so they need to be differentiated but with quarkus there is only one web app so `/` makes sense).

marcosgopen (approved):
Thanks for your contribution @tomjenkinson

PR #1990, opened 2022-06-23
JBTM-3607 force participant and coordinator to use distinct stores

mmusgrov:
https://issues.redhat.com/browse/JBTM-3607

!JDK8 JDK11

!CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_JACORB !QA_JTS_JDKORB !QA_JTS_OPENJDKORB !PERF LRA !DB_TESTS mysql db2 postgres oracle

Thanks @jmfinelli for the pom changes

### DIFF
--- a/rts/lra/test/pom.xml
+++ b/rts/lra/test/pom.xml
@@ -114,6 +114,8 @@
         <lra.coordinator.container.qualifier>lra-coordinator-server</lra.coordinator.container.qualifier>
         <!-- Qualifier of the lra-coordinator of the deployment -->
         <lra.coordinator.deployment.qualifier>lra-coordinator</lra.coordinator.deployment.qualifier>
+        <!-- The path where the Object Store of the coordinator should be saved -->
+        <lra.coordinator.object.store.path>${project.build.directory}/wfly_lra_objectstore</lra.coordinator.object.store.path>
         <!-- the AppServerCoordinatorDeploymentObserver creates the deployment named 'lra-coordinator' and deploys it to the app server,
              the last part of the path with value 'lra-coordinator' is defined by LRA coordinator JAX-RS endpoint -->
         <lra.coordinator.url>http://${lra.coordinator.host}:${lra.coordinator.port}/lra-coordinator/lra-coordinator</lra.coordinator.url>
@@ -130,6 +132,7 @@
              standalone.xml file provided in WildFly (${env.JBOSS_HOME}/standalone/configuration/standalone.xml). Please, see the plugins
              section of this profile (specifically, maven-antrun-plugin) for further info -->
         <lra.participant.xml.filename>lra-participant.xml</lra.participant.xml.filename>
+        <server.jvm.object.store.arg>-DObjectStoreEnvironmentBean.objectStoreDir=${lra.coordinator.object.store.path}</server.jvm.object.store.arg>
         <skip.wildfly.plugin>true</skip.wildfly.plugin>
         <!-- where the WFLY test application will be started at (the coordinator is war deployed for the same app server at the same port) -->
         <test.application.port>8080</test.application.port>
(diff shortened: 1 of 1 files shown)


rts/lra/test/pom.xml
```diff
@@ -27,7 +27,6 @@
 
   <properties>
 
-    <lra.coordinator.debug.params></lra.coordinator.debug.params>
```
tomjenkinson, line 30: @mmusgrov do you know what happens when this is read in at: https://github.com/jbosstm/narayana/blob/master/rts/lra/test/pom.xml#L359 - similar for the other removed ones. I presume it would be OK but I don't know. Also, I think it should go in a separate commit?

rts/lra/test/pom.xml
```diff
@@ -232,6 +231,10 @@
                 </goals>
                 <phase>pre-integration-test</phase>
                 <configuration>
+		  <java-opts>
+		    <java-opt>${server.jvm.object.store.arg}</java-opt>
+		    <!--<java-opt>-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005</java-opt>-->
```
tomjenkinson, line 236: I don't think this line should be in the JBTM-3607 commit, it could even be removed

jmfinelli: Maybe, instead of removing it completely, we could introduce a property to debug WildFly on-demand. For example, a property named `<wildfly-debug>` with a comment like "This property can be used to debug WildFly when it is appended to WildFly's JVM's arguments. Common use cases are: `maven-wildfly-plugin` and `arquillian's WildFly container`"

mmusgrov: I'll create a JIRA for it

rts/lra/test/pom.xml
mmusgrov: It won't do anything because `<argument>${lra.coordinator.debug.params}</argument>` would then correspond to no argument.

I will push another commit that does not do any clean up (and only does what's necessary to fix the bug).

jmfinelli: My 2c:
- Initially, this and the other properties were defined in the [properties](https://github.com/jbosstm/narayana/blob/master/rts/lra/test/pom.xml#L28-L55) section of the `lra-test` pom. Then, for some reason, they were moved in the profiles where they were actually utilised
- Removing these properties from the [properties](https://github.com/jbosstm/narayana/blob/master/rts/lra/test/pom.xml#L28-L55) of the `lra-test` pom does not have any effect on the `properties` sections in profiles, i.e. things still work
- About `lra.coordinator.debug.params` in particular, I think it was a mistake: `lra.coordinator.debug.params` should have been defined in the `properties` section of the [this](https://github.com/jbosstm/narayana/blob/master/rts/lra/test/pom.xml#L330) profile. For example, the `debug.lra.coordinator` [profile](https://github.com/jbosstm/narayana/blob/master/rts/lra/test/pom.xml#L304) defines `debug.lra.coordinator` in its `properties` section.

jmfinelli: As a consequence, I think that we can go for either removing [lra.coordinator.debug.params](https://github.com/jbosstm/narayana/blob/master/rts/lra/test/pom.xml#L359) from the `start.lra.coodinator.before.it.tests` profile or defining that property in the profile's `properties` section. I would +1 the first option

rts/lra/test/pom.xml
mmusgrov: https://issues.redhat.com/browse/JBTM-3619

rts/lra/test/pom.xml
mmusgrov: > @mmusgrov do you know what happens when this is read in at: https://github.com/jbosstm/narayana/blob/master/rts/lra/test/pom.xml#L359 - similar for the other removed ones. I presume it would be OK but I don't know. Also, I think it should go in a separate commit?

@tomjenkinson I have reworked the PR to only do what is necessary to achieve the separation of the store between the two instances.

jmfinelli: OK, I see why all those properties were left blank now (and then activated through profiles)...but we should write a README to use this pom :rofl:

mmusgrov:
@tomjenkinson and @jmfinelli can this one be merged now that I have addressed each of your comments?

jmfinelli:
LGTM :+1:

tomjenkinson (approved):
LGTM - thanks

PR #1992, opened 2022-06-29
CI SCRIPT: Trigger LRA on more than just rts/lra changes and rebase t…

tomjenkinson:
NO_TEST


.github/workflows/lra-test.yml
```diff
@@ -5,10 +5,22 @@ on:
     branches: [ master ]
     paths:
       - "rts/lra/**"
+      - "rts/pom.xml"
+      - "ArjunaCore/arjuna/**"
+      - "ArjunaCore/pom.xml"
+      - "pom.xml"
+      - "scripts/hudson/narayana.sh"
+      - ".github/workflows/lra-test.yml"
   pull_request:
```
mmusgrov, line 14: This and the following spec look identical. The manual says we can combine them using:
```
on: [push, pull_request]
```

mmusgrov (approved):
Approved with a suggested simplification.

.github/workflows/lra-test.yml
tomjenkinson: Thanks, @mmusgrov I have incorporated combination into the changeset. I didn't understand the notation at first but I think I have it working for this part now

PR #1993, opened 2022-06-30
JBTM-3624 JBTM-3626 JacORB (and don't provide OpenJDK to JacORB compiler classpath JBTM-3590)

tomjenkinson:
Consolidate the versions of JacORB: https://issues.redhat.com/browse/JBTM-3624

CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA
!QA_JTS_JACORB
!QA_JTS_JDKORB !QA_JTS_OPENJDKORB !PERF !LRA !DB_TESTS mysql db2 postgres oracle

Also includes a commit to deal with modular.jdk.args which should no longer be required since https://issues.redhat.com/browse/JBTM-3590 but the way I changed it also changes QA_JTS_JDKORB QA_JTS_OPENJDKORB. That change should be tested here and https://github.com/jbosstm/narayana/pull/1995


qa/run-tests.xml
```diff
@@ -73,19 +73,6 @@
   <property name="testoutput.zip.includes" value="testoutput/**,TEST*"/>
   <property name="profile" value=""/>
 
-  <condition property="modular.jdk.args" value="--add-modules=java.corba --upgrade-module-path ../ArjunaJTS/jtax/target/upgrademodulepath --add-exports java.corba/com.sun.corba.se.spi.orb=ALL-UNNAMED --add-exports java.corba/com.sun.corba.se.spi.ior=ALL-UNNAMED --add-exports java.corba/com.sun.corba.se.impl.ior=ALL-UNNAMED --add-opens=java.base/java.io=java.corba">
```
mmusgrov, line 76: The QA tests fail because `modular.jdk.args` is not set and ${modular.jdk.args} gets passed as the last element of the java command line so java wants it to be a classname to run. It should work if you just set `modular.jdk.args=` (ie the empty string) or maybe just get rid of `modular.jdk.args` completely - it get's added in run-tests.xml as an additional argument:

```<sysproperty key="additional.elements" value="${modular.jdk.args} @{additional.elements} ${objectStoreElements}"/>
```

tomjenkinson: Thanks @mmusgrov I will put the PR on hold so it can be reviewed when it is passing

mmusgrov: ah, I reviewed it too early, sorry about that

tomjenkinson: BTW, the QA_JTS worked - thanks for the help

ArjunaJTS/jtax/tests/classes/TransactionManagerGetStatusDifferenceDemonstrator.java
```diff
@@ -94,7 +94,7 @@ public void test() throws InvalidName, SystemException,
                     // com.arjuna.orbportability.internal.orbspecific.jacorb.orb.implementations.jacorb_2_0
 					assertTrue(
 							"Status: " + getStatusSync.getTransactionManagerGetStatus(),
-							getStatusSync.getTransactionManagerGetStatus() == Status.STATUS_NO_TRANSACTION);
```
tomjenkinson, line 97: @mmusgrov please can you take a look at this? When consolidating other parts of the build system to use JacORB 3.7 (something partially done in https://github.com/jbosstm/narayana/pull/1364/) it shows this has changed and now for JacORB when a Jakarta Transactions `Synchronization` is invoked in this manner it gets `Status.STATUS_COMMITTED` rather than `Status.STATUS_NO_TRANSACTION`. JacORB is then now consistent with the other ORBs and local JTA mode.

tomjenkinson: In regards to using 3.7 of JacORB, do you know if that version no longer needed whatever the the special patched version mentioned here had?: https://github.com/jbosstm/narayana/blob/master/ArjunaCore/scripts/jts-jacorb-setup-env.sh#L48

mmusgrov: I don't think the patched version is necessarily specific to narayana rather they were general fixes maintained by jboss/nick. I've looked at some of the patches in [1] and they seem like general fixes. I don't know whether or not Nick pushed the fixes to the jacorb project. That said, we only seem to use the bundled jacorb.jar in [2,3] and I don't know what those services are for (they are either historical or are used for the JTS + Corba Name Service in a docker container work that Gytis did [4] but, without checking it is just a guess.

Regarding the question about the Synchronization returning Status.STATUS_COMMITTED rather than Status.STATUS_NO_TRANSACTION: my input here is that since the transaction context is undefined in the after synch (according to the spec) the actual return value can be either. I would assume that we changed it because returning the actual status is more informative and was requested.

But in summary, since we are about to move JacORB and since the only issue seems to be what the after synch returns, I think we can explain this difference and move on, WDYT?

[1] https://repository.jboss.org/jacorb/
[2] ./ArjunaCore/arjuna/services/config/recoveryservice.conf
[3] ./ArjunaJTS/jts/services/config/transactionservice.conf
[4] https://github.com/jboss-dockerfiles/narayana

mmusgrov (approved):
I am provisionally approving this PR subject to the understanding that:

#1 The four test failures are due to the change in jacorb semantics of the after synchroniztion and that this PR will be quickly followed up with a PR from @jmfinelli  to remove JacORB completely;
#2 The PR is re-tested on the CORE axis. I tried running it on my laptop but it failed while Eclipse transforming the `XTS/localjunit/unit` using the command `./build.sh clean install -Prelease,community` hitting the issue [1] - however I am retesting to verify that it wasn't an issue at my end (I'm popping out now and will check my build in an hour or so);
#3 And an observation - the `Requirements` section in the top level README.md still says that to build narayana it's not sufficient to use a JRE because the build requires access to an IDL compiler so ensure that the JAVA_HOME environment variable is set correctly ...

[1] [main] INFO Transformer - Input does not exist [ /home/mmusgrov/src/forks/narayana/narayana/XTS/localjunit/unit/target/classes ] [ /home/mmusgrov/src/forks/narayana/narayana/XTS/localjunit/unit/target/classes ]

ArjunaJTS/jtax/tests/classes/TransactionManagerGetStatusDifferenceDemonstrator.java
mmusgrov: @jmfinelli This comment provides more context about why would like to quickly follow up with your PR.

jmfinelli:
> The four test failures are due to the change in jacorb semantics of the after synchroniztion and that this PR will be quickly followed up with a PR from @jmfinelli to remove JacORB completely;

I was holding the PR to remove JacORB as there were a couple of PRs interfering wth the modifications needed to remove JacORB. Those PRs got merged today so I can now rebase my branch on top of `master` and submit my PR.

ArjunaJTS/jtax/tests/classes/TransactionManagerGetStatusDifferenceDemonstrator.java
tomjenkinson: The bundled JacORB is what is used by standalone distributions: https://github.com/jbosstm/narayana/blob/master/narayana-full/src/main/assembly/bin.xml and during `qa/` tests: https://github.com/jbosstm/narayana/blob/master/qa/TaskImpl.properties.template#L40 (though this PR changes both of these to use the versions defined in poms now)

jmfinelli:
> We should consider closing this PR and merging [#2001](https://github.com/jbosstm/narayana/pull/2001) (assuming changes from this PR are in there) so that the removal can be merged in the same PR

+1

PR #2009, opened 2022-07-18
JBTM-3646 Upgrade the jacocoant uber jar

mmusgrov:
https://issues.redhat.com/browse/JBTM-3646

I updated the shaded jar only (obtained from https://github.com/jacoco/jacoco/releases).
I couldn't find the shaded jar in https://mvnrepository.com/artifact/org.jacoco
I'd be interested to know if anyone knows how to avoid hard copying it.

!JACOCO !QA_JTS_OPENJDKORB
!CORE !TOMCAT !AS_TESTS !RTS !XTS !QA_JTA QA_!JTS_OPENJDKORB !PERF !LRA !DB_TESTS !mysql !db2 !postgres !oracle


jmfinelli:
[This](https://github.com/jbosstm/narayana/blob/master/qa/pom.xml#L146-L194) might be a good starting point

mmusgrov:
> [This](https://github.com/jbosstm/narayana/blob/master/qa/pom.xml#L146-L194) might be a good starting point EDIT: Forgot to link the [maven repository](https://mvnrepository.com/artifact/org.jacoco/org.jacoco.ant/0.8.8)

That isn't the shaded jar.

jmfinelli:
Why do we need the shaded jar of JaCoCo ant? It seems to me that we only use it in the `qa` module, don't we? AFAIK, the `qa` module is an `ant` project and it should not interfere with other maven dependencies...what am I missing?

mmusgrov:
If you try adding a non-shaded jar you will get errors, we need more than one of the jacoco artifacts to run the qa code coverage on the test suite. i didn't quite follow your answer or question but the qa test suite only works if we use the shaded jar which is only available via the download.

jmfinelli:
Re-reading my message...I figured out that I should have used more words/sentences (i.e. I was a bit cryptic) :sweat_smile:

Basically, why do we need to use a shaded JaCoCo Ant jar? Is there any particular reason? I asked that because it is possible to download an [uber JaCoCo Ant jar](https://www.jacoco.org/jacoco/trunk/doc/repo.html) instead of the shaded jar. This would include all dependencies and it seems to work with the `qa` module (I am currently running a test...it's been running for a while [without errors] so I hope I will get a result soon). AFAIK, shaded jars exist to avoid conflict between dependencies of the shaded jar and the actual maven project where the shaded jar is used. As our `qa` module is not a maven project (i.e. does not download any JaCoCo artifact) I was wondering if there is a particular reason to go for the shaded JaCoCo Ant jar instead of the JaCoCo Ant jar with `nodeps` classifer.

jmfinelli:
I did it again! I forgot to add in my previous message that, in my on-going test, I downloaded the uber JaCoCo Ant jar using:
```
<dependency>
  <groupId>org.jacoco</groupId>
  <artifactId>org.jacoco.ant</artifactId>
  <version>0.8.8</version>
  <classifer>nodeps</classifer>
</dependency>
```
Copied it into `qa/ext` using `maven-dependency-plugin` and changed the `run-tests.xml` file to point to the new jar file (`maven-dependency-plugin` copies it with a different name)

mmusgrov:
Yes the uber jar is what we want but I couldn't find a maven artifact for that, where is it?

ext/pom.xml
```diff
@@ -12,53 +13,56 @@
...
+  <packaging>pom</packaging>
+  <name>Narayana: ext</name>
+  <description>Narayana: Ext All</description>
 
-    <properties>
-        <sortpom.skip>true</sortpom.skip>
-    </properties>
+  <properties>
+    <sortpom.skip>false</sortpom.skip>
+  </properties>
```
jmfinelli, line 30: Thanks for applying sortpom to this module +1
Instead of switching sortpom off (i.e. `sortpom.skip = false`), just remove this property

mmusgrov: lol, I forgot, sortpom is more of a hindrance than and an aid.

mmusgrov: I only turned it on because I couldn't be bothered to format it manually since I had done a fair bit of cut and pasting into the pom.

mmusgrov: And it makes it hard to see what the PR actually changed.

mmusgrov: Although in this PR I only changed the plugin so it should be easy for you to see the functional changes.

jmfinelli: > Although in this PR I only changed the plugin so it should be easy for you to see the functional changes.

+1

code-coverage/pom.xml
```diff
@@ -257,11 +257,17 @@
...
                       <fileset dir="../ArjunaJTS/jts/target/idlj-surefire-reports" includes="**/*.xml"/>
+                      -->
                       <fileset dir="../ArjunaJTS/jts/target/idlj-openjdk-surefire-reports" includes="**/*.xml"/>
+                      <!--
                       <fileset dir="../ArjunaJTS/orbportability/target/idlj-surefire-reports" includes="**/*.xml"/>
+                      -->
                       <fileset dir="../ArjunaJTS/orbportability/target/idlj-openjdk-surefire-reports" includes="**/*.xml"/>
+                      <!--
                       <fileset dir="../ArjunaJTS/jtax/target/${surefire.reports.directory}" includes="**/*.xml"/>
+                      -->
```
jmfinelli, line 270: Yeah, I think those should have been removed in JBTM-3590...So WDYT if we remove them in this PR? (e.g. renaming this commit on the line `[JBTM-3590] left-over`)

ext/pom.xml
jmfinelli: Could you please remove this property from the pom? Sortpom is activated by default in the [main pom](https://github.com/jbosstm/narayana/blob/master/pom.xml#L187-L188)

ext/pom.xml
```diff
@@ -12,53 +13,56 @@
...
+                  <goal>copy</goal>
+                </goals>
+                <phase>package</phase>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+  </profiles>
```
jmfinelli, line 67: Is there any reason to split the removal of `maven-install-plugin` and the employment of `maven-dependency-plugin` in two commits? WDYT if we squash these two commits together?

jmfinelli (changes requested):
Logically, this looks good to me :-)
I only have a couple of minor changes I would like to discuss...but nothing that would stop this PR

ext/pom.xml
mmusgrov: There were three separate issues fixed in this PR, hence the reason I've kept them separate.

code-coverage/pom.xml
mmusgrov: They have been removed/disabled (I used `<!-- .. -->` XML style comments).

I left them as comments so that its clear we don't produce reports for the modules, I want them there as place holders just in case we are not including some tests in the report.

jmfinelli: IIRC, the `idlj` module was removed completely, hence `idlj-surefire-reports` cannot be produced any longer. These lines might introduce confusion as they point to a module that is only available in the git history.

ext/pom.xml
jmfinelli: Fair point, thanks for clarifying this :-)

mmusgrov:
The sortpom change triggered a build of the QA_JTS_OPENJDKORB axis which is a waste of resource so I deleted the run.

code-coverage/pom.xml
mmusgrov: The problem is that the JACOCO profile was never ran with the PRs that removed the code coverage profiles so we have only discovered the problem now. Guessing is the wrong option so I prefer to comment them out and check later whether we have the correct set of code coverage reports.

jmfinelli: OK, that makes sense. I will create a JIRA/add this to my ToDo list

jmfinelli (approved):

code-coverage/pom.xml
jmfinelli: [JBTM-3651](https://issues.redhat.com/browse/JBTM-3651)

PR #2015, opened 2022-07-22
JBTM-3648 LRA not usable in user deployments

xstefank:
https://issues.redhat.com/browse/JBTM-3648

!CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERF LRA !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/rts/lra/client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
+++ b/rts/lra/client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
@@ -1,2 +1,1 @@
 org.jboss.resteasy.client.jaxrs.internal.ResteasyClientBuilderImpl
-org.jboss.resteasy.microprofile.client.impl.MpClientBuilderImpl
--- a/rts/lra/client/pom.xml
+++ b/rts/lra/client/pom.xml
@@ -31,11 +31,6 @@
     </dependency>
     <dependency>
       <groupId>org.jboss.resteasy</groupId>
-      <artifactId>resteasy-client</artifactId>
-      <version>${version.org.jboss.resteasy}</version>
-    </dependency>
-    <dependency>
-      <groupId>org.jboss.resteasy</groupId>
       <artifactId>resteasy-jackson2-provider</artifactId>
       <version>${version.org.jboss.resteasy}</version>
     </dependency>


xstefank:
I think we can resolve [JBTM-3648](https://issues.redhat.com/browse/JBTM-3648) because it is verified manually. Test is orthogonal. I know it's just my word for now but as you said I am trying to convince @marcosgopen to use direct narayana dependencies in the quickstars so they would verify this case. Actually I found that I don't need this merged for my Quarkus LRA work so no worries for now, just needs to go to next narayana release.

mmusgrov (commented):
I think the change looks okay and if it passes the TCK then I'd be happy to see it merged.

But I believe @jmfinelli made the commit that you are reverting (and I don't know the backstory behind it) so he should make the final call.

jmfinelli (approved):
It looks good to me (and it passes tests :-) - IIRC, JBTM-3479 was created because tests were not passing...so +1 from me )

Context: unfortunately, I do not remember more than what I wrote in the description of JBTM-3479. I recall I had a chat about this with Ondra and @xstefank but that has been flushed ;-) sorry if I am not able to provide more info about this issue.

xstefank:
@jmfinelli haha, I don't remember what all we talk about too :). But the problem here is that WFLY doesn't show error, the endpoint is just not invoked so it's hard to find what is causing it. But if the dependencies were needed somewhere, we will find it soon enough. Note also that the TCK passes as Tom showed above, and I can also say that now we pass the TCK in Quarkus where I soon make a PR once we have a new release of narayana. So I am sufficiently confident that it's OK to do this change now.

rts/lra/client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
```diff
@@ -1,2 +1 @@
 org.jboss.resteasy.client.jaxrs.internal.ResteasyClientBuilderImpl
-org.jboss.resteasy.microprofile.client.impl.MpClientBuilderImpl
```
tomjenkinson, line 2: Do we know if it might be appropriate to remove `org.jboss.resteasy.microprofile.client.impl.MpClientBuilderImpl` from https://github.com/jbosstm/narayana/blob/e6f02e25950677ab06493839997447248e98c408/rts/lra/jaxrs/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder too?

rts/lra/client/src/main/resources/META-INF/services/javax.ws.rs.client.ClientBuilder
```diff
@@ -1,2 +1 @@
 org.jboss.resteasy.client.jaxrs.internal.ResteasyClientBuilderImpl
```
tomjenkinson, line 1: Perhaps as the resteasy-client Jar is removed this classname should not be here as I think it may come from the jar being removed? https://github.com/resteasy/Resteasy/blob/main/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ResteasyClientBuilderImpl.java

PR #2017, opened 2022-07-22
JBTM-3656 Race condition between adding and looking up WSBA recovery …

mmusgrov:
https://issues.redhat.com/browse/JBTM-3656

XTS
!CORE !TOMCAT !AS_TESTS !RTS !JACOCO !QA_JTA !QA_JTS_OPENJDKORB !PERF !LRA !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/XTS/WS-T/dev/src/org/jboss/jbossts/xts/recovery/participant/at/XTSATRecoveryManager.java
+++ b/XTS/WS-T/dev/src/org/jboss/jbossts/xts/recovery/participant/at/XTSATRecoveryManager.java
@@ -1,6 +1,7 @@
 package org.jboss.jbossts.xts.recovery.participant.at;
 
 import com.arjuna.ats.arjuna.common.Uid;
+import com.arjuna.webservices.logging.WSTLogger;
 
 import java.util.NoSuchElementException;
 
@@ -23,12 +24,16 @@ public abstract class XTSATRecoveryManager {
      */
     public static XTSATRecoveryManager getRecoveryManager()
     {
-        int i = 0;
-
-        while (theRecoveryManager == null && i++ < 2) {
-            try {
-                Thread.sleep(1000);
-            } catch (InterruptedException e) {
+        // use a lock to get the recovery manager to avoid a race with the code that sets it
+        synchronized (lock) {
+            // theRecoveryManager must eventually be set to a non-null value
+            while (theRecoveryManager == null) {
+                try {
+                    lock.wait();
+                } catch (InterruptedException e) {
+                    // reset the interrupted status
+                    Thread.currentThread().interrupt();
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


XTS/WS-T/dev/src/org/jboss/jbossts/xts/recovery/participant/at/XTSATRecoveryManager.java
```diff
@@ -23,12 +25,25 @@ public abstract class XTSATRecoveryManager {
...
+                    lock.wait(MAX_INIT_TIMEOUT_MILLIS);
+                } catch (InterruptedException e) {
+                    // reset the interrupted status
+                    Thread.currentThread().interrupt();
+                }
+
+                if (System.currentTimeMillis() > end) {
+                    if (theRecoveryManager == null) {
+                        WSTLogger.logger.warnf("%s recoveryManager has still not been set after %d ms",
+                                XTSBARecoveryManager.class, MAX_INIT_TIMEOUT_MILLIS);
```
tomjenkinson, line 43: @mmusgrov I think this should be XTSATRecoveryManager and the import on line 5 removed, do you agree?

XTS/WS-T/dev/src/org/jboss/jbossts/xts/recovery/participant/at/XTSATRecoveryManager.java
```diff
@@ -166,4 +191,6 @@ public static XTSATRecoveryManager setRecoveryManager(XTSATRecoveryManager recov
      */
 
     protected static XTSATRecoveryManager theRecoveryManager = null;
+    protected final static Object lock = new Object();
+    private final static int MAX_INIT_TIMEOUT_MILLIS = 10000;
```
tomjenkinson, line 195: I think we should just wait with no particular timeout, presumably when setRecoveryManager is not called that is some symptom of a significant misconfiguration?

tomjenkinson: If there is a good reason to not wait without timeout then I think some of the callers of this method need to be made more able to deal with null: https://github.com/jbosstm/narayana/search?p=1&q=getRecoveryManager

XTS/WS-T/dev/src/org/jboss/jbossts/xts/recovery/participant/at/XTSATRecoveryManager.java
mmusgrov: My IDE needs intelligent cut and paste ;-)

XTS/WS-T/dev/src/org/jboss/jbossts/xts/recovery/participant/at/XTSATRecoveryManager.java
mmusgrov: The server will hang - I thought that failing with a exception would be kinder.
But I agree that not setting it or setting it to zero would be an XTS implementation error (assuming that the application cannot set the recovery manager).

tomjenkinson (approved):
LGTM - thanks

PR #2018, opened 2022-07-22
Update PULL_REQUEST_TEMPLATE.md

tomjenkinson:
NO_TEST

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -7,3 +7,11 @@ Please refer to our [guidelines for making contributions](https://github.com/jbo
 The build axis can be controlled by prefixing a ! on the following as appropriate:
 
 CORE TOMCAT AS_TESTS RTS JACOCO XTS QA_JTA QA_JTS_OPENJDKORB PERF LRA DB_TESTS mysql db2 postgres oracle
+
+By default the pull request will run with `JDK11`, if you prefix a `!` to `JDK11` then it will not run with `JDK11`.
+
+By default the PR will not compile and run with JDK 17. To override this you need to concatenate the two sets of characters: `JDK` `17` as a single word and provide the result into the description.
+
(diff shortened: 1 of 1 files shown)


marcosgopen (approved):
It seems clear to me, thanks for improving the template.

marcosgopen:
@tomjenkinson Could we make the comment "NO_TEST" prevent the github actions from running?

PULL_REQUEST_TEMPLATE.md
```diff
@@ -7,3 +7,7 @@ Please refer to our [guidelines for making contributions](https://github.com/jbo
 The build axis can be controlled by prefixing a ! on the following as appropriate:
 
 CORE TOMCAT AS_TESTS RTS JACOCO XTS QA_JTA QA_JTS_OPENJDKORB PERF LRA DB_TESTS mysql db2 postgres oracle
+
+By default the pull request will run with `JDK11`, if you prefix a `!` to `JDK11` then it will not run with `JDK11`.
+
+To run with JDK 17 then you need to concatenate the two sets of characters: `JDK` `17` as a single word and provide the result into the description.
```
mmusgrov, line 13: ".... Provide the result into the description" is a bit clunky, how about "To run with JDK 17 make sure the string JDK17 is included in the PR description".

mmusgrov: Ah !JDK17 could be assumed to mean don't run with JDK17. So I would reword it as follows:

"By default the PR will not compile and run with JDK 17. To override this default ensure that the text `JDK17` is included in the PR description."

tomjenkinson: But if I put even that in the template (and someone forgets to remove the template wording) then JDK17 will be in the description even if not intentional. I wonder how often people don't remove the template though.

I am going to change a bit the start to the wording though to be what you have written: "By default the PR will not compile and run with JDK 17. To override this"

mmusgrov: lol, okay

PR #2020, opened 2022-07-25
RELEASE SCRIPT: Allow the list of REPOS to be tagged to be overridden

tomjenkinson:
NO_TEST

Unfortunately it is not fully complete (I think) as I haven't work out how to pass a REPO that needs to be renamed in the clone to the REPOS variable.

E.g. `REPOS='a "b c"' ./scripts/pre-release.sh` would assume three variables, 'a' '"b' and 'c"' rather than two: 'a' and 'b c'

That said, if you just want to tag say narayana on it's own or along with something like document or quickstarts this approach should work.


mmusgrov (changes requested):
Thanks for the note.
@tomjenkinson can we close (not merge) this one now.

PR #2021, opened 2022-07-25
JBTM-3654 Refactor JAX-RS in LRA Coordinator to use discovery

xstefank:
https://issues.redhat.com/browse/JBTM-3654

!CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERF LRA !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
+++ b/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
@@ -58,10 +58,8 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
 import org.eclipse.microprofile.faulttolerance.Bulkhead;
@@ -135,20 +133,6 @@ public class Coordinator extends Application {
 
     private final LRAService lraService;
     private final RecoveryCoordinator recoveryCoordinator;
-
-    @Override
-    public Set<Class<?>> getClasses() {
-        HashSet<Class<?>> classes = new HashSet<>();
-        classes.add(Coordinator.class);
-        return classes;
-    }
-
-    @Override
-    public Set<Object> getSingletons() {
-        Set<Object> containerFilters = new HashSet<>();
-        containerFilters.add(new CoordinatorContainerFilter());
-        return containerFilters;
-    }
 
     public Coordinator() {
         lraService = LRARecoveryModule.getService();
--- a/rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
+++ b/rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
@@ -326,6 +326,16 @@ public class LRATest {
         }
     }
 
+    @ApplicationPath("/")
+    public static class LRACoordinator extends Application {
+        @Override
+        public Set<Class<?>> getClasses() {
+            HashSet<Class<?>> classes = new HashSet<>();
+            classes.add(Coordinator.class);
+            return classes;
+        }
+    }
+
     @BeforeClass
     public static void start() {
         System.setProperty("lra.coordinator.url", TestPortProvider.generateURL('/' + COORDINATOR_PATH_NAME));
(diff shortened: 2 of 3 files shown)


xstefank:
@tomjenkinson this is reintroducing changes from https://github.com/jbosstm/narayana/pull/1966 as we discussed. The current Coordinator JAX-RS application class is manually specifying resources so the quickstart can't work since the Coordinator has no knowledge of the user JAX-RS resources. Switching to dynamic scanned loading of JAX-RS resources fixes this issue.

xstefank:
I personally, never seen REST resource extensing Application before. So this is why I don't think it's a good idea. Application represent a portable means of **configuration**, REST resource is an exposed REST API. So I believe two instances are created nevertheless and this mixing seems potentially problematic. But I will try to refactor to keep the Coordinator a sublcass of Application as I understand you don't want to break public API even thought I don't think someone uses is directly so the impact shouldn't exist.

mmusgrov:
I think the change looks okay.
Does it fix the quickstart failure that @marcosgopen was trying to fix?

xstefank:
@mmusgrov yes, it does. Now I am a little confuse whether I should try to do what Tom proposes or are you ok merging it like this?

mmusgrov:
Tom has done the review, I only skimmed it so go with his recommendations (also edited my last comment).

xstefank:
Yes, the spec says that the app should be scanned but if you deploy application class with manual configuration then you are choosing which resources to expose. In Quarkus generally usage of Application class is discouraged (Quarkus creates one for the user) and only recommended if you want to use the ApplicationPath or OpenAPI annotations like we do in lra-coordinator.

xstefank:
@tomjenkinson @mmusgrov adjusted as requested.

tomjenkinson (approved):
I am OK with the change to Coordinator and so adding approval. I note that the test has been modified to use an Application that declares specific classes - was that required from change to Coordinator? If so, will it affect user deployments somehow and if so, would it be expected to be covered in documentation? /cc @mmusgrov

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/api/Coordinator.java
```diff
@@ -136,20 +134,6 @@ public class Coordinator extends Application {
...
-        HashSet<Class<?>> classes = new HashSet<>();
-        classes.add(Coordinator.class);
-        return classes;
-    }
-
-    @Override
-    public Set<Object> getSingletons() {
-        Set<Object> containerFilters = new HashSet<>();
-        containerFilters.add(new CoordinatorContainerFilter());
-        return containerFilters;
```
marcosgopen, line 150: Thanks @xstefank for the PR, you safely removed this because the auto discovery see the @Provider annotation in CoordinatorContainerFilter. Am I right?

xstefank: Yes, we also specify them in the service file.

marcosgopen: Perfect, thanks!

xstefank:
@tomjenkinson I think we can merge this as is. Just FYI for another day, the Coordinator extending Application is currently also problem with native compilation in Quarkus :) This problem keeps coming up again and again.

mmusgrov:
> I am OK with the change to Coordinator and so adding approval. I note that the test has been modified to use an Application that declares specific classes - was that required from change to Coordinator? If so, will it affect user deployments somehow and if so, would it be expected to be covered in documentation? /cc @mmusgrov

The main validation of the functionality of adding user deployments is the embedded quickstart and since that works with this PR I have no objections to merging this one - unless there is further input we will merge it today,

marcosgopen:
Thanks @xstefank , I confirm your PR made work the two quickstarts I was working on [here](https://github.com/jbosstm/quickstart/pull/399) (cdi-participant and cdi-embedded) : http://narayanaci1.eng.hst.ams2.redhat.com/job/btny-pulls-narayana-quickstart-matrix/jdk=jdk17.latest,label=linux/51/
N.B.  The job failure is related to a separate issue (affecting other quickstarts).

marcosgopen:
> I am not certain it does work with the embedded quickstart - I tried a run on [jbosstm/quickstart#402](https://github.com/jbosstm/quickstart/pull/402) that failed but it was to do with Jersey client which IIUC seems like it's missing the dependencies so IMO not likely caused by this change. I am trying again with the "embedded" commit from https://github.com/jbosstm/quickstart/pull/399/commits too but that doesn't seem to change what goes on with the dependencies so it might need some more changes? /cc @marcosgopen

@tomjenkinson thanks for checking the integration with the two quickstarts. As discussed [here](https://github.com/jbosstm/quickstart/pull/399#discussion_r932126480) that was a missing dependency (causing that error) that I added later. 
As confirm in my last comment those quickstarts are now working.

PR #2023, opened 2022-07-25
Update pre-release.sh

tomjenkinson:
NO_TEST

I don't think the Narayana version is written in the files currently in https://github.com/jboss-dockerfiles/narayana and it is passed in the release script when building: https://github.com/jbosstm/narayana/blob/master/narayana-release-process.sh#L215


mmusgrov (approved):
Agreed, thanks

PR #2024, opened 2022-07-28
Upgrade Quarkus to 2.11.1.Final

tomjenkinson:
!CORE !TOMCAT !AS_TESTS RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERF LRA !DB_TESTS mysql db2 postgres oracle

PR #2026, opened 2022-07-29
Wildfly 26

mmusgrov:
CORE !TOMCAT AS_TESTS RTS !JACOCO XTS !QA_JTA !QA_JTS_OPENJDKORB !PERF LRA !DB_TESTS mysql db2 postgres oracle

JDK17 JDK11

### DIFF
--- a/scripts/hudson/narayana.sh
+++ b/scripts/hudson/narayana.sh
@@ -353,53 +353,37 @@ function build_narayana {
 }
 
 function clone_as {
-  echo "Cloning AS sources from https://github.com/jbosstm/jboss-as.git"
+  repo="https://github.com/wildfly/wildfly.git" # was https://github.com/jbosstm/jboss-as.git
+  AS_BRANCH="26.x"
+  echo "Cloning AS sources from $repo"
 
   cd ${WORKSPACE}
   if [ -d jboss-as ]; then
     echo "Updating existing checkout of WildFly"
     cd jboss-as
 
-    git remote | grep upstream
-    if [ $? -ne 0 ]; then
-      git remote add upstream https://github.com/wildfly/wildfly.git
-    fi
     #Abort any partially complete rebase
     git rebase --abort
-    git checkout 5_BRANCH
-    [ $? -eq 0 ] || fatal "git checkout 5_BRANCH failed"
+    git checkout $AS_BRANCH
+    [ $? -eq 0 ] || fatal "git checkout $AS_BRANCH failed"
     git fetch
-    [ $? -eq 0 ] || fatal "git fetch https://github.com/jbosstm/jboss-as.git failed"
-    git reset --hard jbosstm/5_BRANCH
-    [ $? -eq 0 ] || fatal "git reset 5_BRANCH failed"
+    [ $? -eq 0 ] || fatal "git fetch $repo failed"
     git clean -f -d -x
     [ $? -eq 0 ] || fatal "git clean failed"
     git rebase --abort
     rm -rf .git/rebase-apply
   else
     echo "First time checkout of WildFly"
-    git clone https://github.com/jbosstm/jboss-as.git -o jbosstm
-    [ $? -eq 0 ] || fatal "git clone https://github.com/jbosstm/jboss-as.git failed"
+    git clone https://github.com/wildfly/wildfly.git jboss-as
+    [ $? -eq 0 ] || fatal "git clone $repo failed"
 
     cd jboss-as
-
-    git remote add upstream https://github.com/wildfly/wildfly.git
+    git checkout $AS_BRANCH
   fi
+  git remote add upstream https://github.com/wildfly/wildfly.git
 
-  [ -z "$AS_BRANCH" ] || git fetch jbosstm +refs/pull/*/head:refs/remotes/jbosstm/pull/*/head
-  [ $? -eq 0 ] || fatal "git fetch of pulls failed"
-  [ -z "$AS_BRANCH" ] || git checkout $AS_BRANCH
-  [ $? -eq 0 ] || fatal "git fetch of pull branch failed"
-  [ -z "$AS_BRANCH" ] || echo "Using non-default AS_BRANCH: $AS_BRANCH"
-
-  git fetch upstream
-  echo "This is the JBoss-AS commit"
-  echo $(git rev-parse upstream/main)
-  echo "This is the AS_BRANCH $AS_BRANCH commit"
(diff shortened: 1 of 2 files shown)


scripts/hudson/narayana.sh
```diff
@@ -986,8 +971,13 @@ export ANT_OPTS="$ANT_OPTS $IPV6_OPTS"
...
 
 [ $NARAYANA_BUILD = 1 ] && build_narayana "$@"
-[ $AS_CLONE = 1 ] && clone_as "$@"
-[ $AS_BUILD = 1 ] && build_as "$@"
+
+if [ "$_jdk" -ge 17 ]; then
+  [ $AS_BUILD = 1 ] && AS_DOWNLOAD=1 # 26.x branch won't build on JDK 17 so download it instead
+else
+  [ $AS_CLONE = 1 ] && clone_as "$@" # why would we not set AS_BUILD=1 here
+  [ $AS_BUILD = 1 ] && build_as "$@" # should really set AS_DOWNLOAD=0 - not sure why we don't do that
```
tomjenkinson, line 979: I think this is script is not doing what you say as it may not have been seen necessary before and we rely on the PROFILE (or environment) being set "correctly" (such as someone not introducing a profile with AS_BUILD=1 and AS_DOWNLOAD=1 or someone not exporting AS_CLONE=1 and not exporting AS_BUILD=1 at the same time as neither of those permutations would seem to be useful).

What you propose to set AS_BUILD=1 or AS_DOWNLOAD=0 would be reasonable changes I think (as would instead some earlier validation and informing the person executing the script it is not right to set inappropriate combinations). If you do change any variables for the user, please consider sending a message to console anyway to let them know what you are doing.

scripts/hudson/narayana.sh
```diff
@@ -353,53 +353,37 @@ function build_narayana {
...
-    [ $? -eq 0 ] || fatal "git clone https://github.com/jbosstm/jboss-as.git failed"
+    git clone https://github.com/wildfly/wildfly.git jboss-as
+    [ $? -eq 0 ] || fatal "git clone $repo failed"
 
     cd jboss-as
-
-    git remote add upstream https://github.com/wildfly/wildfly.git
+    git checkout $AS_BRANCH
   fi
+  git remote add upstream https://github.com/wildfly/wildfly.git
```
tomjenkinson, line 383: This is an interesting change, the implication of no "-d jboss-as" is that the upstream remote would not exist and so we add the upstream remote if the directory was not present but it's not guaranteed that the directory would exist and the remote would not. If you want to move from the previous assumption I think you are going to need to set this adding of the remote into something like:
```git remote -v | grep upstream | grep https://github.com/wildfly/wildfly.git
  if [ $? = 1 ]; then
    git remote add upstream https://github.com/wildfly/wildfly.git
  fi````

Comment (tomjenkinson) [reply scripts/hudson/narayana.sh:383]: sorry I think I have misread this block

File: scripts/hudson/narayana.sh [hunk 3]
```diff
@@ -353,53 +353,37 @@ function build_narayana {
...
+  repo="https://github.com/wildfly/wildfly.git" # was https://github.com/jbosstm/jboss-as.git
+  AS_BRANCH="26.x"
+  echo "Cloning AS sources from $repo"
 
   cd ${WORKSPACE}
   if [ -d jboss-as ]; then
     echo "Updating existing checkout of WildFly"
     cd jboss-as
 
-    git remote | grep upstream
```
tomjenkinson, line 363: Why didn't you just leave these lines here and the lines 385-386? Was it cleanup? If so my comment around 383 I think is relevant still in case there already is a jboss-as directory

scripts/hudson/narayana.sh
tomjenkinson: I think is still be generally relevant but I had misread a bit and now I realised 386 was in the block where the directory did not exist )though as commented higher up around 363 there should be handling for the case where there may or may not be an "upstream" remote already

scripts/hudson/narayana.sh
```diff
@@ -986,8 +971,13 @@ export ANT_OPTS="$ANT_OPTS $IPV6_OPTS"
 # run the job
 
 [ $NARAYANA_BUILD = 1 ] && build_narayana "$@"
-[ $AS_CLONE = 1 ] && clone_as "$@"
-[ $AS_BUILD = 1 ] && build_as "$@"
+
+if [ "$_jdk" -ge 17 ]; then
+  [ $AS_BUILD = 1 ] && AS_DOWNLOAD=1 # 26.x branch won't build on JDK 17 so download it instead
```
tomjenkinson, line 976: Using a downloaded AS will just run the tests using the Narayana that is already in there rather than the Narayana that we are building so won't verify that changes in Narayana have not broken anything.

scripts/hudson/narayana.sh
mmusgrov: We need an upstream so that we can rebase against 26.x hence `git remote add upstream https://github.com/wildfly/wildfly.git`.

scripts/hudson/narayana.sh
mmusgrov: I didn't follow that comment.
But I do agree that we should bail out if the script requires a WildFly build.

scripts/hudson/narayana.sh
tomjenkinson: My point is that repeatedly running this script (say locally) would try to do the `git remote add upstream https://github.com/wildfly/wildfly.git` again and that should be protected. If you just run it without trying to see if it might be there already:
```
$ git remote add upstream https://github.com/wildfly/wildfly.git
error: remote upstream already exists.
```

scripts/hudson/narayana.sh
tomjenkinson: I was trying to explain why it probably wasn't doing what you were asking/unsure about but re-reading I can see that is not clear - I will edit the comment. I wasn't speaking about bailing out actually

tomjenkinson: (First comment edited now)

tomjenkinson: [BTW in https://github.com/jbosstm/narayana/pull/2024/files I am experimenting with an option that would use AS_CLONE=1 (so we get the app server integration tests) and AS_DOWNLOAD=1 (so we wouldn't build the app server - useful for JDK 17)]

mmusgrov:
I've removed my approval pending more discussion about what to do with JDK 17 builds

PR #2028, opened 2022-08-02
Wildfly jdk17

mmusgrov:
Build and test narayana using the most recent WildFly version that supports JDK 17 and javaee.

CORE RTS LRA XTS AS_TESTS
JDK17 JDK11
!TOMCAT !JACOCO XTS !QA_JTA !QA_JTS_OPENJDKORB !PERF !DB_TESTS mysql db2 postgres oracle

### DIFF
--- a/scripts/hudson/narayana.sh
+++ b/scripts/hudson/narayana.sh
@@ -189,7 +189,7 @@ function init_test_options {
     elif [[ $PROFILE == "LRA" ]]; then
         if [[ ! $PULL_DESCRIPTION_BODY == *!LRA* ]]; then
           comment_on_pull "Started testing this pull request with LRA profile: $BUILD_URL"
-          export AS_BUILD=0 AS_CLONE=0 AS_DOWNLOAD=1 AS_TESTS=0 NARAYANA_BUILD=0 NARAYANA_TESTS=0 XTS_AS_TESTS=0 XTS_TESTS=0 TXF_TESTS=0 txbridge=0
+          export AS_BUILD=1 AS_CLONE=1 AS_DOWNLOAD=0 AS_TESTS=0 NARAYANA_BUILD=1 NARAYANA_TESTS=0 XTS_AS_TESTS=0 XTS_TESTS=0 TXF_TESTS=0 txbridge=0
           export RTS_AS_TESTS=0 RTS_TESTS=0 JTA_CDI_TESTS=0 QA_TESTS=0 JAC_ORB=0 JTA_AS_TESTS=0
           export TOMCAT_TESTS=0 LRA_TESTS=1
         else
@@ -352,55 +352,23 @@ function build_narayana {
   return 0
 }
 
+# clone WildFly main
+# note that this version does not use 5_BRANCH
+# if we require changes to WildFly sources then we should either
+# a) raise a WildFly PR, or
+# b) or re-instate the part of the script that rebases 5_BRANCH
+# If the second option is needed it will be straightforward to reinstate rebasing against 5_BRANCH
 function clone_as {
-  echo "Cloning AS sources from https://github.com/jbosstm/jboss-as.git"
+  REPO="https://github.com/wildfly/wildfly.git"
+  echo "Cloning AS sources from $REPO"
 
-  cd ${WORKSPACE}
+  cd $WORKSPACE
+
   if [ -d jboss-as ]; then
-    echo "Updating existing checkout of WildFly"
-    cd jboss-as
-
-    git remote | grep upstream
-    if [ $? -ne 0 ]; then
-      git remote add upstream https://github.com/wildfly/wildfly.git
-    fi
-    #Abort any partially complete rebase
-    git rebase --abort
-    git checkout 5_BRANCH
-    [ $? -eq 0 ] || fatal "git checkout 5_BRANCH failed"
-    git fetch
-    [ $? -eq 0 ] || fatal "git fetch https://github.com/jbosstm/jboss-as.git failed"
-    git reset --hard jbosstm/5_BRANCH
-    [ $? -eq 0 ] || fatal "git reset 5_BRANCH failed"
-    git clean -f -d -x
-    [ $? -eq 0 ] || fatal "git clean failed"
-    git rebase --abort
-    rm -rf .git/rebase-apply
-  else
-    echo "First time checkout of WildFly"
-    git clone https://github.com/jbosstm/jboss-as.git -o jbosstm
-    [ $? -eq 0 ] || fatal "git clone https://github.com/jbosstm/jboss-as.git failed"
-
-    cd jboss-as
-
-    git remote add upstream https://github.com/wildfly/wildfly.git
+    rm -rf jboss-as # start afresh
   fi
 
-  [ -z "$AS_BRANCH" ] || git fetch jbosstm +refs/pull/*/head:refs/remotes/jbosstm/pull/*/head
-  [ $? -eq 0 ] || fatal "git fetch of pulls failed"
-  [ -z "$AS_BRANCH" ] || git checkout $AS_BRANCH
-  [ $? -eq 0 ] || fatal "git fetch of pull branch failed"
-  [ -z "$AS_BRANCH" ] || echo "Using non-default AS_BRANCH: $AS_BRANCH"
-
-  git fetch upstream
-  echo "This is the JBoss-AS commit"
-  echo $(git rev-parse upstream/main)
-  echo "This is the AS_BRANCH $AS_BRANCH commit"
-  echo $(git rev-parse HEAD)
-
-  echo "Rebasing the wildfly upstream/main on top of the AS_BRANCH $AS_BRANCH"
-  git pull --rebase upstream main
-  [ $? -eq 0 ] || fatal "git rebase failed"
+  git clone $REPO jboss-as || fatal "git clone $REPO failed"
 
   if [ $REDUCE_SPACE = 1 ]; then
     echo "Deleting git dir to reduce disk usage"
@@ -411,9 +379,12 @@ function clone_as {
 }
 
 function build_as {
-  echo "Building JBoss EAP/WildFly"
+  AS_BRANCH=27.0.0.Alpha1
+  echo "Building WildFly $AS_BRANCH"
 
   cd $WORKSPACE/jboss-as
+
+  git checkout $AS_BRANCH || fatal "git checkout $AS_BRANCH failed"
 
   # building WildFly
   [ "$_jdk" -lt 17 ] && export MAVEN_OPTS="-XX:MaxMetaspaceSize=512m -XX:+UseConcMarkSweepGC $MAVEN_OPTS"
@@ -453,8 +424,10 @@ function download_as {
   # clean up any previously downloaded zip files (this will not clean up old directories)
   rm -f artifacts.zip wildfly-*.zip
 
-  # download the latest wildfly nighly build (which we know supports Java 11)
+  # download the latest wildfly nighly build (which we know supports Java 17)
+  # re-enable downloading main when narayana migrates to jakarta - see JBTM-3588
   AS_LOCATION=${AS_LOCATION:-https://ci.wildfly.org/guestAuth/repository/downloadAll/WF_Nightly/.lastSuccessful/artifacts.zip}
+  #AS_LOCATION="https://ci.wildfly.org/guestAuth/repository/downloadAll/WF_26xNightlyJobs_Nightly/.lastSuccessful/artifacts.zip"
   wget -nv ${AS_LOCATION}
   ### The following sequence of unzipping wrapping zip files is a way how to process the WildFly nightly build ZIP structure
   ### which is changing time to time
@@ -475,7 +448,7 @@ function download_as {
 
   [ -d "${JBOSS_HOME}" ] || fatal "After unzipping the file '$zip', '${JBOSS_HOME}' does not exist"
 
-  echo JBOSS_HOME=$JBOSS_HOME
+  echo "JBOSS_HOME=$JBOSS_HOME"
 
   # clean up downloaded zip files
   rm -f wildfly-*.zip artifacts.zip
@@ -543,7 +516,6 @@ function rts_tests {
 
 function lra_tests {
   echo "#0. LRA Test"
-  ./build.sh install -pl ArjunaCore/arjunacore,rts -am -DskipTests -Pcommunity "$@"
   cd ./rts/lra/
   echo "#0. Running LRA tests using $ARQ_PROF profile"
   PRESERVE_WORKING_DIR=true ../../build.sh -fae -B -Pcommunity -P$ARQ_PROF $CODE_COVERAGE_ARGS $ENABLE_LRA_TRACE_LOGS -Dlra.test.timeout.factor="${LRA_TEST_TIMEOUT_FACTOR:-1.5}" "$@"
--- a/pom.xml
+++ b/pom.xml
@@ -262,7 +262,7 @@
     <version.org.sonatype.plugins.nexus-staging-maven-plugin>1.6.6</version.org.sonatype.plugins.nexus-staging-maven-plugin>
     <version.org.wildfly.arquillian>3.0.1.Final</version.org.wildfly.arquillian>
     <version.postgresql>42.3.1</version.postgresql>
-    <version.quarkus>2.10.0.Final</version.quarkus>
+    <version.quarkus>2.11.1.Final</version.quarkus>
     <version.smallrye-converter-api>1.0.10</version.smallrye-converter-api>
     <version.sortpom>3.0.0</version.sortpom>
     <version.sun.jdk>jdk</version.sun.jdk>


scripts/hudson/narayana.sh
```diff
@@ -411,20 +411,27 @@ function clone_as {
...
   [ "$_jdk" -ge 17 ] && export MAVEN_OPTS="-XX:MaxMetaspaceSize=512m $MAVEN_OPTS"
-  JAVA_OPTS="-Xms1303m -Xmx1303m -XX:MaxMetaspaceSize=512m $JAVA_OPTS" ./build.sh clean install -B -DskipTests -Dts.smoke=false $IPV6_OPTS -Dversion.org.jboss.narayana=${NARAYANA_CURRENT_VERSION} "$@"
-  [ $? -eq 0 ] || fatal "AS build failed"
+  JAVA_OPTS="-Xms1303m -Xmx1303m -XX:MaxMetaspaceSize=512m $JAVA_OPTS" ./build.sh clean install -B -DskipTests -Dts.smoke=false $IPV6_OPTS -Dversion.org.jboss.narayana=${NARAYANA_CURRENT_VERSION} "$@" || fatal "AS build failed"
 
   WILDFLY_VERSION_FROM_JBOSS_AS=`awk '/wildfly-parent/ { while(!/<version>/) {getline;} print; }' ${WORKSPACE}/jboss-as/pom.xml | cut -d \< -f 2|cut -d \> -f 2`
   echo "AS version is ${WILDFLY_VERSION_FROM_JBOSS_AS}"
-  JBOSS_HOME=${WORKSPACE}/jboss-as/build/target/wildfly-${WILDFLY_VERSION_FROM_JBOSS_AS}
-  export JBOSS_HOME=`echo  $JBOSS_HOME`
+  export JBOSS_HOME=${WORKSPACE}/jboss-as/build/target/wildfly-${WILDFLY_VERSION_FROM_JBOSS_AS}
```
tomjenkinson, line 434: Am I right that this is tidy up?

scripts/hudson/narayana.sh
```diff
@@ -411,20 +411,27 @@ function clone_as {
...
+  cd jboss-as
+
+  git checkout $AS_BRANCH || fatal "git checkout $AS_BRANCH failed"
 
   # building WildFly
   [ "$_jdk" -lt 17 ] && export MAVEN_OPTS="-XX:MaxMetaspaceSize=512m -XX:+UseConcMarkSweepGC $MAVEN_OPTS"
   [ "$_jdk" -ge 17 ] && export MAVEN_OPTS="-XX:MaxMetaspaceSize=512m $MAVEN_OPTS"
-  JAVA_OPTS="-Xms1303m -Xmx1303m -XX:MaxMetaspaceSize=512m $JAVA_OPTS" ./build.sh clean install -B -DskipTests -Dts.smoke=false $IPV6_OPTS -Dversion.org.jboss.narayana=${NARAYANA_CURRENT_VERSION} "$@"
-  [ $? -eq 0 ] || fatal "AS build failed"
+  JAVA_OPTS="-Xms1303m -Xmx1303m -XX:MaxMetaspaceSize=512m $JAVA_OPTS" ./build.sh clean install -B -DskipTests -Dts.smoke=false $IPV6_OPTS -Dversion.org.jboss.narayana=${NARAYANA_CURRENT_VERSION} "$@" || fatal "AS build failed"
```
tomjenkinson, line 430: Is this tidy up?

scripts/hudson/narayana.sh
```diff
@@ -468,15 +477,14 @@ function download_as {
   zip=$(ls wildfly-*-SNAPSHOT.zip) # example the current latest is wildfly-preview-27.0.0.Beta1-SNAPSHOT.zip
 
   export JBOSS_HOME=${JBOSS_HOME:-"${PWD}/${zip%.*}"}
+  echo "JBOSS_HOME=$JBOSS_HOME"
```
tomjenkinson, line 480: I think this is tidy up too, perhaps to state earlier in the build what JBOSS_HOME is set to? That seems a good thing to do in a tidy up commit - were the "" required?

scripts/hudson/narayana.sh
```diff
@@ -985,10 +993,10 @@ export ANT_OPTS="$ANT_OPTS $IPV6_OPTS"
 
 # run the job
 
-[ $NARAYANA_BUILD = 1 ] && build_narayana "$@"
-[ $AS_CLONE = 1 ] && clone_as "$@"
+[ $NARAYANA_BUILD = 1 ] && build_narayana "$@" # NB we should do this after building the AS since it takes longer
```
tomjenkinson, line 996: I don't understand the comment but I don't think there is a functional change here yet but what I would say is building Narayana after the AS would be wrong because then the AS would not be able to see (and include) the built Narayana

scripts/hudson/narayana.sh
```diff
@@ -985,10 +993,10 @@ export ANT_OPTS="$ANT_OPTS $IPV6_OPTS"
...
+[ $AS_CLONE = 1 ] && AS_BUILD=1 # keep it simple and assume that we are cloning the AS because we want to build it
+[ $AS_DOWNLOAD = 1 ] && AS_BUILD=1 # keep it simple and build it from scratch - we can revisit it later if required
```
tomjenkinson, line 998: I think this is confusing, I think you could just "re-program" the PROFILE for those that have AS_DOWLOAD=1 to be AS_CLONE=1 and AS_BUILD=1 instead

tomjenkinson: It seems it's only the LRA PROFILE: https://github.com/jbosstm/narayana/blob/master/scripts/hudson/narayana.sh#L192

tomjenkinson: If you change
`export AS_BUILD=0 AS_CLONE=0 AS_DOWNLOAD=1 AS_TESTS=0 NARAYANA_BUILD=0 NARAYANA_TESTS=0 XTS_AS_TESTS=0 XTS_TESTS=0 TXF_TESTS=0 txbridge=0`
to:
`export AS_BUILD=1 AS_CLONE=1 AS_DOWNLOAD=0 AS_TESTS=0 NARAYANA_BUILD=0 NARAYANA_TESTS=0 XTS_AS_TESTS=0 XTS_TESTS=0 TXF_TESTS=0 txbridge=0` I think it would accomplish what you are trying to do with this change

scripts/hudson/narayana.sh
```diff
@@ -453,8 +460,10 @@ function download_as {
   # clean up any previously downloaded zip files (this will not clean up old directories)
   rm -f artifacts.zip wildfly-*.zip
 
-  # download the latest wildfly nighly build (which we know supports Java 11)
-  AS_LOCATION=${AS_LOCATION:-https://ci.wildfly.org/guestAuth/repository/downloadAll/WF_Nightly/.lastSuccessful/artifacts.zip}
+  # download the latest wildfly nighly build (which we know supports Java 17)
+  # re-enable downloading main when narayana migrates to jakarta - see JBTM-3588
+  #AS_LOCATION=${AS_LOCATION:-https://ci.wildfly.org/guestAuth/repository/downloadAll/WF_Nightly/.lastSuccessful/artifacts.zip}
+  AS_LOCATION="https://ci.wildfly.org/guestAuth/repository/downloadAll/WF_26xNightlyJobs_Nightly/.lastSuccessful/artifacts.zip"
```
tomjenkinson, line 466: Given you aren't intending to use `AS_LOCATION="https://ci.wildfly.org/guestAuth/repository/downloadAll/WF_26xNightlyJobs_Nightly/.lastSuccessful/artifacts.zip"` from `AS_LOCATION=${AS_LOCATION:-https://ci.wildfly.org/guestAuth/repository/downloadAll/WF_Nightly/.lastSuccessful/artifacts.zip}` is not necessary. You could still add the comment you proposed

scripts/hudson/narayana.sh
```diff
@@ -411,20 +411,27 @@ function clone_as {
 }
 
 function build_as {
-  echo "Building JBoss EAP/WildFly"
+  REPO="https://github.com/wildfly/wildfly.git"
+  AS_BRANCH=27.0.0.Alpha1
```
tomjenkinson, line 415: I will import this tag into the Narayana fork

tomjenkinson: It's there now: https://github.com/jbosstm/jboss-as/tree/27.0.0.Alpha1

tomjenkinson: Something to consider is that 27.0.0.Alpha1 is no longer supported by WildFly I expect

scripts/hudson/narayana.sh
```diff
@@ -189,7 +189,7 @@ function init_test_options {
     elif [[ $PROFILE == "LRA" ]]; then
         if [[ ! $PULL_DESCRIPTION_BODY == *!LRA* ]]; then
           comment_on_pull "Started testing this pull request with LRA profile: $BUILD_URL"
-          export AS_BUILD=0 AS_CLONE=0 AS_DOWNLOAD=1 AS_TESTS=0 NARAYANA_BUILD=0 NARAYANA_TESTS=0 XTS_AS_TESTS=0 XTS_TESTS=0 TXF_TESTS=0 txbridge=0
+          export AS_BUILD=0 AS_CLONE=0 AS_DOWNLOAD=1 AS_TESTS=0 NARAYANA_BUILD=1 NARAYANA_TESTS=0 XTS_AS_TESTS=0 XTS_TESTS=0 TXF_TESTS=0 txbridge=0
```
tomjenkinson, line 192: If you want to use a cloned and build version of the AS you could change `AS_BUILD=0 AS_CLONE=0 AS_DOWNLOAD=1` to `AS_BUILD=1 AS_CLONE=1 AS_DOWNLOAD=0`

scripts/hudson/narayana.sh
mmusgrov: I didn't understand the point of it, so yes I'd say it is.

scripts/hudson/narayana.sh
mmusgrov: It is a simplification to reduce the cognitive overload, if that's clean up then yes.

scripts/hudson/narayana.sh
mmusgrov: The bash script runs with `set -x` so it just adds noise.

scripts/hudson/narayana.sh
mmusgrov: Yes, I noticed that, the last push did something similar to what you suggested but I will take another look ...

scripts/hudson/narayana.sh
mmusgrov: Yes I spotted that one too and my last push corrected it.

scripts/hudson/narayana.sh
mmusgrov: Hopefully we won't need this tag for long if we quickly move to support native jakarta.
My primary goal is being able to create narayana releases.
I am cloning WildFly directly and ignoring 5_BRANCH (but like I say hopefully we won't need my workarounds for long).

scripts/hudson/narayana.sh
mmusgrov: Okay but my script changes makes the `AS_CLONE` variable redundant since I always build WildFly if the build needs to test against WildFly.

scripts/hudson/narayana.sh
```diff
@@ -543,7 +551,6 @@ function rts_tests {
 
 function lra_tests {
   echo "#0. LRA Test"
-  ./build.sh install -pl ArjunaCore/arjunacore,rts -am -DskipTests -Pcommunity "$@"
```
tomjenkinson, line 546: I think this should be a separate commit to the WFLY changes but I think it's fair to do it (so long as NARAYANA_BUILD=1 is set for the LRA profile - which I see it is proposed to be)

scripts/hudson/narayana.sh
tomjenkinson: I am not sure if I understand what you are responding to, I could imagine your response being why this type of line would be removed (rather than moved).

scripts/hudson/narayana.sh
```diff
@@ -985,10 +992,10 @@ export ANT_OPTS="$ANT_OPTS $IPV6_OPTS"
 
 # run the job
 
-[ $NARAYANA_BUILD = 1 ] && build_narayana "$@"
-[ $AS_CLONE = 1 ] && clone_as "$@"
+[ $NARAYANA_BUILD = 1 ] && build_narayana "$@" # NB we should do this after building the AS since it takes longer
+[ $AS_CLONE = 1 ] && AS_BUILD=1 # keep it simple and assume that we are cloning the AS because we want to build it
```
tomjenkinson, line 996: I have found a scenario where AS_CLONE=1 and AS_BUILD=0 seems to be useful so I don't think we should prevent that. I am also not spotting any existing PROFILE in https://github.com/jbosstm/narayana/blob/master/scripts/hudson/narayana.sh set as `AS_BUILD=0 AS_CLONE=1` so I don't think it should be needed to changed for this pull request.

scripts/hudson/narayana.sh
tomjenkinson: I think reconfiguring the PROFILE here would be clearer

scripts/hudson/narayana.sh
tomjenkinson: I think any changes related to cloning the AS should go into the clone_as function: https://github.com/jbosstm/narayana/blob/master/scripts/hudson/narayana.sh#L355

scripts/hudson/narayana.sh
tomjenkinson: We use the pattern `[ $? -eq 0 ] || fatal` many places in https://github.com/jbosstm/narayana/blob/master/scripts/hudson/narayana.sh. If a change is necessary, I think they should all be proposed to be changed at once but I would suggest to discuss the change on Zulip and give time to receive feedback.

scripts/hudson/narayana.sh
tomjenkinson: I think it looks better as you propose it, but I think the tidy up should go in a separate commit

scripts/hudson/narayana.sh
mmusgrov: What was the scenario?

scripts/hudson/narayana.sh
mmusgrov: I put some comments about the complexity/performance trade-off and why simplification should trump generality. But if it is going to help getting the PR passed then I can re-instate some of the complexity.

scripts/hudson/narayana.sh
tomjenkinson: To be able to run the tests in the WildFly repo against a downloaded WildFly - that being said to make that useful some further changes in narayana.sh are needed to overwrite the Narayana jars in the downloaded WildFly.

If you want to go this route I think it separate to this pull request. I don't expect that would cause problems for this pull request because I don't think any of the profiles today set AS_CLONE=1 and AS_BUILD=0

scripts/hudson/narayana.sh
mmusgrov: I'm not aware of any scenario where one would want to clone the AS and not build it.
But if you require it then I will look into re-instating AS_CLONE.

scripts/hudson/narayana.sh
```diff
@@ -985,10 +992,10 @@ export ANT_OPTS="$ANT_OPTS $IPV6_OPTS"
...
+[ $AS_DOWNLOAD = 1 ] && AS_BUILD=1 # keep it simple and build it from scratch - we can revisit it later if required
 [ $AS_BUILD = 1 ] && build_as "$@"
-[ $AS_DOWNLOAD = 1 ] && download_as "$@"
```
tomjenkinson, line 991: If you want to remove the ability to download the AS I think you should propose to remove more than this one line. But I think that should be discussed and proposed separate to this pull request. To prevent any of the current PROFILE from trying to download I think you should just change the LRA PROFILE to change AS_DOWNLOAD=1 to AS_DOWNLOAD=0 (as well as changing both AS_BUILD and AS_CLONE to 1)

scripts/hudson/narayana.sh
tomjenkinson: If clone_as function remains then I think adding cloning functionality to the build_as function will add complexity.

scripts/hudson/narayana.sh
tomjenkinson: To be clear, the other things to remove would include the download_as function, as well as any reference to AS_DOWNLOAD in the script

mmusgrov: @tomjenkinson I have argued elsewhere in this PR about the complexity/performance trade-off and why simplification should trump generality. AS_DOWNLOAD adds no significant, if any, functional improvements, only performance, and I'd like to see the difference between downloading from team city versus building it from source. I've also argued that we can re-instate some of the complexity once we've migrated to jakarta APIs if that's what the community wants. Note that I am getting some heat from the quarkus community, who have no dependency on WildFly, about the next release of Narayana and I can't do that for them with the scripts in their current state.

scripts/hudson/narayana.sh
tomjenkinson: What I am suggesting at this particular line is to change "AS_BUILD=0 AS_CLONE=0 AS_DOWNLOAD=1" to "AS_BUILD=1 AS_CLONE=1 AS_DOWNLOAD=0" (and to keep your change "NARAYANA_BUILD=1")

scripts/hudson/narayana.sh
tomjenkinson: If none of the profiles use "AS_DOWNLOAD" then removing this line would have no affect. The reason it is somehow having an effect at the moment is the way you have AS_DOWNLOAD=1 in the LRA PROFILE

scripts/hudson/narayana.sh
```diff
@@ -411,20 +379,21 @@ function clone_as {
 }
 
 function build_as {
-  echo "Building JBoss EAP/WildFly"
+  AS_BRANCH=27.0.0.Alpha1
+  echo "Building WildFly $AS_BRANCH"
 
-  cd $WORKSPACE/jboss-as
+  cd jboss-as
```
tomjenkinson, line 385: I reckon this could still say `cd $WORKSPACE/jboss-as` to minimize the change set

scripts/hudson/narayana.sh
```diff
@@ -927,7 +896,7 @@ function perf_tests {
 
   comment_on_pull "$PERF_OUTPUT" "$BUILD_URL"
 
-  [ $res = 0 ] || fatal "there were regressions in one or more of the benchmarks (see previous PR comment for details"
+  [ $res = 0 ] || fatal "there were regressions in one or more of the benchmarks (see previous PR comment for details)"
```
tomjenkinson, line 899: I think this tidy up could go in a separate commit

scripts/hudson/narayana.sh
tomjenkinson: If you do decide not to include this line here, then you might also want to keep your removal of the `echo JBOSS_HOME=$JBOSS_HOME` but if you do that should be in a commit that tidies up narayana.sh rather than connecting it to WildFly

scripts/hudson/narayana.sh
mmusgrov: I will remove them. But discussing on Zulip is overkill, micro-managing software updates seems wrong and will turn people off from reading our chat so let's limit such discussions to the PR's.

But, at risk of repeating myself, anything we can do to simplify this behemoth of a script should be encouraged, so when I find the time I will raise a PR to simplify it - we have all been guilty of misunderstanding how this script works.

scripts/hudson/narayana.sh
mmusgrov: It's mad code, but I will reinstate the original!

scripts/hudson/narayana.sh
mmusgrov: > I am not sure if I understand what you are responding to, I could imagine your response being why this type of line would be removed (rather than moved).

The `echo JBOSS_HOME=$JBOSS_HOME` statement is superfluous because the script runs with `set -x` which already logs the value and, moreover, it logs it at the point it was set.

scripts/hudson/narayana.sh
mmusgrov: lol, ok

scripts/hudson/narayana.sh
```diff
@@ -352,55 +352,23 @@ function build_narayana {
   return 0
 }
 
+# clone WildFly main
+# note that this version does not use 5_BRANCH
+# if we require changes to WildFly sources then we should either
+# a) raise a WildFly PR, or
```
jmfinelli, line 358: Just to be picky, the `or` could be removed

...but feel free to ignore this comment ;-)

jmfinelli (approved):
LGTM! Thank you very much Michael +1

scripts/hudson/narayana.sh
mmusgrov: suitably ignored ;-)

I'll consider it if I ever get time to clean the whole script

mmusgrov:
> @mmusgrov do you think the XTS profile failure on JDK17 is similar (different test though) to https://issues.redhat.com/browse/JBTM-3667?

1. We do periodically see the XTS test failure in the XTS crash recovery tests, see [1] for example, and is unlikely to be related to this particular PR, more likely a test artefact (remark it passed on the last run with both JDK's which lends weight to it being a transient failure).

Since the purpose of the PR is to release an alpha, I don't think that this transient failure needs to block the PR and, indeed, the problem with this kind of heisenbug related to test timing issues is that they are notoriously difficult to track down. We should monitor the regular CI job and if it starts happening more regularly we'd need to resolve it.

2. The AS-TESTS failure is unrelated and the failure is due to a timeout breach:
  Integration - MicroProfile (org.wildfly.test.integration.microprofile.reactive.messaging.kafka.serializer.ReactiveMessagingKafkaSerializerTestCase).
The test passed on the last run of the axis.

[1] https://issues.redhat.com/browse/JBTM-2038?jql=project%20%3D%20JBTM%20AND%20component%20%3D%20XTS%20and%20text%20~%20%22was%20not%20killed%20by%20Byteman%22%20ORDER%20BY%20priority%20DESC%2C%20updated%20DESC

PR #2029, opened 2022-08-02
JBTM-3664 Change the use of AS_TEST to AS_TESTS

tomjenkinson:
!CORE !TOMCAT !AS_TESTS RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERF !LRA !DB_TESTS mysql db2 postgres oracle


mmusgrov (approved):
Well spotted.
The profile failed as expected due to the WildFly migration so this one is good to merge.

PR #2039, opened 2022-09-09
Remove jboss-corba-ots-spi dependency

marcosgopen:
https://issues.redhat.com/browse/JBTM-3698
NO_TEST 
I followed Wildfly implementation https://github.com/wildfly/wildfly/tree/main/transactions/src/main/java/org/jboss/iiop/tm  and I kept the original package (org.jboss.iiop.tm) used in wildfly for the new interface created InboundTransactionCurrent (previously present in dependency jboss-corba-ots-spi).


mmusgrov (approved):
Nice work!

PR #2040, opened 2022-09-09
JBTM-3588 remove eclipse transformed jakarta artefacts

mmusgrov:
Now that we are natively supporting jakarta the equivalent artefacts that we created using the eclipse bytecode transformer are not needed. This PR removes them.


mmusgrov:
I'm merging this one, the various job failures are anticipated.

PR #2043, opened 2022-09-14
Db2tests

tomjenkinson:
!CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS QA_JTA !QA_JTS_OPENJDKORB !PERF !LRA !DB_TESTS mysql db2 postgres oracle

Although this does make a change that would add QA_JTS_OPENJDKORB, it is to trigger the tests for that JTS profile (via "ci-tests") but if that causes problems after merge, hopefully we can deal with that later. The QA_JTA tests trigger the "ci-tests-nojts" in that profile.

The pull request also deletes some config (in a separate commit) that should no longer be used.


mmusgrov (approved):
Yes, I notice that the db2 tests ran and that the `junit-jdbc-testsuite` passed.
I think we should merge this one, can you do that @tomjenkinson ?

PR #2045, opened 2022-10-06
[JBTM-3710] QA's building system updated

jmfinelli:
This PR addresses [JBTM-3710](https://issues.redhat.com/browse/JBTM-3710)

JDK17 CORE  QA_JTA QA_JTS_OPENJDKORB
!TOMCAT !AS_TESTS !RTS !JACOCO !XTS !PERF !LRA !DB_TESTS !mysql !db2 !postgres !oracle

PR #2049, opened 2022-10-18
[JBTM-3631] Removed the dependencies between QA and both ArjunaJTS and Narayana-full

jmfinelli:
This PR addresses [JBTM-3631](https://issues.redhat.com/browse/JBTM-3631)

To test:
JDK17 CORE QA_JTA QA_JTS_OPENJDKORB

Skipped:
!TOMCAT !AS_TESTS !RTS !JACOCO !XTS !PERF !LRA !DB_TESTS !mysql !db2 !postgres !oracle


qa/TaskImpl.properties.template
```diff
@@ -30,18 +30,9 @@ COMMAND_LINE_1=-classpath
 # file we use command line override of some properties.
 #
 COMMAND_LINE_2=\
-  dist${file.separator}narayana-full-5.13.1.Final-SNAPSHOT${file.separator}etc${file.separator}\
```
tomjenkinson, line 33: @jmfinelli is it fine to remove `etc/` folder?

jmfinelli: Adding the actual jars to the classpath should take care of the `etc/` folder

qa/tests/build.xml
```diff
@@ -231,14 +239,14 @@
...
-            <src path="../../common/tests"/>
+            <src path="${external.libraries}/lib/jts-test-sources/com/"/>
+            <src path="${external.libraries}/lib/common-test-sources/com/"/>
 
             <exclude name="**/javaidl/**/*.java" unless="is.sunorb"/>
 
             <exclude name="org/jboss/jbossts/qa/ArjunaCore/Common/ClassPathParserTest.java" />
-            <exclude name="com/hp/mwtests/ts/jts/recovery/RecoveredTransactionUnitTest.java" />
-            <exclude name="com/hp/mwtests/ts/jts/recovery/RecoveredServerTransactionUnitTest.java" />
+            <exclude name="hp/mwtests/ts/jts/recovery/RecoveredTransactionUnitTest.java" />
```
tomjenkinson: How come the com/ could be removed from this line and the one below?

ArjunaJTS/idl/idlj-openjdk/pom.xml
```diff
@@ -81,6 +81,7 @@
             <configuration>
               <sources>
                 <source>${project.build.directory}/generated-sources/idl</source>
+                <source>${project.basedir}/../src/main/</source>
```
tomjenkinson, line 84: @jmfinelli is this definitely needed I am not spotting a reference to this path in the diff?

qa/pom.xml
```diff
@@ -33,38 +40,124 @@
...
+      <groupId>org.jboss.openjdk-orb</groupId>
+      <artifactId>openjdk-orb</artifactId>
+      <version>${version.org.jboss.openjdk-orb}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jboss.openjdk-orb</groupId>
+      <artifactId>openjdk-orb</artifactId>
+      <version>${version.org.jboss.openjdk-orb}</version>
+      <classifier>sources</classifier>
+    </dependency>
```
tomjenkinson, line 100: I think this block is just moved from below and would have been more suitable in the commit https://github.com/jbosstm/narayana/pull/2049/commits/55e3ae899f10e29e8ba50b9883bbc5df55e4435a

ArjunaJTS/idl/idlj-openjdk/pom.xml
tomjenkinson: Ah, it's https://github.com/jbosstm/narayana/pull/2049/commits/dde6621d981b608faa6d4b2f307919dae134a869#diff-d4f42289a362c7d991ef7305ebf659ea59d80c9fcb0e3ee42fc80f79f457aa92R59

qa/tests/build.xml
jmfinelli: IIRC, packing and unpacking source code ended up with a bit of headache for `ant`. To fix this, I had to remove `com` from the path. I can double check this though and see if I manage to introduce `com` back in the path

qa/pom.xml
tomjenkinson: This block would be lines 90-100 above

ArjunaJTS/idl/idlj-openjdk/pom.xml
jmfinelli: Yes, it is. Basically, this line is packing `idl` files in the source jar

qa/pom.xml
jmfinelli: +1

qa/TaskImpl.properties.template
tomjenkinson: The etc folder from the dist would normally have in it:
https://github.com/jbosstm/narayana/blob/master/narayana-full/src/main/assembly/bin.xml#L195
https://github.com/jbosstm/narayana/blob/master/narayana-full/src/main/assembly/bin.xml#L206
https://github.com/jbosstm/narayana/blob/master/narayana-full/src/main/assembly/bin.xml#L212

Is it fine for qa/ not to have those files?

qa/tests/build.xml
jmfinelli: I don't think this is a problem but I'll double check if I can restore `com` in the path. Thanks Tom

qa/TaskImpl.properties.template
tomjenkinson: Thanks @jmfinelli - I think the last one is the most interesting as by having the file on the file system it could be readily changed to try different things but thanks again for checking.

jmfinelli: You're welcome @tomjenkinson :-)

If users want to use `jbossts-properties.xml` for trying things out, they have the option to unzip `narayana-jts-idlj.jar` (command = `unzip narayana-jts-idlj.jar -d narayana-jts-idlj`) and load `jbossts-properties.xml` directly in the classpath. Nevertheless, I think that keeping the file in the jar is the best solution; in fact, as this PR and #2045 prove, QA has not been updated for years

tomjenkinson: I think I was thinking more if a developer wanted to try with different jbossts-properties.xml (this wouldn't show in commit log)

qa/tests/build.xml
tomjenkinson: Thank you

qa/pom.xml
```diff
@@ -33,38 +40,124 @@
     <dependency>
       <groupId>org.jboss.spec.javax.transaction</groupId>
       <artifactId>jboss-transaction-api_1.2_spec</artifactId>
-      <version>1.0.0.Final</version>
+      <version>${version.org.jboss.spec.javax.transaction}</version>
```
marcosgopen, line 43: Thanks @jmfinelli for the PR! Is this dependency still needed?  if yes why?

jmfinelli: TBH, I haven't checked if `jboss-transaction-api_1.2_spec` is still needed as the goal of this PR is to detach QA from both ArjunaJTS and Narayana-full. I guess we can remove this dependency (if not needed) during our migration to Jakarta. WDYT?

qa/TaskImpl.properties.template
jmfinelli: I think the above procedure (i.e. extracting `jbossts-properties.xml`) should work also for developers in case they want to try different configurations.

> this wouldn't show in commit log

That `jbossts-properties.xml` should be related to `narayana-jts-idlj.jar`, shouldn't it? If this is the case, developers/users who want to use a different configuration should extract `narayana-jts-idlj.jar` in a folder and add that folder to the classpath.

tomjenkinson: What I was getting at really by `this wouldn't show in commit log` is to respond to `QA has not been updated for years` where if a developer had needed to use the flexibility to add a jbossts-properties.xml to test something in `qa/` locally that would not have shown in changes in `qa/`. The capability (which is being removed) to do that could be considered easier when there is already a folder in this template that would have added a folder to the classpath but I don't suppose it has to be in the template, someone can just manually add it to their own TaskImpl.properties.

qa/pom.xml
tomjenkinson: I would presume to expect it is needed as https://github.com/jbosstm/narayana/pull/2049/files/55e3ae899f10e29e8ba50b9883bbc5df55e4435a#diff-98ba2d8741d0f457de06355f6c6951c198db3ca7d470aa35c97b861c6f1c2b94R115 only makes this dependency provided https://github.com/jbosstm/narayana/blob/master/ArjunaJTS/narayana-jts-idlj/pom.xml#L159 so I expect `qa/` needs to provide it.

ArjunaJTS/jts/pom.xml
```diff
@@ -192,6 +192,24 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
```
tomjenkinson, line 195: @jmfinelli please can you point out where this is used in the build after this change?

common/pom.xml
```diff
@@ -63,6 +63,24 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
```
tomjenkinson, line 66: @jmfinelli please can you point out where this is used in the build after this change?

ArjunaJTS/jts/pom.xml
jmfinelli: The test jar gets extracted [here](https://github.com/jbosstm/narayana/pull/2049/files#diff-d4f42289a362c7d991ef7305ebf659ea59d80c9fcb0e3ee42fc80f79f457aa92R100)

common/pom.xml
jmfinelli: The test jar gets extracted [here](https://github.com/jbosstm/narayana/pull/2049/files#diff-d4f42289a362c7d991ef7305ebf659ea59d80c9fcb0e3ee42fc80f79f457aa92R98)

qa/.gitignore
```diff
@@ -2,9 +2,13 @@
 .project
 dbdrivers
 dist
+ext
 idl-compiler.*
 build
+*.properties
```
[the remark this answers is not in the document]
jmfinelli: As I needed to remove `qa/ext` from the repository, I thought it was good to remove other files from `.gitignore`. I will group the other modifications in a separate commit
BTW, `*.properties` is also for two files that get created during testing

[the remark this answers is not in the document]
jmfinelli: Done :-)

jmfinelli:
Please, do not consider the `Hold` label when reviewing. I have used this label only to stop the CI testing this PR

qa/pom.xml
marcosgopen: Thank you both! I agree with you @jmfinelli, the removal of the dependency can be done during the migration.

ArjunaJTS/jts/pom.xml
tomjenkinson: Although I think your link is not going to a location that I can find, I do see the use of `>jts<` for `<classifier>test-sources</classifier>` in qa/pom.xml but I can't see `<classifier>sources</classifier>` - is it definitely needed here?

common/pom.xml
tomjenkinson: Although I think your link is not going to a location that I can find, I do see the use of `>common<` for `<classifier>test-sources</classifier>` in qa/pom.xml but I can't see `<classifier>sources</classifier>` - is it definitely needed here?

tomjenkinson: For a bit of clarity on the specific xml I am querying:
```
          <execution>
            <id>attach-sources</id>
            <goals>
              <goal>jar-no-fork</goal>
            </goals>
          </execution>
```
That applies for `jts` comment too in it's respective pom.xml.

tomjenkinson: I can find some unzipping in https://github.com/jbosstm/narayana/blob/20ada636f438ae8bbaf550b10c38490bc5384b23/qa/tests/build.xml but I don't think that `>jts<` or `>common< ` `sources` (rather than `test-sources`) are used in there but perhaps the reason you are providing these is to keep it generating sources for these modules as well as the test sources? I am not sure so clarification would be great, thank you.

ArjunaJTS/jts/pom.xml
jmfinelli: Basically, I tried to recreate the behaviour I noticed before adding `<classifier>sources</classifier>`. In other words, the source jar was already generated and put in the `target` folder, that is why I added the `<classifier>sources</classifier>` to the plugin.

common/pom.xml
jmfinelli: > ... perhaps the reason you are providing these is to keep it generating sources for these modules as well as the test sources? I am not sure so clarification would be great, thank you.

+1 that is indeed the reason :-)

tomjenkinson: Thank you for confirming!

qa/.gitignore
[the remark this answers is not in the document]
tomjenkinson: Thanks for considering this point

tomjenkinson (approved):

PR #2052, opened 2022-10-19
5.11 updates

tomjenkinson:
Given that sra and lra use an old version of Quarkus on this branch, they are being disabled from the build.

A later version of Byteman is used.

The Undertow dependencies and not directly added (`master` does not add them either - https://github.com/jbosstm/narayana/blob/master/rts/at/tx/pom.xml)

!JACOCO !AS_TESTS !XTS !RTS (these should use the app server but WildFly 23.x (https://github.com/jbosstm/narayana/blob/5.11/scripts/hudson/narayana.sh#L393) is no longer maintained. !LRA (it downloads the latest nightly)


mmusgrov (approved):
Changes look good, thanks

PR #2057, opened 2022-11-07
[SCRIPT_ONLY] Replaced git with https when 'git clone'

jmfinelli:
Replaced `git` with `https` when using `git clone`

NO_TEST


marcosgopen (approved):
Thanks @jmfinelli for your PR.

PR #2066, opened 2022-11-24
[JBTM-3656] XTS TestBA race condition fix

marcosgopen:
https://issues.redhat.com/browse/JBTM-3656
https://issues.redhat.com/browse/JBTM-3612
https://issues.redhat.com/browse/JBTM-3616

JDK17 AS_TESTS
!CORE !TOMCAT  !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERF !LRA !DB_TESTS !mysql !db2 !postgres !oracle


XTS/WS-T/dev/src/org/jboss/jbossts/xts/recovery/participant/ba/XTSBARecoveryManager.java
```diff
@@ -25,16 +25,19 @@ public abstract class XTSBARecoveryManager {
...
             // theRecoveryManager must eventually be set to a non-null value
             while (theRecoveryManager == null) {
                 try {
-                    lock.wait();
+                    //syncrhronized block inside the loop for efficiency
+                    synchronized (lock) {
+
+                        lock.wait();
+                    
+                    }
```
jmfinelli, line 36: Could you please share your reasoning about this?

jmfinelli: Thanks for you reply @marcosgopen

> I think using the while loop inside the synchronized block is useless

Why? To me it seems to be very logical. Moreover, using a synchronized block inside a loop (as you changed) goes exactly against your comment `syncrhronized block inside the loop for efficiency`. Even though it might appear like a single instruction, acquiring a lock with `synchronized` is a very expensive operation and it shouldn't be used in a loop.

> ...**most of the time** the recoveryManager is already set.

That "most of the time" doesn't sound right. Synchronisation is used for concurrency -> "most of the time" doesn't cover when two (or more) processes try to modify the same variable.

> Additionally you might think that the synchronized block prevent other threads from changing the theRecoveryManager value, which is true inside the block, but actually the value can be changed by other threads before or after returning theRecoveryManager (https://github.com/jbosstm/narayana/blob/master/XTS/WS-T/dev/src/org/jboss/jbossts/xts/recovery/participant/ba/XTSBARecoveryManager.java#L40) because it is a reference to the object

I'm not sure I follow this but I guess it can be answered with my previous comment about "most of the time".

marcosgopen: Thanks again Manuel for your answer. I think this conversation is very helpful.
`acquiring a lock with synchronized is a very expensive operation and it shouldn't be used in a loop`
Correct me if I am wrong: from my understanding threads enter in that loop only once, where they wait until lock.notifyAll is invoked (https://github.com/jbosstm/narayana/blob/master/XTS/WS-T/dev/src/org/jboss/jbossts/xts/recovery/participant/ba/XTSBARecoveryManager.java#L59) and the while condition is not true anymore.
 `That "most of the time" doesn't sound right.`
 Sorry for that, I think this change created confusion. The change in XTSBARecoveryManager is not needed for a concurrency issue, this change is for efficiency only. This is why I said 'most of the time'.

jmfinelli: > from my understanding threads enter in that loop only once, where they wait until lock.notifyAll is invoked (https://github.com/jbosstm/narayana/blob/master/XTS/WS-T/dev/src/org/jboss/jbossts/xts/recovery/participant/ba/XTSBARecoveryManager.java#L59) and then the while condition (https://github.com/jbosstm/narayana/blob/master/XTS/WS-T/dev/src/org/jboss/jbossts/xts/recovery/participant/ba/XTSBARecoveryManager.java#L30) is not true anymore.

This "could" be true (i.e. it depends on many factors like compilation optimisations, the state of the variable in a precise moment, runtime optimisations, etc. ). It "is" true only if `synchronized` is used to protect the entire loop

jmfinelli: Mind I haven't studied how `theRecoveryManager` is used in any detail apart from reading the code. If you have a reasoning stronger than "efficiency", could you please share it (for my convenience)?

mmusgrov (commented):
I agree with the synchronisation on `recoveryModules` but I don't see the point of synchronising the getter and setter methods for subordinateCoordinateRecoveryStarted.

mmusgrov:
> I agree with the synchronisation on `recoveryModules` but I don't see the point of synchronising the getter and setter methods for subordinateCoordinateRecoveryStarted.

Actually I don't think we need synchronization around updates to recoveryModules either since they are read by taking a copy of the collection before iterating using the following strategy:

```
        synchronized (recoveryModules) {
            recoveryModulesCopy = new ArrayList<XTSBARecoveryModule>(recoveryModules);
        }
```

marcosgopen:
Thanks @mmusgrov , initially I got the idea from the XTS at module (https://github.com/jbosstm/narayana/blob/master/XTS/recovery/src/org/jboss/jbossts/xts/recovery/participant/at/XTSATRecoveryManagerImple.java#L302). Now I see it is not the fix for this error.

XTS/WS-T/dev/src/org/jboss/jbossts/xts/recovery/participant/at/XTSATRecoveryManager.java
```diff
@@ -29,6 +29,10 @@ public static XTSATRecoveryManager getRecoveryManager()
             // theRecoveryManager must eventually be set to a non-null value
             while (theRecoveryManager == null) {
                 try {
+                    
+                    if (Thread.currentThread().isInterrupted()) {
+                        break;
```
mmusgrov, line 34: This change means that a null value could be returned but the method is designed to guarantee that a non-null value is returned so other things will break.

Something must eventually call setRecoveryManager() which will call notifyAll causing this thread to return from the call to `lock.wait();` and loop test will be re-evaluated.

The thread synchronisation is enforced by running the relevant code inside a `synchronized (lock)` code block.

XTS/WS-T/dev/src/org/jboss/jbossts/xts/recovery/participant/at/XTSATRecoveryManager.java
```diff
@@ -24,11 +24,19 @@ public abstract class XTSATRecoveryManager {
...
+
         // use a lock to get the recovery manager to avoid a race with the code that sets it
         synchronized (lock) {
             // theRecoveryManager must eventually be set to a non-null value
             while (theRecoveryManager == null) {
                 try {
+
+                    if (currentTimeout - System.currentTimeMillis() < 0)
+                    {
+                        break ;
```
mmusgrov, line 37: This change means that a null value could be returned but the method is designed to guarantee that a non-null value is returned so other things will break.

Something must eventually call setRecoveryManager() which will call notifyAll causing this thread to return from the call to lock.wait(); and the loop condition will be re-evaluated.

The thread synchronisation is enforced by running the relevant code inside a synchronized (lock) code block.

XTS/WS-T/dev/src/org/jboss/jbossts/xts/recovery/participant/ba/XTSBARecoveryManager.java
```diff
@@ -24,11 +24,15 @@ public abstract class XTSBARecoveryManager {
...
     public static XTSBARecoveryManager getRecoveryManager()
     {
+        
+        
+
         // use a lock to get the recovery manager to avoid a race with the code that sets it
         synchronized (lock) {
             // theRecoveryManager must eventually be set to a non-null value
-            while (theRecoveryManager == null) {
+            if (theRecoveryManager == null) {
```
mmusgrov, line 33: `lock.wait` can return for a number of reasons so you need to retest it each time it returns.

mmusgrov: The existing logic looks fine, what scenario do you think it doesn't cover. But if you are unsure about it and want to change it then you'd need a reproducer for the failure in order to verify whatever change you introduce.

mmusgrov: But I do notice that can improve things by applying the "Double-Checked Locking" strategy. Will you update your PR to do that.

marcosgopen: Thanks for the review @mmusgrov ! I updated the PR implementing the "Double-Checked Locking" strategy.

mmusgrov (approved):
Looks good, hopefully this will finally/fully resolve the issue, thanks.

XTS/WS-T/dev/src/org/jboss/jbossts/xts/recovery/participant/at/XTSATRecoveryManager.java
marcosgopen: Thanks for the review @mmusgrov . I removed that block of code, now the method guarantees that a non-null value is returned

mmusgrov:
@marcosgopen I removed the hold label. I'll leave this PR for you to merge.

marcosgopen:
The 'theRecoveryManager' variable has been declared **volatile**  because under concurrency situation it must be detected to be set by other threads otherwise the thread might read the old value of the variable and enter in the synchronized block and not exiting.
Looking at the stacktrace [1]  we can see it is possible that XTS did start, but then the deployment failed as it missed seeing the recovery manager due to this volatile issue and so after some time the test might try to stop and is hanging waiting for this deployment that can no longer complete so locks up. Eventually this might cause byteman to fail to kill the JVM (as shown in the error message of the console log [2]).
Even if reported failures occurred during XTSBA tests the thread might be blocked in XTSATRecoveryManager class because wildfly extension undertow registers it during context initialization (https://github.com/jbosstm/narayana/blob/master/XTS/localjunit/xtstest/src/org/jboss/jbossts/xts/servicetests/bean/XTSServiceTestRunnerBean.java#L86), see the stacktrace for more info [1]

[1] attachment in https://issues.redhat.com/browse/JBTM-3656
`"ServerService Thread Pool -- 83" #292 prio=5 os_prio=0 cpu=6.28ms elapsed=78302.90s tid=0x00007f0da81bd800 nid=0xa4f in Object.wait()  [0x00007f0b18b91000]
   java.lang.Thread.State: WAITING (on object monitor)
	at java.lang.Object.wait(java.base@11.0.14/Native Method)
	- waiting on <0x00000000c989c0a0> (a java.lang.Object)
	at java.lang.Object.wait(java.base@11.0.14/Object.java:328)
	at org.jboss.jbossts.xts.recovery.participant.at.XTSATRecoveryManager.getRecoveryManager(XTSATRecoveryManager.java:32)
	- waiting to re-lock in wait() <0x00000000c989c0a0> (a java.lang.Object)
	at org.jboss.jbossts.xts.servicetests.service.recovery.TestATRecoveryModule.register(TestATRecoveryModule.java:38)
	at org.jboss.jbossts.xts.servicetests.bean.XTSServiceTestRunnerBean.start(XTSServiceTestRunnerBean.java:86)
	at org.jboss.jbossts.xts.servicetests.bean.XTSServiceTestRunnerBean.contextInitialized(XTSServiceTestRunnerBean.java:49)
	at io.undertow.servlet.core.ApplicationListeners.contextInitialized(ApplicationListeners.java:187)
	at io.undertow.servlet.core.DeploymentManagerImpl$1.call(DeploymentManagerImpl.java:219)
	at io.undertow.servlet.core.DeploymentManagerImpl$1.call(DeploymentManagerImpl.java:187)
	at io.undertow.servlet.core.ServletRequestContextThreadSetupAction$1.call(ServletRequestContextThreadSetupAction.java:42)
	at io.undertow.servlet.core.ContextClassLoaderSetupAction$1.call(ContextClassLoaderSetupAction.java:43)
	at org.wildfly.extension.undertow.deployment.UndertowDeploymentInfoService$UndertowThreadSetupAction.lambda$create$0(UndertowDeploymentInfoService.java:1435)
...`
[2] see issue description in https://issues.redhat.com/browse/JBTM-3656

marcosgopen:
An additional check has been added to variable 'theRecoveryManager' [3] following the double-checked locking strategy. The check reduces the overhead of acquiring the lock.

[3] https://github.com/jbosstm/narayana/pull/2066/files#diff-f0fb04f6db8942f15d985b0b8e05d07795f97495b80338fa4e0fc5af77ede450R28

PR #2067, opened 2022-11-28
JBTM-3723 Remove contention point on TransactionManager#transactionManager

Sanne:
Fix proposal for https://issues.redhat.com/browse/JBTM-3723

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/jta/TransactionManager.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/jta/TransactionManager.java
@@ -66,7 +66,7 @@ public class TransactionManager
      * @return The singleton transaction manager.  Can return null if the instantiation failed.
      */
 	
-    public synchronized static javax.transaction.TransactionManager transactionManager ()
+    public static javax.transaction.TransactionManager transactionManager ()
     {
 		return jtaPropertyManager.getJTAEnvironmentBean().getTransactionManager();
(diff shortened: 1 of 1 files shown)


mmusgrov (approved):
I think it's fine to remove the keyword.
We already perform double checked locking when looking up the transaction manager [1]

Way back in Jul 2006 we used to do the lookup directly:

```
_transactionManager = (javax.transaction.TransactionManager) Thread.currentThread().getContextClassLoader().loadClass(JNDIManager.getTransactionManagerImplementationClassname()).newInstance();
```
hence we needed the synchronization, but now we do it via our environment bean

```
return jtaPropertyManager.getJTAEnvironmentBean().getTransactionManager();
```

which does the double lock check.

[1] https://github.com/jbosstm/narayana/blob/master/ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java#L192

mmusgrov:
@jmfinelli this looks like a good change, do you mind if I merge it?

jmfinelli:
> @jmfinelli this looks like a good change, do you mind if I merge it?

Not at all! LGTM as well!

jmfinelli:
> FWIW I'll leave backporting choices to you all, but bear in mind it would be great to have this fix available in EAP8.

It will be included in our next release, i.e. Narayana 6.0 (which will be part of EAP8)

PR #2083, opened 2023-02-03
upgrade smallrye-converter-api dependency

mmusgrov:
Our smallrye dependency is way out of date. I have upgraded it to match what quarkus use but note that we only use it in the `cdi` module and even there the code path that could potentially use it is never exercised

CORE
!TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERF !LRA !DB_TESTS !mysql !db2 !postgres !oracle


marcosgopen (approved):
Thanks @mmusgrov for your PR. ILGTM

jmfinelli:
This PR breaks Narayana.

PR #2102, opened 2023-03-16
[TEMPLATE] Removed PERFORMANCE from test axis

jmfinelli:
NO_TEST

This PR is needed as our CI is not able (at the moment) to use different agents for different type of tests. In other words, pull requests can only be tested with one agent.


jmfinelli:
Sorry, wrong assumption

PR #2117, opened 2023-04-04
Set defaultValue for sraId

marcosgopen:
Update sra readme and fix sra demo 
CORE RTS

### DIFF
--- a/rts/sra/src/main/java/io/narayana/sra/demo/api/FlightController.java
+++ b/rts/sra/src/main/java/io/narayana/sra/demo/api/FlightController.java
@@ -57,7 +57,7 @@ public class FlightController extends SRAParticipant {
     @Path("/book")
     @Produces(MediaType.APPLICATION_JSON)
     @SRA(SRA.Type.REQUIRED)
-    public Booking bookFlight(@HeaderParam(RTS_HTTP_CONTEXT_HEADER) String sraId,
+    public Booking bookFlight(@HeaderParam(RTS_HTTP_CONTEXT_HEADER) @DefaultValue("") String sraId,
                               @QueryParam(FLIGHT_NUMBER_PARAM) @DefaultValue("") String flightNumber,
                               @QueryParam(FLIGHT_SEATS_PARAM) @DefaultValue("1") Integer seats,
                               @QueryParam("mstimeout") @DefaultValue("500") Long timeout) {
--- a/rts/sra/src/main/java/io/narayana/sra/demo/api/HotelController.java
+++ b/rts/sra/src/main/java/io/narayana/sra/demo/api/HotelController.java
@@ -56,7 +56,7 @@ public class HotelController extends SRAParticipant {
     @Path("/book")
     @Produces(MediaType.APPLICATION_JSON)
     @SRA(SRA.Type.REQUIRED)
-    public Booking bookRoom(@HeaderParam(RTS_HTTP_CONTEXT_HEADER) String sraId,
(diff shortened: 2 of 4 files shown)


marcosgopen:
As discussed on zulip (https://narayana.zulipchat.com/#narrow/stream/323715-developers) we cannot have a default SRA id. Therefore I closed this PR, readme changes will be included in a separate PR.

rts/sra/src/main/java/io/narayana/sra/demo/api/FlightController.java
```diff
@@ -57,7 +57,7 @@ public class FlightController extends SRAParticipant {
     @Path("/book")
     @Produces(MediaType.APPLICATION_JSON)
     @SRA(SRA.Type.REQUIRED)
-    public Booking bookFlight(@HeaderParam(RTS_HTTP_CONTEXT_HEADER) String sraId,
+    public Booking bookFlight(@HeaderParam(RTS_HTTP_CONTEXT_HEADER) @DefaultValue("") String sraId,
```
moulalis, line 60: We cant set a default value for sraId as per Tom, because a transaction can't have a "default" ID. It needs to be allocated by the transaction manager.

rts/sra/src/main/java/io/narayana/sra/demo/api/HotelController.java
```diff
@@ -56,7 +56,7 @@ public class HotelController extends SRAParticipant {
     @Path("/book")
     @Produces(MediaType.APPLICATION_JSON)
     @SRA(SRA.Type.REQUIRED)
-    public Booking bookRoom(@HeaderParam(RTS_HTTP_CONTEXT_HEADER) String sraId,
+    public Booking bookRoom(@HeaderParam(RTS_HTTP_CONTEXT_HEADER) @DefaultValue("") String sraId,
```
moulalis, line 59: We cant set a default value for sraId as per Tom, because a transaction can't have a "default" ID. It needs to be allocated by the transaction manager.

rts/sra/src/main/java/io/narayana/sra/demo/api/TripController.java
```diff
@@ -112,7 +112,7 @@ void finiController() {
     @Produces(MediaType.APPLICATION_JSON)
     // end = false because we want the SRA to be associated with a booking until the user confirms the booking
     @SRA(value = SRA.Type.REQUIRED, end = false)
-    public Response bookTrip( @HeaderParam(RTS_HTTP_CONTEXT_HEADER) String sraId,
+    public Response bookTrip( @HeaderParam(RTS_HTTP_CONTEXT_HEADER) @DefaultValue("") String sraId,
```
moulalis, line 115: We cant set a default value for sraId as per Tom, because a transaction can't have a "default" ID. It needs to be allocated by the transaction manager.

moulalis (changes requested):

mmusgrov (changes requested):
The RTS_HTTP_CONTEXT_HEADER header is how we propagate the transaction context, it needs to contain an actual transaction id representing a running transaction that the microservice can use to participate in the transaction. The enlistment of the service/participant happens in the JAX-RS filter (ServerSRAFilter) which makes calls to the transaction coordinator (in the demo I believe it uses the RTS WildFly subsystem to run the coordinator) so if the header is null then the participant has not been enlisted in the transaction (by the filter). So setting a default value is not correct.

Note that the demo code was written before the Jakarta migration so I'd expect there to be some changes to get it working. I can take a look tomorrow at getting it working if there is no progress before then.

PR #2118, opened 2023-04-07
LRA testsuite needs microprofile-config-api

marcosgopen:
CORE AS_TESTS 
see error https://ci-jenkins-csb-narayana.apps.ocp-c1.prod.psi.redhat.com/job/narayana-AS_TESTS/PROFILE=AS_TESTS,jdk=openJDK11,label=jnlp-agent/130/artifact/jboss-as/testsuite/integration/microprofile-tck/lra/target/wildfly/standalone/log/server.log
in build https://ci-jenkins-csb-narayana.apps.ocp-c1.prod.psi.redhat.com/job/narayana-AS_TESTS/PROFILE=AS_TESTS,jdk=openJDK11,label=jnlp-agent/130 (requires VPN)

"Caused by: java.lang.ClassNotFoundException: org.eclipse.microprofile.config.ConfigProvider"

### DIFF
--- a/narayana-bom/pom.xml
+++ b/narayana-bom/pom.xml
@@ -24,7 +24,6 @@
         <groupId>org.apache.activemq</groupId>
         <artifactId>artemis-journal</artifactId>
         <version>${version.org.apache.activemq}</version>
-        <scope>provided</scope>
         <exclusions>
           <exclusion>
             <groupId>io.netty</groupId>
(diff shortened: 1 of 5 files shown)


marcosgopen:
In my last commit I removed dependencies' scopes from the bom. It think it would be better to let the single projects to choose the scope needed (i.e. runtime, provided etc.). BTW not setting the scope is equal to set the compile scope. 
I will open a zulip topic about this PR.

marcosgopen:
As described in the zulip topic "https://narayana.zulipchat.com/#narrow/stream/323715-developers/topic/AS_TESTS.20and.20lra-coordinator's.20module.2Exml" the testsuite error is due to the update of the lra-coordinator. In order to fix it a Wildfly PR is needed in order to update lra-coordinator module.xml (https://github.com/wildfly/wildfly/blob/main/microprofile/galleon-common/src/main/resources/modules/system/layers/base/org/jboss/narayana/rts/lra-coordinator/main/module.xml). Therefore this PR is not needed anymore. Removal of dependencies' scopes in the bom will be done in a separate PR.

PR #2079, opened 2023-01-23
[JBTM-3739] Create Narayana BOM

moulalis:
Issue: https://issues.redhat.com/browse/JBTM-3739

CORE TOMCAT AS_TESTS RTS JACOCO XTS QA_JTA QA_JTS_OPENJDKORB LRA

Skipped
!PERF !DB_TESTS !mysql !db2 !postgres !oracle

PR #2121, opened 2023-04-13
[JBTM-3776] Remove dependencies' scope in the Narayana bom

marcosgopen:
https://issues.redhat.com/browse/JBTM-3776

### DIFF
--- a/narayana-bom/pom.xml
+++ b/narayana-bom/pom.xml
@@ -24,7 +24,6 @@
         <groupId>org.apache.activemq</groupId>
         <artifactId>artemis-journal</artifactId>
         <version>${version.org.apache.activemq}</version>
-        <scope>provided</scope>
         <exclusions>
           <exclusion>
             <groupId>io.netty</groupId>
@@ -36,38 +35,32 @@
         <groupId>io.mashona</groupId>
         <artifactId>mashona-logwriting</artifactId>
         <version>${version.io.mashona}</version>
-        <scope>provided</scope>
       </dependency>
       <dependency>
         <groupId>org.jboss.logging</groupId>
         <artifactId>jboss-logging</artifactId>
         <version>${version.org.jboss.logging.jboss-logging}</version>
-        <scope>provided</scope>
       </dependency>
       <dependency>
         <groupId>org.jboss.logging</groupId>
         <artifactId>jboss-logging-processor</artifactId>
         <version>${version.org.jboss.logging.jboss-logging-processor}</version>
-        <scope>provided</scope>
       </dependency>
       <dependency>
         <groupId>org.jboss.narayana</groupId>
         <artifactId>common</artifactId>
         <version>${project.version}</version>
         <classifier>tests</classifier>
-        <scope>test</scope>
       </dependency>
       <dependency>
         <groupId>org.jboss.byteman</groupId>
         <artifactId>byteman-bmunit</artifactId>
         <version>${version.org.jboss.byteman}</version>
-        <scope>test</scope>
       </dependency>
       <dependency>
         <groupId>com.h2database</groupId>
         <artifactId>h2</artifactId>
         <version>${version.com.h2database}</version>
-        <scope>test</scope>
       </dependency>
       <dependency>
         <groupId>org.jboss.narayana.arjunacore</groupId>
@@ -113,7 +106,6 @@
         <groupId>jakarta.annotation</groupId>
         <artifactId>jakarta.annotation-api</artifactId>
         <version>${version.jakarta.annotation.jakarta-annotation-api}</version>
-        <scope>provided</scope>
       </dependency>
       <dependency>
         <groupId>jakarta.ejb</groupId>
@@ -124,7 +116,6 @@
         <groupId>io.smallrye.reactive</groupId>
         <artifactId>smallrye-reactive-converter-api</artifactId>
         <version>${version.smallrye-converter-api}</version>
-        <scope>provided</scope>
       </dependency>
       <dependency>
         <groupId>org.jboss.arquillian.core</groupId>
@@ -147,7 +138,6 @@
         <groupId>org.jboss.naming</groupId>
         <artifactId>jnpserver</artifactId>
         <version>${version.jnpserver}</version>
-        <scope>test</scope>
         <exclusions>
           <exclusion>
             <groupId>org.jboss.logging</groupId>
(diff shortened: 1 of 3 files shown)


jmfinelli:
I think our CI is not behaving well today. Although you selected all testing axis, only some of them started. Could you please start the other testing axis please?

narayana-bom/pom.xml
```diff
@@ -24,7 +24,6 @@
         <groupId>org.apache.activemq</groupId>
         <artifactId>artemis-journal</artifactId>
         <version>${version.org.apache.activemq}</version>
-        <scope>provided</scope>
```
mmusgrov, line 27: If you remove the scope here then you need to check that the scope is present in the poms that use the dependency (such as https://github.com/jbosstm/narayana/blob/main/narayana-full/pom.xml#L155).

I think you need to check every usage to verify that the scope is right otherwise the dependency will use the default scope.

marcosgopen: Thanks Michael for your review. My idea is to have the default scope (compile) for every dependency without the scope set. In the case you mentioned (https://github.com/jbosstm/narayana/blob/main/narayana-full/pom.xml#L155) do you have any specific reason to set the scope as provided? 
I checked that dependency. The scope there has never been set. It had had compile scope until the bom was created and set it as provided.  With this PR the scope will be 'compile' again.

marcosgopen:
AS_TESTS failures are not related to this PR.

narayana-bom/pom.xml
mmusgrov: The "previews scope" is what we had prior to pr/2079 ( [JBTM-3739] Create Narayana BOM #2079) so we'd need to check that PR in order to reinstate the original scope, unless the scope is still set in each module pom - perhaps @moulalis could provide us with some guidance.

marcosgopen: I confirm the scope is still set in each module pom (as it was before PR https://github.com/jbosstm/narayana/pull/2079), is it right @moulalis ?

jmfinelli:
WDYT about squashing together commits from `f686843db1102fa27fc8cd3c862da3b0c35ba29e` to `2fb47ea906bc06ae014f059a3ae8fda1403e6502`? Also, I think that the h2 version bump should go into another PR.

mmusgrov (changes requested):
This is one PR for 3 separate and unrelated issues: dependency scope modifications, CMR test changes updating JDBC urls, and JAX-RS resource changes.

I'd like to be able to approve it but it will set a precedent if we allow it - in particular the JAX-RS resource changes are unrelated and do not have any means of associating the change with any problem).

Also the associated JIRA (JBTM-3776) has a blank description (the tool should be making this a mandatory field).

mmusgrov:
> WDYT about squashing together commits from `f686843db1102fa27fc8cd3c862da3b0c35ba29e` to `2fb47ea906bc06ae014f059a3ae8fda1403e6502`? Also, I think that the h2 version bump should go into another PR.

@jmfinelli  Would that be issue https://issues.redhat.com/browse/JBTM-3736 (Upgrade h2 dependency)?

marcosgopen:
Sorry @mmusgrov and @jmfinelli . I didn't want to include h2 dependency upgrade in this PR. It will go in a separate PR, therefore I removed those commits. Regarding the JAX-RS resource changes those are needed after the removal of the scopes from the bom (otherwise we would have some failures on CI). I will double-check those changes and write a comment about those.

marcosgopen:
> 

I will write the description in the associated JIRA (JBTM-3776). In particular I will describe why JAX-RS resource changes are needed. Thanks for reminding me @mmusgrov

rts/lra/proxy/api/pom.xml
```diff
@@ -59,7 +59,6 @@
     <dependency>
       <groupId>org.jboss.resteasy</groupId>
       <artifactId>resteasy-client</artifactId>
-      <scope>test</scope>
```
marcosgopen, line 62: I needed to remove the scope 'test' otherwise the compilation fails after removing scope from dependencies' bom.

rts/lra/proxy/test/pom.xml
```diff
@@ -44,6 +44,10 @@
       <artifactId>resteasy-client</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
marcosgopen, line 47: I needed to add jakarta.ws.rs-api dependency with scope 'compile' (default one) otherwise the compilation fails after removing scope from dependencies' bom.

jmfinelli (changes requested):
Could you please squash all commits together?

marcosgopen:
> Could you please squash all commits together?

Done

mmusgrov:
> > @jmfinelli Would that be issue https://issues.redhat.com/browse/JBTM-3736 (Upgrade h2 dependency)?
> 
> Yep, I think that JBTM-3736 is a good candidate for bumping the h2 version. BTW, WDYT to explicitly add in Narayana's `CONTRIBUTING.md` file that minor changes (e.g. version bumps, rephrasing a sentence, code style, etc.) do _not_ need to be tracked with a JIRA? As far as I understand, JIRAs can help to communicate technical details problems to contributors and keep track of the motivations behind a change in the code. In other words, connecting a PR to a JIRA makes sense (AFAIU) when the change in the code is big enough (and/or complex enough) that an explanation is needed.

1. A version bump is a replacement of functionality that we depend on, so without inspecting the release notes can we really infer that it's a minor change? But I agree that, in general, we can skip the issue tracker (notwithstanding the fact that our parent project, WildFly, asks for component upgrade trackers).

2. We explicitly prohibit PR's that only change the style of the code.

3. Rephrasing sentences seems a bit of a vague statement so let's keep tracking such changes in the issue tracker.

4. And I'm not sure we want to add a catch all of `etc.` since it opens up potential for us to become embroiled in all sorts of random discussions.

Hence I'd vote no to such a change to `CONTRIBUTING.md`.

narayana-bom/pom.xml
moulalis: > I confirm the scope is still set in each module pom (as it was before PR #2079), is it right @moulalis ?

Yes @marcosgopen, I didn't change scope in any of the module so it is as it was before.

mmusgrov (approved):

narayana-bom/pom.xml
mmusgrov: @jmfinelli from my point of view I consider this conversation as resolved but I will not press the `Resolve conversation` button since it it requires both of us to "agree to commit" to it!

mmusgrov:
> This is one PR for 3 separate and unrelated issues: dependency scope modifications, CMR test changes updating JDBC urls, and JAX-RS resource changes.
> 
> I'd like to be able to approve it but it will set a precedent if we allow it - in particular the JAX-RS resource changes are unrelated and do not have any means of associating the change with any problem).
> 
> Also the associated JIRA (JBTM-3776) has a blank description (the tool should be making this a mandatory field).

Your later updates have resolved my issues, thanks.

rts/lra/proxy/test/pom.xml
mmusgrov: +1

PR #2085, opened 2023-02-08
[PR_TEMPLATE] Users are now notified that our CI doesn't trigger test axis

jmfinelli:
This PR notifies users that our CI isn't working properly as test axis aren't triggered automatically.

NO_TEST

PR #2123, opened 2023-04-14
Update CONTRIBUTING.md

tomjenkinson:
NO_TEST


marcosgopen (approved):
Thanks Tom for your PR

jmfinelli (approved):

CONTRIBUTING.md
```diff
@@ -60,6 +60,7 @@ But the following guidelines provide a more detailed set of requirements that we
...
 3. Engineers are not allowed to submit PRs which only contain formatting changes. The guidance on formatting code are covered in the [Coding Guidelines](#coding-guidelines) section below.
 4. New PRs are tested against multiple Jenkins CI axes. If you know that a change only affects particular axes then you can disable the ones that aren't required, ask a team member for clarification if you're not sure (or look at how the `PROFILE` variable is used in the CI script `scripts/hudson/narayana.sh`. When you first create a PR, the PR description field contains basic instructions about how to disable a CI test axis.
+5. If the pull request is not ready to be reviewed yet, you must add the label "Hold" - this can be added when the pull request is being created.
```
mmusgrov, line 63: Can we also add something to deter users from using our CI to test their code changes as a way of avoiding their responsibility of doing the testing locally first.

jmfinelli: Yes please! +1 from me!

tomjenkinson: I would prefer to deal with that separate to this pull request which just raises the attention on the Hold label and so I expect it to be a different decision.

If the project doesn't want people to use CI for pre-testing we could make the poller configuration ignore pull requests which have the "Hold" label and describe that restriction in this CONTRIBUTING.md. If that is what the project would like I still think that could be follow up once it's agreed by the project.

marcosgopen: Personally I find the Ci testing useful because I find running all profiles locally is quite time-consuming. I agree the core profile should be run locally first then. Just to mention the CI poller is ignoring all pull request at the moment (https://github.com/jbosstm/narayana/pull/2085)

marcosgopen: What about if we ask users to check that github actions (LRA github action https://github.com/jbosstm/narayana/blob/main/.github/workflows/lra-test.yml) pass on their branch before creating the PR?

marcosgopen: BTW I agree with Tom this can be done in a separate PR.

mmusgrov: > I would prefer to deal with that separate to this pull request (which just raises the attention on the Hold label) - I expect it to be a different decision.
> 
> If the project doesn't want people to use CI for pre-testing we could make the poller configuration ignore pull requests which have the "Hold" label and describe that restriction in this CONTRIBUTING.md. If that is what the project would like I still think that could be follow up once it's agreed by the project.

Well it gives the wrong message, if I read that I would infer that the team is okay with people using our CI as a testbed for their development. And if you are telling them to use the Hold label wisely then using the label just to get some free testing seems like an unwise use of the label. So telling them under what circumstances they can use the Hold label needs to include our caveat, ie the caveat belongs with the statement you are proposing in this PR.

tomjenkinson: I like the idea of the GitHub action from @marcosgopen but I think we should add a CORE version rather than use the LRA one personally.

tomjenkinson: @jmfinelli @mmusgrov please can you update your reviews to request changes if you would like the pull request changing?

jmfinelli: > Personally I find the Ci testing useful because I find running all profiles locally is quite time-consuming. I agree the core profile should be run locally first then. Just to mention the CI poller is ignoring all pull requests at the moment (#2085)

I would like to mention that we don't have unlimited resources on our CI. There is a maximum number of agents that can run in parallel. In other words, if contributors use our CI to do the initial testing of their PRs, our CI has to deal with a workload that may be over the allocated resources. The reason we use this limitation in our CI comes from the assumption that users raise PRs only when they know (with a fair level of confidence) that their PR is ready to be merged.

Moreover, although this might not be everyone's concern, using our CI to run the first round of testing for a new PR is not a green option. What I mean is that testing on our CI still means  that there is a CPU running somewhere, which means that we're wasting energy (and money). Being conscious about how we use our CI is another way to be conscious about the environment.

mmusgrov (changes requested):
Can we also add something to deter users from using our CI to test their code changes as a way of avoiding their responsibility of doing the testing locally first.

CONTRIBUTING.md
mmusgrov: So @tomjenkinson something like the following could be used:

```
5. Although the expectation is that the engineer raising the PR has tested his changes prior to submitting the request to merge changes, there are some circumstances where this may not be possible in which case you must add the label "Hold" and update the PR description indicating why it isn't ready for review just yet. This policy is a signal to reviewers that the changes are not yet ready to be reviewed.
```

mmusgrov:
> We can, please can you propose some wording? And would you like it to be enforced by the mechanism I mentioned whereby (when working) the poller would ignore pull requests with `Hold` label. This would then lead more to `Hold` being for specifically requested peer review of code rather than pre-testing.

I proposed some additional text.

If the engineer doesn't want the PR to be tested then omitting the axes/using NO_TEST in the description will suffice so no need to add additional logic to the poller.

jmfinelli (approved):

CONTRIBUTING.md
marcosgopen: Regarding the CORE github action see https://github.com/jbosstm/narayana/pull/2124

mmusgrov (approved):

PR #2124, opened 2023-04-17
core test github action to be approved

marcosgopen:
core-test.yml file has been pushed on main branch: https://github.com/jbosstm/narayana/commit/b66a81ee1725fef91752fb0ca9513eabe7c9dc2f

the core-test tests the CORE profile and is now working: https://github.com/jbosstm/narayana/actions/runs/4719819517

If you agree to have the core-test github action tested in Narayana I can close this PR.


mmusgrov:
@marcosgopen  How does this test complement the one running on CI and why can we not have one or the other rather than both.

tomjenkinson (approved):

marcosgopen:
> This pull request itself might not the best place to discuss the point I am going to make but something I think we could consider (if we reconsidered adding more GitHub actions) is if we can just make them run on-demand by the developer (is it `workflow_dispatch`?)?

Thanks Tom for the review. The workflow_dispatch is to run it manually too. If we want to avoid the action to run it automatically we need to use something like:
`on:
  push:
    # Sequence of patterns matched against refs/heads
    branches-ignore:    
      - '*'`

PR #2131, opened 2023-04-27
Declare no-arg constructor to TMApplication

marcosgopen:
From Jakarta EE cdi specification:
_If a bean class does not explicitly declare a constructor using @Inject, the constructor that accepts
no parameters is the bean constructor_.

This is needed only for bean classes (i.e. annotated as @ApplicationScoped ) and the TMApplication is not annotated. However it is possible that a project uses 'all' as bean-discovery-mode and 'not having a no-arg constructor or a constructor with @Inject annotation' might be a problem. Therefore I recommend having a no-arg constructor  and/or an @Inject constructor. 

related issue is https://github.com/eclipse/lsp4jakarta/issues/191

Note 1: The existing constructor cannot be annotated with @Inject because the parameters must be injectable as well according to the [spec](https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html#instantiationl)

Note 2: only recently the default bean-discovery-mode has changed to 'annotated' (CDI 4 https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#_jakarta_contexts_and_dependency_injection_4_0 )

Note 3: I got an error on a quarkus-resteasy quickstart related to this issue:
```
Build step
io.quarkus.resteasy.server.common.deployment.ResteasyServerCommonProcessor#build
threw an exception: java.lang.RuntimeException: Unable to handle class:
org.jboss.jbossts.star.service.TMApplication

[ERROR] Caused by: java.lang.NoSuchMethodException:
org.jboss.jbossts.star.service.TMApplication.<init>()
[ERROR]         at
java.base/java.lang.Class.getConstructor0(Class.java:3585)
[ERROR]         at
java.base/java.lang.Class.getConstructor(Class.java:2271)
[ERROR]         at
io.quarkus.resteasy.server.common.deployment.ResteasyServerCommonProcessor.getAllowedClasses(ResteasyServerCommonProcessor.java:1010)
```

CORE RTS LRA

### DIFF
--- a/rts/at/tx/src/main/java/org/jboss/jbossts/star/resource/RESTRecord.java
+++ b/rts/at/tx/src/main/java/org/jboss/jbossts/star/resource/RESTRecord.java
@@ -150,8 +150,7 @@ public class RESTRecord extends AbstractRecord implements Comparable {
     private void check_suspend(Fault f) {
         if (fault.equals(f))
         {
-            try
-            {
+            try {
                 log.infof("%s: for 10 seconds", f);
                 Thread.sleep(10000);
             } catch (InterruptedException e) {
@@ -212,8 +211,7 @@ public class RESTRecord extends AbstractRecord implements Comparable {
         if (prepareURI == null || txId == null)
             return TwoPhaseOutcome.PREPARE_READONLY;
 
-        try
-        {
+        try {
             String body = new TxSupport().httpRequest(new int[] {HttpURLConnection.HTTP_OK}, this.prepareURI, "PUT",
                     TxMediaType.TX_STATUS_MEDIA_TYPE, TxSupport.toStatusContent(TxStatus.TransactionPrepared.name()));
 
@@ -308,8 +306,7 @@ public class RESTRecord extends AbstractRecord implements Comparable {
         String commitUri = (nextState == TxStatus.TransactionCommittedOnePhase && this.commitOnePhaseURI != null)
             ? this.commitOnePhaseURI : this.commitURI;
 
-        try
-        {
+        try {
             if (log.isTraceEnabled())
                 log.tracef("committing %s", commitUri);
 
@@ -374,8 +371,7 @@ public class RESTRecord extends AbstractRecord implements Comparable {
                     return false;
                 }
 
-                try
-                {
+                try {
                     TxSupport.getStatus(new TxSupport().httpRequest(new int[] {HttpURLConnection.HTTP_OK},
                             uri, "PUT", TxMediaType.TX_STATUS_MEDIA_TYPE,
                             TxSupport.toStatusContent(nextState.name())));
@@ -407,8 +403,7 @@ public class RESTRecord extends AbstractRecord implements Comparable {
      * @return true if the participant did move
      */
     private boolean hasParticipantMoved() {
-        try
-        {
+        try {
             if (log.isTraceEnabled())
                 log.tracef("seeing if participant has moved: %s  recoveryURI: %s", coordinatorID, recoveryURI);
 
@@ -476,8 +471,7 @@ public class RESTRecord extends AbstractRecord implements Comparable {
     }
 
     public boolean save_state(OutputObjectState os, int t) {
-        try
-        {
+        try {
             os.packString(txId);
             os.packBoolean(prepared);
             os.packString(participantURI);
@@ -501,8 +495,7 @@ public class RESTRecord extends AbstractRecord implements Comparable {
     }
 
     public boolean restore_state(InputObjectState os, int t) {
-        try
-        {
+        try {
             txId = os.unpackString();
             prepared = os.unpackBoolean();
             participantURI = os.unpackString();
--- a/rts/at/tx/src/main/java/org/jboss/jbossts/star/service/TMApplication.java
+++ b/rts/at/tx/src/main/java/org/jboss/jbossts/star/service/TMApplication.java
@@ -47,6 +47,16 @@ public class TMApplication extends Application {
     Set<Class<?>> classes = new HashSet<Class<?>> ();
 
     public TMApplication(Class<?> ... extraClasses) {
+        this();
+        
+        try
+        {
+            Collections.addAll(classes, extraClasses);
+        } catch (Throwable e) {
+          RESTATLogger.atI18NLogger.warn_jaxrsTM(e.getMessage(), e);
+        }
+    }
+    public TMApplication() {
 //        singletons.addAll(Arrays.asList(resources));
         try
         {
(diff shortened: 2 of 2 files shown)


rts/at/tx/src/main/java/org/jboss/jbossts/star/service/TMApplication.java
```diff
@@ -46,7 +46,17 @@ public class TMApplication extends Application {
     HashSet<Object> singletons = new HashSet<Object>();
     Set<Class<?>> classes = new HashSet<Class<?>> ();
 
+
```
mmusgrov, line 49: You added an extra blank line, the policy is that we should match the existing format in the file.

rts/at/tx/src/main/java/org/jboss/jbossts/star/service/TMApplication.java
```diff
@@ -46,7 +46,17 @@ public class TMApplication extends Application {
...
     public TMApplication(Class<?> ... extraClasses) {
+        this();
+        try
```
mmusgrov, line 52: I'm pretty sure the style says we need a blank line before try, if not then the policy is that we should match the existing format in the file.

rts/at/tx/src/main/java/org/jboss/jbossts/star/service/TMApplication.java
```diff
@@ -46,7 +46,17 @@ public class TMApplication extends Application {
...
+        {
```
mmusgrov, line 53: The style guidelines say that we should match the existing format in the file. Most of the other files implementing REST-AT don't put the opening brace on a separate line. When I wrote it they all had this style but other PRs have used a different style in places.

mmusgrov (approved):
There were some styling issues, once those are fixed I'm ok for this one to be merged.

rts/at/tx/src/main/java/org/jboss/jbossts/star/service/TMApplication.java
marcosgopen: I've just added the blank line before try. Thanks

rts/at/tx/src/main/java/org/jboss/jbossts/star/service/TMApplication.java
marcosgopen: Personally I prefer not to put the opening brace on a separate line, but I did so in order to match the existing format of the file (see https://github.com/jbosstm/narayana/blob/main/rts/at/tx/src/main/java/org/jboss/jbossts/star/service/TMApplication.java#L51-L52)

mmusgrov: The module contains a mix of positioning of braces. And REST-AT is a module containing many source files and the general style is not to put the brace on its own line. So if the second SHA contains style changes only then leave it in a separate commit.

marcosgopen:
I added a second commit for code style. I edited the 'try {'  format. Now all the try- catch blocks in restat api follow the following rules:
- no opening brace on a separate line
- space between 'try' and opening brace

rts/at/tx/src/main/java/org/jboss/jbossts/star/service/TMApplication.java
marcosgopen: Could you confirm the code style commit I pushed is ok with you @mmusgrov ?

rts/at/tx/src/main/java/org/jboss/jbossts/star/resource/RESTRecord.java
```diff
@@ -212,8 +211,7 @@ public int topLevelPrepare() {
         if (prepareURI == null || txId == null)
```
mmusgrov, line 211: This one needs braces as well

mmusgrov: Actually this isn't in our style rules so I guess keep it as it is.

But should we include it in our checkstyle rules?

rts/at/tx/src/main/java/org/jboss/jbossts/star/resource/RESTRecord.java
```diff
@@ -150,8 +150,7 @@ public int nestedPrepare() {
     private void check_suspend(Fault f) {
         if (fault.equals(f))
         {
```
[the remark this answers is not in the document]
mmusgrov: Our rule is:

```
        <module name="LeftCurly">
            <property name="option" value="eol"/>
            <property name="tokens" value="CLASS_DEF,CTOR_DEF,INTERFACE_DEF,METHOD_DEF,LITERAL_SWITCH,STATIC_INIT"/>
        </module>
```

so I wonder why this isn't being reported when we run checkstyle over it?

rts/at/tx/src/main/java/org/jboss/jbossts/star/resource/RESTRecord.java
mmusgrov: I added the `LITERAL_IF` token (https://checkstyle.sourceforge.io/config_blocks.html#LeftCurly) to the ruleset and it didn't report the missing curly:

```
         <module name="LeftCurly">
             <property name="option" value="eol"/>
-            <property name="tokens" value="CLASS_DEF,CTOR_DEF,INTERFACE_DEF,METHOD_DEF,LITERAL_SWITCH,STATIC_INIT"/>
+            <property name="tokens" value="LITERAL_IF,CLASS_DEF,CTOR_DEF,INTERFACE_DEF,METHOD_DEF,LITERAL_SWITCH,STATIC_INIT"/>
         </module>
```

mmusgrov (approved):
I'm thinking the whole module is inconsistent with regards to styling so I am still approving your fix for the no-arg constructor. I will leave it up to you as to whether or not to ignore my styling change request.

rts/at/tx/src/main/java/org/jboss/jbossts/star/resource/RESTRecord.java
marcosgopen: I changed only the try-catch blocks. I wouldn't change the checkstyle configuration because we would need to change  LITERAL_TRY  LITERAL_IF LITERAL_ELSE and so on and I am afraid we would appear too strict. I'd rather contributors were coherent with the module code style/format.

mmusgrov:
@marcosgopen Is there a JBTM for this, if not will you create one, link this PR and mark the PR as closed on the issue and update the `Fix Version/s` field to `6.next`.

PR #2138, opened 2023-05-26
JBTM-3779: Move out the test artifacts in narayana-bom

marcosgopen:
https://issues.redhat.com/browse/JBTM-3779

CORE TOMCAT AS_TESTS RTS JACOCO XTS QA_JTA QA_JTS_OPENJDKORB PERFORMANCE LRA

### DIFF
--- a/ArjunaJTS/narayana-jts-idlj/pom.xml
+++ b/ArjunaJTS/narayana-jts-idlj/pom.xml
@@ -18,11 +18,6 @@
   <description>JTS uber jar with idlj stubs</description>
   <dependencies>
     <dependency>
-      <groupId>org.apache.ant</groupId>
-      <artifactId>ant</artifactId>
-      <scope>provided</scope>
-    </dependency>
-    <dependency>
       <groupId>jakarta.annotation</groupId>
       <artifactId>jakarta.annotation-api</artifactId>
       <scope>provided</scope>
--- a/narayana-bom/pom.xml
+++ b/narayana-bom/pom.xml
@@ -36,11 +36,6 @@
         </exclusions>
       </dependency>
       <dependency>
-        <groupId>io.mashona</groupId>
-        <artifactId>mashona-logwriting</artifactId>
-        <version>${version.io.mashona}</version>
-      </dependency>
-      <dependency>
         <groupId>org.jboss.logging</groupId>
         <artifactId>jboss-logging</artifactId>
         <version>${version.org.jboss.logging.jboss-logging}</version>
@@ -55,16 +50,6 @@
         <artifactId>common</artifactId>
         <version>${project.version}</version>
         <classifier>tests</classifier>
-      </dependency>
-      <dependency>
-        <groupId>org.jboss.byteman</groupId>
-        <artifactId>byteman-bmunit</artifactId>
-        <version>${version.org.jboss.byteman}</version>
-      </dependency>
-      <dependency>
-        <groupId>com.h2database</groupId>
-        <artifactId>h2</artifactId>
-        <version>${version.com.h2database}</version>
       </dependency>
       <dependency>
         <groupId>org.jboss.narayana.arjunacore</groupId>
@@ -126,16 +111,6 @@
         <artifactId>arquillian-core-api</artifactId>
         <version>${version.org.jboss.arquillian.core}</version>
       </dependency>
-      <dependency>
-        <groupId>org.jboss.arquillian.protocol</groupId>
-        <artifactId>arquillian-protocol-servlet-jakarta</artifactId>
-        <version>${version.org.jboss.arquillian.core}</version>
-      </dependency>
-      <dependency>
-        <groupId>org.jboss.arquillian.extension</groupId>
-        <artifactId>arquillian-extension-byteman</artifactId>
-        <version>${version.arquillian.byteman}</version>
-      </dependency>
 
       <!-- Standalone JNDI server -->
       <dependency>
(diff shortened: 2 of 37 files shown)


marcosgopen:
@zhfeng Is this what you were looking for? I created the narayana-bom-test with test dependencies. It has been implemented as a submodule of the narayana-bom in this PR.

zhfeng:
Thanks @marcosgopen for working on this PR. I just wonder why we need `narayana-test-bom` here. It seems this test-bom can only be used internally.

narayana-bom/narayana-bom-test/pom.xml
zhfeng, line 1: Is it possible to manage these test dependencies in the top `pom.xml`? There is no strong opinion to have a `test-bom` since it is only used internally.

marcosgopen: I think we can move those test dependencies in the root pom.xml . Let me share this on Zulip https://narayana.zulipchat.com/#narrow/stream/323715-developers/topic/JBTM-3735.20Create.20narayana.20boms

marcosgopen: As discussed on Zulip it would be better to have a test BOM "to ensure that we use the same version [for a dependency] in all projects and modules."

ArjunaJTS/narayana-jts-idlj/pom.xml
```diff
@@ -24,11 +24,6 @@
   <name>Narayana: ArjunaJTS narayana-jts-idlj</name>
   <description>JTS uber jar with idlj stubs</description>
   <dependencies>
-    <dependency>
-      <groupId>org.apache.ant</groupId>
-      <artifactId>ant</artifactId>
-      <scope>provided</scope>
-    </dependency>
```
jmfinelli, line 31: I guess this is not used, isn't it?

narayana-bom/pom.xml
```diff
@@ -294,7 +219,7 @@
         <version>${version.jakarta.xml.ws.jakarta.xml.ws-api}</version>
       </dependency>
 
-      <!-- Narayan full -->
+      <!-- Narayana full -->
```
jmfinelli, line 222: This one goes in a different commit

jmfinelli (changes requested):

narayana-bom/narayana-bom-test/pom.xml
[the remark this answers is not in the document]
marcosgopen: Thanks @jmfinelli for your review. This is a good question. Actually I thought that when we need narayana-bom-test we also need narayana-bom IMO and this is why narayana-bom-test is a submodule of narayana-bom so that it inherits its dependencies. In fact when you need narayana-bom-test you don't need to import also narayana-bom. Said that, we could also move the narayana-bom-test as you suggested if it is better to do so.

ArjunaJTS/narayana-jts-idlj/pom.xml
marcosgopen: Yes, exactly. It is a test dependency and it is not used there. So instead using the narayana-bom-test I thought it would be better to remove that test dep.

narayana-bom/pom.xml
```diff
@@ -11,6 +11,9 @@
   <packaging>pom</packaging>
   <name>Narayana Bom</name>
   <description>Narayana BOM</description>
+  <modules>
+    <module>narayana-bom-test</module>
+  </modules>
```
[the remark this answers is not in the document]
marcosgopen: See my previous comment. As I wrote there it is ok with me to have independent BOMs, but I think narayana-bom-test always goes together with narayana-bom.

narayana-bom/pom.xml
marcosgopen: Thanks Manuel, I agree and I will update my PR.

narayana-bom/pom.xml
```diff
@@ -929,6 +665,11 @@
         <artifactId>wrapper</artifactId>
         <version>${version.tanukisoft}</version>
       </dependency>
+      <dependency>
+        <groupId>io.quarkus</groupId>
+        <artifactId>quarkus-resteasy</artifactId>
+        <version>${version.quarkus}</version>
+      </dependency>
```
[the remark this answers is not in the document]
marcosgopen: It was added because it is present in Narayana (https://github.com/jbosstm/narayana/blob/main/rts/at/tx/pom.xml#L162) but it has not been included in the bom yet.

marcosgopen:
I set the Hold label to this PR because I want to prepare performance and quickstart PRs that are affected by this PR before merging this.

marcosgopen:
> Good thinking Marco. Will that be using the `depends on pr/xxx` directive (or whatever the correct syntax is)?

Sure, I will add the "depends on" syntax.

narayana-bom/narayana-bom-test/pom.xml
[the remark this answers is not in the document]
jmfinelli: I don't think creating a dependency between the main bom and the test bom is worth saving a couple of lines in the poms that use both boms.

jmfinelli, line 1: I don't think creating a dependency between the main bom and the test bom is worth saving a couple of lines in the poms that use both boms.

jmfinelli (changes requested):

narayana-bom/narayana-bom-test/pom.xml
marcosgopen: Ok, thanks @jmfinelli! I will them make them independent. What about test scope in the narayana-bom-test? Should I use test scope for test dependencies or should I let the projects choose what scope should they use (so having compile scope in the bom? IMO this will affect less external projects using Narayana test dependencies.

jmfinelli: It makes sense to me to use the `test` scope in the test bom

marcosgopen: ok, thanks

jmfinelli (approved):
LGTM. Just a question, can't the version of the bom be implied from the overall version of the project? i.e. ${project.version}

marcosgopen:
> LGTM. Just a question, can't the version of the bom be implied from the overall version of the project? i.e. ${project.version}

Yes, it is also possible to use the ${project.version} annotation. I do not have any preference about using one or the other.

marcosgopen:
Thanks for your review @jmfinelli !

jmfinelli:
> > LGTM. Just a question, can't the version of the bom be implied from the overall version of the project? i.e. ${project.version}
> 
> Yes, it is also possible to use the ${project.version} annotation. I do not have any preference about using one or the other.

In that case, my personal preference would be to replace all versions with ${project.version}. Less boilerplate

marcosgopen:
@jmfinelli I addressed your comment. Now merging the PR. Thanks again for your review.

PR #2141, opened 2023-05-30
Gitignore

marcosgopen:
add /bin/ folders in gitignore

NO_TEST


.gitignore
```diff
@@ -26,3 +26,4 @@ tomcat-deployment/
 narayana-tomcat/
 CosServices.cfg
 audit.log
+**/bin/
```
mmusgrov, line 29: This will ignore the following files which need to be under revision control: https://github.com/jbosstm/narayana/tree/main/ArjunaJTS/services/bin

mmusgrov (changes requested):

.gitignore
marcosgopen: Thanks Mike for your review. What about this then:
`**/bin/
!ArjunaJTS/services/bin/*`

mmusgrov: If it works then OK, but it does make things less minimal and will need to be maintained.

But why do you have a bin directory in your clone - most developers will not be creating bin directories, in fact the gitignore file is for things produced by the build, not for files that the user creates in his private clone.

marcosgopen: It is probably a wrong configuration in my environment creating /bin/ folder for .class files. As it is not a common problem I am going to close this PR. Thanks for your answer Mike.

PR #2149, opened 2023-06-27
[JBTM-3790] Add creationTimeMillis as bean property to replace internal birthDate local variable

graben:
[JBTM-3790](https://issues.redhat.com/browse/JBTM-3790)

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
@@ -512,6 +512,16 @@ public class StateManager
     }
 
     /**
+     * @return the time the object was created in milliseconds since midnight, January 1, 1970 UTC
+     * @see System.currentTimeMillis
+     */
+
+    public final long getCreationTimeMillis()
+    {
+        return creationTimeMillis;
+    }
+
+    /**
      * Destroy the object (e.g., remove its state from the persistent store.)
      * Calls to destroy for volatile objects (ones not maintained within the
      * volatile object store) are ignored, and FALSE is returned.
@@ -688,14 +698,13 @@ public class StateManager
             else
                 UidHelper.packInto(Uid.nullUid(), os);
             
-            long birthDate = System.currentTimeMillis();
-            
             if (tsLogger.logger.isTraceEnabled()) {
                 tsLogger.logger.trace("StateManager.packHeader for object-id " + get_uid()
-                        + " birth-date " + birthDate);
+                        + " birth-date " + creationTimeMillis);
+            }
             }
             
-            os.packLong(birthDate);
+            os.packLong(creationTimeMillis);
         }
         catch (IOException ex)
         {
@@ -748,11 +757,11 @@ public class StateManager
                 throw new IOException(tsLogger.i18NLogger.get_StateManager_15());
             }
             
-            long birthDate = os.unpackLong();
+            creationTimeMillis = os.unpackLong();
             
             if (tsLogger.logger.isTraceEnabled()) {
                 tsLogger.logger.trace("StateManager.unpackHeader for object-id " + get_uid()
-                        + " birth-date " + birthDate);
+                        + " birth-date " + creationTimeMillis);
             }
             
             hdr.setTxId(txId);
@@ -1474,6 +1483,7 @@ public class StateManager
     private int currentStatus = ObjectStatus.PASSIVE;
     private int initialStatus = ObjectStatus.PASSIVE;
     private int myType;
+    private long creationTimeMillis = System.currentTimeMillis();
     private ParticipantStore participantStore = null;
     private String storeRoot = null;
     private ReentrantLock mutex = new ReentrantLock();


ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
```diff
@@ -511,6 +511,15 @@ public final Uid get_uid ()
         return objectUid;
     }
 
+    /**
+     * @return the object's birth date
```
mmusgrov, line 515: Will you indicate the time units stating that it returns the time the object was created in units of milliseconds since midnight, January 1, 1970 UTC (and include a reference to @see <a href="https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/System.html#currentTimeMillis()">System.currentTimeMillis</a>) or something similar

mmusgrov (changes requested):

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
graben: Javadoc is updated

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
```diff
@@ -511,6 +511,16 @@ public final Uid get_uid ()
         return objectUid;
     }
 
+    /**
+     * @return the time the object was created in milliseconds since midnight, January 1, 1970 UTC
+     * @see System.currentTimeMillis
+     */
```
mmusgrov, line 517: I'm going to approve the PR but I'm interested to know if your IDE added this blank line or yourself.
Note that at some point we are likely to update our checkstyle rules to disallow it.

mmusgrov (approved):
Thanks for finding this bug and for raising the PR.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
graben: I followed the design of the file. Everywhere newlines between method and javadoc.

mmusgrov:
@jmfinelli Will this one be ok to merge after CI finishes?

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
```diff
@@ -1468,6 +1476,7 @@ final void destroyed ()
     protected Hashtable usingActions = null;
     protected final Uid objectUid;
     protected int objectModel = ObjectModel.SINGLE;
+    protected long birthDate = System.currentTimeMillis();
```
mmusgrov, line 1479: @graben since this is now an instance variable with greater visibility what do you think about renaming it to `creationTime` which is more standard than birthDate (which will mean changing the name of the `setter` as well).

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
mmusgrov: Oh yes, you're correct, sorry about that.

mmusgrov (changes requested):
On further thought I have a change request relating to the new instance variable that this PR introduces: https://github.com/jbosstm/narayana/pull/2149#discussion_r1248760555

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
graben: Good idea, will change it as soon as possible. :-)

graben:
@mmusgrov : I renamed the variable to `creationTimeMillis` to avoid conflicts with overriding classes that return `String `for method `getCreationTime()`. I also slightly reworked the `TRACE` logging during `packHeader`/`unpackHeader` to make it look more 21st century like. ;)

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
mmusgrov: > Good idea, will change it as soon as possible. :-) Btw getCreationTime should not return String as mentioned in your examples?

Why not? The actual returned string is `new SimpleDateFormat("E, dd MMM yyyy HH:mm:ss Z").format(new Date(birthDate));`

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
```diff
@@ -748,11 +757,11 @@ protected void unpackHeader (InputObjectState os, Header hdr)
...
             }
             
-            long birthDate = os.unpackLong();
+            creationTimeMillis = os.unpackLong();
             
             if (tsLogger.logger.isTraceEnabled()) {
-                tsLogger.logger.trace("StateManager.unpackHeader for object-id " + get_uid()
-                        + " birth-date " + birthDate);
+                tsLogger.logger.tracef("StateManager.unpackHeader for object-id %s creationTime %s",
+                        get_uid(), Instant.ofEpochMilli(creationTimeMillis));
```
mmusgrov, line 764: I'm not convinced that changing the trace message from printing a long to printing a date is related to this change, normally we'd require justification for adding extra changes beyond just the fix. But since this is "just" a trace message and is unlikely to impact users I'll still approve. @jmfinelli are you still ok to approve with these  extra changes to the PR?

mmusgrov (approved):

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
graben: The question was before I decided to rename to the variable. I think it's better to return the long value rather than the String representation if further execution is done with the date value!

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/StateManager.java
jmfinelli: WDYT if we include both of them?
```suggestion
tsLogger.logger.tracef("StateManager.unpackHeader for object-id %s creationTime %s (creationTime in milliseconds: %s)",
                        get_uid(), Instant.ofEpochMilli(creationTimeMillis), creationTimeMillis);
```

graben: I see no value in outputting the long value. I think most people will ask new Date(...) to resolve what the log is telling. :-)

mmusgrov: long values can be easily compared using scripts but a formatted date string cannot.
I don't see any benefit in producing a verbose date (which is kind of already available since log messages are timestamped). Besides, the change to the log message is unrelated to the issue at hand, if you do need to see a different message then maybe add it in an new PR.

graben: Okay, I reverted any log format specific change

mmusgrov (approved):
Thanks for the bug report and fix.

PR #2151, opened 2023-07-08
JBTM-3790 fix syntax error introduced by last PR

mmusgrov:
https://issues.redhat.com/browse/JBTM-3790

CORE

Fix for syntax error caused by pr/2149.


mmusgrov:
The CI run for the CORE profile shows that main now compiles so I will merge it early.
Running the QA_JTA profile was accidental so I'm not waiting for that.

mmusgrov:
@jmfinelli I merged this without a review since it is fixing a syntax error which stops all of our axes from running.

FYI we need automated CI testing, the error this PR fixes was caused by PR #2149 - I never re-ran CI after the final change to that PR.

PR #2152, opened 2023-07-10
[JBTM:3785] Upgrade Servlet API from 5.0.0 to 6.0.0 in the BOM

moulalis:
Issue: https://issues.redhat.com/browse/JBTM-3785

CORE TOMCAT AS_TESTS RTS !JACOCO XTS QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE LRA !DB_TESTS

jmfinelli: Testing axis modified

### DIFF
--- a/narayana-bom/pom.xml
+++ b/narayana-bom/pom.xml
@@ -34,7 +34,7 @@
     <version.jakarta.persistence.jakarta.persistence-api>3.1.0</version.jakarta.persistence.jakarta.persistence-api>
     <version.jakarta.platform.jakarta.jakartaee-api>10.0.0</version.jakarta.platform.jakarta.jakartaee-api>
     <version.jakarta.resource.jakarta-resource-api>2.0.0</version.jakarta.resource.jakarta-resource-api>
-    <version.jakarta.servlet.jakarta-servlet-api>5.0.0</version.jakarta.servlet.jakarta-servlet-api>
+    <version.jakarta.servlet.jakarta-servlet-api>6.0.0</version.jakarta.servlet.jakarta-servlet-api>
     <version.jakarta.transaction.jakarta-transaction-api>2.0.1</version.jakarta.transaction.jakarta-transaction-api>
     <version.jakarta.ws.rs.jakarta-ws-rs-api>3.1.0</version.jakarta.ws.rs.jakarta-ws-rs-api>
     <version.jakarta.xml.bind.jakarta-xml-bind-api>4.0.0</version.jakarta.xml.bind.jakarta-xml-bind-api>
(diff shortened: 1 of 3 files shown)


mmusgrov:
@moulalis Hello, I noticed that the CORE profile is failing with `NoClassDefFoundError: jakarta/servlet/SingleThreadModel` which seems related, will you be able to find out what's causing it?

moulalis:
> @moulalis Hello, I noticed that the CORE profile is failing with `NoClassDefFoundError: jakarta/servlet/SingleThreadModel` which seems related, will you be able to find out what's causing it?

Sure I will look into it.

moulalis:
@mmusgrov 

Removals, deprecations or backwards incompatible changes

Issue 225 Deprecated wrapped response handling in the doHead method in favour of container implementation of HEAD method behavior
Remove API classes and methods that were deprecated in Servlet 5.0 and earlier. This includes removing the SingleThreadModel and HttpSessionContext interfaces and the HttpUtils class as well as various deprecated methods. There is not that class in the new api So dependencies that are expecting it fail i.e. undertow.

https://mvnrepository.com/artifact/io.undertow/undertow-servlet/2.3.7.Final

I can see last version is using servlet api 6, which is good to know.

Have to check if resteasy-undertow last version is using servlet 6. Will investigate it more and update.

Thanks.

mmusgrov:
Ah, right. Good info, thanks for digging into it.

marcosgopen:
Please @moulalis could you retest it upgrading the resteasy version to 6.2.5.Final and removing these 2 lines: https://github.com/jbosstm/narayana/blob/main/rts/at/tx/pom.xml#L27-L28. These line are not needed anymore since all dependencies would use servlet api 6.0

mmusgrov:
@moulalis Is there any update on this PR?

mmusgrov:
@moulalis does this PR need the `Hold` label on it?

moulalis:
> @moulalis does this PR need the `Hold` label on it?

I will put a hold label until Investigated.

moulalis:
> Please @moulalis could you retest it upgrading the resteasy version to 6.2.5.Final and removing these 2 lines (https://github.com/jbosstm/narayana/blob/main/rts/at/tx/pom.xml#L27-L28) and this line (https://github.com/jbosstm/narayana/blob/main/rts/at/tx/pom.xml#L51) . These lines are not needed anymore since all dependencies would use servlet api 6.0

Hi @mmusgrov @marcosgopen I've just updated PR to above changes, could you please trigger CI tests?

mmusgrov:
@marcosgopen do you think that we can merge this PR now?

PR #2155, opened 2023-08-21
comment out automatic website update

marcosgopen:
This PR disables automatic update of the website source.
The Narayana team is currently manually managing the update of the website
A new PR with the new script will follow as soon as it is tested and working.

NO_TESTS

### DIFF
--- a/build-release-pkgs.xml
+++ b/build-release-pkgs.xml
@@ -152,20 +152,8 @@
       <arg value="develop"/>
     </exec>
 
-    <!-- Build web page and push to jboss.org -->
-    <exec executable="bash" dir="${workdir}/narayana.io/">
-      <arg value="-c"/>
-      <arg value="${awestruct.executable} -P production"/>
-    </exec>
-    <antcall target="rsync.web"/>
-  </target>
-
-  <target name="rsync.web" if="${rsync.enabled}">
-    <exec executable="bash" dir="${workdir}/narayana.io/" failonerror="true">
-      <arg value="-c"/>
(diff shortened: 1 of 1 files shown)


mmusgrov (changes requested):
Instead of simply commenting out those targets why not delete them but replace the comment
`<!-- Build web page and push to jboss.org -->` with `We no longer build web pages, nor push to the file server` perhaps adding some text indication of how the release coordinator should perform the equivalent step.

We have a record of the deleted lines in github and the extra comment I've suggested will be a prompt to the reader to look back over the git history to see how we used to do it.

marcosgopen:
> Instead of simply commenting out those targets why not delete them but replace the comment `<!-- Build web page and push to jboss.org -->` with `We no longer build web pages, nor push to the file server` perhaps adding some text indication of how the release coordinator should perform the equivalent step.
> 
> We have a record of the deleted lines in github and the extra comment I've suggested will be a prompt to the reader to look back over the git history to see how we used to do it.

Maybe I was unclear with this PR. I would like to comment out these lines because I am currently manually managing the update of the website. I will create a PR replacing these commented out lines with proper scripts needed to automatically build and push/update the website.

marcosgopen:
> Instead of simply commenting out those targets why not delete them but replace the comment `<!-- Build web page and push to jboss.org -->` with `We no longer build web pages, nor push to the file server` perhaps adding some text indication of how the release coordinator should perform the equivalent step.
> 
> We have a record of the deleted lines in github and the extra comment I've suggested will be a prompt to the reader to look back over the git history to see how we used to do it.

Ok, I will change that as you suggested.

mmusgrov (approved):

PR #2169, opened 2023-09-26
Describe Narayana's BOMs usage and function

marcosgopen:
NO_TEST

This PR addresses Laird's forum question in 'narayana-users' conversation "[Why does the Narayana BOM feature third-party dependencies?](https://groups.google.com/g/narayana-users/c/CCWr8ZBIe_Q)"
I added comment for Narayana BOMs in order to clarify their function

### DIFF
--- a/narayana-bom-test/pom.xml
+++ b/narayana-bom-test/pom.xml
@@ -15,7 +15,12 @@
   <artifactId>narayana-bom-test</artifactId>
   <packaging>pom</packaging>
   <name>Narayana Bom Test</name>
-  <description>Narayana BOM for test dependencies</description>
+  <description>Narayana BOM is for centralising dependency management for development purposes (test scoped dependencies)</description>
+
+<!--
+	This Bill of Materials (BOM) is for centralising dependency management across the various modules of Narayana.
+    The BOM is intended for development purposes within Narayana.
+-->
 
   <properties>
     <version.ant-contrib>1.0b3</version.ant-contrib>
--- a/narayana-bom/pom.xml
+++ b/narayana-bom/pom.xml
@@ -14,7 +14,12 @@
   <artifactId>narayana-bom</artifactId>
   <packaging>pom</packaging>
   <name>Narayana Bom</name>
-  <description>Narayana BOM</description>
+  <description>Narayana BOM is for centralising dependency management for development purposes</description>
+
+<!--
+    This Bill of Materials (BOM) is for centralising dependency management across the various modules of Narayana.
+    The BOM is intended for development purposes within Narayana.
+-->
 
   <properties>
     <version.com.sybase.jConnect>6.0</version.com.sybase.jConnect>


narayana-bom/pom.xml
```diff
@@ -16,6 +16,13 @@
   <name>Narayana Bom</name>
   <description>Narayana BOM</description>
 
+<!--
+    This BOM is for developers' use across the Narayana repos rather than end users'.
+    This BOM's purpose is to collect all dependencies (with default scope, which is 'provided') needed in Narayana repos and centralize their update
+    In this way when a dependency is updated it is updated in all Narayana repositories
+    and we could immediately test the change everywhere and evaluate its impact 
+-->
+
```
jmfinelli, line 25: Thanks Marco. WDYT if we go for something shorter? Something like:
`This Bill of Materials (BOM) is for centralising dependency management across the various modules of Narayana.`

marcosgopen: Yes, I think something shorter would be better. I would underline anyway that this is for " developers' use across the Narayana repos rather than end users'." What about keeping that too?

marcosgopen:
@jmfinelli I updated the comment, what about it now?

tomjenkinson (dismissed):

jmfinelli (approved):

mmusgrov (changes requested):
@marcosgopen If the purpose is to address Laird's forum question then you should update the <description> tag.

narayana-bom-test/pom.xml
```diff
@@ -17,6 +17,11 @@
   <name>Narayana Bom Test</name>
   <description>Narayana BOM for test dependencies</description>
 
+<!--
+	This Bill of Materials (BOM) is for centralising dependency management across the various modules of Narayana.
+    The BOM is intended for development purposes within Narayana. End users should not use it for integrating Narayana into their projects.
```
mmusgrov, line 22: Why should end users not use it?

marcosgopen: Thanks for you review @mmusgrov , I removed the last sentence because I agree it has a different meaning than I intended. My point was that the BOM is intended for development purpose, so the previous sentence should be enough.

marcosgopen:
> @marcosgopen If the purpose is to address Laird's forum question then you should update the tag.

Thanks @mmusgrov for the review, I updated the PR description. If needed I will update the commit message too.

narayana-bom-test/pom.xml
mmusgrov: And users should still be okay to use it. If they want to use different versions of a dependency then they use the `<exclusions>` tag.

mmusgrov:
> > @marcosgopen If the purpose is to address Laird's forum question then you should update the tag.
> 
> Thanks @mmusgrov for the review, I updated the PR description. If needed I will update the commit message too.

I meant the `<description>` tag of the pom

narayana-bom-test/pom.xml
marcosgopen: Yes, I agree.

marcosgopen:
> > > @marcosgopen If the purpose is to address Laird's forum question then you should update the tag.
> > 
> > 
> > Thanks @mmusgrov for the review, I updated the PR description. If needed I will update the commit message too.
> 
> I meant the `<description>` tag of the pom

Very nice point! I updated the pom's description tag.

tomjenkinson (approved):

mmusgrov (approved):

PR #2180, opened 2023-10-17
JBTM-3820 Use fixed version of dependencies

marcosgopen:
NO_TESTS

This PR is for 5.9 branch

https://issues.redhat.com/browse/JBTM-3820

### DIFF
--- a/rts/at/tx/pom.xml
+++ b/rts/at/tx/pom.xml
@@ -213,13 +213,13 @@
     <dependency>
         <groupId>io.undertow</groupId>
         <artifactId>undertow-servlet</artifactId>
-        <version>[2.0.21,)</version>
+        <version>2.2.26.Final</version>
         <scope>test</scope>
     </dependency>
     <dependency>
         <groupId>io.undertow</groupId>
         <artifactId>undertow-core</artifactId>
-        <version>[2.0.21,)</version>
+        <version>2.2.26.Final</version>
         <scope>test</scope>
     </dependency>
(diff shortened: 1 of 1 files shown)


marcosgopen:
Thanks @mmusgrov for your review. After this PR I am going to remove 5.9 from CI as not supported anymore.
IIRC @tomjenkinson and @jmfinelli expressed the intention to close this PR. I can understand that this branch will not be maintained anymore, so I am ok with that. But since we know about its compilation failure I would merge this PR. Please let me know if you are ok with this choice.

mmusgrov:
> @marcosgopen I would be hesitant to maintain this branch any more and would feel like rejecting it still.
> 
> We talked about something elsewhere that I will bring here as a topic of branches being left open. We could discuss a process to close off old branches in detail somewhere (zulip?) but I think maybe just tagging the tip of the branch as `archived-<branchName>` and then delete the branch could work? We could do it with the 5.2, 5.5 and 5.9 branches in the states they are in now (instead of merging this pull request) /cc @mmusgrov @jmfinelli

I agree with Tom's proposal to tag them and then delete the branch and any jobs that depend on the branch.

marcosgopen:
Ok, so I am closing the PR because this 5.9 branch will be deleted/archived.

mmusgrov:
FYI I created the tags but decided to rename the branches to archived-5.2, archived-5.5 and archived-5.9 rather than delete them.

PR #2181, opened 2023-10-31
JBTM-3821 Report the reason for failing to set the nodeIdentifier pro…

mmusgrov:
https://issues.redhat.com/browse/JBTM-3821

CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !LRA !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/TxControlUnitTest.java
+++ b/ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/TxControlUnitTest.java
@@ -5,7 +5,9 @@
 
 package com.hp.mwtests.ts.arjuna.atomicaction;
 
+import static com.arjuna.ats.arjuna.coordinator.TxControl.NODE_NAME_SIZE;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/TxControlUnitTest.java
```diff
@@ -56,4 +58,13 @@ public void testGetSet () throws Exception
         
         assertTrue(TxControl.getXANodeName().equals(nodeName));
     }
+
+    @Test
+    public void testNodeName () {
+        try {
+            TxControl.setXANodeName(new String(new char[NODE_NAME_SIZE + 1]));
+        } catch (IllegalArgumentException e) {
+            assertNotNull("IllegalArgumentException has no detail message.", e.getMessage());
```
marcosgopen, line 67: Thanks @mmusgrov for your PR! Could we use the assertEqual rather then assertNotNull now?

mmusgrov: I don't see why we would need to do that, so no.

mmusgrov: We could change it when we move to junit 5

marcosgopen: Ok, I will approve it then.

mmusgrov: I think you are looking at the [javadoc for junit.framework](https://javadoc.io/static/junit/junit/4.13.2/junit/framework/Assert.html) which has been deprecated in favour of org.junit which is the "new" framework.

marcosgopen: Actually I was referring to org.junit. But I wrote assertEqual whilst it is [assertEquals](https://junit.org/junit4/javadoc/4.13/org/junit/Assert.html#assertEquals(java.lang.String,%20java.lang.Object,%20java.lang.Object)) with the final 's'. It was already imported at line 8, now 9.

PR #2188, opened 2023-12-13
[JBTM-3823] fix ArrayIndexOutOfBoundsException in QA

marcosgopen:
https://issues.redhat.com/browse/JBTM-3823

Use synchronizedList because it is thread safe.
CORE !TOMCAT !AS_TESTS !RTS !JACOCO! XTS QA_JTA QA_JTS_OPENJDKORB !PERFORMANCE !LRA !DB_TESTS JDK17


mmusgrov:
The change looks good, very nice. We should wait for the QA_JTA profile though since that is the axis effected by this fix. I'm not sure what the issue is with unzip'ing wildfly, but I'll say again that it's doesn't make sense to be unzipping WildFly in the Core profile to run integration tests against a particular server, those tests need to moved into one of the other axes - core is to test the base functionality of arjuna/narayana.

mmusgrov (approved):
Looks good and thanks for raising MWQEINFRA-3960

PR #2199, opened 2024-01-08
BUILD SCRIPT ONLY: Do not support rebase on WildFly upstream or overr…

tomjenkinson:
NO_TESTS


marcosgopen (approved):
Thanks Tom for your PR!

PR #2201, opened 2024-01-12
JBTM-3830 ensure that shutdown won't hang due to missed notification

shawkins:

- [x] Pull Request title is properly formatted: JBTM-XYZ Subject
- [x] Pull Request contains a link to the JIRA issue(s) and that they contain sufficient information for the reviewer to be able to gauge whether or not the proposed changes correctly address the issue

See https://issues.redhat.com/browse/JBTM-3830 - this ensures it takes more than a spurious wakeup for shutdown to continue.

I still need to also confirm that we're not entirely missing the removal in this case as well.

CORE PERFORMANCE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !LRA !DB_TESTS !mysql !db2 !postgres !oracle

By default the pull request will run with `JDK11`, if you prefix a `!` to `JDK11` then it will not run with `JDK11`.

By default the PR will not compile and run with JDK 17. To override this you need to concatenate the two sets of characters: `JDK` `17` as a single word and provide the result into the description.

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java
@@ -715,22 +715,24 @@ public class TransactionReaper
             _inShutdown = true;
 
             /*
-                * If the caller does not want to wait for the normal transaction timeout
-                * periods to elapse before terminating, then we first start by enabling
-                * our time machine!
-                */
-
-            if (!waitForTransactions) {
-                _reaperElements.setAllTimeoutsToZero();
-            }
-
-            /*
                 * Wait for all of the transactions to
                 * terminate normally.
                 */
             while (!_reaperElements.isEmpty()) {
+                /*
+                 * If the caller does not want to wait for the normal transaction timeout
+                 * periods to elapse before terminating, then we first start by enabling
+                 * our time machine!
+                 */
+
+                if (!waitForTransactions) {
+                    _reaperElements.setAllTimeoutsToZero();
+                    nextDynamicCheckTime.set(0);
+                    notifyAll();
+                }
+                
                 try {
-                    this.wait();
+                    this.wait(1000);
                 }
                 catch (final Exception ex) {
                 }
--- a/ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/reaper/ReaperTestCase3.java
+++ b/ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/reaper/ReaperTestCase3.java
@@ -229,7 +229,14 @@ public class ReaperTestCase3  extends ReaperTestCaseControl
         // byteman rules will ensure that the reaper and reaperworker rendezvous gte deleted
         // under this call
 
+        long start = System.currentTimeMillis();
+
         TransactionReaper.terminate(false);
+
+        long duration = System.currentTimeMillis() - start;
+
+        // should complete quickly
+        assertTrue(duration < 500);
 
         assertEquals(0, reaper.numberOfTransactions());
 
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/coordinator/ReaperElement.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/coordinator/ReaperElement.java
@@ -323,4 +323,8 @@ public class ReaperElement implements Comparable<ReaperElement>
     public long getTransactionTimeoutAbsoluteMillis() {
         return _transactionTimeoutAbsoluteMillis;
     }
+
+    void setTransactionTimeoutAbsoluteMillis(long transactionTimeoutAbsoluteMillis) {
+        this._transactionTimeoutAbsoluteMillis = transactionTimeoutAbsoluteMillis;
+    }
 }
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/coordinator/ReaperElementManager.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/coordinator/ReaperElementManager.java
@@ -100,6 +100,7 @@ public class ReaperElementManager
         flushPending();
         for(ReaperElement reaperElement : elementsOrderedByTimeout) {
             reaperElement.setNextCheckAbsoluteMillis(0);
+            reaperElement.setTransactionTimeoutAbsoluteMillis(0);
         }
     }
 
--- a/ArjunaJTS/jtax/pom.xml
+++ b/ArjunaJTS/jtax/pom.xml
@@ -217,6 +217,11 @@
       <build>
         <plugins>
           <plugin>
+            <inherited>false</inherited>
+            <groupId>org.sonatype.plugins</groupId>
+            <artifactId>nexus-staging-maven-plugin</artifactId>
+          </plugin>
+          <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-install-plugin</artifactId>
             <inherited>false</inherited>


shawkins:
Unfortunately this fix does not resolve the situation - running the scenario again I see that when the shutdown is hung the _reaperElements is still populated with an entry:

```
ReaperElement < BasicAction: 0:ffff7f000001:aeb5:65a18923:1a status: ActionStatus.RUNNING, 300, RUN, null >
```

So there's still something outstanding preventng shutdown.

mmusgrov:
> Unfortunately this fix does not resolve the situation - running the scenario again I see that when the shutdown is hung the _reaperElements is still populated with an entry:
> 
> ```
> ReaperElement < BasicAction: 0:ffff7f000001:aeb5:65a18923:1a status: ActionStatus.RUNNING, 300, RUN, null >
> ```
> 
> So there's still something outstanding preventng shutdown.

In that case do you mind if we close this PR, there is insufficient information to make progress with it.

shawkins:
> In that case do you mind if we close this PR, there is insufficient information to make progress with it.

You should consider incorporating one or both of these changes as there is currently a race condition between the shutdown and the removeElementClient methods, even if it does not resolve our issue.

mmusgrov:
I'm not keen on the timed wait but I suppose the other change kind of looks reasonable, the problem being that it incurs the taking of a lock on every pass even though we only need to do the check during shutdown which could be days/months/years apart. I'm also concerned that we don't have a reproducer, Steven can you provide one that we can easily test with and if the code is actually buggy then we will need something that we can unit test with. @jmfinelli , will you review Steven's fix to see if the analysis (without a reproducer nor a test) is valid and whether it is the only solution.

mmusgrov:
@jmfinelli will you also check why `removeElementClient` doesn't appear in the thread dump and why the transaction isn't timing out - @shawkins  this is why we need a reproducer so that we can be sure we aren't guessing at the problem

shawkins:
> I'll do some more tracing to find out what logic is starting a transaction in question - more than likely it should be set to rollback only by some higher level code by this point.

At least for one probablematic run, the left over transaction is being started by https://github.com/keycloak/keycloak/blob/d7f26bf2f1963e89efc6d1e9ea00358fa163e1e3/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/transaction/TransactionalSessionHandler.java#L41

It looks like it should normally be closed by org.keycloak.quarkus.runtime.integration.jaxrs.CloseSessionHandler with a underlying call to https://github.com/keycloak/keycloak/blob/d7f26bf2f1963e89efc6d1e9ea00358fa163e1e3/services/src/main/java/org/keycloak/services/DefaultKeycloakSession.java#L398 - but that doesn't seem to be happening due to the abnormal startup.

So another avenue here downstream is to make sure transactions like these are more assuredly dealt with so they aren't still in a running state.

cc @vmuzikar

mmusgrov:
> > the problem being that it incurs the taking of a lock on every pass even though we only need to do the check during shutdown which could be days/months/years apart
> 
> You'll have spurious wakeups more often than thoses intervals. Is there a wait timeout you would be comfortable with as a tradeoff for being more defensive? It's a minimal check that is being performed before you would enter the wait again - microseconds vs a 1 second wait won't be much load / locking.

I meant that the original code did not take a lock if "_inShutdown" is false but now, with your change, you take the lock every time a transaction is removed from the reaper. But since the reaper may run for days/months/years you have introduced a locking overhead on every transaction - over the lifetime of the reaper that can result in a lot of extra locking which is an overhead.

> 
> > I'm also concerned that we don't have a reproducer, Steven can you provide one that we can easily test with and if the code is actually buggy then we will need something that we can unit test with.
>
> I'm working on it. By tweaking the narayana build it seems like it's becoming clearer what is happening.

Thanks, do you think a simple test/reproducer will be possible?

> 
> > Steven, do we know if the transaction actually times out.
>
> I can now confirm that it does not - in a timely fashion. This may come down to what the behavior of waitForTransactions is supposed to be. The integration logic in quarkus is calling it with a false value:

The JTA spec does not provide realtime guarantees so the definition of "timely" is not part of the spec.

> 
> https://github.com/quarkusio/quarkus/blob/1b7c88e789ebf059969e9354d7435b1882b0aa0f/extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java#L189
> 
> It's documented on TransactionReaper.terminate that waitForTransactions false is equivalent to setting all of the transactions to setRollbackOnly and having them terminate.
> 
> The actual shutdown implementation though is just calling _reaperElements.setAllTimeoutsToZero() a single time for the false case. It is not adjusting the nextDynamicCheckTime nor calling notify on itself - meaning that the reaper thread won't wake up, or even if it does may just go back to sleep.

The reaper thread runs on a for loop waking up periodically and looks at the reaper queue and finds them all eligible for rollback (because of the call to _reaperElements.setAllTimeoutsToZero) and it does so.

> 
> If I correct the logic to sucessfully let the check method run further, the transaction will simply enter the TRACE status, then exit back into RUN - it will not be terminated early as ReaperElement._transactionTimeoutAbsoluteMillis has not been adjusted and will still be greater than _nextCheckAbsoluteMillis (which has been set to set to 0 by setAllTimeoutsToZero).
> 
> For clarification is needed on:
> 
> * I'll do some more tracing to find out what logic is starting a transaction in question - more than likely it should be set to rollback only by some higher level code by this point.
> * how should waitForTransactions behave?

The [javadoc](https://github.com/jbosstm/narayana/blob/6.0.1.Final/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java#L960) says the "reaper will wait until all transactions have terminated (or been terminated by it)" and since the timeout has been set to zero they are all eligible for rollback and the ReaperWorkerThread should issue the actual cancel() calls.

> * the current logic does not guard against simultaneous calls to shutdown and insert. Setting aside the other issues listed above, not synchronzing in insert and just calling setAllTimeoutsToZero a single time leaves open the case where a transaction is inserted after setAllTimeoutsToZero has been called, which throws of the intent of that logic by introducing a new non-zero _nextCheckAbsoluteMillis

I'd need to study the code in detail on that although I don't see it happening given my current understanding of it. But since your observation, if true, does not seem to be a possible cause of your behaviour I will not investigate whether what you say is or is not possible.

shawkins:
> I meant that the original code did not take a lock if "_inShutdown" is false but now, with your change, you take the lock every time a transaction is removed from the reaper.

Sorry I somehow thought you were complaining about the overhead of the 1000 ms wait as it would more often reaquire the lock.

The locked checking of _inShutdown is simply to properly ensuring there is a memory barrier, you could also consider making it volatile - however if you are okay with introducing the timed wait in shutdown, the synchronization change isn't strictly necessary.

> The JTA spec does not provide realtime guarantees so the definition of "timely" is not part of the spec.

To clarify timely - the current behavior is that the transcaion is not canceled and can simply continue, which does not seem to be the intent of waitForTransactions=false 

> The [javadoc](https://github.com/jbosstm/narayana/blob/6.0.1.Final/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java#L960) says the "reaper will wait until all transactions have terminated (or been terminated by it)" and since the timeout has been set to zero they are all eligible for rollback and the ReaperWorkerThread should issue the actual cancel() calls.

It is not canceled proactively.

> I'd need to study the code in detail on that although I don't see it happening given my current understanding of it. But since your observation, if true, does not seem to be a possible cause of your behaviour I will not investigate whether what you say is or is not possible.

It does appear to be true - in fact it's reproducible by the existing test ReaperTestCase3.testReaperWait. That makes it clear that even the original timeout is not honored.

mmusgrov:
Rather than performing static analysis of the code, since it is quite complicated/tricky, can we see the logs. Initially the standard ones (which log at WARN) which you should have (but if those don't indicate why the transaction is still running we'd need trace logs).

shawkins:
> Rather than performing static analysis of the code, since it is quite complicated/tricky

Did you try running ReaperTestCase3.testReaperWait? This behavior is clearly a regression.

> can we see the logs. Initially the standard ones (which log at WARN) which you should have (but if those don't indicate why the transaction is still running we'd need trace logs).

There are no transaction related logs. I'll work upstream to identify exactly why we're not proactively completing the transaction, but the transaction lifecycle is happening as part of the jaxrx / resteasy session filter / handler logic - so it may not be in our direct control there either.

Let me try to clearly summarize the three related threads that have been brought up here:
- ReaperTestCase3.testReaperWait is a clear reproducer for this behavior - I'll update the JIRA with this.
- This pr has identified several code weaknesses, but they do not directly contribute to what we're seeing with JBTM-3830. I'm open to refining this pr, or if you would rather start with a proper fix for JBTM-3830, that's fine too.
- Upstream during an abnormal startup a transaction started for a request session is not getting completed.

mmusgrov:
> > Rather than performing static analysis of the code, since it is quite complicated/tricky
> 
> Did you try running ReaperTestCase3.testReaperWait? This behavior is clearly a regression.

No I didn't, that test has always passed, and is still passing.

> 
> > can we see the logs. Initially the standard ones (which log at WARN) which you should have (but if those don't indicate why the transaction is still running we'd need trace logs).
> 
> There are no transaction related logs. I'll work upstream to identify exactly why we're not proactively completing the transaction, but the transaction lifecycle is happening as part of the jaxrx / resteasy session filter / handler logic - so it may not be in our direct control there either.

In that case you may need to enable logging.

> 
> Let me try to clearly summarize the three related threads that have been brought up here:
> 
> * ReaperTestCase3.testReaperWait is a clear reproducer for this behavior - I'll update the JIRA with this.

This is news to me, you will need to be more explicit about your thinking processes.

> * This pr has identified several code weaknesses, but they do not directly contribute to what we're seeing with JBTM-3830. I'm open to refining this pr, or if you would rather start with a proper fix for JBTM-3830, that's fine too.

I disagree, you have performed a static visual code analysis and come to the conclusion that you think the code is suspicious.

> * Upstream during an abnormal startup a transaction started for a request session is not getting completed.

Not sure what this means without definitions of abnormal or what a request session is.

shawkins:
> No I didn't, that test has always passed, and is still passing.

Can you clarify how you know that it is passing if you haven't run it?

> This is news to me, you will need to be more explicit about your thinking processes.

I have been explicit about how the code is currently not functioning as one would expect. I have also been explicit that if you run that test case, it will hang. Note that if you simply do "mvn clean install" that test will not be run - that is the case with all tests annotated with `@RunWith(BMUnitRunner.class)`. If you know of some job or something that should be exercising those tests, please double check it.

> I disagree, you have performed a static visual code analysis and come to the conclusion that you think the code is suspicious.

Again it is of course up to you what you do with this pr, but to deny that you have code here that is not defensive or in places incorrect is not helpful.

> Not sure what this means without definitions of abnormal or what a request session is.

Let's decouple the thread of what is happening in keycloak from this entirely at this point and focus instead on the regression with waitForTransactions.

shawkins:
My appologies - I meant to reference testReaperForce, not testReaperWait. Force is the one passing false. I corrected the link on the JIRA as well.

mmusgrov:
> My appologies - I meant to reference testReaperForce, not testReaperWait. Force is the one passing false. I corrected the link on the JIRA as well.

I see, that was significant. I will try it with that test instead then ...

mmusgrov:
> > No I didn't, that test has always passed, and is still passing.
> 
> Can you clarify how you know that it is passing if you haven't run it?

I ran it successfully on our main branch. But I didn't want to get into a long discussion about running it with byteman on the 6.x branch, how did you run the test?

> 
> > This is news to me, you will need to be more explicit about your thinking processes.
> 
> I have been explicit about how the code is currently not functioning as one would expect. I have also been explicit that if you run that test case, it will hang. Note that if you simply do "mvn clean install" that test will not be run - that is the case with all tests annotated with `@RunWith(BMUnitRunner.class)`. If you know of some job or something that should be exercising those tests, please double check it.
>
> > I disagree, you have performed a static visual code analysis and come to the conclusion that you think the code is suspicious.
> 
> Again it is of course up to you what you do with this pr, but to deny that you have code here that is not defensive or in places incorrect is not helpful.

Again, you have performed a visual analysis and decided that the code doesn't look correct.

> 
> > Not sure what this means without definitions of abnormal or what a request session is. 
> 
> Let's decouple the thread of what is happening in keycloak from this entirely at this point and focus instead on the regression with waitForTransactions.

shawkins:
> I ran it successfully on our main branch. But I didn't want to get into a long discussion about running it with byteman on the 6.x branch, how did you run the test?

I ran it from the the IDE. At least my JRE requires adding the argument -Djdk.attach.allowAttachSelf=true. 

Just to make things clearer I've seen two different hanging behaviors. The first in which the test still "passes" is that the transactions will timeout by the original timeout - which is not expected. If you change the timeout of one of the transactions to something much longer, this should be very clear.

Other runs I have produced even longer hangs than the timeouts, but that does not seem to be as reproducible. 

So at best there is no difference between the wait and the force behavior. In keycloak I think the default timeout for transaction created by the session logic was three minutes, so this is very noticable.

mmusgrov:
Ok that seems like a reproducer that I can work with, give me, or @jmfinelli , a couple of days and we'll see if we can reproduce the behaviour. Is this only on the 6.x branch (I ran all the tests in ReaperTestCase3 a couple of times on the main branch with no failures)?

shawkins:
> Is this only on the 6.x branch (I ran all the tests in ReaperTestCase3 a couple of times on the main branch with no failures)?

No this is happening on main. Also the test is written current in such a way as to not check elapsed time, so as long as the transactions eventually timeout, the test will still pass. I've updated the test in this pr to make the bad behavior clearer.

On the less reproducible behavior I have seen both what looks like a test harness related situation where the byteman stuff is waiting to do a rendezvous, and also a state where the timeouts are not eventually enforced. The latter could be relevant, but I probably wont' spend much time on it until this initial issue is resolved.

shawkins:
@mmusgrov since I haven't heard anything on JBTM-3830, and no solution has yet been identified on https://github.com/quarkusio/quarkus/issues/38164 I've updated this pr to include possible changes for this.

As far as I can tell we need to update both the wakeup and the timeout on the remaining elements when waitForTransactions is false. We also don't want to only set the timeouts to 0 a single time as this opens up a timing issue that an insert happening after / simultaneously with shutdown will cause that timeout to be honored. And finally for safety reasons discussed above, the main wait loop of the shutdown method should be timed as to not risk missed notifications - in some rare circumstances we are still seeing what looks like indefinite shutdown hangs well past any timeouts on the transactions.

jmfinelli:
Sure, no problem. I will review Steven's PR ASAP :-)

mmusgrov (approved):
I am happy with this fix but let's also wait for Manuel to provide his feedback before merging.

shawkins:
> @shawkins Is that an accurate summary of the content of your PR?

Yes, let me add some more thoughts.

> the shutdown method forcibly sets all transaction timeouts to zero (via the new method setTransactionTimeoutAbsoluteMillis)

@jmfinelli mentioned this may need to be phase dependent, that we can only force the timeout if not already prepared.

> then loops waiting at 1 sec intervals for those transactions to terminate. The periodic one second wait during shutdown is not a regression since normally the code would be waiting for longer than that for transactions to naturally time out.

There are a couple of aspects to this. Setting the nextDynamicCheckTime to 0 and doing a notifyAll will ensure that the reaper thread wakes up immediately and doesn't just go back to sleep. Without those changes, the reaper will wait until the transaction with the most immediate original timeout reaches it's original timeout before cleaning things up. 

The other part about doing this in the loop with a timed wait is a safety mechanism to ensure that we're forcing cleanup even if there's been some other concurrency issue - remove didn't notify the transaction reaper, or another transaction was inserted concurrently to or after shutdown. 

If you want to tweak this timed wait behavior, that's fine with me too. It could use a longer / indefinite wait when waitForTransactions=true and/or use an exponential wait.

mmusgrov:
> > @jmfinelli mentioned this may need to be phase dependent, that we can only force the timeout if not already prepared.
> 
> We don't have to do any special handling here since, according to the JTA state model, it isn't possible to cancel a transaction that isn't already running (ie Status.STATUS_ACTIVE or Status.STATUS_MARKED_ROLLBACK), i.e. the existing code will work no matter what state the transaction is in.

Unless Manuel is thinking about the shutdown returning while a transaction is still in its' prepare phase. Although messy, that case would be handled by recovery which would find no transaction log so it would instruct any prepared participants to rollback (this is the presumed abort optimisation).

mmusgrov:
@shawkins do you know if your PR solves the original quarkus issue https://github.com/quarkusio/quarkus/issues/38164?

shawkins:
> @shawkins do you know if your PR solves the original quarkus issue [quarkusio/quarkus#38164](https://github.com/quarkusio/quarkus/issues/38164)?

It solves the behavior of hanging on the main thread, which is good enough for us. It of course does not address the keycloak / quarkus issue with jaxrs that we end up with this leftover transactions in the first place.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java
```diff
@@ -714,23 +714,27 @@ private final void shutdown(boolean waitForTransactions)
...
                 */
             while (!_reaperElements.isEmpty()) {
+                /*
+                 * If the caller does not want to wait for the normal transaction timeout
+                 * periods to elapse before terminating, then we first start by enabling
+                 * our time machine!
+                 */
+
+                if (!waitForTransactions) {
+                    _reaperElements.setAllTimeoutsToZero();
```
tomjenkinson, line 729: Please can I ask why this block was moved into the `!_reaperElements.isEmpty()` condition?

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java
```diff
@@ -714,23 +714,27 @@ private final void shutdown(boolean waitForTransactions)
...
+                    if (_dynamic) {
+                        nextDynamicCheckTime.set(0);
```
tomjenkinson, line 731: Please can I ask the intention by only setting this when _dynamic is true? For example it is [used](https://github.com/jbosstm/narayana/blob/main/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java#L97) in the check method outside even when the reaper is not dynamic?

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/reaper/ReaperTestCase3.java
```diff
@@ -229,8 +229,15 @@ public void testReaperForce() throws Exception
         // byteman rules will ensure that the reaper and reaperworker rendezvous gte deleted
         // under this call
 
+        long start = System.currentTimeMillis();
```
tomjenkinson, line 232: Given the comment `force a termination and don't wait for the normal timeout periods` I personally would be fine with your additional check of how long the timeout takes and making sure it's shorter than the timeout in this test but please can you clarify why you needed to modify the timeouts of the inserts above? Hopefully reapable3 being 4 seconds would generally be long enough on it's own to be able to show that the reaper was terminated in a time less than the timeout of a Reapable? We might consider allowing just over a second for the theoretical scenario where the wait(1000) takes the full length of time.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java
shawkins: @tomjenkinson my reading of https://github.com/jbosstm/narayana/blob/64b02cc09db97498219bef1269cf98eed4eba309/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java#L571 and the checkPeriod method made the intention of nextDynamicCheckTime seem to only apply in the dynamic case. 

But as you say there is inconsistent usage of it in check/reinsertElement, and removeElementReaper. So it's probably best to omit the _dynamic check here as well.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java
shawkins: Because there are no guarntees on the handling of reaper elements added simultanously or after shutdown is called.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/reaper/ReaperTestCase3.java
shawkins: > but please can you clarify why you needed to modify the timeouts of the inserts above?

Because that change was made initially to highlight the issue.

As far as what the could or should be, I don't have too strong of an opinion. There are two things that need accounted for - that reaping begins immediately after shutdown and does not wait for the first timeout (this is the effect of setting  nextDynamicCheckTime to 0 and notifying all), and that you do not wait any of the remaining transactions to fully timeout.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java
tomjenkinson: Well there is a comment https://github.com/jbosstm/narayana/blob/main/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java#L929-L931 which is in the the public documentation of the API so hopefully client applications would not be adding things in the situations you mentioned (I want to thank @jmfinelli who mentioned something in this area)?

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/reaper/ReaperTestCase3.java
tomjenkinson: We should have the test removed from the exclusion in https://github.com/jbosstm/narayana/blob/main/ArjunaCore/arjuna/pom.xml#L61 (thanks @jmfinelli for the guidance) so builds and CI runs it /cc @mmusgrov

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java
shawkins: I'll leave it up to you guys if you want to forego the added safety of moving this into the loop. If it gets moved back out, I'll re-examine the quakus logic - at first glance it didn't seem to ensure that a transaction could not be started concurrently with shutdown due to the async nature of the events in play.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java
shawkins: Changed to remove the _dynamic check. A follow-up issue should be considered on making its usage more consistent.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/reaper/ReaperTestCase3.java
shawkins: Changed the timeouts back.

tomjenkinson: Thank you!

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java
tomjenkinson: I think Quarkus should be made to respect this requirement if there is indeed a problem with that. From my reading of the API we should not need to change the location of this code. I will raise a pull request on your branch @shawkins. With that moved back, personally I would be ready to approve the pull request (assuming it passes CI).

tomjenkinson: I raised https://github.com/shawkins/narayana/pull/1/ to try to get this change into your JBTM-3830 branch pre-completing the review /cc @mmusgrov @jmfinelli

tomjenkinson (approved):

mmusgrov:
I received a private message from Manuel saying that he is okay with this PR so I will merge it now.

PR #2209, opened 2024-02-19
fix: periodic shutdown hangs

shawkins:
Follow on to https://github.com/jbosstm/narayana/pull/2201#discussion_r1491384185 - the inconsistent usage of the dynamic next time will cause the reaper to not shutdown in periodic mode. 

I'm not sure if this needs another JIRA, or if you want to have an additional non-byteman testcase for this. Like what @jmfinelli was working on, it would be nice to just have a test matrix of shutdown combinations.

- [ ] Pull Request contains a link to the JIRA issue(s) and that they contain sufficient information for the reviewer to be able to gauge whether or not the proposed changes correctly address the issue

By default the pull request will run with `JDK11`, if you prefix a `!` to `JDK11` then it will not run with `JDK11`.

By default the PR will not compile and run with JDK 17. To override this you need to concatenate the two sets of characters: `JDK` `17` as a single word and provide the result into the description.

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TransactionReaper.java
@@ -568,7 +568,7 @@ public class TransactionReaper
             throw new IllegalStateException(tsLogger.i18NLogger.get_coordinator_TransactionReaper_1());
         }
 
-        if (_dynamic && reaperElement.getNextCheckAbsoluteMillis() < nextDynamicCheckTime.get()) {
+        if (reaperElement.getNextCheckAbsoluteMillis() < nextDynamicCheckTime.get()) {
             updateCheckTimeForEarlierInsert(reaperElement.getNextCheckAbsoluteMillis());
         }
     }
--- a/ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/reaper/ReaperTestCase4.java
+++ b/ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/reaper/ReaperTestCase4.java
@@ -1,0 +1,87 @@
+/*
+   Copyright The Narayana Authors
+   SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.hp.mwtests.ts.arjuna.reaper;
+
+import org.junit.Test;
+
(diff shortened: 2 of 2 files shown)


mmusgrov:
@shawkins we need a reproducer for this, does the test case in your PR address?
Why wasn't this case addressed in your previous PR, is it a new issue you have just found?

shawkins:
> is the test case in your PR sufficient?

Yes, back out the change to TransactionReaper and you'll see an indefinite hang.

> Why wasn't this case addressed in your previous PR

We aren't using periodic mode, and my interest was in getting the other pr merged rather than subjecting it to further review.

> is it a new issue you have just found and if so we need a new JBTM?

That's up to you whether you want to treat this as a new distinct issue.

> Did we merge the other one too early?

No, this is a related issue for periodic mode noticed in working the other issue. The next dynamic time is being used in places that aren't checking _dynamic. So the simplest fix is to just remove the _dynamic check from there as well.

mmusgrov:
We need a JBTM for it with the details so that future engineers can determine what the issue is. TBH I would have preferred a complete fix to the pr/2201 (shutdown hang) rather than a staged one which is an unusual way of fixing thing.

shawkins:
> TBH I would have preferred a complete fix to the pr/2201 (shutdown hang) rather than a staged one which is an unusual way of fixing thing.

Sure these are highly related cases. Just to clarify further the two cases are:
- dynamic forced shutdown - that affects quarkus / keycloak. I didn't test it, but periodic forced may have had an issue as well.
- periodic wait shutdown - this doesn't affect us, but it seems good to address.

> We need a JBTM for it with the details so that future engineers can determine what the issue is. 

Sure, I'll open something.

mmusgrov:
> > TBH I would have preferred a complete fix to the pr/2201 (shutdown hang) rather than a staged one which is an unusual way of fixing thing.
> 
> Sure these are highly related cases. Just to clarify further the two cases are:
> 
> * dynamic forced shutdown - that affects quarkus / keycloak. I didn't test it, but periodic forced may have had an issue as well.
> * periodic wait shutdown - this doesn't affect us, but it seems good to address.
> 
> > We need a JBTM for it with the details so that future engineers can determine what the issue is.
> 
> Sure, I'll open something.

Or you can add a note to the original JBTM.
I will approve this one now since it has already passed the CORE profile CI tests.

mmusgrov (approved):

PR #2214, opened 2024-03-05
[JBTM-3843] Rollback is no longer invoked in all cases

jmfinelli:
[JBTM-3843](https://issues.redhat.com/browse/JBTM-3843)

Full pass:
!CORE !RTS !QA_JTA !QA_JTS_OPENJDKORB !DB_TESTS !XTS !LRA  !AS_TESTS 

No needed:
!JACOCO !PERFORMANCE

### DIFF
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
@@ -212,13 +212,11 @@ public class XAResourceRecord extends AbstractRecord implements ExceptionDeferre
 			 * XAER_PROTO.
 			 */
 
-			if (_rollbackOptimization) // won't have rollback called on it
-				removeConnection();
-
 			switch (e1.errorCode)
 			{
-			case XAException.XAER_RMERR:
-			case XAException.XAER_RMFAIL:
+			// No txn -> assume rollback
+			case XAException.XAER_NOTA:
+			// No txn -> assume rollback
 			case XAException.XA_RBROLLBACK:
 			case XAException.XA_RBEND:
 			case XAException.XA_RBCOMMFAIL:
@@ -227,10 +225,15 @@ public class XAResourceRecord extends AbstractRecord implements ExceptionDeferre
 			case XAException.XA_RBOTHER:
 			case XAException.XA_RBPROTO:
 			case XAException.XA_RBTIMEOUT:
+				// This will avoid calling rollback when aborting
+				_rolledBack = true;
+				return TwoPhaseOutcome.PREPARE_NOTOK;
+			// We might need to roll back the XA resource for these error codes
 			case XAException.XAER_INVAL:
 			case XAException.XAER_PROTO:
-			case XAException.XAER_NOTA: // resource may have arbitrarily rolled back (shouldn't, but ...)
-				return TwoPhaseOutcome.PREPARE_NOTOK;  // will not call rollback
+			case XAException.XAER_RMERR:
+			case XAException.XAER_RMFAIL:
+				return TwoPhaseOutcome.PREPARE_NOTOK;
 			default:
 				return TwoPhaseOutcome.HEURISTIC_HAZARD; // we're not really sure (shouldn't get here though).
 			}
@@ -239,9 +242,6 @@ public class XAResourceRecord extends AbstractRecord implements ExceptionDeferre
 		{
             jtaLogger.i18NLogger.warn_resources_arjunacore_preparefailed(XAHelper.xidToString(_tranID),
                     _theXAResource.toString(), "-", e2);
-
-			if (_rollbackOptimization) // won't have rollback called on it
-				removeConnection();
 
 			return TwoPhaseOutcome.PREPARE_NOTOK;
 		}
@@ -297,13 +297,17 @@ public class XAResourceRecord extends AbstractRecord implements ExceptionDeferre
 						&& (e1.errorCode < XAException.XA_RBEND))
 				    {
 					/*
-					 * Has been marked as rollback-only. We still
-					 * need to call rollback.
+					 * Has been marked as rollback-only. If the branch has not been
+					 * already rolled back (i.e. _rolledBack == false), we still need
+					 * to call rollback.
 					 */
 					 
 				    } else if ((e1.errorCode == XAException.XAER_RMERR) || (e1.errorCode == XAException.XAER_RMFAIL)){
 				    	    try {
-				    	    	    _theXAResource.rollback(_tranID);
+				    	      if (!this._rolledBack) {
+				    	        _theXAResource.rollback(_tranID);
+				    	        this._rolledBack = true;
+				    	      }
 				    	    } catch (XAException e2)
 				    	    {
 				    	       addDeferredThrowable(e2);
@@ -335,7 +339,10 @@ public class XAResourceRecord extends AbstractRecord implements ExceptionDeferre
 
 				try
 				{
-					_theXAResource.rollback(_tranID);
+				  if (!this._rolledBack) {
+				    _theXAResource.rollback(_tranID);
+				    this._rolledBack = true;
+				  }
 				}
 				catch (XAException e1)
 				{
@@ -376,6 +383,11 @@ public class XAResourceRecord extends AbstractRecord implements ExceptionDeferre
 						case XAException.XA_RBOTHER:
 						case XAException.XA_RBPROTO:
 						case XAException.XA_RBTIMEOUT:
+							/** The resource manager has rolled back the
+							 * transaction branch’s work and has released
+							 * all held resources. Nothing left to do
+							 */
+							this._rolledBack = true;
 							break;
 						default:
 							return TwoPhaseOutcome.FINISH_ERROR;
@@ -487,6 +499,7 @@ public class XAResourceRecord extends AbstractRecord implements ExceptionDeferre
 						case XAException.XA_RBTRANSIENT:
 						case XAException.XAER_RMERR:
 	                                        case XAException.XAER_PROTO:  // XA spec implies rollback
+							_rolledBack = true;
 							_heuristic = TwoPhaseOutcome.HEURISTIC_ROLLBACK;
 							return TwoPhaseOutcome.HEURISTIC_ROLLBACK;
 						case XAException.XA_HEURMIX:
@@ -584,6 +597,7 @@ public class XAResourceRecord extends AbstractRecord implements ExceptionDeferre
 	    {
             jtaLogger.i18NLogger.warn_resources_arjunacore_opcnulltx("XAResourceRecord.1pc");
 
+            _rolledBack = true;
 	        return TwoPhaseOutcome.ONE_PHASE_ERROR;  // rolled back!!
 	    }
 	    else
@@ -678,6 +692,9 @@ public class XAResourceRecord extends AbstractRecord implements ExceptionDeferre
 	                    _theXAResource.commit(_tranID, true);
 	                else {
 	                    _theXAResource.rollback(_tranID);
+	                    // _rolledBack set to true even though
+	                    // it's not needed at this stage
+	                    this._rolledBack = true;
 	                    throw endRBOnly;
 	                }
 	            }
@@ -704,6 +721,7 @@ public class XAResourceRecord extends AbstractRecord implements ExceptionDeferre
 	                    forget();
 	                    break;
 	                case XAException.XA_HEURRB:
+	                       _rolledBack = true;
 	                       forget();
 	                       return TwoPhaseOutcome.ONE_PHASE_ERROR;
 	                case XAException.XA_RBROLLBACK:
@@ -715,6 +733,11 @@ public class XAResourceRecord extends AbstractRecord implements ExceptionDeferre
 	                case XAException.XA_RBTIMEOUT:
 	                case XAException.XA_RBTRANSIENT:
 	                case XAException.XAER_RMERR:
+	                    /** These codes imply that the RM rolled back the branch
+	                     * Setting _rolledBack to true is not strictly needed here,
+	                     * but we'll set it anyway for consistency
+	                     */
+	                    this._rolledBack = true;
 	                    return TwoPhaseOutcome.ONE_PHASE_ERROR;
 	                case XAException.XAER_NOTA:
 	                    _heuristic = TwoPhaseOutcome.HEURISTIC_HAZARD;
@@ -728,6 +751,7 @@ public class XAResourceRecord extends AbstractRecord implements ExceptionDeferre
 	                    return TwoPhaseOutcome.HEURISTIC_HAZARD;
 	                case XAException.XA_RETRY:  // XA does not allow this to be thrown for 1PC!
 	                case XAException.XAER_PROTO:
+	                    this._rolledBack = true;
 	                    return TwoPhaseOutcome.ONE_PHASE_ERROR; // assume rollback
 	                case XAException.XAER_RMFAIL: // This was modified as part of JBTM-XYZ - although RMFAIL is not clear there is a rollback/commit we are flagging this to the user
 	                    _heuristic = TwoPhaseOutcome.HEURISTIC_HAZARD;
@@ -750,7 +774,10 @@ public class XAResourceRecord extends AbstractRecord implements ExceptionDeferre
 	            }
 	        }
 	        else
-	            return TwoPhaseOutcome.ONE_PHASE_ERROR;
+	        {
+	          this._rolledBack = true;
+	          return TwoPhaseOutcome.ONE_PHASE_ERROR;
+	        }
 	    }
 
 	    if (commit)
@@ -1308,6 +1335,7 @@ public class XAResourceRecord extends AbstractRecord implements ExceptionDeferre
 	
 	private TransactionImple _theTransaction;
     private boolean _recovered = false;
+    private boolean _rolledBack = false;
 
     // extra metadata from the wrapper, if present
     private String _productName;
@@ -1315,8 +1343,6 @@ public class XAResourceRecord extends AbstractRecord implements ExceptionDeferre
     private String _jndiName;
     private static final XAResourceRecordWrappingPlugin _xaResourceRecordWrappingPlugin =
             jtaPropertyManager.getJTAEnvironmentBean().getXAResourceRecordWrappingPlugin();
-
-	private static final boolean _rollbackOptimization = jtaPropertyManager.getJTAEnvironmentBean().isXaRollbackOptimization();
 
     /*
      * WARNING: USE WITH EXTEREME CARE!!
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
@@ -449,10 +449,11 @@ public class XAResourceRecord extends AbstractRecord implements ExceptionDeferre
                             case XAException.XA_RBPROTO:
                             case XAException.XA_RBTIMEOUT:
                             case XAException.XA_RBTRANSIENT:
-                            case XAException.XAER_RMERR:
                             case XAException.XAER_PROTO:
-                                // could really do with an ABORTED status in TwoPhaseOutcome to differentiate
                                 // XA spec implies rollback
+                            case XAException.XAER_RMERR:
+                                // could really do with an ABORTED status in TwoPhaseOutcome to differentiate
+
                                 _rolledBack = true;
                                 _heuristic = TwoPhaseOutcome.HEURISTIC_ROLLBACK;
                                 return TwoPhaseOutcome.HEURISTIC_ROLLBACK;
@@ -615,8 +616,9 @@ public class XAResourceRecord extends AbstractRecord implements ExceptionDeferre
                      * Not strictly necessary since calling commit will
                      * do the rollback if end failed as above.
                      */
-                    if (endHeuristic != null)
+                    if (endHeuristic != null) {
                         throw endHeuristic;
+                    }
 
                     if (commit) {
                         _theXAResource.commit(_tranID, true);
--- a/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
+++ b/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
@@ -3,8 +3,6 @@
    SPDX-License-Identifier: Apache-2.0
  */
 
-
-
 package com.arjuna.ats.internal.jta.resources.arjunacore;
 
 import com.arjuna.ats.arjuna.ObjectType;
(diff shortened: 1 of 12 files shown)


marcosgopen:
Thank you Manuel for opening this PR! Let me know when this is ready for review.

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -319,14 +321,19 @@ public void rollback() throws org.omg.CORBA.SystemException,
...
 				    } else if ((e1.errorCode == XAException.XAER_RMERR) || (e1.errorCode == XAException.XAER_RMFAIL)){
 				    	    try {
-				    	    	    _theXAResource.rollback(_tranID);
+							if (!this._rolledBack)
+							{
+				    	    	_theXAResource.rollback(_tranID);
+								this._rolledBack = true;
+							}
 				    	    } catch (XAException e2)
-				    	    {
```
tomjenkinson, line 329: @jmfinelli a bit of code formatting is here too, please can I ask you to move it out of this specific commit?

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
```diff
@@ -586,30 +585,6 @@ public void setXaResourceOrphanFilters(List<XAResourceOrphanFilter> xaResourceOr
...
-    /**
-     * Returns if connections associated to XAResources that fail during prepare should be cleaned up immediately.
-     * TODO move to JDBC module as it's only for our own connection manager?
-     *
-     * Default: false.
-     * Equivalent deprecated property: com.arjuna.ats.jta.xaRollbackOptimization
-     *
-     * @return true for cleanup during prepare, false for cleanup during phase two rollback.
-     */
-    public boolean isXaRollbackOptimization()
```
tomjenkinson, line 598: Public API change implies Narayana 8 should be discussed. Maybe we want to consider if some of this should be in a separate commit?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
```diff
@@ -678,6 +694,9 @@ public int topLevelOnePhaseCommit()
 	                    _theXAResource.commit(_tranID, true);
 	                else {
 	                    _theXAResource.rollback(_tranID);
+						// _rolledBack set to true even though
+						// it's not needed at this stage
+						this._rolledBack = true;
```
tomjenkinson: Why would you do this? Would it not be set on line 309 or 346 above? Or is there another path?

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -746,6 +761,9 @@ public void commit_one_phase() throws HeuristicHazard,
 	                    _theXAResource.commit(_tranID, true);
 	                else {
 	                    _theXAResource.rollback(_tranID);
+						// _rolledBack set to true even though
+						// it's not needed at this stage
+						this._rolledBack = true;
```
tomjenkinson: Is it not set by 333/358? Is there another path?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
```diff
@@ -1308,6 +1337,7 @@ public boolean isForgotten() {
 	
 	private TransactionImple _theTransaction;
     private boolean _recovered = false;
+	private boolean _rolledBack = false;
```
tomjenkinson: I am not sure if it's important or not but I do not think you are making this part of the persistent state of the object so after recovery I suppose an extra rollback could be called?

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -1469,14 +1498,15 @@ private List<SerializableXAResourceDeserializer> getXAResourceDeserializers() {
...
-	
+
 	protected XAResource _theXAResource;
 
 	private RecoverableXAConnection _recoveryObject;
 	private Xid _tranID;
 	private boolean _prepared;
 	private boolean _committed;
 	private boolean _valid;
+	private boolean _rolledBack;
```
tomjenkinson: I am not sure if it's important or not but I do not think you are making this part of the persistent state of the object so after recovery I suppose an extra rollback could be called?

ArjunaJTS/integration/src/test/java/com/arjuna/ats/jta/distributed/SimpleIsolatedServers.java
```diff
@@ -1065,7 +1065,8 @@ public void testMigrateTransactionSubordinateTimeout() throws Exception {
 			Thread.currentThread().setContextClassLoader(classLoader);
 		}
 		assertTrue("" + completionCounter.getRollbackCount("2000"), completionCounter.getRollbackCount("2000") == 1);
-		assertTrue("" + completionCounter.getRollbackCount("1000"), completionCounter.getRollbackCount("1000") == 2);
+		// The second rollback invocation (from the proxyXAResource) is ignored because of JBTM-XXXX, that's why the following condition is == 1
```
tomjenkinson: @jmfinelli please can you add the full JBTM number here?

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
tomjenkinson: BTW the Javadoc gives a clue actually as to why this was here it seems something for transactional driver perhaps

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
```diff
@@ -318,6 +322,7 @@ public int topLevelAbort()
 				    }
 				    else
 				    {
+						// TODO: Is this in the right catch block?
```
tomjenkinson: @jmfinelli please can you elaborate on the TODO? Does it need resolving?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
```diff
@@ -328,14 +333,18 @@ public int topLevelAbort()
 				}
                 catch(RuntimeException e)
                 {
+					// TODO: Is this in the right catch block?
```
tomjenkinson: @jmfinelli please can you elaborate on the TODO? Does it need resolving?

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/common/FailureXAResource.java
```diff
@@ -9,146 +9,166 @@
...
     }
 
-    public void rollback(Xid xid) throws XAException
-    {
-        if ((_locale == FailLocation.rollback) || (_locale == FailLocation.prepare_and_rollback))
-        {
-            if (_type == FailType.normal)
-                throw new XAException(XAException.XA_HEURMIX);
-            
-            if (_type == FailType.heurcom)
```
tomjenkinson, line 121: What prompted the change from the `if (_type...` to a `switch` style of comparison?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
```diff
@@ -227,10 +225,15 @@ public int topLevelPrepare()
...
+				_rolledBack = true;
+				return TwoPhaseOutcome.PREPARE_NOTOK;
+			// We might need to roll back the XA resource for these error codes
 			case XAException.XAER_INVAL:
 			case XAException.XAER_PROTO:
-			case XAException.XAER_NOTA: // resource may have arbitrarily rolled back (shouldn't, but ...)
-				return TwoPhaseOutcome.PREPARE_NOTOK;  // will not call rollback
+			case XAException.XAER_RMERR:
+			case XAException.XAER_RMFAIL:
+				return TwoPhaseOutcome.PREPARE_NOTOK;
```
tomjenkinson: Maybe we need a comment something like:
// We do not know the outcome of the resource manager conclusively and so we will try to call rollback again

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/common/FailureXAResource.java
```diff
@@ -9,146 +9,166 @@
...
-            
-            if (_type == FailType.timeout)
-                throw new XAException(XAException.XA_RBTIMEOUT);
-            
-            if (_type == FailType.XA_RBCOMMFAIL)
-                throw new XAException(XAException.XA_RBCOMMFAIL);
+
+    public void commit(Xid id, boolean onePhase) throws XAException {
+        if (_locale == FailLocation.commit) {
+            switch (_type) {
```
tomjenkinson, line 52: What prompted the change from the if (_type... to a switch style of comparison?

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -251,9 +251,14 @@ public org.omg.CosTransactions.Vote prepare() throws HeuristicMixed,
...
+				// This will avoid calling rollback when aborting
+				_rolledBack = true;
+				return Vote.VoteRollback;
+				// We might need to roll back the XA resource for these error codes
 			case XAException.XAER_INVAL:
 			case XAException.XAER_PROTO:
-			case XAException.XAER_NOTA: // resource may have arbitrarily rolled back (shouldn't, but ...)
+			case XAException.XAER_RMERR:
+			case XAException.XAER_RMFAIL:
 				return Vote.VoteRollback;
```
tomjenkinson: Maybe we need a comment something like:
// We do not know the outcome of the resource manager conclusively and so we will try to call rollback again

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/common/FailureXAResource.java
jmfinelli: I've moved this change into the STYLE commit. Thanks Tom

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/common/FailureXAResource.java
```diff
@@ -9,146 +9,166 @@
 import javax.transaction.xa.XAResource;
 import javax.transaction.xa.Xid;
 
-public class FailureXAResource implements XAResource
```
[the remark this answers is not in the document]
jmfinelli: I've moved this change into the STYLE commit. Thanks Tom

ArjunaJTS/integration/src/test/java/com/arjuna/ats/jta/distributed/SimpleIsolatedServers.java
jmfinelli: Done. Thanks

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -835,8 +860,11 @@ public void commit_one_phase() throws HeuristicHazard,
 	            }
 	        }
 	        else
+			{
+				this._rolledBack = true;
 	            throw new TRANSACTION_ROLLEDBACK();
-	    }
+	        }
+		}
```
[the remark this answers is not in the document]
jmfinelli: Hopefully, the new commits resolve this comment.

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
jmfinelli: Solved

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
jmfinelli: This is not the best of the explanation, isn't it :sweat_smile: ?

What I wanted to discuss here is:
```
Why is ArjunaJTA returning the message with ID 16045
(i.e. "attempted rollback of {0} ({1}) failed with exception code {2}")
when we're not catching error codes from a rollback attempt?
```
Here, we're catching exceptions generated by an `end` attempt. Although we're passing TMFAIL as parameter, AFAIU, the XA specs says that:
```
A resource manager might choose to mark a transaction branch
as rollback-only at this point. In fact, a transaction manager does so
for the global transaction. If a resource manager chooses to do so also,
xa_end() returns one of the [XA_RB∗] values.
```
i.e. ArjunaJTA should call rollback on that RM itself when the previous `end` invocation fails. Thus, It seems to me that the two messages with ID 160450 shouldn't be used here (i.e. line 326 and 337) Please, could you help me understand this part of the code? @tomjenkinson

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
jmfinelli: I'm not sure I understand this comment. Why would topLevelOnePhaseCommit depend on topLevelAbort?

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
jmfinelli: > Public API change implies Narayana 8 should be discussed. Maybe we want to consider if some of this should be in a separate commit?

I see your point. On the other hand, `isXaRollbackOptimization` hasn't been working for a long while and nobody complained...so it is likely that nobody is using this configuration. @mmusgrov do you think we can include this change in a minor or (as Tom suggested) should we create a new JBTM for removing it and wait the next major?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
jmfinelli: ^^^^

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
jmfinelli: I'm not sure I understand this comment. Why would commit_one_phase depend on rollback?

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
tomjenkinson: If the functionality is proven not to be working, could we consider to remove the "implementation" and log errors if someone tries to use the public API parts (setter and getter)? I am wondering now, does this pull request somehow prove the feature is not working?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
tomjenkinson: I will review!

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
tomjenkinson: Sorry, assuming that was there before I missed that. I will resolve this.

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
tomjenkinson: Sorry, assuming that was there before I missed that. I will resolve this.

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
tomjenkinson: Sorry - I misunderstood. I will resolve this!

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
tomjenkinson: I will take a look into this but I think it will be tomorrow at least now - thanks for asking

ArjunaJTS/integration/src/test/java/com/arjuna/ats/jta/distributed/SimpleIsolatedServers.java
tomjenkinson: I think it's possible that this is still saying JBTM-XXXX (at least from viewing https://github.com/jbosstm/narayana/pull/2214/files#diff-f273d7bba38eaa0e38556825398072748c28164b6f6bb0a255792c70d7346745R1069)

jmfinelli: Duh

jmfinelli: Checking the code, it seems that JBTM-XXXX has been changed to JBTM-3843. Could you please let me know if you still see `JBTM-XXXX`? Thanks

tomjenkinson: I don't think so  - please take a look at https://github.com/jmfinelli/narayana/blob/JBTM-3843/ArjunaJTS/integration/src/test/java/com/arjuna/ats/jta/distributed/SimpleIsolatedServers.java#L1069 for example?

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
jmfinelli: Nope, I didn't develop any test for that configuration property as I planned to remove it. Nevertheless, it should be possible to test it using the two tests I introduced with this PR (when I was confirming Wladimir's [analysis](https://groups.google.com/g/narayana-users/c/o6L7uMFmkvo), I used the test in `XAResourceRecordUnitTest` to check if the rollback optimization was working)

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/common/FailureXAResource.java
jmfinelli: I preferred the switch style because it fits better this kind of logic (i.e. the `switch` statement can have a number of possible execution paths...which here was simulated with a sequence of `if` statements)

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
tomjenkinson: Please can you elaborate on why you think a RuntimeException shouldn't trigger this?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
tomjenkinson: Regarding the `else` - did you consider the >= XA_RBBASE in the first catch? (https://github.com/jmfinelli/narayana/blob/JBTM-3843/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java#L263)?

jmfinelli: Yep, the subsequent rollback [invocation](https://github.com/jmfinelli/narayana/blob/JBTM-3843/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java#L305) should cover that case

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
jmfinelli: Catching `RuntimeException` there is not the problem. What I think should be changed is the message with ID 160450. That message covers cases where `rollback` fails. In this catch block, it seems that ArjunaJTA is handling a failure of `end`.

ArjunaJTS/integration/src/test/java/com/arjuna/ats/jta/distributed/SimpleIsolatedServers.java
jmfinelli: Double duh...the STYLE commit changed this back to JBTM-XXXX. Copy and paste mistake :face_with_head_bandage: I will redo the STYLE commit to be sure that I didn't make any other silly mistakes. Thanks for double checking Tom

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
tomjenkinson: I see, it is within a call regarding abort though. Personally I expect I would leave it as is given it is in the abort path. If there was no connection to rollback at this point I hope I would feel differently.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
tomjenkinson: I think I understand from the other thread that you are concerned it's an xa_end call rather than a rollback and so rather than an XA problem or additional logging, you are thinking about the message talking about rollback rather than xa_end. As the logging is within a call regarding abort, personally I expect I would leave it as (if there was no connection to rollback at this point I hope I would feel differently).

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
```diff
@@ -227,10 +225,15 @@ public int topLevelPrepare()
 			case XAException.XA_RBOTHER:
 			case XAException.XA_RBPROTO:
 			case XAException.XA_RBTIMEOUT:
+				// This will avoid calling rollback when aborting
+				_rolledBack = true;
```
tomjenkinson: I am wondering if these _rolledBack should be replaced with calls to removeConnection() or at least augmented with them. I am not super sure of it but it looks like removeConnection() is intended to disconnect the XAResource from the transaction which then is somehow conditionally used to avoid doing topLevelAbort. That being said topLevelAbort would still execute in some cases due to the conditionality and we don't want that I think from an XA point of view (and when it is suppressed, it would return FINISH_OK which I am not sure if it's always correct but that is how it is written today).

By only setting this new variable _rolledBack rather than calling removeConnection(), do you know if the _recoveryObject will be closed in the same way as it was for all cases?

tomjenkinson: I might be running into a similar problem with a pull request I have: https://github.com/jbosstm/narayana/pull/2215/files#r1516032102 (in my case I am avoiding a call to destroyState() [which would call removeConnection()])

ArjunaJTS/integration/src/test/java/com/arjuna/ats/jta/distributed/SimpleIsolatedServers.java
jmfinelli: OK, now it seems that I managed to change the JBTM-XXXX :relieved:

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
jmfinelli: OK, I'm gonna resolve this discussion. Thanks Tom

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
mmusgrov: @jmfinelli Just mark it deprecated (and fill in the release note field on the JIRA) and since the default is false there should be no difference between removing it and not removing it. Then if someone is relying on it to remove the connection there won't be any surprises for them.

I would *not* spend any time writing tests for a piece of broken functionality which, even if it did work, is no needed anyway.

mmusgrov:
@jmfinelli and @tomjenkinson I'm not convinced that so much reformatting is warranted in this case: just because we have a rule that allows it does not mean we have to do so. We added that rule because one engineer regularly kept pushing us to do it but there are a number of engineers who'd prefer that people suppress the urge.

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
tomjenkinson: @jmfinelli please you add a comment here summarising if you feel it's important (or not) to (at least try to, as it won't always be possible if various crash takes place) persist the information of whether the rollback took place or whether you know it would not be useful in a crash recovery situation?

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
jmfinelli: I'll do that. Thanks @mmusgrov

jmfinelli: I added `JTAEnvironmentBean:xaRollbackOptimization is deprecated with the intention of removing it in a future major release` to `Release Note Text`. Should I also modify `Release Note Status`?

tomjenkinson: I would expect it's worth pointing out that invoking the function is no longer expected to have any affect on the behaviour of Narayana at runtime (IOW it's basically a no-op now).

jmfinelli:
> @jmfinelli and @tomjenkinson I'm not convinced that so much reformatting is warranted in this case: just because we have a rule that allows it does not mean we have to do so. We added that rule because one engineer regularly kept pushing us to do it but there are a number of engineers who'd prefer that people suppress the urge.

I think that a reformatting is very helpful for reading the code modified in this PR. Going through the code was a pain as different formatting guidelines were used over the years (for example, tabs vs spaces, and different indentation rules)

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
jmfinelli: Release Note Text updated.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/common/FailureXAResource.java
tomjenkinson: I will add the note here that I don't know if I think that the rule we have (https://github.com/jbosstm/narayana/blob/main/CONTRIBUTING.md#coding-guidelines) permits the change from `if, then, else` to `switch`.

mmusgrov:
> > @jmfinelli and @tomjenkinson I'm not convinced that so much reformatting is warranted in this case
> 
> Given we have a rule for this in place I am fine with it being used here for general formatting (and I think it's working well to have it in a separate commit). If a change to the rule is needed, I think we should discuss that on Zulip (and a change to CONTRIBUTING.md could be proposed).

It might look nice but how do we diff this code when chasing down difficult regressions or changes - then we might find it is not "working well".

I couldn't stomach another round of discussions on zulip on this topic and it would not be a good use of our time since people will insist on being able to do wholesale reformatting - the problem is that just because a rule exists people assume that they have to apply it. With such changes it's next to impossible to detect bugs introduced by such reformatting especially, as you have detected, if the author is slipping in changes to control primitives (ie 'if, then, else' to 'switch').

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
```diff
@@ -586,30 +585,6 @@ public void setXaResourceOrphanFilters(List<XAResourceOrphanFilter> xaResourceOr
         }
     }
 
-    /**
-     * Returns if connections associated to XAResources that fail during prepare should be cleaned up immediately.
```
tomjenkinson, line 590: @jmfinelli I think the Javadoc of the setter and getter should be updated as the behaviour described is no longer expected to take any effect at runtime.

It should also be `@deprecated` in the Javadoc and @Deprecated in the code

marcosgopen:
> I can't recall why we decided to make a rule that permitted reformatting the files one by one. Perhaps we could discuss (again - I expect) reformatting all the code in one go?

I would be favorable to 'reformatting all the code in one go'.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
tomjenkinson: It is possible I was running into this kind of problem, my change had previously prevented removeConnection from being executed. I have fixed it in a commit I have made here https://github.com/jbosstm/narayana/pull/2215/commits/a4f7f5153b65a72ba466c4e49af0d7032b296952

I might be squashing this before long though so I will include the point here.

I had something like:
```
					if (_prepared)
						if (destroyState)
							destroyState();
					else
						removeConnection();
```
And I think it should be:
```
					if (_prepared && destroyState)
						destroyState();
					else
						removeConnection();
```

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/common/FailureXAResource.java
```diff
@@ -9,157 +9,131 @@
...
-                throw new XAException(XAException.XA_HEURRB);
-            
-            if (_type == FailType.timeout)
-                throw new XAException(XAException.XA_RBTIMEOUT);
-            
-            if (_type == FailType.XA_RBCOMMFAIL)
-                throw new XAException(XAException.XA_RBCOMMFAIL);
+    public void end(Xid xid, int flags) throws XAException {
+        if (_locale == FailLocation.end) {
+            switch (_type) {
```
tomjenkinson, line 59: Noting here 2/3 changes from if -> switch

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/common/FailureXAResource.java
```diff
@@ -9,157 +9,131 @@
...
-                xae.initCause(new Throwable("test message"));
-                throw xae;
-            } else if(_type == FailType.XA_HEURHAZ) {// XA spec invalid error code
-                throw new XAException(XAException.XA_HEURHAZ);
-            } else if(_type == FailType.XA_RBINTEGRITY) {
-                throw new XAException(XAException.XA_RBINTEGRITY);
-            } else {
-                throw new XAException(XAException.XAER_INVAL);
+    public int prepare(Xid xid) throws XAException {
+        if ((_locale == FailLocation.prepare) || (_locale == FailLocation.prepare_and_rollback)) {
```
tomjenkinson, line 82: Noting here 3/3 changes from if -> switch

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
mmusgrov: If you are already making changes related to this logic, probably doesn't even need to be a change to this particular statement but in the logic of how connections are handled, then I don't think our recommended policies would prohibit you from making that change, ie you would not need to wait for a reformat of the whole file. 

@tomjenkinson In other words, if your change clarifies the area that you are fixing then we are flexible enough to see it as of benefit and related to the fix and we would block the PR using arguments based on rigid application of our re-formatting rules.

mmusgrov: But to answer your question, I think the way the author probably intended to code the boolean condition the way you coded it. So I agree that it's a good change and an example of why the braces should be mandatory when coding up this kind of logic.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/common/FailureXAResource.java
jmfinelli: I rolled back this change.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/common/FailureXAResource.java
jmfinelli: I rolled back this style modification.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/common/FailureXAResource.java
jmfinelli: I rolled back this change

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
jmfinelli: Yep, I also think that a situation like:
```
Prepare -> rollback (from us or from the resource) -> crash
```
can be a problem. Basically, if the RM initiated/received a rollback  but `_rolledBack` was not set to true, we will request rollback again on the same RM when we recover. However, even if we put some code in `save_state`/`restore_state` to try to save `_rolledBack` in the object store, there is still a chance to crash before  saving that variable...and we will need to request rollback again on the same resource. Having said that, the scenario where we request rollback again is not a problem as we should receive an `XAER_NOTA`, which we already handle correctly. Nevertheless, I introduced some code to try to save `_rolledBack` in the object store to reduce the chances to request rollback again after crashes.

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
jmfinelli: Done

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
jmfinelli: Replied ^^^

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
```diff
@@ -837,6 +864,14 @@ public boolean save_state(OutputObjectState os, int t)
...
 			os.packInt(_heuristic);
 			os.packBoolean(_committed);
+
+			/*
+			 * Even though there's still a chance of crashing before saving _rolledBack,
+			 * it's still worth trying to reduce the chances of invoking rollback again
+			 * on the same XA resource
+			 */
+
+			os.packBoolean(_rolledBack);
```
tomjenkinson: +1

tomjenkinson: Sorry, now I have thought about this a bit more I wonder if it's worth putting this as a separate improvement? The reason I say this is because otherwise the object store will be incompatible when someone upgrades. Maybe as this is nice to have it could be moved to a "version 8" of Narayana?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
```diff
@@ -212,13 +212,11 @@ public int topLevelPrepare()
 			 * XAER_PROTO.
 			 */
 
-			if (_rollbackOptimization) // won't have rollback called on it
-				removeConnection();
```
tomjenkinson, line 216: @jmfinelli when_rolledBack = true, from topLevelPrepare() are you sure that removeConnection() (perhaps it will be https://github.com/jbosstm/narayana/pull/2214/commits/5f0fdfeb59f119f604ceb215baa2a72609294eb8#diff-6ba9616ad040b9a0f852604d2d443821d786bb2982a9d6a1ad0bd6df048f8730R773) is called? If it's not called a transactional driver `close()` opportunity is lost so I would like to know.

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -236,13 +238,11 @@ public org.omg.CosTransactions.Vote prepare() throws HeuristicMixed,
 			 * XAER_PROTO.
 			 */
 
-			if (_rollbackOptimization) // won't have rollback called on it
-				removeConnection();
```
tomjenkinson, line 240: @jmfinelli it's essentially the same question as I had above but for this other class and so the control flow might be slightly different. So for this class when`_rolledBack = true` during topLevelPrepare() are you sure that removeConnection() is called (in topLevelAbort() somewhere perhaps)?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
jmfinelli: I think so: `removeConnection()` should cover all cases.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
jmfinelli: I'm not fully following why the OS would be incompatible when someone upgrades :-/

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
tomjenkinson: Thank you (BTW I am going to respond with more discussion but inline there rather than here and will resolve this thread)

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
tomjenkinson: Thank you!

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
tomjenkinson: Thank you!

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
tomjenkinson: If you imagine Narayana "7.0.2.Final" (let's say, but anything after this pull request were merged) were to be ran with an existing object store from a version prior then when doing a restore state of one of these objects it would not find a valid (as I suppose it would read some of the bytes of the XidImple [next thing to be written] as one) boolean to use and would then be basically in error. Please can you let me know if you would like me to expand on this?

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
tomjenkinson: Thank you!

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
tomjenkinson: I think this looks like the last issue (to be clear, affecting both XAResourceRecord even if I resolved the other conversation) in this commit. If you agree and don't think it should break transaction integrity, I think an option to consider is to back out the save and restore capability of the XAResourceRecord and to consider to raise that as a separate JBTM for the next major release adding a release note to that enhancement that it will require users to use a fresh object store or something like that?

If you do think it would break transaction integrity to not save this, we need to consider including it further but I am not sure where to go with it then to not break existing applications. If we haven't already got this feature we could consider to introduce at the same time writing a version for record types into the object store for records in order to help with compatibility in the future but it won't help with breaking applications this time I suppose (except maybe it will because maybe we can somehow detect if they don't have a version we read in the old way)...

jmfinelli: > If you imagine Narayana "7.0.2.Final" (let's say, but anything after this pull request were merged) were to be ran with an existing object store from a version prior then when doing a restore state of one of these objects it would not find a valid (as I suppose it would read some of the bytes of the XidImple [next thing to be written] as one) boolean to use and would then be basically in error. Please can you let me know if you would like me to expand on this?

Thanks Tom, that clarifies it. On the other hand though, I don't see why a user would have txns in the OS during a Narayana upgrade. Maybe we should advice users to end/solve all txns before a Narayana upgrade (in our documentation)?

tomjenkinson: If it's not already there, that would seem like an important thing to add to our documentation. I have added a JBTM to investigate this https://issues.redhat.com/browse/JBTM-3856

jmfinelli: Added a reply in that JBTM.

tomjenkinson: Thank you, much appreciated!

tomjenkinson: I agree with this being part of the fix. That said, it does break compatibility and only minimises a window.

I think as this is a known breaking change:
* JBTM-3856 should be considered a blocker to merging the change as is
* JBTM-3843 should be targeted for the next release that can have breaking changes. I presume that might be Narayana 8 or perhaps even Narayana 7.1 depending on what the project allows for
* If JBTM-3843 is backported to an earlier version of Narayana, it should be considered whether the breaking change part can be backed out as the inconvenience of a user dealing with a changed object store between versions does not seem to outweigh the benefit of preventing a spurious call to xa_rollback for this case

I think a useful enhancement to consider for Narayana would be that object store records are versioned somehow and designed to be forwards compatible such that if a later version of Narayana adds a new field, that version can still read the old format records if the changes would not affect data integrity.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
```diff
@@ -918,6 +953,7 @@ public boolean restore_state(InputObjectState os, int t)
 	    {
 	        _heuristic = os.unpackInt();
 	        _committed = os.unpackBoolean();
+					_rolledBack = os.unpackBoolean();
```
tomjenkinson: @jmfinelli would you agree to add a test that verifies this _rolledBack behaviour is working as expected after recovery, please? I can work on this instead if you prefer?

tomjenkinson: I have created a test for review, https://github.com/jmfinelli/narayana/pull/3. Though I would mention that _rolledBack seems only useful to be saved for serializable XAResources

mmusgrov: I doubt many people will be using serializable XAResources so is thi PRs fixing a corner case or is there more to this issue than it seems?

tomjenkinson: Most of the pull request is fixing the case where we are calling erroneously calling xa_rollback when a resource manager returned a rollback condition during prepare and that is beneficial to all users.

Persisting the value is logical and thus appropriate to include, but it does not seem to benefit all XAResource implementations and breaks compatibility with the object store format for this record type so that would need describing and considering for any release that does not imply starting with a fresh object store.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
tomjenkinson: Please can you clarify why it was safe to remove a conditional call to removeConnection?

tomjenkinson: So I expect it is close to something like _rollbackOptimization was default to false, and given it's no longer configurable it's expected to not be needed. Personally I don't think that deprecating something, implies removing the functionality.

tomjenkinson: Also, in your comment you said, removeConnection should cover all the case where the record hasn't been prepared. Please can you point out the code where removeConnection is called when the prepare has not completed for the record?

tomjenkinson: I do understand though, we can't call removeConnection if we get XAER_INVAL, XAER_PROTO, XAER_RMERR or XAER_RMFAIL. So maybe this removeConnection just moves to after the line below where `_rolledBack = true;`?

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
```diff
@@ -595,19 +596,22 @@ public void setXaResourceOrphanFilters(List<XAResourceOrphanFilter> xaResourceOr
...
+     * Deprecated: This method will be removed in the next major release
+     *
      * Sets if failed resources should be cleaned up during prepare or during phase two.
      *
      * @param xaRollbackOptimization true for immediate cleanup, false for phase two cleanup.
      */
+    @Deprecated
     public void setXaRollbackOptimization(boolean xaRollbackOptimization)
     {
-        this.xaRollbackOptimization = xaRollbackOptimization;
```
tomjenkinson, line 610: If removing the functionality does go ahead, I think logging a warning might (needs a bit more thought) be appropriate to add here

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
tomjenkinson: BTW when running the current changes and the test removeConnection did not get called (for either TestResource)

tomjenkinson: Looking at the code, I am not seeing how removeConnection is called for prepared resources in the rollback case (already today) as it is guarded by !_prepared (unlike the commit case which does not check for whether it was _prepared or not)

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
```diff
@@ -297,13 +297,17 @@ public int topLevelAbort()
 						&& (e1.errorCode < XAException.XA_RBEND))
 				    {
 					/*
-					 * Has been marked as rollback-only. We still
-					 * need to call rollback.
+					 * Has been marked as rollback-only. If the branch has not been
+					 * already rolled back (i.e. _rolledBack == true ), we still need
```
tomjenkinson, line 301: I think it might be clearer to give the code example in the inverse of how it is here, in other words `(i.e. _rolledBack = false)` - please let me know what you think? The text before and after the example appear to be describing that case.

jmfinelli: Done, thanks @tomjenkinson

qa/tests/src/org/jboss/jbossts/qa/CrashRecovery12Resources/Resource01.java
```diff
@@ -87,6 +87,14 @@ public int prepare(Xid xid) throws XAException
...
 		if (crashIn == ROLLBACK)
 		{
+			/*
+			 * Since JBTM-3843, rollback is no longer invoked
+			 * during the second phase of the 2PC protocol as
+			 * the Resource Manager should have rolled back
+			 * the branch during the first phase
+			 */
+			myLog(ERR, "Crashing in prepare");
+			myLog(OUT, "Passed");
```
tomjenkinson, line 97: I understand the need to change this compared to the existing code, that being said I think the change to this test could be a little more than the lines here because (I presume) this test is testing a crash situation in the resource where the resource has rolled back in prepare but actually the Resource is no longer crashing, rather it is returning XAException.XA_RBROLLBACK out of prepare (as it was doing so before) so I am not sure what to suggest. I presume we maybe should be System.exit(0) in here but not sure what other changes might then be needed.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
tomjenkinson: Transactional driver seems to be able to release connections even when removeConnection is not called. I will resolve this

qa/tests/src/org/jboss/jbossts/qa/CrashRecovery12Resources/Resource01.java
jmfinelli: Modified. Thanks Tom

ArjunaJTS/integration/src/test/java/com/arjuna/ats/jta/distributed/TestResource.java
```diff
@@ -121,37 +136,46 @@ public TestResource(String nodeName, boolean b, boolean fatalCommit) {
...
+				return XA_OK;
 			}
-			return XA_OK;
 		}
 	}
 
+
 	public synchronized void commit(Xid id, boolean onePhase) throws XAException {
-		System.out.println("[" + Thread.currentThread().getName() + "] TestResource (" + serverId + ")      XA_COMMIT  [" + id + "]");
+		System.out.println("[" + Thread.currentThread().getName() + "] TestResource (" + serverId + ")      XA_COMMIT  [" + xid + "]");
```
tomjenkinson, line 178: Hi Manuel the id part of the logging should go in the second commit in the pull request rather than this one please. Same with line 191 a little below. It should be `this.fatalXid = id;` in this commit please.

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
jmfinelli: We removed the breaking change thus I think we can merge (and backport) this PR as is. @tomjenkinson, WDYT?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
jmfinelli: We removed the braking change so this can be resolved. Thanks

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
jmfinelli: Thanks Tom. Done

ArjunaJTS/integration/src/test/java/com/arjuna/ats/jta/distributed/TestResource.java
jmfinelli: Done. Thanks Tom

ArjunaJTS/integration/src/test/java/com/arjuna/ats/jta/distributed/TestResource.java
```diff
@@ -38,10 +40,28 @@ public class TestResource implements XAResource {
...
     private transient boolean fatalCommit;
 
+	private transient boolean fatalPrepare;
+	private Xid fatalXid;
+
 	public TestResource(String serverId, boolean readonly) {
+		this(serverId, readonly, false);
+	}
+
+	public TestResource(String serverId, boolean b, boolean fatalCommit) {
```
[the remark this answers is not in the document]
jmfinelli: Done. Thanks Done

ArjunaJTA/jta/classes/com/arjuna/ats/jta/common/JTAEnvironmentBean.java
tomjenkinson: Thank you for adding the warning

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
```diff
@@ -750,7 +774,10 @@ public int topLevelOnePhaseCommit()
 	            }
 	        }
 	        else
-	            return TwoPhaseOutcome.ONE_PHASE_ERROR;
+	        {
+	          this._rolledBack = true;
```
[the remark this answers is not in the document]
jmfinelli: Yep, that is the reason. Thanks Tom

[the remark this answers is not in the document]
tomjenkinson: Thank you for your response, I will resolve the conversation.

ArjunaJTS/integration/src/test/java/com/arjuna/ats/jta/distributed/TestResource.java
tomjenkinson: Thank you

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/XAResourceRecord.java
tomjenkinson: I am going to resolve this thread. (and I definitely think we can backport the functional commit of the pull request, I would need to consider the benefit of backporting the commit which renamed the variables).

Thank you

tomjenkinson (approved):
LGTM - thank you!

jmfinelli:
Thank you Tom :-)

PR #2215, opened 2024-03-06
JBTM-3838 Review JTS and JTA way to handle RMFAIL in xa_rollback

tomjenkinson:

CORE !TOMCAT AS_TESTS !RTS !JACOCO !XTS QA_JTA QA_JTS_OPENJDKORB PERFORMANCE !LRA DB_TESTS mysql db2 postgres oracle

### DIFF
--- a/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
+++ b/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
@@ -343,6 +343,8 @@ public class XAResourceRecord extends com.arjuna.ArjunaOTS.OTSAbstractRecordPOA
 				    }
 				}
 
+				// This is used to ensure that if xa_rollback fails
+				boolean destroyState = true;
 				try
 				{
 					_theXAResource.rollback(_tranID);
@@ -401,6 +403,14 @@ public class XAResourceRecord extends com.arjuna.ArjunaOTS.OTSAbstractRecordPOA
 								destroyState();
 								break;
 							}
+						case XAException.XAER_RMFAIL:
+							if (_prepared) {
+								// We do not want to delete the state on disk because that would prevent the recovery
+								// manager from reattempting recovery and XAER_RMFAIL out of xa_rollback is transient
+								// failure
+								destroyState = false;
+							}
+							break;
 						default:
 							destroyState();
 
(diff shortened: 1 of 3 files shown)


marcosgopen:
Thanks @tomjenkinson for your PR. I noticed some QA tests related to database (db2,oracle and postgres) failed. Are you aware of this? i.e JDBCResources01_abstract_Test and JDBCResources02_abstract_Test 03, 04,17 and 27.
`jdbcresources01_oracle_thin_jndi JDBCResources01_abstract_Test003 Fail (0m18.928s)
jdbcresources01_oracle_thin_jndi JDBCResources01_abstract_Test004 Fail (0m24.574s)
jdbcresources01_oracle_thin_jndi JDBCResources01_abstract_Test017 Fail (0m22.801s)
jdbcresources01_oracle_thin_jndi JDBCResources01_abstract_Test027 Fail (0m23.766s)
jdbcresources02_oracle_thin_jndi JDBCResources02_abstract_Test003 Fail (0m18.076s)
jdbcresources02_oracle_thin_jndi JDBCResources02_abstract_Test004 Fail (0m17.921s)
jdbcresources02_oracle_thin_jndi JDBCResources02_abstract_Test017 Fail (0m21.058s)
jdbcresources02_oracle_thin_jndi JDBCResources02_abstract_Test027 Fail (0m23.067s)`

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -422,7 +432,8 @@ public void rollback() throws org.omg.CORBA.SystemException,
 				{
 				    _phaseTwoStarted = true;
 					if (_prepared)
-						destroyState();
+						if (destroyState)
+							destroyState();
```
tomjenkinson, line 436: I think I can see a problem with this, by not calling destroyState we are then not going to call removeConnection which is going to keep at least the JDBC transactional driver associated with this XAResourceRecord. I am not sure if that is the cause of the failures that @marcosgopen shared though because I am not sure if it's expected a resource manager in the test (not necessarily the database) would be returning RMFAIL.

tomjenkinson: I think it was a missing call to `removeConnection()` but it wasn't quite for the reason I thought. By adding the code as I added it, the `removeConnection()` had moved to being an else clause on the `if (destroyState)` rather than being the else of the `if (_prepared)`

marcosgopen (approved):
Thank you @tomjenkinson for your fix and your test. Much appreciated!!

PR #2220, opened 2024-03-11
[JBTM-3847] Enable checkstyle by default

marcosgopen:
https://issues.redhat.com/browse/JBTM-3847 

Enable checkstyle by default
CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !LRA !DB_TESTS

### DIFF
--- a/ArjunaJTA/jdbc/pom.xml
+++ b/ArjunaJTA/jdbc/pom.xml
@@ -16,6 +16,9 @@
   <name>Narayana: ArjunaJTA jdbc</name>
   <description>transactional driver</description>
 
+  <properties>
+    <checkstyle.skip>true</checkstyle.skip>
+  </properties>
   <dependencyManagement>
     <dependencies>
       <dependency>
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -101,8 +101,7 @@ We follow [Open source participation guidelines](https://www.redhat.com/en/resou
 
 ### Coding Guidelines
 
-We use the maven `checkstyle` and `sortpom` plugins which are skipped by default for older modules
-and enabled for newer ones.
+We use the maven `checkstyle` and `sortpom` plugins which are enabled by default and skipped in certain cases for older modules.
 
 For older modules where checkstyle is disabled, when editing a file you should either
 a) follow the style currently used in the file you have edited or,
--- a/ArjunaCore/pom.xml
+++ b/ArjunaCore/pom.xml
@@ -18,6 +18,7 @@
   <name>Narayana: Arjunacore</name>
   <description>Narayana: Arjunacore All</description>
   <properties>
+    <checkstyle.skip>true</checkstyle.skip>
     <sortpom.skip>true</sortpom.skip>
   </properties>
 
--- a/ArjunaJTA/cdi/pom.xml
+++ b/ArjunaJTA/cdi/pom.xml
@@ -18,8 +18,6 @@
 
   <properties>
 
-    <!-- using Narayana checkstyle during maven build -->
-    <checkstyle.skip>false</checkstyle.skip>
     <ipv6.server.jvm.args>-Djboss.bind.address=[::1] -Djboss.bind.address.management=[::1]
       -Djboss.bind.address.unsecure=[::1] -Djava.net.preferIPv4Stack=false -Djava.net.preferIPv6Addresses=true</ipv6.server.jvm.args>
     <server.jvm.args>${jvm.args.other} ${jvm.args.memory} ${jvm.args.debug} ${jvm.args.modular}</server.jvm.args>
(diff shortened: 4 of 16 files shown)


rts/pom.xml
```diff
@@ -21,7 +21,6 @@
   <properties>
     <version.resteasy-client>${version.org.jboss.resteasy}</version.resteasy-client>
     <version.narayana>${project.version}</version.narayana>
-    <sortpom.skip>true</sortpom.skip>
```
tomjenkinson, line 24: @marcosgopen rts/pom.xml is not a new module, I found this module `org.jboss.narayana.rts:rest-tx` in https://github.com/jbosstm/narayana/blob/5.5.34.Final/rts/pom.xml. I think you should revert the removal of `<sortpom.skip>false</sortpom.skip>`  for this module (and remove the commit with the sorted pom for this module). Is that OK for you, please?

CONTRIBUTING.md
```diff
@@ -101,16 +101,13 @@ We follow [Open source participation guidelines](https://www.redhat.com/en/resou
...
+We use the maven `checkstyle` and `sortpom` plugins which are enabled by default, but skipped for older modules.
 
 For older modules where checkstyle is disabled, when editing a file you should either
-a) follow the style currently used in the file you have edited or,
-b) reformat it according to our checkstyle ruleset (by temporarily enabling checkstyle
-on the affected module) but separate functional changes from formatting changes into different commits
-in the same PR (this rule facilitates traceability).
-In other words a file (in a module with checkstyle disabled) may only be re-formatted if it also
-contains a functional change and is made in a separate commit.
+a) reformat it according to our checkstyle ruleset (by temporarily enabling checkstyle on the affected module) but separate functional changes from formatting changes into different commits in the same PR (this rule facilitates traceability).
```
tomjenkinson, line 107: @marcosgopen I think this part of the changeset should be reverted as I don't think we should be altering the ordering of the options. Please can I ask if you are OK to revert this part of the change?

rts/pom.xml
marcosgopen: Sure, thanks for your review.

CONTRIBUTING.md
marcosgopen: mmh, I am ok with your review. But could I ask you to read the https://issues.redhat.com/browse/JBTM-3302 description and tell me if you still have the same opinion as my change is aligned to that issue.

marcosgopen: However I think you are right and this is not the right place where to change it, so I have just removed it as you asked and updated the PR.

ArjunaJTA/jdbc/pom.xml
```diff
@@ -16,6 +16,9 @@
   <name>Narayana: ArjunaJTA jdbc</name>
   <description>transactional driver</description>
 
+  <properties>
+    <checkstyle.skip>true</checkstyle.skip>
```
tomjenkinson, line 20: Does this need `<sortpom.skip>true</sortpom.skip>` adding too? If you look at the diff for this pull request and where `<checkstyle.skip>true</checkstyle.skip>` I think it's possible (unless it is inherited somehow for the various modules) that those modules which we have checkstyle.skip being true should also have `<sortpom.skip>true</sortpom.skip>` too. Do you agree that seems a correct default assumption?

marcosgopen: Not sure @tomjenkinson , strangly enough the sortpom was already enabled by default: https://github.com/jbosstm/narayana/blob/main/pom.xml#L177-L178
So I think modules that were not skipping before it shouldn't skip it now.

marcosgopen: At least now checkstyle and sortpom will be consistent, meaning that they will be both enabled by default.

CONTRIBUTING.md
```diff
@@ -101,8 +101,7 @@ We follow [Open source participation guidelines](https://www.redhat.com/en/resou
 
 ### Coding Guidelines
 
-We use the maven `checkstyle` and `sortpom` plugins which are skipped by default for older modules
-and enabled for newer ones.
+We use the maven `checkstyle` and `sortpom` plugins which are enabled by default, but skipped for older modules.
```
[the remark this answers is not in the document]
tomjenkinson: maybe "and skipped in certain cases for older modules" - WDYT?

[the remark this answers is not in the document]
marcosgopen: Oh, I see what you mean. Yes, I agree and update the [CONTRIBUTING.md](https://github.com/jbosstm/narayana/pull/2220/files/b539a0db5ad21c30e459925ffeb098c6b46b8438#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055)

tomjenkinson (approved):
LGTM - thank you @marcosgopen!

PR #2221, opened 2024-03-14
JBTM-3056 xts-test-servlet is not needed in the full distribution

mmusgrov:
https://issues.redhat.com/browse/JBTM-3056

xts-test-servlet does not need to be in the full distro

XTS !CORE !TOMCAT !AS_TESTS !RTS !JACOCO !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !LRA !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/narayana-full/pom.xml
+++ b/narayana-full/pom.xml
@@ -85,11 +85,6 @@
       <classifier>participant</classifier>
       <type>war</type>
     </dependency>
-    <!-- This dependency should be removed in JBTM-3056 -->
-    <dependency>
-      <groupId>org.jboss.narayana.xts</groupId>
-      <artifactId>xts-test-servlet</artifactId>
-    </dependency>
     <dependency>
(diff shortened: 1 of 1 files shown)


marcosgopen (approved):
Thanks @mmusgrov for your PR, could I ask you what is the use of 'xts-test-servlet'? Is it still even used?
If fact, as i wrote to the Jira comments, 'when building it we get " skip non existing resourceDirectory", "No sources to compile", "No tests to run. JAR will be empty - no content was marked for inclusion!" and  "No sources in project. Archive not created."'

mmusgrov:
I don't know. I wasn't able to duplicate what you reported in your comment - I built the project and I built the narayana-full jar but did not see what you saw in the output. The jar, though, contains just the pom which I think probably matches up with your comment. The module hasn't changed since it was migrated from svn over a decade ago so I suspect it really is just some left over cruft.

PR #2224, opened 2024-03-14
JBTM-3848

tomjenkinson:
https://issues.redhat.com/browse/JBTM-3848

AS_TESTS (directly benefit from the change)
RTS (for rts_as_tests)
XTS (for xts_as_tests tx_bridge_tests compensations_tests)

!CORE !TOMCAT !JACOCO !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !LRA !DB_TESTS mysql db2 postgres oracle


scripts/hudson/narayana.sh
```diff
@@ -416,13 +416,9 @@ function build_as {
 
 function tests_as {
   # running WildFly testsuite if configured to be run by axis AS_TESTS
-  if [ $AS_TESTS_TRACE ]; then
```
tomjenkinson, line 419: the change removes the ability to set trace output when running the AS_TESTS axis as the server is not built and ran in two steps now. But the test passes when it's ran in one step though rather than seeming to cause every build to fail

scripts/hudson/narayana.sh
```diff
@@ -438,6 +434,11 @@ function init_jboss_home {
   grep 'connection-timeout' "${CONF}"
   #Enable remote debugger
   echo JAVA_OPTS='"$JAVA_OPTS -agentlib:jdwp=transport=dt_socket,address=8787,server=y,suspend=n"' >> "$JBOSS_HOME"/bin/standalone.conf
+
+  if [ $AS_TESTS_TRACE ]; then
```
tomjenkinson, line 438: This should allow other axis to benefit from being able to enable AS trace still

marcosgopen (approved):
Thanks @tomjenkinson for spotting that the 'one step' build and test is working. Your change LGTM! Additionally I can also see that 'install -DallTests' is the documented way to  "run all the tests"  https://github.com/wildfly/wildfly/blob/main/README.md?plain=1#L96-L98

PR #2226, opened 2024-03-14
[JBTM-3853] Performance test hanging

marcosgopen:
https://issues.redhat.com/browse/JBTM-3853

CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !LRA DB_TESTS mysql db2 postgres oracle

### DIFF
--- a/tools/src/main/java/io/narayana/perf/Measurement.java
+++ b/tools/src/main/java/io/narayana/perf/Measurement.java
@@ -385,9 +385,11 @@ public class Measurement<T> implements Serializable {
         } catch (BrokenBarrierException e) {
             opts.incrementErrorCount(); // ? exactly how many errors were there?
             opts.setCancelled(true);
+            throw new RuntimeException(e);
         } catch (TimeoutException e) {
             opts.incrementErrorCount(); // ? exactly how many errors were there?
             opts.setCancelled(true);
+            throw new RuntimeException(e);
         }
 
         for (Future<Measurement<T>> t : tasks) {


tomjenkinson (commented):
Thank you for the pull request and a fix in this area will be great - I do want to ask if you know if the test will fail if the timeout fires or will the framework just think there is low performance?

marcosgopen:
> Thank you for the pull request and a fix in this area will be great - I do want to ask if you know if the test will fail if the timeout fires or will the framework just think there is low performance?

Thanks @tomjenkinson for your question, it should fail throwing a runtimeException: https://github.com/jbosstm/narayana/blob/main/tools/src/main/java/io/narayana/perf/Measurement.java#L371-L384

marcosgopen:
Now that you mention this Tom, I think this line should be removed: https://github.com/jbosstm/narayana/blob/main/tools/src/main/java/io/narayana/perf/Measurement.java#L367
I'll update the PR.

tools/src/main/java/io/narayana/perf/Measurement.java
```diff
@@ -364,8 +364,6 @@ public Measurement<T> call() throws Exception {
         long start = System.nanoTime();
 
         try {
-            cyclicBarrier.await(); // wait for each thread to arrive at the barrier
```
tomjenkinson, line 367: I think I noticed this code myself, do you think it is definitely wrong? It doesn't somehow let them all sync up in one place before the test starts in full or something?

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/performance/Performance1.java
```diff
@@ -22,8 +22,9 @@ public void test() {
         int threadCount =  1;
         int batchSize = 100;
 
+        // Using 10 minutes timeout (maxTestTime), it should be lower than 1 minute
         Measurement measurement = new Measurement.Builder(getClass().getName() + "_test1")
-                .maxTestTime(0L).numberOfCalls(numberOfTransactions)
+                .maxTestTime(600000L).numberOfCalls(numberOfTransactions)
```
mmusgrov, line 27: The timeout controls how long to wait for all threads to reach the barrier. But all threads should reach the barrier and if one of them doesn't reach it then that is the bug ie you need to find out which thread is blocked from hitting the barrier - setting a non zero timeout is bypassing the problem.

Also note that the measurement framework was written before JMH was available and the tests that use it should be rewritten to use JMH, there aren't many of them left so that should be our priority after which we can delete the framework.

mmusgrov (changes requested):
The change does not get at the root cause of the hang.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/performance/Performance1.java
tomjenkinson: IIUC Marco spotted this (at least most recently) when databases weren't working so basically the test was not going to be successful (presumably for any thread basically). I understand what you are saying though and I think it is possible to consider revering these timeouts (even if I am not super comfortable with that as they give a bit more protection that what I am about to say).

So following on from that, @marcosgopen please can you confirm if the addition (in the Measurement.java elsewhere in your changeset) of one of the new `throw new RuntimeException(e);` would handle the same situation the same anyway and thus the timeouts don't make a material improvement and could therefore be reverted? Was it the BrokenBarrierException clause?

marcosgopen: @tomjenkinson is right @mmusgrov , the problem is that the database returned an error and the test hung instead of failing.

marcosgopen: > @marcosgopen please can you confirm if the addition (in the Measurement.java elsewhere in your changeset) of one of the new throw new RuntimeException(e); would handle the same problem as the timeouts anyway and thus the timeouts don't make a material improvement and could therefore be reverted? In other words (and from a bit of presumption), was it the BrokenBarrierException clause that would be considered the more direct fix when the database is not working?

I see what you mean @tomjenkinson, I think you are right. Probably the fact that the exception is caught and not re thrown is the problem. So I can remove the changes to the Performance* classes and keep the Measurement change only. That would be enough to prevent the test from hanging. Thanks

tomjenkinson (approved):
LGTM [BTW, I don't expect the TimeoutException clause needed the RuntimeException to be thrown to fix the reported scenario, but I do hope it could anyway provide a useful extra protection if getMaxTestTime was greater than 0 and the CyclicBarrier::await(long,TimeUnit) was therefore used and having received a TimeoutException we end up at t.get() which seems it could maybe block then].

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/performance/Performance1.java
mmusgrov: @tomjenkinson and @marcosgopen  Why don't you just port them to JMH and move them to the perf repo and then delete the measurement framework? Porting these perf tests isn't particularly difficult.

tomjenkinson: We should do that @mmusgrov I agree. It seems a good next step in this particular area (I think fixing possible hangs is still a good win in the meantime).

For the porting side of things, I think I found the issue [https://issues.redhat.com/browse/JBTM-2621] - I think it needs re-opening as I think it might not have been completed from a quick look at the linked pull request, do you agree with my read or do you think we need a new issue? The hanging test was `com.hp.mwtests.ts.arjuna.performance.Performance1` which is different to the ones mentioned explicitly in JBTM-2621 but then JBTM-2621 talks about `if there are no others then once ported the old performance benchmark harness can be removed. `

marcosgopen:
> I approved it even though I suggested porting it to JMH and moving it to the perf repo.

Yes, I will work on 'porting it to JMH', those tests also need to be moved to the performance project IIRC.

marcosgopen:
I will keep this PR open until all tests finish.

PR #2228, opened 2024-03-14
JBTM-3855 Deprecate some code for removal

tomjenkinson:
https://issues.redhat.com/browse/JBTM-3855

!CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !LRA !DB_TESTS mysql db2 postgres oracle


mmusgrov (commented):
@tomjenkinson  The JIRA does not say why the code is no longer required so it's unclear whether or not something would need to replace it. Did you try removing the deprecated methods to see if it still builds and that the tests still pass? If so then I would be willing to approve the PR.

mmusgrov (approved):
Thanks for the explanations, I didn't spot that it was a subtask and the parent task provides more context. Since it still builds when you remove the deprecated items I think it is safe to mark it deprecated, we can always un-deprecate if we find that the IDL compiler is needed.

PR #2231, opened 2024-03-20
Jbtm 1801 Move all test properties into root pom.xml

marcosgopen:
https://issues.redhat.com/browse/JBTM-1801

CORE !TOMCAT AS_TESTS RTS !JACOCO XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !LRA !DB_TESTS mysql db2 postgres oracle

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -156,6 +156,8 @@
     <buildproperty.date>${maven.build.timestamp}</buildproperty.date>
     <!-- Checkstyle is disabled by default, but it could be enabled for default by a particular module -->
     <checkstyle.skip>true</checkstyle.skip>
+    <ipv6.server.jvm.args>-Djboss.bind.address=[::1] -Djboss.bind.address.management=[::1]
+      -Djboss.bind.address.unsecure=[::1] -Djava.net.preferIPv4Stack=false -Djava.net.preferIPv6Addresses=true</ipv6.server.jvm.args>
     <jvm.args.byteman>-Dorg.jboss.byteman.verbose
                         -Djboss.modules.system.pkgs=org.jboss.byteman
                         -Dorg.jboss.byteman.transform.all
@@ -173,7 +175,7 @@
     <maven.compiler.source>11</maven.compiler.source>
     <maven.compiler.target>11</maven.compiler.target>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <server.jvm.args>${jvm.args.other} ${jvm.args.byteman} ${jvm.args.memory} ${jvm.args.debug} ${jvm.args.modular}</server.jvm.args>
+    <server.jvm.args>${jvm.args.other} ${jvm.args.memory} ${jvm.args.debug} ${jvm.args.modular}</server.jvm.args>
     <!-- Sortpom is enabled by default, but it could be disabled in sub-modules setting sortpom.skip = true -->
     <sortpom.skip>false</sortpom.skip>
     <!-- The surefire/failsafe test execution logs goes to a file by default -->
--- a/rts/at/webservice/pom.xml
+++ b/rts/at/webservice/pom.xml
@@ -51,12 +51,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>asm</groupId>
-      <artifactId>asm-all</artifactId>
-      <version>${version.asm}</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <scope>test</scope>
--- a/ArjunaJTA/cdi/pom.xml
+++ b/ArjunaJTA/cdi/pom.xml
@@ -16,13 +16,6 @@
   <name>Narayana: ArjunaJTA cdi</name>
   <description>Narayana: ArjunaJTA cdi</description>
 
-  <properties>
-
-    <ipv6.server.jvm.args>-Djboss.bind.address=[::1] -Djboss.bind.address.management=[::1]
-      -Djboss.bind.address.unsecure=[::1] -Djava.net.preferIPv4Stack=false -Djava.net.preferIPv6Addresses=true</ipv6.server.jvm.args>
-    <server.jvm.args>${jvm.args.other} ${jvm.args.memory} ${jvm.args.debug} ${jvm.args.modular}</server.jvm.args>
-  </properties>
-
   <dependencyManagement>
     <dependencies>
       <dependency>
--- a/ArjunaJTA/jta/pom.xml
+++ b/ArjunaJTA/jta/pom.xml
@@ -21,7 +21,6 @@
     <!-- java.net.preferIPv6Addresses=false is a workaround to (not) solve the bug in arquillian-extension-byteman -->
     <ipv6.server.jvm.args>-Djboss.bind.address=[::1] -Djboss.bind.address.management=[::1]
       -Djboss.bind.address.unsecure=[::1] -Djava.net.preferIPv4Stack=false -Djava.net.preferIPv6Addresses=false</ipv6.server.jvm.args>
-    <server.jvm.args>${jvm.args.other} ${jvm.args.memory} ${jvm.args.debug} ${jvm.args.modular}</server.jvm.args>
   </properties>
 
   <dependencyManagement>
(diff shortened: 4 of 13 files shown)


marcosgopen:
AS_TESTS failure is not related to this PR.

mmusgrov:
The PR changes don't match up with the JIRA which seems specific to XTS.
The "Affects Version/s:" field on the issue is not set so we cannot tell what file and what properties the JBTM was referring to.

You should probably create a new JBTM for this and then mark JBTM-1801 as out of date.

marcosgopen:
> The PR changes don't match up with the JIRA which seems specific to XTS. The "Affects Version/s:" field on the issue is not set so we cannot tell what file and what properties the JBTM was referring to.
> 
> You should probably create a new JBTM for this and then mark JBTM-1801 as out of date.

I am ok with marking JBTM-1801 as out of date, though I don't think it refers to XTS, but I think XTS property was just an example. In fact in the title it says 'Move all test properties into root pom.xml', not XTS properties. Concerning the properties the JBTM was referring to I am pretty sure he refers to [this property](https://github.com/jbosstm/narayana/blame/a0a74fa36d215a5e033f34392f885bec52f7eb2e/XTS/localjunit/pom.xml#L22) and similar, in fact that was the version at the time the Jira was opened.

mmusgrov:
I think you are inferring too much, although you could well be correct the problem is that:

- the Description field refers to XTS
- the description says "... come up with a common set of properties that works in all cases", note the plural in this quote
  but your PR only refers to a single property ("server.jvm.args")
- the description does try to indicate which properties the JBMT should cover (https://github.com/jbosstm/narayana/blob/master/XTS/localjunit/pom.xml#L22)
  but because the "Affects Version/s" field is blank we cannot ascertain which properties the author intended

I guess this issue tracker is an exemplar of the importance of having well written JBTMs.

mmusgrov:
The date the JIRA was created indicates that the description is referring to SHA 9da763d16 and using this assumption the code link points at the following properties:
    `ipv4.server.jvm.args, ipv6.server.jvm.args, server.debug.args`.

Does your PR cover all of these?

mmusgrov:
I also notice that the XTS pom is overriding the property in the parent pom. Does this mean that if you move the property up to the parent pom then all poms will get the same value - ie will the child poms still have their intended values?

marcosgopen:
> The date the JIRA was created indicates that the description is referring to SHA [9da763d](https://github.com/jbosstm/narayana/commit/9da763d168c488538fb2c414d474b597e4dd11ce) and using this assumption the code link points at the following properties: `ipv4.server.jvm.args, ipv6.server.jvm.args, server.debug.args`.
> 
> Does your PR cover all of these?

Yes, it does. Please note that 'ipv4.server.jvm.args' does not exist anymore.

marcosgopen:
> I also notice that the XTS pom is overriding the property in the parent pom. Does this mean that if you move the property up to the parent pom then all poms will get the same value - ie will the child poms still have their intended values?
> 
> And did you "... come up with a common set of properties that works in all cases" as the JBTM is requesting?

Yes, I did. And I put them in the root pom. Please note that I interpreted that sentence as "... come up with a common set of properties that works in *most* cases" because there are some module that need custom configuration (e.g. module testing with byteman).

marcosgopen:
> I think you are inferring too much, although you could well be correct the problem is that:
> 
>     * the Description field refers to XTS
> 
>     * the description says "... come up with a common set of properties that works in all cases", note the plural in this quote
>       but your PR only refers to a single property ("server.jvm.args")
> 
>     * the description does try to indicate which properties the JBMT should cover (https://github.com/jbosstm/narayana/blob/master/XTS/localjunit/pom.xml#L22)
>       but because the "Affects Version/s" field is blank we cannot ascertain which properties the author intended
> 
> 
> I guess this issue tracker is an exemplar of the importance of having well written JBTMs.

I agree with you, just a little correction: my PR refers to these properties "server.jvm.args" and "ipv6.server.jvm.args" (note that 'ipv4.server.jvm.args' does not exist anymore).

mmusgrov:
> > I think you are inferring too much, although you could well be correct the problem is that:
> > ```
> > * the Description field refers to XTS
> > 
> > * the description says "... come up with a common set of properties that works in all cases", note the plural in this quote
> >   but your PR only refers to a single property ("server.jvm.args")
> > 
> > * the description does try to indicate which properties the JBMT should cover (https://github.com/jbosstm/narayana/blob/master/XTS/localjunit/pom.xml#L22)
> >   but because the "Affects Version/s" field is blank we cannot ascertain which properties the author intended
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I guess this issue tracker is an exemplar of the importance of having well written JBTMs.
> 
> I agree with you, just a little correction: my PR refers to these properties "server.jvm.args" and "ipv6.server.jvm.args" (note that 'ipv4.server.jvm.args' does not exist anymore).

Can you update the JBTM description to address these three bullets, the "Affects Version/s" field can be found by looking at SHA https://github.com/jbosstm/narayana/commit/9da763d168c488538fb2c414d474b597e4dd11ce

mmusgrov:
@jmfinelli Will you review this PR, in particular checking that each module still gets it's expected set of property values - eg each module that uses them used to override the parent values so the review would need to check that the modules still see the expected property values after they have been moved to the parent pom.

pom.xml
```diff
@@ -173,7 +175,7 @@
     <maven.compiler.source>11</maven.compiler.source>
     <maven.compiler.target>11</maven.compiler.target>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <server.jvm.args>${jvm.args.other} ${jvm.args.byteman} ${jvm.args.memory} ${jvm.args.debug} ${jvm.args.modular}</server.jvm.args>
```
jmfinelli, line 176: What's the reason for removing `jvm.args.byteman`?

rts/at/webservice/pom.xml
```diff
@@ -50,12 +50,6 @@
       <artifactId>narayana-jta</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>asm</groupId>
-      <artifactId>asm-all</artifactId>
-      <version>${version.asm}</version>
-      <scope>test</scope>
-    </dependency>
```
jmfinelli, line 58: I'll add the comment here but this message is valid for all `asm` removals: is this dependency removal related to JBTM-1801?

jmfinelli: I just noticed you used a different commit to remove `asm`, which is alright for me.

pom.xml
marcosgopen: Thanks for your question Manuel, I think that the byteman argument is specific to only some modules. So I would opt to include it only in modules that need it, and not in the root pom. WDYT? Please note that the jvm.args.byteman variable is defined in the root pom too, so it would be injected in all projects using server.jvm.args even when not using byteman.

marcosgopen: I see your point, thanks. I will update the PR.

mmusgrov: Which modules use byteman? If there aren't many then my vote would be to have each module set their own system properties since they add noise to command lines.

marcosgopen: There are quite a few modules using byteman: arjuna, ArjunaJTA, ArjunaJTS, compensation, rts, txbridge and XTS (byteman_support). Though not all sub modules needs it . However I investigated a bit more and apparently some modules (e.g. ArjunaJTA/jta) fail when 'jvm.args.byteman' is set in the the 'server.jvm.args' argument which error `org.jboss.arquillian.container.spi.client.container.LifecycleException: The java process starting the managed server exited unexpectedly with code [1]`, which is , I think, due to the fact that byteman start an agent with a listener (which needs a port, see line https://github.com/jbosstm/narayana/blob/main/pom.xml#L162). I couldn't go deeper in the error though. Anyway I'd rather not use such a configuration generally because it starts an agent which might not be needed and it consumes more resources and time. After this additional consideration, do you agree with me @jmfinelli that it would be better not include it by default as an server argument ?

mmusgrov: Thanks

jmfinelli: Thanks @marcosgopen. It looks like we need to investigate further how we activate byteman in different modules. WDYT about opening a new JBTM to track this task? Nevertheless, I agree with you guys that activating byteman from the root pom is a waste of resources thus I'll approve this PR as is.

jmfinelli (approved):

pom.xml
marcosgopen: I agree with you Manuel, a new JBTM for tracking the usage of Byteman would be helpful. Thanks

PR #2232, opened 2024-03-21
JBTM-3861 Ensure that the failsafe plugin is bound to an execution

tomjenkinson:
https://issues.redhat.com/browse/JBTM-3861

!CORE !TOMCAT !AS_TESTS !RTS !JACOCO XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !LRA !DB_TESTS mysql db2 postgres oracle

### DIFF
--- a/XTS/localjunit/WSTX11-interop/pom.xml
+++ b/XTS/localjunit/WSTX11-interop/pom.xml
@@ -47,6 +47,14 @@
 
           </systemProperties>
         </configuration>
+        <executions>
+          <execution>
+            <goals>
+              <goal>integration-test</goal>
+              <goal>verify</goal>
+            </goals>
+          </execution>
+        </executions>
       </plugin>
     </plugins>
   </build>


marcosgopen:
FYI @tomjenkinson the build fails because "org.jboss.narayana.compensations", and the reason for that is https://github.com/wildfly/wildfly/pull/17669

mmusgrov:
> FYI @tomjenkinson the build fails because " org.jboss.modules.ModuleNotFoundException: org.jboss.narayana.compensations", and the reason for that is [wildfly/wildfly#17669](https://github.com/wildfly/wildfly/pull/17669) Meaning that compensation tests cannot be run with WildFly anymore. We might remove them or disable them (until we can test them against another AS). WDYT @mmusgrov ? See PR: #2233

Yes let's remove them rather than enabling them, note that the endpoints needed to run the tests require buy-in from other companies that implemented the WS specs.

PR #2246, opened 2024-04-22
[JBTM-3879] Xa recovery scan cursor handling

marcosgopen:
https://issues.redhat.com/browse/JBTM-3879

Patch XA recovery scan cursor handling
Add unit test to test the XAResourceWrapper and Recovery scan status

CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !LRA !DB_TESTS

### DIFF
--- a/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleScanUnitTest.java
+++ b/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleScanUnitTest.java
@@ -1,0 +1,150 @@
+/*
+   Copyright The Narayana Authors
+   SPDX-License-Identifier: Apache-2.0
+ */
+
+
+
+package com.hp.mwtests.ts.jta.recovery;
+
+import static org.junit.Assert.assertEquals;
+
+import java.util.Arrays;
+import java.util.List;
+
+import javax.transaction.xa.XAException;
+import javax.transaction.xa.XAResource;
+import javax.transaction.xa.Xid;
+
+import org.jboss.tm.XAResourceWrapper;
+import org.junit.Test;
+
+import com.arjuna.ats.arjuna.common.Uid;
+import com.arjuna.ats.internal.jta.recovery.arjunacore.JTANodeNameXAResourceOrphanFilter;
+import com.arjuna.ats.internal.jta.recovery.arjunacore.NodeNameXAResourceOrphanFilter;
+import com.arjuna.ats.internal.jta.recovery.arjunacore.XARecoveryModule;
+import com.arjuna.ats.jta.common.jtaPropertyManager;
+import com.arjuna.ats.jta.recovery.XAResourceRecoveryHelper;
+import com.arjuna.ats.jta.xa.XidImple;
+
+public class XARecoveryModuleScanUnitTest
+{
+
+    /*
+     * XAResource.TMENDRSCAN should be invoked during periodicWorkFirstPass
+     *
+     * This guarantees that no idle connections are left after periodicWorkFirstPass
+     */
+    @Test
+    public void testEndRecoveryScanIsCalled() throws Exception {
+        int orphanSafetyInterval = jtaPropertyManager.getJTAEnvironmentBean().getOrphanSafetyInterval();
+        List<String> xaRecoveryNodes = jtaPropertyManager.getJTAEnvironmentBean().getXaRecoveryNodes();
+        jtaPropertyManager.getJTAEnvironmentBean().setOrphanSafetyInterval(0);
+        jtaPropertyManager.getJTAEnvironmentBean()
+                .setXaRecoveryNodes(Arrays.asList(new String[] { NodeNameXAResourceOrphanFilter.RECOVER_ALL_NODES }));
+        XARecoveryModule xarm = new XARecoveryModule();
+        xarm.addXAResourceOrphanFilter(new JTANodeNameXAResourceOrphanFilter());
+        XaResourceWrapperImpl xaResImpl = new XaResourceWrapperImpl();
+        XAResourceRecoveryHelper xaRRH = new XAResourceRecoveryHelper() {
+
+            XAResource[] xares = new XAResource[] { xaResImpl };
+            @Override
+            public boolean initialise(String p) throws Exception {
+                return false;
+            }
+
+            @Override
+            public XAResource[] getXAResources() throws Exception {
+                return xares;
+            }
+        };
+        xarm.addXAResourceRecoveryHelper(xaRRH);
+        assertEquals(XAResource.TMNOFLAGS, xaResImpl.getTmScanStatus());
+        xarm.periodicWorkFirstPass();
+        assertEquals(XAResource.TMENDRSCAN, xaResImpl.getTmScanStatus());
+        xarm.periodicWorkSecondPass();
+        assertEquals(XAResource.TMENDRSCAN, xaResImpl.getTmScanStatus());
+        jtaPropertyManager.getJTAEnvironmentBean().setOrphanSafetyInterval(orphanSafetyInterval);
+        jtaPropertyManager.getJTAEnvironmentBean().setXaRecoveryNodes(xaRecoveryNodes);
+    }
+
+    class XaResourceWrapperImpl implements XAResourceWrapper {
+
+        @Override
+        public XAResource getResource() {
+            return null;
+        }
+
+        @Override
+        public String getProductName() {
+            return null;
+        }
+
+        @Override
+        public String getProductVersion() {
+            return null;
+        }
+
+        @Override
+        public String getJndiName() {
+            return "test";
+        }
+        Xid xid = new XidImple(new Uid());
+        int tmScanStatus = 0;
+        boolean rolledback = false;
+
+        public int getTmScanStatus() {
+            return tmScanStatus;
+        }
+
+        public void setTmScanStatus(int tmScanStatus) {
+            this.tmScanStatus = tmScanStatus;
+        }
+
+        @Override
+        public void commit(Xid xid, boolean b) throws XAException {
+        }
+
+        @Override
+        public void end(Xid xid, int i) throws XAException {
+        }
+
+        @Override
+        public void forget(Xid xid) throws XAException {
+        }
+
+        @Override
+        public int getTransactionTimeout() throws XAException {
+            return 0;
+        }
+
+        @Override
+        public boolean isSameRM(XAResource xaResource) throws XAException {
+            return false;
+        }
+
+        @Override
+        public int prepare(Xid xid) throws XAException {
+            return 0;
+        }
+
+        @Override
+        public Xid[] recover(int i) throws XAException {
+            setTmScanStatus(i);
+            return new Xid[] { xid };
+        }
+
+        @Override
+        public void rollback(Xid xid) throws XAException {
+            rolledback = true;
+        }
+        @Override
+        public boolean setTransactionTimeout(int i) throws XAException {
+            return false;
+        }
+
+        @Override
+        public void start(Xid xid, int i) throws XAException {
+        }
+    }
+}
--- a/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
+++ b/ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
@@ -400,16 +400,21 @@ public class XARecoveryModuleUnitTest
             }
         });
 
+        int oldCount = recoverCalled;
+        // on the error free path
+        // we expect periodicWorkFirstPass to make two recover calls (TMSTARTRSCAN and
+        // TMENDRSCAN),
+        // and we expect periodicWorkSecondPass to make no recover calls
         xarm.periodicWorkFirstPass();
-        assertEquals(recoverCalled, 1);
+        assertEquals(recoverCalled, oldCount + 2);
         xarm.periodicWorkSecondPass();
-        assertEquals(recoverCalled, 2);
+        assertEquals(recoverCalled, oldCount + 2);
         xarm.periodicWorkFirstPass();
-        assertEquals(recoverCalled, 3);
+        assertEquals(recoverCalled, oldCount + 4);
         xarm.periodicWorkFirstPass();
-        assertEquals(recoverCalled, 5);
+        assertEquals(recoverCalled, oldCount + 6);
         xarm.periodicWorkSecondPass();
-        assertEquals(recoverCalled, 6);
+        assertEquals(recoverCalled, oldCount + 6);
     }
 
     @Test
@@ -441,7 +446,6 @@ public class XARecoveryModuleUnitTest
                 public String getJndiName() {
                     return "test";
                 }
-
                 int count = 0;
                 Xid xid = new XidImple(new Uid());
 
@@ -478,20 +482,17 @@ public class XARecoveryModuleUnitTest
                 @Override
                 public Xid[] recover(int i) throws XAException {
                     count++;
-                    if (count == 1 || count == 5) {
-                        return new Xid[]{xid};
+                    if (count == 1 || count == 3 || count == 5) {
+                        return new Xid[] { xid }; // TMSTARTRSCAN
                     } else if (count > 5) {
-                        return new Xid[0];
+                        return new Xid[0]; // TMSENDRSCAN when count == 6 so the recovery pass succeeds
                     } else {
-                        throw new XAException();
+                        throw new XAException(); // if count is 2 or 4 ie the TMSENDRSCAN call
                     }
                 }
 
                 @Override
                 public void rollback(Xid xid) throws XAException {
-                    if (count == 1) { // This comes from the first end scan
-                        throw new XAException(XAException.XA_RETRY);
-                    }
                     rolledback = true;
                 }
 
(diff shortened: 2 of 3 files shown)


ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
```diff
@@ -400,16 +400,24 @@ public void start(Xid xid, int i) throws XAException {
             }
         });
 
+        int oldCount = recoverCalled;
+        callFirstPass(xarm, oldCount);
```
mmusgrov, line 404: We have lost some test precision here, before we asserted the actual number of recover calls but now we are just testing that the number is incrementing on the first path and I'd expect the test would fail if you called callSecondPass twice in a row. Since the expectation on the error free path is two recover calls on the first pass and no recover calls on the second pass, the following would be more maintainable:

```
        int oldCount = recoverCalled;
        callFirstPass(xarm, oldCount + 2);
        callSecondPass(xarm, oldCount + 2);
        callFirstPass(xarm, oldCount + 4);
        callFirstPass(xarm, oldCount + 6);
        callSecondPass(xarm, oldCount + 6);
```

or better still, because the following makes it obvious what the test is doing:

```
        int oldCount = recoverCalled;
        // on the error free path
        // we expect periodicWorkFirstPass to make two recover calls,
        // and we expect periodicWorkSecondPass to make no recover calls
        xarm.periodicWorkFirstPass();
        assertEquals(recoverCalled, oldCount + 2);
        xarm.periodicWorkSecondPass();
        assertEquals(recoverCalled, oldCount + 2);
        xarm.periodicWorkFirstPass();
        assertEquals(recoverCalled, oldCount + 4);
        xarm.periodicWorkFirstPass();
        assertEquals(recoverCalled, oldCount + 6);
        xarm.periodicWorkSecondPass();
        assertEquals(recoverCalled, oldCount + 6);
```

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
```diff
@@ -525,20 +533,17 @@ public XAResource[] getXAResources() throws Exception {
         // 1st pass: returns one xid (count is 1)
         xarm.periodicWorkFirstPass();
         // 2nd pass: throws an exception (count is 2)
```
mmusgrov, line 535: These comments are no longer valid because the second pass does not call recover(). Also the following comments stating the value of count are no longer valid. And instead of incrementing `recoverCount` by 2 each time sometimes it will only increment by 1 or zero if exceptions are thrown. Also it might be worthwhile using `recoverCount` instead of `count` in the dummy XAResource so that the test can continue asserting that the recover calls are called the correct number of times.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
```diff
@@ -525,20 +533,17 @@ public XAResource[] getXAResources() throws Exception {
...
-        xarm.periodicWorkSecondPass();
-        assertTrue(xarm.getContactedJndiNames().contains("test"));
```
mmusgrov, line 529: Which bit of code caused this assertion to fail?
Or have you removed it for some other reason?

mmusgrov (changes requested):

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleScanUnitTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+import com.arjuna.ats.arjuna.common.Uid;
+import com.arjuna.ats.internal.jta.recovery.arjunacore.JTANodeNameXAResourceOrphanFilter;
+import com.arjuna.ats.internal.jta.recovery.arjunacore.NodeNameXAResourceOrphanFilter;
+import com.arjuna.ats.internal.jta.recovery.arjunacore.XARecoveryModule;
+import com.arjuna.ats.jta.common.jtaPropertyManager;
+import com.arjuna.ats.jta.recovery.XAResourceRecoveryHelper;
+import com.arjuna.ats.jta.xa.XidImple;
+
+public class XARecoveryModuleScanUnitTest
```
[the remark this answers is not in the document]
marcosgopen: Thanks, I created it @mmusgrov : https://github.com/agroal/agroal/pull/103

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
marcosgopen: I think this test is now very confusing and may be deleted. However I will try to change it.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
marcosgopen: It fails because now the counts are not aligned to the logic implemented in the XAResourceWrapper. Let me fix that.

marcosgopen:
I tried to change the unit tests with the new commit as you suggested @mmusgrov. WDYT? 
I'll then squash the commits.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
```diff
@@ -518,28 +521,25 @@ public XAResource[] getXAResources() throws Exception {
             }
         });
 
-
-        // The first two recovery cycles do nothing with the resource (because phase two is getting the exception)
+        // During the periodicWorkFirstPass recover method is called twice
```
mmusgrov, line 524: The logic of this test is no longer testing the same behaviour, in particular the final assert `assertFalse(rolledback);` has been switched, the original test did `assertTrue(rolledback);` which is what we want because on the third recovery cycle there is no exception and it is only then that we detect that it's an orphan (because the resource says it doesn't know about the xid on the 3rd cycle) and recovery rolls it back and hence `assertTrue(rolledback);` will pass. The code you need, in addition to the resource recover method change, is as follows (note that I have included calls to periodicWorkSecondPass otherwise the revovery module thinks it is `BETWEEN_PASSES` and runs different logic (but we want the test to test the standard logic):

```

        // The first two recovery cycles do nothing with the resource (because phase one is getting the exception)
        // When count reaches 6 it sees that the xid has gone and presumes abort so calls rollback and hence
        // assertTrue(rolledback) passes
        xarm.periodicWorkFirstPass(); // recoverCount is now 2 (and the TMENDRSCAN will have thrown an exception)
        xarm.periodicWorkSecondPass(); // complete the cycle
// I didn't debug why the jndi names assert is failing (I think you figured it out though on the first review)
//        assertTrue(xarm.getContactedJndiNames().contains("test"));
        assertFalse(rolledback); // since recover threw an exception there is no rollback yet

        xarm.periodicWorkFirstPass(); // recoverCount count is now 4 (and the TMENDRSCAN will have thrown an exception)
        xarm.periodicWorkSecondPass(); // complete the cycle
//        assertFalse(xarm.getContactedJndiNames().contains("test"));
        assertFalse(rolledback); // since recover threw an exception there is no rollback yet

        xarm.periodicWorkFirstPass(); // recoverCount count is now 6. On the TMENDRSCAN it will have returned 1 xid and none on the TMSENDRSCAN
        xarm.periodicWorkSecondPass(); // complete the cycle
//        assertTrue(xarm.getContactedJndiNames().contains("test"));
        // since there was a successful recovery cycle (ie no exceptions) which returned an empty list of xids
        // it should have rolled back as an orphan
        assertTrue(rolledback);
```

mmusgrov (changes requested):
The main change request is the method that tests the recovery module with exceptions ie `testRecoverPassFailure`.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
```diff
@@ -57,6 +57,7 @@
 public class XARecoveryModuleUnitTest
 {
     private boolean rolledback;
+    private int recoverCount = 0;
```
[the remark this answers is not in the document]
mmusgrov: I'm also doubting that recoverCount needs to be declared as an instance variable: the asserts for the value don't add anything since recoverCount is internal to how the test operates, the code comments about recoverCount should be sufficient.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
```diff
@@ -477,10 +482,11 @@ public int prepare(Xid xid) throws XAException {
...
-                    count++;
-                    if (count == 1 || count == 5) {
+                    recoverCount++;
+                    if (recoverCount == 1 || recoverCount == 2) {
                         return new Xid[]{xid};
-                    } else if (count > 5) {
+                    }
+                    else if (recoverCount > 3) {
                         return new Xid[0];
                     } else {
```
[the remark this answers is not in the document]
mmusgrov: BTW in my other comment I suggested an alternative implementation of the recover method.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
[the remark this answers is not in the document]
marcosgopen: Agreed, so I will restore the previous variable (count) and just rename it to 'recoverCount'.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
```diff
@@ -477,10 +482,11 @@ public int prepare(Xid xid) throws XAException {
 
                 @Override
                 public Xid[] recover(int i) throws XAException {
-                    count++;
-                    if (count == 1 || count == 5) {
+                    recoverCount++;
+                    if (recoverCount == 1 || recoverCount == 2) {
```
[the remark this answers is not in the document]
marcosgopen: Thanks @mmusgrov , PR updated

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
```diff
@@ -477,10 +482,11 @@ public int prepare(Xid xid) throws XAException {
...
                         return new Xid[]{xid};
-                    } else if (count > 5) {
+                    }
```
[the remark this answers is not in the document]
marcosgopen: Thanks, updated.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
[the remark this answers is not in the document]
marcosgopen: Thanks, updated

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
```diff
@@ -518,28 +521,25 @@ public XAResource[] getXAResources() throws Exception {
             }
         });
 
-
-        // The first two recovery cycles do nothing with the resource (because phase two is getting the exception)
+        // During the periodicWorkFirstPass recover method is called twice
         // When count reaches 6 it sees that the xid has gone and presumes abort so calls rollback and hence assertTrue(rolledback) passes
 
-        // 1st pass: returns one xid (count is 1)
+        // returns one xid (count is 2)
```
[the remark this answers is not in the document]
marcosgopen: I agree, that it quite confusing. I updated the comments as you suggested, thanks.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
marcosgopen: Updated, thanks!

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
marcosgopen: the assertion failed because now the exception is thrown during the  periodicWorkFirstPass, where the list of Jndi is cleaned: see https://github.com/jbosstm/narayana/blob/main/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java#L187

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
```diff
@@ -441,8 +446,7 @@ public String getProductVersion() {
                 public String getJndiName() {
                     return "test";
                 }
-
-                int count = 0;
+                int recoverCount = 0;
```
mmusgrov, line 449: Now that you have gone back to the original use of this variable I think it should keep the original name - in the future it will make code diffs simpler to analyse.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
```diff
@@ -477,21 +481,18 @@ public int prepare(Xid xid) throws XAException {
...
-                    if (count == 1 || count == 5) {
-                        return new Xid[]{xid};
-                    } else if (count > 5) {
-                        return new Xid[0];
-                    } else {
-                        throw new XAException();
+                    recoverCount++;
+                    if (recoverCount == 1 || recoverCount == 3 || recoverCount == 5) {
+                        return new Xid[] { xid }; // TMSTARTRSCAN
+                    }else if (recoverCount > 5) {
```
mmusgrov, line 487: The code reformat has dropped the space between `}` and `else` which makes it inconsistent with the rest of the code in this file.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
```diff
@@ -519,26 +520,31 @@ public XAResource[] getXAResources() throws Exception {
...
-        assertFalse(rolledback);
-        // 1st pass: returns an empty list of xids (count is 5)
-        xarm.periodicWorkFirstPass();
-        // 2nd pass: returns an empty list of xids (count is 6)
-        xarm.periodicWorkSecondPass();
+        assertFalse(rolledback); // since recover threw an exception there is no rollback yet
+        xarm.periodicWorkFirstPass(); // recoverCount count is now 4 (and the TMENDRSCAN will have thrown an
+                                      // exception)
+        xarm.periodicWorkSecondPass(); // complete the cycle
+        // 'test' was cleaned from the list before the exception is thrown
```
mmusgrov, line 538: Nice, thanks

mmusgrov (changes requested):
I had two minor change requests but the PR is looking good.

Can you also squash the four commits for the test changes - in an ideal world we'd want the tests to be with the functional change but I think it is useful taking Jonathan's commit as is.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
marcosgopen: Thanks, updated

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
marcosgopen: Thanks, updated.

marcosgopen:
I have created the [JBTM](https://issues.redhat.com/browse/JBTM-3879) (feel free to add comments/update the description) for it and squash the commits as suggested.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
```diff
@@ -441,7 +446,6 @@ public String getProductVersion() {
                 public String getJndiName() {
                     return "test";
                 }
-
                 int count = 0;
```
mmusgrov, line 449: Thanks for changing it back to count. Quite a few comments still refer to `recoverCount`, will you change those to count too please.

mmusgrov (changes requested):

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/XARecoveryModuleUnitTest.java
marcosgopen: Sure, I forgot about that. Updated, thanks

mmusgrov (approved):

PR #2248, opened 2024-04-25
Refactor BeanPopulator to work with Quarkus native mode

zhfeng:

- [ ] Pull Request contains a link to the JIRA issue(s) and that they contain sufficient information for the reviewer to be able to gauge whether or not the proposed changes correctly address the issue

Pull requests build and run with JDK11 and JDK17. Add the prefix `!` to disable these defaults. Include the text `JDK21` to build and run with JDK21


mmusgrov:
@zhfeng So this PR tells graalvm not to initialize our properties at build time. graalvm provides "two command line flags to explicitly specify when a class should be initialized: --initialize-at-build-time and --initialize-at-run-time". Why can users not use those?

zhfeng:
@mmusgrov 
> So this PR tells graalvm not to initialize our properties at build time

No, these propreties are still initilize at build time. But when the application starts up, we use `@RecomputeFieldValue` to set `beanInstances` with `null`. And at the same time, we set `TransactionImple` and `LastResourceRecord` to initilize at runtime, then it will force to create a new Map and reload the properties during initlize `LAST_RESOURCE_OPTIMISATION_INTERFACE`

```java
private static final Class LAST_RESOURCE_OPTIMISATION_INTERFACE = jtaPropertyManager.getJTAEnvironmentBean()
            .getLastResourceOptimisationInterface();
```

Now `LAST_RESOURCE_OPTIMISATION_INTERFACE` can get the right value what we pass by command line.

zhfeng:
The orignal codes in [BeanPopulator.java](https://github.com/jbosstm/narayana/blob/main/common/classes/com/arjuna/common/internal/util/propertyservice/BeanPopulator.java#L32-L39)
```java
public static <T> T getDefaultInstance(Class<T> beanClass) throws RuntimeException {
        T instance = (T) beanInstances.get(beanClass.getName());

        if (instance != null)
           return instance;

        return getNamedInstance(beanClass, null, null);
    }
```

beacause `beanInstances` is initlized at build time. So it always return the `instance` in the Map.

common/classes/com/arjuna/common/internal/util/propertyservice/BeanPopulator.java
```diff
@@ -29,8 +29,12 @@ public class BeanPopulator
 {
     private static final ConcurrentMap<String, Object> beanInstances = new ConcurrentHashMap<String, Object>();
 
+    private static ConcurrentMap<String, Object> getBeanInstances() {
+        return beanInstances;
+    }
```
zhfeng, line 34: @mmusgrov It should only introduce `getBeanInstances` method in `narayana` and I think we can use `@Substitute` in Quarkus extension to rewrite this method to check the `null` value.

In `narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/graal/PropertiesFactorySubstitution.java`
```java
    @Substitute
    private static ConcurrentMap<String, Object> getBeanInstances() {
        if (beanInstances == null) {
            synchronized (BeanPopulator.class) {
                if (beanInstances == null) {
                    beanInstances = new ConcurrentHashMap<String, Object>();
                }
            }
        }
        return beanInstances;
    }

```
Does it make sense and could be accepeted by Narayana?

zhfeng:
It seems that only need to remove the object in `beanInstances` but not the whole map. I will try to find an another way. So please hold on this PR.

PR #2249, opened 2024-04-25
BUILD SCRIPT ONLY: Remove the restat-web-*.war if it exists

mmusgrov:
This change adds a small correction to commit https://github.com/jbosstm/narayana/commit/ec13ede6937aaecf8ed6923ca2f5c06bb5924eb7 such that the build script won't fail if the target file to be removed does not exist. 

NO_TEST
!CORE !TOMCAT !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !LRA !DB_TESTS !mysql !db2 !postgres !oracle


mmusgrov:
I'm closing since the error is that JBOSS_HOME isn't set so this PR will still fail.

PR #2259, opened 2024-06-04
JBTM-3883 Include an option to configure the nodeIdentifer bytes

mmusgrov:
CORE AS_TESTS RTS JACOCO XTS QA_JTA QA_JTS_OPENJDKORB !PERFORMANCE LRA DB_TESTS mysql db2 postgres oracle
!JDK17


mmusgrov:
FYI @marcosgopen this is the fix for the quarkus workaround that introduced a reduction in the entropy/randomness of the hashing function they are using for the node id.

marcosgopen (approved):
Thanks @mmusgrov for your PR.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/CoreEnvironmentBean.java
```diff
@@ -105,6 +106,25 @@ public void setNodeIdentifier(String nodeIdentifier) throws CoreEnvironmentBeanE
...
 
     	this.nodeIdentifier = nodeIdentifier;
+        this.nodeIdentifierBytes = nodeIdentifier.getBytes(StandardCharsets.UTF_8);
+    }
+
+    public byte[] getNodeIdentifierBytes() {
+        return this.nodeIdentifierBytes;
+    }
+
+    public void setNodeIdentifier(byte[] bytes) throws CoreEnvironmentBeanException
```
jmfinelli, line 116: Wouldn't make sense to use this method as the default one? To be more specific, from `public void setNodeIdentifier(String nodeIdentifier)`, we could invoke directly the new `public void setNodeIdentifier(byte[] bytes)`. Moreover, it would be great if we can specify that `public void setNodeIdentifier(String nodeIdentifier)` uses UTF8 in its javadoc. WDYT @mmusgrov?

mmusgrov: I know that you see two lines repeated in each method but I want to minimise the change set. The goal of the PR is to give the caller the option of explicitly setting the bytes for the purpose of encoding Xids in order to avoid the problems caused by variable width encoding. On the other point, I don't want to force them to use a specific a charset because that is just an encoding and they can decode it in any character set that suits them.

jmfinelli: I see your point :-)

jmfinelli (approved):

PR #2261, opened 2024-06-05
JBTM-3883 Add javadoc and more checks when setting the nodeIdentifer …

mmusgrov:

CORE !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !LRA !DB_TESTS !mysql !db2 !postgres !oracle

This PR follows up on #2259 and adds more input validation and javadoc to CoreEnvironmentBean#setNodeIdentifier(byte[] nodeIdentifierBytes). It also adds an extra test for the new config setting (see ArjPropertyManagerTest.java).

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/CoreEnvironmentBean.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/CoreEnvironmentBean.java
@@ -6,6 +6,8 @@ package com.arjuna.ats.arjuna.common;
 
 import java.io.File;
 import java.nio.charset.StandardCharsets;
+
+import com.arjuna.ats.arjuna.coordinator.TxControl;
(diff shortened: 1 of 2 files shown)


jmfinelli (approved):

mmusgrov:
Thanks @jmfinelli . I added a call from the byte form of the setter to node form of the setter but not the other way around because the byte form needs to do some extra things that the String form can't do.

jmfinelli (approved):

mmusgrov:
> Thanks @jmfinelli . I added a call from the byte form of the setter to node form of the setter but not the other way around because the byte form needs to do some extra things that the String form can't do.

@jmfinelli  I removed the call and validated the length directly in the byte form of the setter.
I also added quite a lot of extra javadoc since these config methods are somewhat subtle.

PR #2262, opened 2024-06-13
BUILD SCRIPT ONLY: Allow the controlling of the nexus plugin separate…

tomjenkinson:
CORE !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !LRA !DB_TESTS mysql db2 postgres oracle
!JDK17

### DIFF
--- a/narayana-release-process.sh
+++ b/narayana-release-process.sh
@@ -204,7 +204,7 @@ then
   echo 1>&2 Could not install narayana
   exit
 fi
-./build.sh clean deploy -Dmaven.repo.local=${PWD}/localm2repo -DskipTests -gs ~/.m2/settings.xml -Dorson.jar.location=$ORSON_PATH -Prelease,community
+./build.sh clean deploy -Dmaven.repo.local=${PWD}/localm2repo -DskipTests -gs ~/.m2/settings.xml -Dorson.jar.location=$ORSON_PATH -Prelease,community -DskipNexusStagingDeployMojo=false
 if [[ $? != 0 ]]
 then
   echo 1>&2 Could not deploy narayana to nexus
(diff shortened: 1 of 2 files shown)


mmusgrov (approved):
Looks good.

So we can now test a release without necessarily deploying to nexus?

mmusgrov:
Thanks, I assume that you had a use case in mind!

marcosgopen:
> With this we can control the deployment to nexus separately so by default a `./build.sh deploy [...]` will not deploy to Nexus. To deploy to nexus we just do `./build.sh deploy -DskipNexusStagingDeployMojo=false [...]`

@tomjenkinson I assume you mean `./build.sh deploy  -DskipMavenDeployMojo=true`

marcosgopen:
> I don't think there is a `skipNexusDeployMojo`? I am going to guess you are referring to `skipMavenDeployMojo`?
> 
> `skipMavenDeployMojo` controls the `maven-deploy-plugin` and `skipNexusStagingDeployMojo` controls the `nexus-staging-maven-plugin`

Thanks Tom, yes I was referring to skipMavenDeployMojo. Ok, I see.

mmusgrov:
> With this we can control the deployment to nexus separately so by default a `./build.sh deploy [...]` will not deploy to Nexus. To deploy to nexus we just do `./build.sh deploy -DskipNexusStagingDeployMojo=false [...]`

Yes the property behaviour is clear, but when would we not want to deploy to nexus, don't we need to do that first in order to have the artifacts synced to maven central.

mmusgrov:
> We do want to be able to release to Nexus, but it's not guaranteed we want to do that every time we do mvn deploy (e.g to deploy to a local file system). This change allows for that flexibility.

Why did you need to deploy to your local filesystem, were you testing the veracity of the release scripts, or was it for other reasons?

PR #2269, opened 2024-07-01
Define how to identify that a change must be in a major version and c…

tomjenkinson:
NO_TEST

(has some relation to https://issues.redhat.com/browse/JBTM-3856)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -128,10 +128,12 @@ Also don't forget the documentation (reference documentation for features, javad
 
 ## Update the issue with the correct release information:
 
+If the change would result in behaviour in Narayana that is incompatible with the current release stream for Narayana, then the relevant JIRA issue `Fix Version/s` must be set to the `<version+1>.next`. For example if the root pom.xml is currently `<version>7.0.3.Final-SNAPSHOT</version>` then the JIRA issue should have a `Fix Version/s` of `8.next` added.
+
 When the github Pull Request has passed all relevant CI checks and has been Approved by a reviewer the code can be merged. If you don't have permission to do this then ping one of the team who will then merge it. Once merged the issue must be updated in the issue tracker (if you don't have permission then a team member will do this):
 
 1. Press the `Pull Request Closed` button (the status will automatically be set to `Done`)
-2. Set the `Fix Version/s` field to `6.next` (if the fix is targeted for a different major version, such as 7.next use that instead).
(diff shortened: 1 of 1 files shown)


marcosgopen (approved):
Thanks @tomjenkinson for your PR

jmfinelli (approved):

mmusgrov (changes requested):
The text needs to be in section that covers JIRA, ie https://github.com/jbosstm/narayana/blob/main/CONTRIBUTING.md#update-the-issue-with-the-correct-release-information

Also the text that refers to `8.later` should be `8.next` since the `next` suffix is referring to the next release whereas the `later` suffix is saying that we don't know which release it is targeted for except that it will be in some 8.x version but not necessarily the next one.

CONTRIBUTING.md
```diff
@@ -128,10 +128,12 @@ Also don't forget the documentation (reference documentation for features, javad
 
 ## Update the issue with the correct release information:
 
+If the change would result in behaviour in Narayana that is incompatible with the current release stream for Narayana, then the relevant JIRA issue `Fix Version/s` must be set to the `<version+1>.next`. For example if the root pom.xml is currently `<version>7.0.3.Final-SNAPSHOT</version>` then the JIRA issue should have a `Fix Version/s` of `8.next` added.
```
mmusgrov, line 131: Remark if I read this example literally then the fix version would need to be set to `8.0.3.Final-SNAPSHOT` but I think contributors will understand your intention.

CONTRIBUTING.md
```diff
@@ -128,10 +128,12 @@ Also don't forget the documentation (reference documentation for features, javad
...
+
 When the github Pull Request has passed all relevant CI checks and has been Approved by a reviewer the code can be merged. If you don't have permission to do this then ping one of the team who will then merge it. Once merged the issue must be updated in the issue tracker (if you don't have permission then a team member will do this):
 
 1. Press the `Pull Request Closed` button (the status will automatically be set to `Done`)
-2. Set the `Fix Version/s` field to `6.next` (if the fix is targeted for a different major version, such as 7.next use that instead).
+2. If the `Fix Version/s` is not set to `<version+1>.next`, set the `Fix Version/s` field to `<version>.next`
```
mmusgrov, line 136: Remark: So this awkward text is accommodating the case where the issue is a blocker.

mmusgrov (approved):
I'm approving because I know what the process is already but I think we should ask someone who is not familiar with the process to review this one, not sure who we'd ask though. But since I've approved it you are at liberty to merge it.

PR #2271, opened 2024-07-04
JBTM-3838 fix with NOTA case

marcosgopen:

Move the default case under the XAException.XAER_NOTA case to avoid conflicts with the XAException.XAER_RMFAIL 'break' 

CORE AS_TESTS !RTS JACOCO XTS QA_JTA QA_JTS_OPENJDKORB !PERFORMANCE !LRA DB_TESTS


mmusgrov (changes requested):
The issue link/reference does not tally with the PR: the issue is about release scripts and the PR is about removing the record from the store.

marcosgopen:
> The issue link/reference does not tally with the PR: the issue is about release scripts and the PR is about removing the record from the store.

Thanks @mmusgrov, I confused JBTM-3888 with JBTM-3838. I will close this PR and open up another one.

PR #2272, opened 2024-07-04
JBTM-3838 fix with NOTA case

marcosgopen:

Move the default case under the XAException.XAER_NOTA case to avoid conflicts with the XAException.XAER_RMFAIL 'break'

CORE AS_TESTS !RTS JACOCO XTS QA_JTA QA_JTS_OPENJDKORB !PERFORMANCE !LRA DB_TESTS

### DIFF
--- a/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
+++ b/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
@@ -397,20 +397,20 @@ public class XAResourceRecord extends com.arjuna.ArjunaOTS.OTSAbstractRecordPOA
 						case XAException.XA_RBTIMEOUT:
 							destroyState();
 							break;
-						case XAException.XAER_NOTA:
+	                    case XAException.XAER_RMFAIL:
+	                        if (_prepared) {
+	                           // We do not want to delete the state on disk because that would prevent the recovery
+	                           // manager from reattempting recovery and XAER_RMFAIL out of xa_rollback is transient
+	                           // failure
+	                           destroyState = false;
+	                        }
+	                        break;
+	                    case XAException.XAER_NOTA:
 							if (_phaseTwoStarted) {
 								// rolled back previously and recovery completed
 								destroyState();
 								break;
 							}
-						case XAException.XAER_RMFAIL:
-							if (_prepared) {
-								// We do not want to delete the state on disk because that would prevent the recovery
-								// manager from reattempting recovery and XAER_RMFAIL out of xa_rollback is transient
-								// failure
-								destroyState = false;
-							}
-							break;
 						default:
 							destroyState();
 


ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -397,27 +397,25 @@ public void rollback() throws org.omg.CORBA.SystemException,
 						case XAException.XA_RBTIMEOUT:
 							destroyState();
 							break;
-						case XAException.XAER_NOTA:
```
tomjenkinson, line 400: I don't think it is this  because NOTA actually goes down a different route due to https://github.com/jbosstm/narayana/pull/2272/files#diff-c0091e065ebb4dea411f68b3ea81f32d4b0ebe2b1349c2876073d576fdc505b8R354

The XAER_NOTA seems like it should not be in here therefore.

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -397,27 +397,25 @@ public void rollback() throws org.omg.CORBA.SystemException,
...
-								throw new org.omg.CosTransactions.HeuristicHazard();
-							else
-								throw new UNKNOWN();
+                        case XAException.XAER_NOTA:
+                            if (_phaseTwoStarted) {
+                                // rolled back previously and recovery completed
+                                destroyState();
+                                break;
+                            }
+                            destroyState();
```
mmusgrov, line 406: The only change introduced by the PR are these 5 lines which have been moved from the default arm into the XAER_NOTA arm of the switch (the reformatting introduced in the PR made that difficult to spot and the default are deleted).

But I think the default arm needs be present (see my previous comment) and, if you agree, then you just need to let the `case XAException.XAER_NOTA` arm of the switch fall through into the default arm, ie the full patch would be:

```
						case XAException.XAER_NOTA:
							if (_phaseTwoStarted) {
								// rolled back previously and recovery completed
								destroyState();
								break;
							}
							// FALL THROUGH
						default:
                                                        etc ...
```

mmusgrov (changes requested):
Please can you also provide a rationale/explanation of why the patch correctly resolves the JBTM issue.

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -397,27 +397,25 @@ public void rollback() throws org.omg.CORBA.SystemException,
...
-							}
-						case XAException.XAER_RMFAIL:
-							if (_prepared) {
-								// We do not want to delete the state on disk because that would prevent the recovery
-								// manager from reattempting recovery and XAER_RMFAIL out of xa_rollback is transient
-								// failure
-								destroyState = false;
-							}
-							break;
-						default:
```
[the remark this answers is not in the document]
marcosgopen: Thanks for your comment @mmusgrov , I reintroduced the default case.

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -397,27 +397,25 @@ public void rollback() throws org.omg.CORBA.SystemException,
...
-							destroyState();
-
-							if (_prepared)
-								throw new org.omg.CosTransactions.HeuristicHazard();
-							else
-								throw new UNKNOWN();
+                        case XAException.XAER_NOTA:
```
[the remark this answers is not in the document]
marcosgopen: Thanks @mmusgrov , I updated the PR.

marcosgopen:
As commented in the issue https://issues.redhat.com/browse/JBTM-3889 and on Zulip (https://narayana.zulipchat.com/#narrow/stream/323715-developers/topic/JBTM-3879.20and.207.2E0.2E2.2EFinal) this PR is not needed anymore so I am going to close it now. cc @mmusgrov @graben

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -397,20 +397,20 @@ public void rollback() throws org.omg.CORBA.SystemException,
...
+	                           destroyState = false;
+	                        }
+	                        break;
+	                    case XAException.XAER_NOTA:
 							if (_phaseTwoStarted) {
 								// rolled back previously and recovery completed
 								destroyState();
 								break;
 							}
-						case XAException.XAER_RMFAIL:
```
mmusgrov, line 406: @marcosgopen why did you move this arm of the switch statement?

marcosgopen:
> I think we should close this pull request as I don't think it is proven to be necessary to reassess the JBTM-3838 implementation. Is that OK with you @marcosgopen ?

Agreed, thanks for your review. We can reopen it if we find proof there is actually something wrong with the current implementation.

PR #2277, opened 2024-07-18
[JBTM-3899] Fixed getNumberOfInflightTransactions

jmfinelli:
[JBTM-3899](https://issues.redhat.com/browse/JBTM-3899)

CORE !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !LRA !DB_TESTS

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
@@ -39,9 +39,29 @@ public class ActionManager
 		_allActions.remove(id);
 	}
 
-    public int getNumberOfInflightTransactions()
+	/**
+	 * <p>
+	 *     This method works out the number of in-flight transactions currently in
+	 *     the system.
+	 * <p>
+	 *     Note: The definition of in-flight (a.k.a. active) transactions can be
+	 *     found in the OMG OTS specification: "[Active is defined as] the state
+	 *     of a transaction when processing is in progress and completion of the
+	 *     transaction has not yet commenced." In addition to the OMG OTS
+	 *     definition, transactions that are marked as abort-only are not
+	 *     considered active.
+	 *     In Arjuna, these considerations are distilled into the following:
+	 *     in-flight transactions are those BasicActions with status() equal to
+	 *     ActionStatus.RUNNING.
+	 *
+	 * @return the number of in-flight transactions currently in the system
+	 */
+	  public int getNumberOfInflightTransactions()
     {
-        return _allActions.size();
+        return (int) _allActions.entrySet()
+          .parallelStream()
+          .filter(entry -> entry.getValue().status() == ActionStatus.RUNNING)
+          .count();
     }
 	
 	private ActionManager()
--- a/ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/TxStatsUnitTest.java
+++ b/ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/TxStatsUnitTest.java
@@ -5,20 +5,25 @@
 
 package com.hp.mwtests.ts.arjuna.atomicaction;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
-
 import java.io.PrintWriter;
 import java.io.StringWriter;
 
-import org.junit.Test;
-
 import com.arjuna.ats.arjuna.AtomicAction;
 import com.arjuna.ats.arjuna.common.arjPropertyManager;
+import com.arjuna.ats.arjuna.coordinator.TwoPhaseOutcome;
 import com.arjuna.ats.arjuna.coordinator.TxStats;
+import com.hp.mwtests.ts.arjuna.resources.BasicRecord;
+import org.junit.Before;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
 
 public class TxStatsUnitTest
 {
+    @Before
+    public void before() {
+        Assertions.assertTrue(TxStats.enabled());
+    }
+
     @Test
     public void test() throws Exception
     {
@@ -31,9 +36,13 @@ public class TxStatsUnitTest
             
             A.begin();
             B.begin();
+
+            Assertions.assertEquals(2, TxStats.getInstance().getNumberOfInflightTransactions());
             
             B.commit();
             A.commit();
+
+            Assertions.assertEquals(0, TxStats.getInstance().getNumberOfInflightTransactions());
         }
         
         for (int i = 0; i < 100; i++)
@@ -41,33 +50,36 @@ public class TxStatsUnitTest
             AtomicAction A = new AtomicAction();
             
             A.begin();
+
+            Assertions.assertEquals(1, TxStats.getInstance().getNumberOfInflightTransactions());
             
             A.abort();
+
+            Assertions.assertEquals(0, TxStats.getInstance().getNumberOfInflightTransactions());
         }
-        
-        AtomicAction B = new AtomicAction();
-        
-        B.begin();
-        
-        assertTrue(TxStats.enabled());
-        assertEquals(100, TxStats.getInstance().getNumberOfAbortedTransactions());
-        assertEquals(100, TxStats.getInstance().getNumberOfApplicationRollbacks());
-        assertEquals(200, TxStats.getInstance().getNumberOfCommittedTransactions());
-        assertEquals(0, TxStats.getInstance().getNumberOfHeuristics());
-        assertEquals(1, TxStats.getInstance().getNumberOfInflightTransactions());
-        assertEquals(100, TxStats.getInstance().getNumberOfNestedTransactions());
-        assertEquals(0, TxStats.getInstance().getNumberOfResourceRollbacks());
-        assertEquals(0, TxStats.getInstance().getNumberOfTimedOutTransactions());
-        assertEquals(301, TxStats.getInstance().getNumberOfTransactions());
+
+        Assertions.assertEquals(100, TxStats.getInstance().getNumberOfAbortedTransactions());
+        Assertions.assertEquals(100, TxStats.getInstance().getNumberOfApplicationRollbacks());
+        Assertions.assertEquals(200, TxStats.getInstance().getNumberOfCommittedTransactions());
+        Assertions.assertEquals(0, TxStats.getInstance().getNumberOfHeuristics());
+        Assertions.assertEquals(100, TxStats.getInstance().getNumberOfNestedTransactions());
+        Assertions.assertEquals(0, TxStats.getInstance().getNumberOfResourceRollbacks());
+        Assertions.assertEquals(0, TxStats.getInstance().getNumberOfTimedOutTransactions());
+        Assertions.assertEquals(300, TxStats.getInstance().getNumberOfTransactions());
         
         PrintWriter pw = new PrintWriter(new StringWriter());
         
         TxStats.getInstance().printStatus(pw);
 
-        // https://issues.jboss.org/browse/JBTM-2643
         for (int i = 0; i < 100; i++) {
             final AtomicAction D = new AtomicAction();
             D.begin();
+
+            D.add(new StatsResource());
+            D.add(new StatsResource());
+
+            Assertions.assertEquals(1, TxStats.getInstance().getNumberOfInflightTransactions());
+
             Thread t1 = new Thread() {
                 @Override
                 public void run() {
@@ -84,10 +96,80 @@ public class TxStatsUnitTest
             t2.start();
             t1.join();
             t2.join();
+            D.removeThread();
         }
 
-        assertEquals(200, TxStats.getInstance().getNumberOfAbortedTransactions());
-        assertEquals(200, TxStats.getInstance().getNumberOfApplicationRollbacks());
-        assertEquals(401, TxStats.getInstance().getNumberOfTransactions());
+        Assertions.assertEquals(200, TxStats.getInstance().getNumberOfAbortedTransactions());
+        Assertions.assertEquals(200, TxStats.getInstance().getNumberOfApplicationRollbacks());
+        Assertions.assertEquals(400, TxStats.getInstance().getNumberOfTransactions());
+
+        for (int i = 0; i < 100; i++)
+        {
+            AtomicAction A = new AtomicAction();
+
+            A.begin();
+
+            // Let's add some resources
+            A.add(new StatsResource());
+            A.add(new StatsResource());
+
+            Assertions.assertEquals(1, TxStats.getInstance().getNumberOfInflightTransactions());
+
+            A.abort();
+
+            Assertions.assertEquals(0, TxStats.getInstance().getNumberOfInflightTransactions());
+        }
+
+        Assertions.assertEquals(300, TxStats.getInstance().getNumberOfAbortedTransactions());
+        Assertions.assertEquals(300, TxStats.getInstance().getNumberOfApplicationRollbacks());
+        Assertions.assertEquals(500, TxStats.getInstance().getNumberOfTransactions());
+
+        for (int i = 0; i < 100; i++)
+        {
+            AtomicAction A = new AtomicAction();
+
+            A.begin();
+
+            A.add(new StatsResource());
+            A.add(new StatsResource());
+
+            Assertions.assertEquals(1, TxStats.getInstance().getNumberOfInflightTransactions());
+
+            A.preventCommit();
+            A.commit();
+
+            Assertions.assertEquals(0, TxStats.getInstance().getNumberOfInflightTransactions());
+        }
+
+        Assertions.assertEquals(400, TxStats.getInstance().getNumberOfAbortedTransactions());
+        Assertions.assertEquals(300, TxStats.getInstance().getNumberOfApplicationRollbacks());
+        Assertions.assertEquals(600, TxStats.getInstance().getNumberOfTransactions());
+    }
+
+    @Test
+    public void testRollbackOnlyStat() {
+        AtomicAction A = new AtomicAction();
+        A.begin();
+        Assertions.assertEquals(1, TxStats.getInstance().getNumberOfInflightTransactions());
+        A.preventCommit();
+        Assertions.assertEquals(0, TxStats.getInstance().getNumberOfInflightTransactions());
+        A.commit();
+    }
+
+    private static class StatsResource extends BasicRecord {
+        
+        @Override
+        public int topLevelPrepare() {
+            // TxStats.getNumberOfInflightTransactions() shouldn't count pending transactions 
+            Assertions.assertEquals(0, TxStats.getInstance().getNumberOfInflightTransactions());
+            return TwoPhaseOutcome.PREPARE_OK;
+        }
+
+        @Override
+        public int topLevelAbort() {
+            // TxStats.getNumberOfInflightTransactions() shouldn't count aborting transactions 
+            Assertions.assertEquals(0, TxStats.getInstance().getNumberOfInflightTransactions());
+            return TwoPhaseOutcome.FINISH_OK;
+        }
     }
 }


jmfinelli:
Please, refer to the second commit (the first commit is #2275 )

mmusgrov:
The need for a change to the stat for the number of in flight transactions is a good catch. I would avoid `<= ActionStatus.PREPARING` since a preparing transaction is not in flight. Instead I'd use `== ActionStatus.RUNNING && !=ActionStatus.ABORT_ONLY ` because the definition of Active is

> The state of a transaction when processing is in progress and completion of the transaction has not yet commenced.

The ABORT_ONLY state is a wee bit nuanced - there was a [jta-dev discussion](https://www.eclipse.org/lists/jta-dev/threads.html#00260) about its meaning a while back. But do update the javadoc for the stat to indicate that we are using that definition.

The existing tests are in TxStatsUnitTest but maybe add a new test class since testing in-flight vs post prepared transactions might be tricky and make those tests harder to read, perhaps you could include a resource that delays the commit while the test asks for the number of stats or some other way.

The semantics of the stat has changed so one could argue about that in two ways:

1. It's a backwardly incompatible change
2. It's a bug fix

I would say it's a bug fix but other stakeholders may have a different view, perhaps when we are all in agreement we can give the community a heads up via zulipchat.

I think this change should be in its own issue tracker and PR (not part of the ActionManager clean up).

jmfinelli:
> The need for a change to the stat for the number of in flight transactions is a good catch. I would avoid `<= ActionStatus.PREPARING` since a preparing transaction is not in flight. Instead I'd use `== ActionStatus.RUNNING && !=ActionStatus.ABORT_ONLY ` because the definition of Active is
> 
> > The state of a transaction when processing is in progress and completion of the transaction has not yet commenced.
> 
> The ABORT_ONLY state is a wee bit nuanced - there was a [jta-dev discussion](https://www.eclipse.org/lists/jta-dev/threads.html#00260) about its meaning a while back. But do update the javadoc for the stat to indicate that we are using that definition.
> 
> The existing tests are in TxStatsUnitTest but maybe add a new test class since testing in-flight vs preparing vs post prepared transactions might be tricky and make those tests harder to read, perhaps you could include a resource that delays the commit and/or the prepare while the test asks for the number of stats or some other way.
> 
> The semantics of the stat has changed so one could argue about that in two ways:
> 
>     1. It's a backwardly incompatible change
> 
>     2. It's a bug fix
> 
> 
> I would say it's both a bug fix and an incompatible change but other stakeholders may have a different view, perhaps when we are all in agreement we can give the community a heads up via zulipchat.
> 
> I think this change should be in its own issue tracker and PR (not part of the ActionManager clean up).

Thank you very much for your review, Michael :-) I will improve this PR following your feedback

jmfinelli:
> The need for a change to the stat for the number of in flight transactions is a good catch. I would avoid `<= ActionStatus.PREPARING` since a preparing transaction is not in flight. Instead I'd use `== ActionStatus.RUNNING && !=ActionStatus.ABORT_ONLY ` because the definition of Active is
> 
> > The state of a transaction when processing is in progress and completion of the transaction has not yet commenced.
> 
> The ABORT_ONLY state is a wee bit nuanced - there was a [jta-dev discussion](https://www.eclipse.org/lists/jta-dev/threads.html#00260) about its meaning a while back. But do update the javadoc for the stat to indicate that we are using that definition.

I have updated this PR to satisfy the above requirement. Thanks @mmusgrov 
 
> The existing tests are in TxStatsUnitTest but maybe add a new test class since testing in-flight vs preparing vs post prepared transactions might be tricky and make those tests harder to read, perhaps you could include a resource that delays the commit and/or the prepare while the test asks for the number of stats or some other way.

I have created a new test in the `TxStatsUnitTest` class as I hope I managed to convey its purpose clearly enough. Please, let me know if the test needs further clarification.
 
> The semantics of the stat has changed so one could argue about that in two ways:
> 
>     1. It's a backwardly incompatible change
> 
>     2. It's a bug fix
> 
> 
> I would say it's both a bug fix and an incompatible change but other stakeholders may have a different view, perhaps when we are all in agreement we can give the community a heads up via zulipchat.

I would tend to think this PR as a bug fix. The reason is that ActionManager was returning the wrong list of in-flight transactions and this PR fixes that error. However, I see your point and I agree that someone might have used this bug in their code. WDYT to settle this discussion by filling the "release notes" field in JBTM-3899?

> I think this change should be in its own issue tracker and PR (not part of the ActionManager clean up).

I have rebased this PR. It should be alright now

jmfinelli:
Thanks for your review @marcosgopen 

@mmusgrov I added a release note to JBTM-3899.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/TxStatsUnitTest.java
```diff
@@ -5,18 +5,21 @@
...
 import com.arjuna.ats.arjuna.coordinator.TxStats;
+import com.hp.mwtests.ts.arjuna.resources.BasicRecord;
+import org.jboss.byteman.contrib.bmunit.BMScript;
+import org.jboss.byteman.contrib.bmunit.BMUnitConfig;
+import org.jboss.byteman.contrib.bmunit.WithByteman;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
 
+@WithByteman
+@BMUnitConfig(debug=true)
```
mmusgrov, line 22: I don't agree with the use of byteman for this since it isn't required. Just begin the transaction, assert the number of inflight transactions and then end it and assert the number of inflight transactions again.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/TxStatsUnitTest.java
```diff
@@ -86,8 +84,42 @@ public void run() {
...
+    @Test
+    @BMScript("TxStatsUnitTest/txStatsUnitTest")
+    public void inFlightTransactionsTest() {
+
+        // There shouldn't be any in-flight transaction now
+        Assertions.assertEquals(0, TxStats.getInstance().getNumberOfInflightTransactions());
+
+        createRendezvous();
+
+        Thread committingThread = new Thread(() -> {
```
mmusgrov, line 101: I don't agree with creating transactions in a separate thread to test the stat. I suspect that the reason you went for this option is that the existing test in this test class is leaving a transaction associated with the thread used to run your test. If you had created a separate class for testing the inflight transactions stat you would not have ran into this problem. The following test should be sufficient for testing the stat, which is easy to understand whereas the one in the PR is difficult to follow:

```
    @Test
    public void test2() {
        Assertions.assertNull(BasicAction.Current());

        Assertions.assertEquals(0, TxStats.getInstance().getNumberOfInflightTransactions());

        AtomicAction txn = new AtomicAction();

        txn.begin();

        Assertions.assertEquals(1, TxStats.getInstance().getNumberOfInflightTransactions());

        txn.commit();

        Assertions.assertEquals(0, TxStats.getInstance().getNumberOfInflightTransactions());
    }
```

I tried it and found that the initial assert will fail because the first test has left a transaction associated with the thread used to run the test. I think some routes through the code will leave abort transactions associated so that the caller can reason about the aborted transaction.

mmusgrov: Also, I thought you were going to test that the stat is reduced once the prepare phase begins and there is no test for that.

mmusgrov (changes requested):

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/TxStatsUnitTest.java
jmfinelli: > The existing tests are in TxStatsUnitTest but maybe add a new test class since testing in-flight vs preparing vs post prepared transactions might be tricky and make those tests harder to read, perhaps you could include a resource that delays the commit and/or the prepare while the test asks for the number of stats or some other way.

I must have misunderstood ^^^ request. TBH, it made sense that we wanted to check how `getNumberOfInflightTransactions()` works just before prepare. I will update this PR following your suggestions. Thanks

jmfinelli: > I tried it and found that the initial assert will fail because the first test has left a transaction associated with the thread used to run the test. I think some routes through the code will leave abort transactions associated so that the caller can reason about the aborted transaction.

This is a bug in the test. Found the solution. Should I fix it with a different JBTM? @mmusgrov

jmfinelli: > Also, I thought you were going to test that the stat is reduced once the prepare phase begins and there is no test for that.

Isn't this enough?
```
Assertions.assertEquals(1, TxStats.getInstance().getNumberOfInflightTransactions());
txn.commit();
Assertions.assertEquals(0, TxStats.getInstance().getNumberOfInflightTransactions());
```
> And a test that includes a test for rollback only transactions to verify that they aren't considered inflight.

Just to double check, you're talking about testing when the transaction is aborted, is that right?

mmusgrov: > > I tried it and found that the initial assert will fail because the first test has left a transaction associated with the thread used to run the test. I think some routes through the code will leave abort transactions associated so that the caller can reason about the aborted transaction.
> 
> This is a bug in the test. Found the solution. Should I fix it with a different JBTM? @mmusgrov

Just put your new test in its own test class and for now just include a comment in the existing TxStatsUnitTest test saying that it leaves the transaction associated. That way fixing the existing test will be low priority, and can happen at our own leisure, and will warn any future engineers to beware.

jmfinelli: I fixed the test already. I was just wondering if you prefer a new JBTM issue tracker for the bug or it is fine to include the fix in this PR

mmusgrov: > > Also, I thought you were going to test that the stat is reduced once the prepare phase begins and there is no test for that.
> 
> Isn't this enough?
> 
> ```
> Assertions.assertEquals(1, TxStats.getInstance().getNumberOfInflightTransactions());
> txn.commit();
> Assertions.assertEquals(0, TxStats.getInstance().getNumberOfInflightTransactions());
> ```
> 
> > And a test that includes a test for rollback only transactions to verify that they aren't considered inflight.
> 
> Just to double check, you're talking about testing when the transaction is aborted, is that right?

The check for inflight transactions is 

> > Also, I thought you were going to test that the stat is reduced once the prepare phase begins and there is no test for that.
> 
> Isn't this enough?
> 
> ```
> Assertions.assertEquals(1, TxStats.getInstance().getNumberOfInflightTransactions());
> txn.commit();
> Assertions.assertEquals(0, TxStats.getInstance().getNumberOfInflightTransactions());
> ```
> 
> > And a test that includes a test for rollback only transactions to verify that they aren't considered inflight.
> 
> Just to double check, you're talking about testing when the transaction is aborted, is that right?

The filter for inflight transactions is:

```
   entry.getValue().status() == ActionStatus.RUNNING &&
       entry.getValue().status() != ActionStatus.ABORT_ONLY)
```

so there should be a check that when the txn is in state ABORT_ONLY it will not be included in the stat and then
 when the txn starts preparing (ie when it makes the transition from RUNNING to ActionStatus.PREPARING) it similarly won't be included in the stat. Those are the only two extra conditions that I'd check for.

mmusgrov: But the stats test is effectively testing the state model so the maximal test would be to test the stats on every state transition and the minimal test would test for just RUNNING, ABORT_ONLY and PREPARING. I'm ok with us going for the minimal, but not less than that, test.

mmusgrov: Actually the test for inflight transactions is wrong,  `== ActionStatus.RUNNING && != ActionStatus.ABORT_ONLY` does not look correct - if it is RUNNING then && != ABORT_ONLY is vacuously true.

mmusgrov: I think` == ActionStatus.RUNNING` is enough to pick out an in flight transaction. Even though an ABORT_ONLY transaction has not started the 2PC protocol it is not considered to be active.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
```diff
@@ -39,9 +39,31 @@ public void remove(Uid id)
...
+	 *     transaction when processing is in progress and completion of the
+	 *     transaction has not yet commenced.
+	 * </p>
+	 *
+	 * @return the number of in-flight transactions currently in the system
+	 */
+	  public int getNumberOfInflightTransactions()
     {
-        return _allActions.size();
+        return (int) _allActions.entrySet()
```
mmusgrov, line 60: I'm including an earlier comment here: I think == ActionStatus.RUNNING is enough to pick out an in flight transaction. Even though an ABORT_ONLY transaction has not started the 2PC protocol it is not considered to be active.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/TxStatsUnitTest.java
mmusgrov: > I fixed the test already. I was just wondering if you prefer a new JBTM issue tracker for the bug or it is fine to include the fix in this PR

Yes, do include it in the PR then.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
jmfinelli: Done. Thanks Michael

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/TxStatsUnitTest.java
jmfinelli: I hope I addressed all the above comments with my last (force) commit. I didn't go for another class as I used the old test method to add the checks you suggested. Please, let me know what you think about the new code. Thanks @mmusgrov

jmfinelli: Forgot to mention that I will squash the two commits together as soon as you approve this PR :-)

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
```diff
@@ -39,9 +39,28 @@ public void remove(Uid id)
...
 
-    public int getNumberOfInflightTransactions()
+	/**
+	 * <p>
+	 *     This method works out the number of in-flight transactions currently in
+	 *     the system.
+	 * </p>
+	 * <p>
+	 *     Note: in-flight transactions are defined as all those transactions that
+	 *     haven't started the second phase of the two-phase commit (2PC) protocol.
```
mmusgrov, line 49: This implies that preparing and prepared transactions will be considered in flight.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/TxStatsUnitTest.java
```diff
@@ -31,34 +28,37 @@ public void test() throws Exception
...
-        assertEquals(100, TxStats.getInstance().getNumberOfApplicationRollbacks());
-        assertEquals(200, TxStats.getInstance().getNumberOfCommittedTransactions());
-        assertEquals(0, TxStats.getInstance().getNumberOfHeuristics());
-        assertEquals(1, TxStats.getInstance().getNumberOfInflightTransactions());
-        assertEquals(100, TxStats.getInstance().getNumberOfNestedTransactions());
-        assertEquals(0, TxStats.getInstance().getNumberOfResourceRollbacks());
-        assertEquals(0, TxStats.getInstance().getNumberOfTimedOutTransactions());
-        assertEquals(301, TxStats.getInstance().getNumberOfTransactions());
+
+        Assertions.assertTrue(TxStats.enabled());
```
mmusgrov, line 53: This previous assertions are using the stats so the statement either needs removing or moved to the start of the test or moved into an @Before method.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/TxStatsUnitTest.java
```diff
@@ -84,10 +87,24 @@ public void run() {
             t2.start();
             t1.join();
             t2.join();
+            D.removeThread();
```
mmusgrov, line 90: Nice, I like it a lot.

mmusgrov (changes requested):
What happened to the tests requested on the issue as discussed in comments [1,2,3]?

[1] https://github.com/jbosstm/narayana/pull/2277#discussion_r1689449770
[2] https://github.com/jbosstm/narayana/pull/2277#discussion_r1688395237
[3] https://github.com/jbosstm/narayana/pull/2277#discussion_r1689426403

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
jmfinelli: It is a rephrased version of "[active/in-flight is defined as] the state of a transaction when processing is in progress and completion of the transaction has not yet commenced." (from the OMG OTS specs). Do you think it should be rephrased?

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/TxStatsUnitTest.java
jmfinelli: Fair point. I'll move this in a @before method. Thanks Michael

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/TxStatsUnitTest.java
jmfinelli: Thanks :-)

jmfinelli:
> What happened to the tests requested on the issue as discussed in comments [1,2,3]?
> 
> [1] [#2277 (comment)](https://github.com/jbosstm/narayana/pull/2277#discussion_r1689449770)
> [2] [#2277 (comment)](https://github.com/jbosstm/narayana/pull/2277#discussion_r1688395237)
>[3] [#2277 (comment)](https://github.com/jbosstm/narayana/pull/2277#discussion_r1689426403)

@mmusgrov 
Please, can you review my last commit? Thanks

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/TxStatsUnitTest.java
```diff
@@ -5,20 +5,25 @@
...
 import com.arjuna.ats.arjuna.AtomicAction;
 import com.arjuna.ats.arjuna.common.arjPropertyManager;
+import com.arjuna.ats.arjuna.coordinator.TwoPhaseOutcome;
 import com.arjuna.ats.arjuna.coordinator.TxStats;
+import com.hp.mwtests.ts.arjuna.resources.BasicRecord;
+import org.junit.Before;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
 
 public class TxStatsUnitTest
```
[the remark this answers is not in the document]
jmfinelli: Added. Thanks Michael

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
jmfinelli: My bad. I'll change this javadoc

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
```diff
@@ -39,9 +39,26 @@ public void remove(Uid id)
...
 	}
 
-    public int getNumberOfInflightTransactions()
+	/**
+	 * <p>
+	 *     This method works out the number of in-flight transactions currently in
+	 *     the system.
+	 * </p>
+	 * <p>
+	 *     Note: The definition of in-flight (AKA active) transactions, can be found
```
tomjenkinson, line 48: I think we could change this to an Arjuna definition.

Note: The definition of in-flight transactions are those BasicActions that have status() of ActionStatus.RUNNING

BTW I wonder if we should be counting CREATED transactions too in the filter? In which case if we change the filter to be that it would be: "The definition of in-flight transactions are those BasicActions that have created and not yet begun, or a status() of ActionStatus.RUNNING"

mmusgrov: That's an unusual definition since it would include transactions that haven't started yet. If we want include those then we would need to add a new stat for it.

jmfinelli: WDYT?

Note: The definition of in-flight (AKA active) transactions, can be found in the OMG OTS specification: [active is defined as] the state of a transaction when processing is in progress and completion of the transaction has not yet commenced. In Arjuna, this definition corresponds to: in-flight transactions are those BasicActions that have created and not yet begun, or a status() of ActionStatus.RUNNING

tomjenkinson: I'll copy @mmusgrov on my contribution here as I think that the change that @jmfinelli pushed to address an earlier comment on how to define in-flight means the discussion is no longer showing here in the diff view (but can be seen https://github.com/jbosstm/narayana/pull/2277#discussion_r1692879991) and my comment is also about definition.

That being said, the wording that @jmfinelli has used is broadly in line with what I said but I still think it's worth to define it from BasicAction point of view (and I am still wondering if we should maybe be counting CREATED status).

mmusgrov: The minimum status is [ActionStatus.RUNNING](https://github.com/jbosstm/narayana/blob/main/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionStatus.java#L24).

mmusgrov: > "Note: The definition of in-flight transactions are those BasicActions that have status() of ActionStatus.RUNNING"

Manuel is [already using that definition](https://github.com/jbosstm/narayana/pull/2277/files#diff-e06a851e6fc1d3ec189ce270be29f80e4e323ffbbc27795fbc9c8cb32e759da6R60).

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/TxStatsUnitTest.java
```diff
@@ -68,6 +75,13 @@ public void test() throws Exception
         for (int i = 0; i < 100; i++) {
             final AtomicAction D = new AtomicAction();
             D.begin();
+
+            // Let's add some resources
+            D.add(new StatsResource());
```
tomjenkinson, line 80: Given this part of the test is also testing part of JBTM-3899 (and not just JBTM-2643) I think the comment just before the for loop could be expanded to clarify that - WDYT @jmfinelli?

tomjenkinson: Or maybe we don't need this bit of the test given you added another test for what I think might be the same thing in the block below where you have ` A.add(new StatsResource());` twice and the assertion??

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
jmfinelli: To include the ABORT_ONLY nuance that Michael suggested, we can go for something like:

Note: The definition of in-flight (AKA active) transactions, can be found in the OMG OTS specification: [active is defined as] the state of a transaction when processing is in progress and completion of the transaction has not yet commenced. In addition to the OMG OTS' definition, transactions that are marked as abort only are not considered to be active as they have not started the 2PC protocol. In Arjuna, these consideration boiled down to: in-flight transactions are those BasicActions that have status() equal to ActionStatus.RUNNING.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/TxStatsUnitTest.java
jmfinelli: I'll remove the comment

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
mmusgrov: And what Manuel is doing in the PR is [consistent with the JTA mapping](https://github.com/jbosstm/narayana/blob/main/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/utils/arjunacore/StatusConverter.java#L15).

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/TxStatsUnitTest.java
mmusgrov: Personally I would just delete the reference to https://issues.jboss.org/browse/JBTM-2643 since it was a bug fix.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
mmusgrov: > "Note: The definition of in-flight transactions are those BasicActions that have status() of ActionStatus.RUNNING"

But if we want to side step the nuances of using the JTA definition (which I believe is consistent with the ArjunaCore definition) then we could just go with this and be done.

tomjenkinson: I am OK with either definition, I'll resolve this therefore. I am going to make the point about CREATED where the filtering is.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
```diff
@@ -39,9 +39,26 @@ public void remove(Uid id)
...
+	 * </p>
+	 *
+	 * @return the number of in-flight transactions currently in the system
+	 */
+	  public int getNumberOfInflightTransactions()
     {
-        return _allActions.size();
+        return (int) _allActions.entrySet()
+          .parallelStream()
+          .filter(entry -> entry.getValue().status() == ActionStatus.RUNNING)
```
tomjenkinson, line 60: I want to point out that this specifically does not include transactions that are CREATED but not yet RUNNING (in case that is important to consider)

jmfinelli: I don't think we should track CREATED BasicActions as in-flight transactions as that status describes BasicActions that haven't been started yet (i.e. they are not active).

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/atomicaction/TxStatsUnitTest.java
jmfinelli: Yeah, I agree with Michael

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
jmfinelli: I'll use this javadoc then:

Note: The definition of in-flight (AKA active) transactions, can be found in the OMG OTS specification: [active is defined as] the state of a transaction when processing is in progress and completion of the transaction has not yet commenced. In addition to the OMG OTS' definition, transactions that are marked as abort only are not considered to be active as they have not started the 2PC protocol. In Arjuna, these consideration boiled down to: in-flight transactions are those BasicActions that have status() equal to ActionStatus.RUNNING.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
mmusgrov: Do we want to include those in the stat though?

jmfinelli: I wouldn't. It doesn't seem to me that counting CREATED BasicAction would add any information to Naryana's stats. Please, let me know if I'm missing something here.

tomjenkinson: It depends what we are wanting to use this method for. If (for example) we want to know when all BasicActions that are in the system are no longer able to be "used" (pre-prepare/abort) then I think we should count them.

I would assume that (at least most, not certain on recovery) CREATED BasicAction's typically becoming RUNNING soon after but might not be RUNNING yet precisely when this method is called.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
```diff
@@ -39,9 +39,31 @@ public void remove(Uid id)
...
+	 *     This method works out the number of in-flight transactions currently in
+	 *     the system.
+	 * </p>
+	 * <p>
+	 *     Note: The definition of in-flight (a.k.a. active) transactions can be
+	 *     found in the OMG OTS specification: "[Active is defined as] the state
+	 *     of a transaction when processing is in progress and completion of the
+	 *     transaction has not yet commenced." In addition to the OMG OTS
+	 *     definition, transactions that are marked as abort-only are not
+	 *     considered active, as they have not started the 2PC protocol.
```
mmusgrov, line 53: > , as they have not started the 2PC protocol

I would remove this since it isn't true.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
```diff
@@ -39,9 +39,31 @@ public void remove(Uid id)
...
+	 *     In Arjuna, these considerations boil down to the following: in-flight
```
mmusgrov, line 54: I wonder if phrasal verbs are generally understandable to non native speakers?

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
mmusgrov: The TM may crash after returning the stat in which case the CREATED transaction will never manifest as a running transaction.

And the stats are always a snapshot of the current state of the system so they are always going to be stale from the callers point of view.

mmusgrov: But a new stat seems like it will satisfy all of our viewpoints.

jmfinelli: I really like the idea of a new stat. I will create a new JBTM to track that effort. Thanks for the suggestion Michael

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
jmfinelli: Fair, I will change "boil down" :-)

mmusgrov: Thanks, in other less formal contexts, like blogs, emails, articles etc, such verbs can be nice, but I felt not so much in formal documentation.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
jmfinelli: https://issues.redhat.com/browse/JBTM-3903

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
jmfinelli: No problem. I used "are distilled into" instead. Thanks Michael

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
jmfinelli: Done :-)

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/ActionManager.java
mmusgrov: This increases the the time spent in the optimised [for throughput] path. Can we make JBTM-3903 dependent on https://issues.redhat.com/browse/JBTM-3901?

mmusgrov: I think it is worthwhile including my last comment in the JIRA so that whoever fixes it will take it into consideration.

mmusgrov: > I want to point out that this specifically does not include transactions that are CREATED but not yet RUNNING (in case that is important to consider)

@tomjenkinson Manuel created a new issue tracker for [this requirement](https://issues.redhat.com/browse/JBTM-3903) as a new statistic since in a crash scenario a CREATED transaction will never manifest as a running transaction. What priority do you want, I would suggest Optional (not that that would reduce its probability of being implemented but it's more reflective of its urgency), it's currently marked as Major.

jmfinelli: Optional sounds good to me. I modified the JBTM accordingly

jmfinelli: I will also make JBTM-3903 depend on JBTM-3901. Thanks Michael

mmusgrov (approved):
Looks great.

PR #2278, opened 2024-07-19
Use the default deploy plugin to disable not needed submodules

marcosgopen:
NO_TESTS

### DIFF
--- a/ArjunaCore/arjuna/services/pom.xml
+++ b/ArjunaCore/arjuna/services/pom.xml
@@ -18,6 +18,15 @@
   <packaging>jar</packaging>
   <build>
     <sourceDirectory>${sourceDirACS}</sourceDirectory>
+    <plugins>
+      <plugin>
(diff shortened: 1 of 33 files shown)


mmusgrov (changes requested):
Can you include a description with this PR, and we probably will also need a JBTM raising for it. The code being removed was added for a reason so we would minimally need it documented somewhere why it's no longer needed and perhaps why it was added in the first place.

marcosgopen:
@tomjenkinson I tested a snapshot release of Narayana at https://repository.jboss.org/nexus/content/repositories/snapshots/org/jboss/narayana/ (date Fri Jul 19 17:31:59 UTC 2024)

@mmusgrov I updated the description and created a JBTM for it. Thanks

tomjenkinson (approved):

mmusgrov (approved):

PR #2281, opened 2024-07-24
JBTM-3883 use a base64 encoder if required

mmusgrov:

I re-opened JBTM-3883 to add this extra check. I added a comment to the issue and to the test in this PR that explains the details of why the base64 encoder might be needed.

CORE !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !LRA !DB_TESTS !mysql !db2 !postgres !oracle


mmusgrov:
I've put this on hold while I think through the impact on crash recovery..

mmusgrov:
I've removed the Hold label because Marco, or Manuel correctly pointed out to me that SHA-224 and base64 encoders are deterministic functions and as such will always produce the same output for a given string, ie assuming that any particular integration gives us the same string we'll be able to detect records that "belong" to the recovery manager.

marcosgopen (approved):
Thanks Mike for your PR, LGTM!

PR #2286, opened 2024-07-31
[JBTM-3883] 'use a base64 encoder if required' fix

marcosgopen:

Update the nodeIdentifierBytes with the trimmed one when needed
Log the proper length of the bytes when error occurs
Check that the string is shorter than NODE_NAME_SIZE (not the bytes length)
! don't confuse TxControl.NODE_NAME_SIZE (which is 28) with NODE_NAME_SIZE (which is 64)

CORE !AS_TESTS !RTS! JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !LRA !DB_TESTS


mmusgrov:
I still need to understand why my port of the fix you applied to the quarkus narayana-jta extension failed before reviewing this PR. We need to do that because the failure is difficult to reproduce. That said I will take a look either today or Monday.

marcosgopen:
I am closing this because I think it does not solve the error shown in the build: https://ci-jenkins-csb-narayana.apps.ocp-c1.prod.psi.redhat.com/job/narayana/280/PROFILE=CORE,jdk=openJDK11,label=jnlp-agent/console
I will keep investigating.

PR #2298, opened 2024-09-23
Update LRAAsyncIT to use thread safe logic

jmfinelli:
[JBTM-3928](https://issues.redhat.com/browse/JBTM-3928)

!CORE !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE LRA !DB_TESTS

!JDK11


tomjenkinson (approved):
LGTM, approving - thank you. It might be worth considering changing lrasToAfterFinish to a thread-safe Collection instead of synchronizing the handling of it here so if you decide you want to go that route I am happy to re-review it. Thanks for the contribution!

jmfinelli:
> LGTM, approving - thank you. It might be worth considering changing lrasToAfterFinish to a thread-safe Collection instead of synchronizing the handling of it here so if you decide you want to go that route I am happy to re-review it. Thanks for the contribution!

I don't think `lrasToAfterFinish` is used concurrently in other places so I'm happy to synchronise around it when needed. Thanks @tomjenkinson

PR #2301, opened 2024-09-26
JBTM-3298 LRA compenstators should compensate in opposite order from …

mmusgrov:
https://issues.redhat.com/browse/JBTM-3298

As a test and a fix for calling LRA participants in the opposite order from which they enlisted.

!CORE !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE LRA !DB_TESTS !mysql !db2 !postgres oracle !JDK11

### DIFF
--- a/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
+++ b/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
@@ -521,8 +521,8 @@ public class LongRunningAction extends BasicAction {
                         failedList = new RecordList();
                     }
 
-                    moveTo(pendingList, preparedList);
-                    moveTo(heuristicList, preparedList);
+                    moveTo(pendingList, preparedList, true);
+                    moveTo(heuristicList, preparedList, true);
 
                     updateState(LRAStatus.Cancelling);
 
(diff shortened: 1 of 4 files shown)


marcosgopen:
Thanks Mike for your PR, could you have a look at the LRA failed job? It looks like the CoordinatorApiIT failed because 'Invalid link URI (<http://complete.url:8080/>; rel="complete"): missing compensator or after LRA callback'

mmusgrov:
The failure was in the FailedLRAIT testsuite on just the JDK17 run which I'm still investigating . We always get `Invalid link URI ...`: we have some tests, not sure which ones though, that provide invalid information and then verifies that the coordinator refuses the request, unfortunately they generate deep stack traces as part of the successful route through the code which make it harder to parse the output for genuine failures.

marcosgopen:
> The failure was in the FailedLRAIT testsuite on just the JDK17 run which I'm still investigating . We always get `Invalid link URI ...`: we have some tests, not sure which ones though, that provide invalid information and then verifies that the coordinator refuses the request, unfortunately they generate deep stack traces as part of the successful route through the code which make it harder to parse the output for genuine failures.

Sorry, you are right. The failing test is 'FailedLRAIT' as you said. I'll have a look.

mmusgrov:
There's no evidence in the logs so I am going to push an update that will include some debug of the failing method (FailedLRAIT#validateStateAndRemove). We can always remove it after the PR re-runs.

mmusgrov:
@marcosgopen I've put it on Hold while I figure out why the changes have effected the FailedLRAIT tests.

mmusgrov:
@marcosgopen Please bear with me while I investigate (it only seems to fail on FailedLRAIT on CI and on OpenJDK 17 so can't debug on my laptop).

I have pushed a debug commit and an experimental commit to give me more information, ie please don't review it further until I've figured out what the problem is.

PR #2309, opened 2024-10-08
JBTM-3356 Verify that unacknowledged AfterLRA callbacks are redelivered

mmusgrov:
https://issues.redhat.com/browse/JBTM-3356

!CORE !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE LRA !DB_TESTS !mysql !db2 !postgres oracle !JDK17

### DIFF
--- a/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
+++ b/rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
@@ -392,8 +392,10 @@ public class LongRunningAction extends BasicAction {
         return status;
     }
 
+    // a recovering LRA is one that is in state Cancelling or Closing or has pending listener notifications
+    // which are maintained on the heuristic list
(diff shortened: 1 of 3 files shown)


marcosgopen:
@mmusgrov I can see that Martin wrote a test too: https://github.com/xstefank/narayana/commit/55a151ad78ef98c82ac3580a6818901a03c2e556
Do you think we can include it too?

mmusgrov:
> @mmusgrov I can see that Martin wrote a test too: [xstefank@55a151a](https://github.com/xstefank/narayana/commit/55a151ad78ef98c82ac3580a6818901a03c2e556) Do you think we can include it too?

It's effectively the same test so having it duplicated would increase our workload. The difference is that all the coordinator tests live with the coordinator so that changes can be tested in the same module. The new test follows the same approach as the other tests in the coordinator test suite.

marcosgopen (approved):
Thanks @mmusgrov for your PR. LGTM

PR #2310, opened 2024-10-10
JBTM-3935 Report all BasicAction object store IOException messages

mmusgrov:
https://issues.redhat.com/browse/JBTM-3935

CORE !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE LRA !DB_TESTS !mysql !db2 !postgres oracle


ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/logging/arjunaI18NLogger.java
```diff
@@ -281,9 +281,9 @@ public interface arjunaI18NLogger {
 	@LogMessage(level = WARN)
 	public void warn_coordinator_BasicAction_21(String arg0);
 
-	@Message(id = 12063, value = "BasicAction.restore_state - error unpacking action status.", format = MESSAGE_FORMAT)
+	@Message(id = 12063, value = "BasicAction.restore_state - error unpacking action status {0}.", format = MESSAGE_FORMAT)
 	@LogMessage(level = WARN)
-	public void warn_coordinator_BasicAction_24();
+	public void warn_coordinator_BasicAction_24(String arg0);
```
tomjenkinson, line 286: Have we got a precedent for editing an old log message that has an in use id? Perhaps we can just add new ones and leave the old ones untouched as maybe (maybe) someone has automation looking for 12063 and it would be unexpected to find different words? I am not super -1 to allowing the change in to the message (especially if we have a precedent) but WDYT to just changing to use warn_coordinator_BasicAction_73 or possibly doing warn_coordinator_BasicAction_24 (so it's not removed) and straight away doing warn_coordinator_BasicAction_73?

mmusgrov: The reason I included it was that the code ARJUNA012063 (warn_coordinator_BasicAction_24) is unchanged and it is that code that identifies the message, plus I tacked it onto the end of the string so anything matching on it would still find it unless they were matching the new line as well.

I have seen it done before but it has depended upon who did the review as to whether it was allowed.

I don't want to include a new message id for it since that would mean having to produce two log messages for the same problem.

PS tagging @marcosgopen since I discussed it with him pointing out that we would need to check with you before including such a change and given your response and my response I will remove it from the PR.

mmusgrov: But two of the main reasons why we would not want to change log messages is because the department that does the localisation might not be geared up for it and any existing documentation, such as [1], would get out of sync.

[1] https://access.redhat.com/webassets/avalon/d/red_hat_jboss_enterprise_application_platform/7.3.Beta/mgmt_model/log-message-reference.html

tomjenkinson (approved):
LGTM - thank you!

PR #2312, opened 2024-10-11
JBTM-3331 Add a check for early timeouts during enlistment

mmusgrov:
https://github.com/mmusgrov/narayana/pull/new/JBTM-3331

!CORE !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE LRA !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
+++ b/rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
@@ -9,12 +9,7 @@ import static io.narayana.lra.LRAConstants.COORDINATOR_PATH_NAME;
 import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_CONTEXT_HEADER;
 import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_RECOVERY_HEADER;
 import static org.hamcrest.Matchers.containsString;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
 
 import java.net.URI;
 import java.net.URISyntaxException;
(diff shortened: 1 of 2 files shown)


mmusgrov:
@marcosgopen I have put this one on Hold while we discuss it. I was unable to reproduce Ondra's original issue - the test I wrote covers what he said but seems to be spec compliant.

The issue also seems related to JBTM-3922 (Investigate the behaviour of LRA when timeout is triggered early) which is assigned to you so I want to discuss the relationship to that issue too.

rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
```diff
@@ -9,12 +9,7 @@
 import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_CONTEXT_HEADER;
 import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_RECOVERY_HEADER;
 import static org.hamcrest.Matchers.containsString;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
```
tomjenkinson: Please can I ask if it was intentionakl to remove these imports in this commit and replace with a .*? I can see you have a separate commit that follows up to remove the .* that is added in this one? https://github.com/jbosstm/narayana/pull/2312/commits/4220b84f76bc2c80fdf69c4cea22ca9003d44c87#diff-709c5ef85ed562bd70894c9781aaaae149a18d85eddc6f7d32902e1c387e0311L12

That commit also changes a lot of `assert[...]` to `Assert.assert[...]` which I don't quite understand? Is it something to move forwards with JUnit versions somehow?

mmusgrov: Look at the second commit where I organise the imports - I thought that was better than updating my settings (which I'll do anyway) but I want to discuss the issue with Marco.

tomjenkinson: Yeah, I had seen the other commit but I am still not clear on why the need to change from (e.g.) assertEquals to Assert.assertEquals (etc)?

mmusgrov: I think the only advantage to using the static calls is to improve readability of the source files. I guess it is a matter of developer preference, but we can avoid using static Asserts if that is what the team prefers.

mmusgrov: I also notice that the hudson script sets `ARQ_PROF=no_arq` - I've not seen that before but it doesn't exist so I'll need to investigate that too.

mmusgrov: > I think the only advantage to using the static calls is to improve readability of the source files. I guess it is a matter of developer preference, but we can avoid using static Asserts if that is what the team prefers.

I'll remove them and if the team wants to allow it then we discuss that separately.

marcosgopen: Thanks for your question Mike, I added the 'no_arq' profile for JDK 11 because we cannot test Wildfly with that JDK only to avoid the arq profile to be enable. I can see now that there is the warning. Maybe it would be better to set ARQ_PROF="" ?

mmusgrov: If ARQ_PROF isn't valid for the JDK11 tests then unsetting ARQ_PROF is best: `unset ARQ_PROF`

mmusgrov:
@tomjenkinson this test requires byteman support adding to the coordinator test suite which is an extra chunk of work that I hadn't anticipated in order to set up the right circumstances to trigger the timeout behaviour. I'll raise it again when I've done that.

PR #2315, opened 2024-10-16
JBTM-3940 NarayanaLRAClient is an internal class so does not need to …

mmusgrov:
https://issues.redhat.com/browse/JBTM-3940

When I moved the client API into an internal package (JBTM-3934) I forgot to remove the @Deprecated annotation. This PR rectifies that oversight.

NO_TEST


marcosgopen (approved):
Thanks @mmusgrov , LGTM!
Do we have a document to describe the deprecation process/workflow in Narayana?

marcosgopen:
> If not, we could propose some text to https://github.com/jbosstm/narayana/blob/main/CONTRIBUTING.md and share it with the community for discussion?

@tomjenkinson  Together with the deprecation process I would be interested in including in the https://github.com/jbosstm/narayana/blob/main/CONTRIBUTING.md a description about the performance expectation and its tolerance range when testing new PRs.

PR #2316, opened 2024-10-18
Quarkus dependency removal

marcosgopen:
Starting from quarkus 3.7 the minimum JDK version is 17.
Because the narayana supports JDK 11 we need to decide if we want to create jdk maven profiles or remove quarkus dependency from Narayana codebase.
I would prefer to move quarkus dependencies in quickstart and use maven jdk profile there if needed.
 
CORE !AS_TESTS RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !LRA !DB_TESTS

### DIFF
--- a/rts/at/tx/src/test/java/org/jboss/jbossts/star/test/BaseTest.java
+++ b/rts/at/tx/src/test/java/org/jboss/jbossts/star/test/BaseTest.java
@@ -13,7 +13,6 @@ import com.arjuna.ats.arjuna.state.InputObjectState;
 import com.arjuna.ats.internal.arjuna.common.UidHelper;
 import com.arjuna.ats.internal.jta.transaction.arjunacore.AtomicAction;
 import com.squareup.okhttp.OkHttpClient;
-import io.quarkus.runtime.QuarkusApplication;
 import io.undertow.Undertow;
 import jakarta.ws.rs.DELETE;
 import jakarta.ws.rs.DefaultValue;
@@ -67,8 +66,7 @@ public class BaseTest {
 
     protected static final ExecutorService executor = Executors.newFixedThreadPool(4);
     protected static boolean USE_NETTY = false;
-    protected static boolean USE_UNDERTOW = false;
-    protected static boolean USE_QUARKUS = true;
+    protected static boolean USE_UNDERTOW = true;
 // jarkarta TODO jersey    private static HttpServer grizzlyServer;
     protected static final String USE_SPDY_PROP = "rts.usespdy";
     protected static final String USE_SSL_PROP = "rts.usessl";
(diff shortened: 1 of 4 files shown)


marcosgopen:
JDK11 tests fail because from Quarkus 3.7 the minimum JDK version is 17, see https://quarkus.io/blog/java-17/

rts/at/tx/src/test/java/org/jboss/jbossts/star/test/BaseTest.java
```diff
@@ -67,8 +66,7 @@ public class BaseTest {
 
     protected static final ExecutorService executor = Executors.newFixedThreadPool(4);
     protected static boolean USE_NETTY = false;
-    protected static boolean USE_UNDERTOW = false;
-    protected static boolean USE_QUARKUS = true;
+    protected static boolean USE_UNDERTOW = true;
```
tomjenkinson, line 69: This is definitely an option for me to move from Quarkus and Undertow to raw Undertow, but if I understood correctly having the Quarkus used helped to test some Quarkus annotation or something? @mmusgrov please can you help share a bit more context on the benefit to running this test with Quarkus?

tomjenkinson (approved):
I am good with this change as it is and approving and it can be merged assuming CI is OK with it. That said, we can re-introduce a Quarkus variant of this test for JDK17 if we decide it is appropriate to do any testing of Quarkus (annotations? /cc @mmusgrov) inside Narayana code base

rts/at/tx/src/test/java/org/jboss/jbossts/star/test/BaseTest.java
mmusgrov: We, Sweta and I, added it because the work was to do with SRA annotations, aka annotations for REST-AT, and that work was designed to facilitate the introduction of a quarkus rest-at extension and having the tests runs on quarkus, the target platform, was the correct thing to do. Such an extension with annotations will give users alternatives to WS-AT.

tomjenkinson: OK, so we can try to run this test twice then, once with Undertow (on JDK 11 and above) and once with Quarkus (when using JDK 17 and above as Quarkus is not supported on JDK 11 if I understood correctly).

tomjenkinson: @marcosgopen please can you raise a JBTM to re-add Quarkus testing of this class for SRA annotation testing on JDK 17 and above and to have the Quarkus dependency updated to a supported one? We can't stay on 3.2.6, 3.8 seems to be the oldest LTS version (https://quarkus.io/security/) and JDK 17 is the minimum since 3.7 (https://quarkus.io/blog/java-17/).

marcosgopen: Created, thanks. https://issues.redhat.com/browse/JBTM-3943

PR #2320, opened 2024-10-22
JBTM-3330 Nested closed LRA is kept in Closing state which is breaking the LRA state model

marcosgopen:
https://issues.redhat.com/browse/JBTM-3330

I think this issue is outdated, but to be sure I added two tests to verify the case is covered in tests.

CORE !AS_TESTS RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE LRA !DB_TESTS

### DIFF
--- a/rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
+++ b/rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
@@ -129,10 +129,12 @@ public class LRATest extends LRATestBase {
         server.deployOldStyle(LRAParticipant.class);
 
         service = LRARecoveryModule.getService();
+        assertNull(testName + ": current thread should not be associated with any LRAs", lraClient.getCurrent());
     }
 
     @After
     public void after() {
+        assertNull(testName + ": current thread should not be associated with any LRAs", lraClient.getCurrent());
         LRALogger.logger.debugf("Finished test %s", testName);
         lraClient.close();
         client.close();
@@ -351,7 +353,6 @@ public class LRATest extends LRATestBase {
         // and since the test then cancelled it directly, ie not via the API, the LRA will still be associated.
         // Therefore, it needs to be cleared otherwise subsequent tests will still see the LRA associated:
         lraClient.clearCurrent(false);
-        assertNull("LRA is still associated with the current thread", lraClient.getCurrent());
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


rts/lra/coordinator/src/test/java/io/narayana/lra/coordinator/domain/model/LRATest.java
```diff
@@ -981,15 +979,40 @@ public void testGrandparentContext() {
...
+        lraClient.closeLRA(parent); // should close everything in the hierarchy
+        // verify the parent and child are closed
+        assertStatus("grandparent", grandParent, null, LRAStatus.Active);
+        assertStatus("parent", parent, null, LRAStatus.Closed);
+        assertStatus("child", child, null, LRAStatus.Closed);
+        lraClient.closeLRA(grandParent); // should close everything in the hierarchy
+    }
+
+    @Test
+    public void testNestedLRAContext() {
```
tomjenkinson, line 1003: The distinction in these tests is super subtle :)

Please can you help explain the difference a bit?

I can see the functional differences to be:
`NestedParticipantIT#testNestedLRAContext grandparent`
`"parent", parent, null, LRAStatus.Active`

Especially I am not clear why `lraClient.closeLRA(parent);` did not close the "parent"

marcosgopen: Thank you @tomjenkinson for pointing that out. That was a mistake, I've fixed it.
The idea is to test that when closing an LRA (also nested LRA) it needs to be is closed status as also all its children.

marcosgopen: There is a misunderstanding @tomjenkinson , when I created this PR the code was correct. I made the wrong change later ( with a force push, see https://github.com/jbosstm/narayana/compare/092374befa34cb8e516ccb0b087c2e6f54029d74..97ec0c0321bb78cac57d3d1e7db3750e819bf1eb). After that force push the build 575 started, but unfortunately another PR wan merged in the code base, so that build failed with "Narayana rebase on main failed. Please rebase it manually". So actually "closing the parent as I had it before" would have resulted in a failed build.

tomjenkinson: Ah, thank you for clarifying!

tomjenkinson (approved):
LGTM - thank you!

marcosgopen:
> Please consider how best to resolve https://issues.redhat.com/browse/JBTM-3330 in JIRA - I don't expect we can mark it as "Done" with 7.next because we don't know if "7.next" fixed it or a different release since "5.10.4.Final" - maybe "Cannot Reproduce" with "7.next"? But please consider it.

Thanks @tomjenkinson , I think close the issue as  'Cannot reproduce' + a comment would be good.

PR #2322, opened 2024-10-28
[JBTM-2833] Part of the OSB tool has been moved to internal packages

jmfinelli:
This PR is a proof of concept to address [JBTM-2833](https://issues.redhat.com/browse/JBTM-2833). The main goal of this PR is to demonstrate that JBTM-2833 is feasible.

CORE AS_TESTS RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE LRA !DB_TESTS


jmfinelli:
Failures are related to WildFly. In fact, this application server needs to update the package where [ObjStoreBrowser lives](https://github.com/wildfly/wildfly/blob/956cd9ad69a0b65b681d14ba76ca4ec79a634e6b/transactions/src/main/java/org/jboss/as/txn/service/ArjunaTransactionManagerService.java#L13)

mmusgrov:
We should avoid merging this one until we have discussed/researched what might be required for external management applications, for example such applications may well operate at the MBean and MXBeanDescription levels so would need the corresponding APIs to be public - I added a note to this effect to the issue tracker.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/tools/osb/mbean/ObjStoreBrowser.java
```diff
@@ -2,7 +2,7 @@
    Copyright The Narayana Authors
    SPDX-License-Identifier: Apache-2.0
  */
-package com.arjuna.ats.arjuna.tools.osb.mbean;
+package com.arjuna.ats.internal.arjuna.tools.osb.mbean;
```
tomjenkinson, line 5: Maybe this one shouldn't be internal? As you found it is the way it seems to be instantiated in WildFly: https://github.com/wildfly/wildfly/blob/956cd9ad69a0b65b681d14ba76ca4ec79a634e6b/transactions/src/main/java/org/jboss/as/txn/service/ArjunaTransactionManagerService.java#L13

jmfinelli: I moved this class to an internal package as it was annotated with `@Deprecated`. Do you think we should move this class, @mmusgrov?

jmfinelli:
@tomjenkinson I will squeeze the two commits together before merging

jmfinelli:
This PR was closed as the OSB tool needs more development to achieve the objective describe in JBTM-2833

PR #2324, opened 2024-10-31
Update jboss dependencies

marcosgopen:
Update plugin dependencies

jboss.logmanager>2.1.19.Final
jboss-logging-processor>3.0.2.Final
jboss-logging>3.6.1.Final
jboss.openjdk-orb>10.1.0.Final
jboss.ws>3.0.0.Final

I removed the plugin version so that it uses the versions defined in the jboss-parent.

CORE AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !LRA !DB_TESTS

### DIFF
--- a/narayana-bom/pom.xml
+++ b/narayana-bom/pom.xml
@@ -57,11 +57,11 @@
     <version.org.jboss.invocation>2.0.0.Final</version.org.jboss.invocation>
     <version.org.jboss.jandex>2.1.1.Final</version.org.jboss.jandex>
     <version.org.jboss.jboss-transaction-spi>8.0.0.Final</version.org.jboss.jboss-transaction-spi>
-    <version.org.jboss.logging.jboss-logging>3.5.0.Final</version.org.jboss.logging.jboss-logging>
-    <version.org.jboss.logging.jboss-logging-processor>2.2.1.Final</version.org.jboss.logging.jboss-logging-processor>
-    <version.org.jboss.logmanager>1.5.1.Final</version.org.jboss.logmanager>
-    <version.org.jboss.openjdk-orb>9.0.1.Final</version.org.jboss.openjdk-orb>
-    <version.org.jboss.ws>2.0.0.Final</version.org.jboss.ws>
+    <version.org.jboss.logging.jboss-logging>3.6.1.Final</version.org.jboss.logging.jboss-logging>
+    <version.org.jboss.logging.jboss-logging-tools>3.0.2.Final</version.org.jboss.logging.jboss-logging-tools>
+    <version.org.jboss.logmanager>2.1.19.Final</version.org.jboss.logmanager>
+    <version.org.jboss.openjdk-orb>10.1.0.Final</version.org.jboss.openjdk-orb>
+    <version.org.jboss.ws>3.0.0.Final</version.org.jboss.ws>
     <version.parsson>1.1.3</version.parsson>
     <version.smallrye-converter-api>2.7.0</version.smallrye-converter-api>
     <version.sortpom>3.0.0</version.sortpom>
@@ -97,7 +97,7 @@
       <dependency>
         <groupId>org.jboss.logging</groupId>
         <artifactId>jboss-logging-processor</artifactId>
-        <version>${version.org.jboss.logging.jboss-logging-processor}</version>
+        <version>${version.org.jboss.logging.jboss-logging-tools}</version>
       </dependency>
       <dependency>
         <groupId>org.jboss.narayana</groupId>
@@ -528,7 +528,7 @@
       <dependency>
         <groupId>org.jboss.logging</groupId>
         <artifactId>jboss-logging-annotations</artifactId>
-        <version>${version.org.jboss.logging.jboss-logging-processor}</version>
+        <version>${version.org.jboss.logging.jboss-logging-tools}</version>
         <optional>true</optional>
       </dependency>
 
--- a/pom.xml
+++ b/pom.xml
@@ -183,14 +183,10 @@
     <!-- The surefire/failsafe test execution logs goes to a file by default -->
     <testLogToFile>true</testLogToFile>
     <version.com.github.spotbugs.spotbugs-maven-plugin>4.8.3.1</version.com.github.spotbugs.spotbugs-maven-plugin>
-    <version.compiler.plugin>3.8.0-jboss-2</version.compiler.plugin>
 
     <version.doxia>1.8.1</version.doxia>
     <version.io.narayana.checkstyle-config>1.0.1.Final</version.io.narayana.checkstyle-config>
     <version.javadoc.plugin>3.5.0</version.javadoc.plugin>
-    <version.maven-failsafe-plugin>3.0.0-M7</version.maven-failsafe-plugin>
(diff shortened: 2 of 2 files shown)


pom.xml
```diff
@@ -10,7 +10,7 @@
   <parent>
     <groupId>org.jboss</groupId>
     <artifactId>jboss-parent</artifactId>
-    <version>39</version>
+    <version>46</version>
```
tomjenkinson, line 13: I think we should try to target the version that WildFly is using: https://github.com/wildfly/wildfly/blob/main/pom.xml#L12

pom.xml
```diff
@@ -183,14 +183,14 @@
     <!-- The surefire/failsafe test execution logs goes to a file by default -->
     <testLogToFile>true</testLogToFile>
     <version.com.github.spotbugs.spotbugs-maven-plugin>4.8.3.1</version.com.github.spotbugs.spotbugs-maven-plugin>
-    <version.compiler.plugin>3.8.0-jboss-2</version.compiler.plugin>
+    <version.compiler.plugin>3.13.0</version.compiler.plugin>
```
tomjenkinson, line 186: Is it at all possible to just remove this line and use the version from the JBoss parent (https://github.com/jboss/jboss-parent-pom/blob/jboss-parent-39/pom.xml)?

pom.xml
```diff
@@ -183,14 +183,14 @@
...
 
     <version.doxia>1.8.1</version.doxia>
     <version.io.narayana.checkstyle-config>1.0.1.Final</version.io.narayana.checkstyle-config>
     <version.javadoc.plugin>3.5.0</version.javadoc.plugin>
-    <version.maven-failsafe-plugin>3.0.0-M7</version.maven-failsafe-plugin>
-    <version.maven-surefire-plugin>3.0.0-M7</version.maven-surefire-plugin>
-    <version.maven.checkstyle-plugin>3.1.2</version.maven.checkstyle-plugin>
+    <version.maven-failsafe-plugin>3.5.1</version.maven-failsafe-plugin>
+    <version.maven-surefire-plugin>3.5.1</version.maven-surefire-plugin>
+    <version.maven.checkstyle-plugin>3.6.0</version.maven.checkstyle-plugin>
```
tomjenkinson, line 193: Maybe these can be the ones from the JBoss Parent too?

narayana-bom/pom.xml
```diff
@@ -57,11 +57,11 @@
...
     <version.org.jboss.jandex>2.1.1.Final</version.org.jboss.jandex>
     <version.org.jboss.jboss-transaction-spi>8.0.0.Final</version.org.jboss.jboss-transaction-spi>
-    <version.org.jboss.logging.jboss-logging>3.5.0.Final</version.org.jboss.logging.jboss-logging>
-    <version.org.jboss.logging.jboss-logging-processor>2.2.1.Final</version.org.jboss.logging.jboss-logging-processor>
-    <version.org.jboss.logmanager>1.5.1.Final</version.org.jboss.logmanager>
-    <version.org.jboss.openjdk-orb>9.0.1.Final</version.org.jboss.openjdk-orb>
-    <version.org.jboss.ws>2.0.0.Final</version.org.jboss.ws>
+    <version.org.jboss.logging.jboss-logging>3.6.1.Final</version.org.jboss.logging.jboss-logging>
+    <version.org.jboss.logging.jboss-logging-processor>3.0.2.Final</version.org.jboss.logging.jboss-logging-processor>
+    <version.org.jboss.logmanager>3.0.6.Final</version.org.jboss.logmanager>
```
tomjenkinson, line 62: I think it's possible that WildFly has not got to this version yet: https://github.com/wildfly/wildfly-core/blob/27.0.0.Beta1/pom.xml#L218

narayana-bom/pom.xml
```diff
@@ -57,11 +57,11 @@
     <version.org.jboss.invocation>2.0.0.Final</version.org.jboss.invocation>
     <version.org.jboss.jandex>2.1.1.Final</version.org.jboss.jandex>
     <version.org.jboss.jboss-transaction-spi>8.0.0.Final</version.org.jboss.jboss-transaction-spi>
-    <version.org.jboss.logging.jboss-logging>3.5.0.Final</version.org.jboss.logging.jboss-logging>
-    <version.org.jboss.logging.jboss-logging-processor>2.2.1.Final</version.org.jboss.logging.jboss-logging-processor>
-    <version.org.jboss.logmanager>1.5.1.Final</version.org.jboss.logmanager>
-    <version.org.jboss.openjdk-orb>9.0.1.Final</version.org.jboss.openjdk-orb>
-    <version.org.jboss.ws>2.0.0.Final</version.org.jboss.ws>
+    <version.org.jboss.logging.jboss-logging>3.6.1.Final</version.org.jboss.logging.jboss-logging>
+    <version.org.jboss.logging.jboss-logging-processor>3.0.2.Final</version.org.jboss.logging.jboss-logging-processor>
```
tomjenkinson, line 61: Maybe we should rename this to version.org.jboss.logging.jboss-logging-tools: https://github.com/wildfly/wildfly-core/blob/27.0.0.Beta1/pom.xml#L218

narayana-bom/pom.xml
```diff
@@ -57,11 +57,11 @@
...
+    <version.org.jboss.logmanager>3.0.6.Final</version.org.jboss.logmanager>
+    <version.org.jboss.openjdk-orb>10.1.0.Final</version.org.jboss.openjdk-orb>
+    <version.org.jboss.ws>3.0.0.Final</version.org.jboss.ws>
```
tomjenkinson, line 64: I might be wrong but I don't _think_ we are using this property anywhere in Narayana directly?

pom.xml
marcosgopen: Yes, good point. I'll test it and update the PR if that works

narayana-bom/pom.xml
marcosgopen: We are using the property to define the jbossws-api version, which is used in multiple places: https://github.com/jbosstm/narayana/blob/main/XTS/localjunit/pom.xml#L54-L58
I'll check what version WildFly is using

narayana-bom/pom.xml
marcosgopen: In WildFly they call it like that because they use the same property for jboss-logging-processors and jboss-logging-annotation artifacts. We can do it too. +1

narayana-bom/pom.xml
marcosgopen: https://github.com/wildfly/wildfly/blob/main/pom.xml#L561 3.0.0.Final used in WildFly

narayana-bom/pom.xml
marcosgopen: You are right, I'll update it to 2.1.19.Final then. Thanks

narayana-bom/pom.xml
tomjenkinson: Sorry, I didn't search well enough and the reference to version.org.jboss.ws is in the narayana-bom itself!

tomjenkinson (approved):

PR #2326, opened 2024-11-01
It may be appropriate to throw an UNKNOWN exception if we don't know …

tomjenkinson:
I haven't raised a JBTM yet.

Issues addressed:
* Possibly we should throw an UNKNOWN if we don't know whether the rollback was successful, similar to how the commit works: https://github.com/jbosstm/narayana/blame/7.0.2.Final/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java#L555
* https://issues.redhat.com/browse/JBTM-3944
* The throwing of the UNKNOWN should not take place in the case of XAER_NOTA

CORE AS_TESTS !RTS !JACOCO !XTS !QA_JTA QA_JTS_OPENJDKORB !PERFORMANCE !LRA !DB_TESTS mysql db2 postgres oracle

!JDK11


jmfinelli (approved):

mmusgrov:
Before this PR gets merged I think we need to document it with a JBTM since the behaviour is somewhat subtle.

mmusgrov (approved):
I agree that reporting a heuristic if the resource throws XAER_RMFAIL on rollback is not correct and that using UNKNOWN, to be consistent with how we handle the same problem during commit (and we really don't know what the resource did), is the correct fix for this issue. And since the PR is addressing a spec non-compliance issue, any existing tests that expect a heuristic to be reported here should be modified to accommodate the fixed behaviour.

PR #2327, opened 2024-11-01
JBTM-3944 Reorder the handling of XAException.XAER_RMFAIL and XAExcep…

tomjenkinson:
https://issues.redhat.com/browse/JBTM-3944

CORE AS_TESTS !RTS !JACOCO !XTS !QA_JTA QA_JTS_OPENJDKORB !PERFORMANCE !LRA !DB_TESTS mysql db2 postgres oracle

!JDK11

### DIFF
--- a/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
+++ b/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
@@ -391,12 +391,6 @@ public class XAResourceRecord extends com.arjuna.ArjunaOTS.OTSAbstractRecordPOA 
                                 _rolledBack = true;
                                 destroyState();
                                 break;
-                            case XAException.XAER_NOTA:
-                                if (_phaseTwoStarted) {
-                                    // rolled back previously and recovery completed
-                                    destroyState();
-                                    break;
-                                }
                             case XAException.XAER_RMFAIL:
                                 if (_prepared) {
                                     /*
@@ -407,6 +401,12 @@ public class XAResourceRecord extends com.arjuna.ArjunaOTS.OTSAbstractRecordPOA 
                                     destroyState = false;
                                 }
                                 break;
+                            case XAException.XAER_NOTA:
+                                if (_phaseTwoStarted) {
+                                    // rolled back previously and recovery completed
+                                    destroyState();
+                                    break;
+                                }
                             default:
                                 destroyState();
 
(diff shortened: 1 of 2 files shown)


jmfinelli (approved):

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -407,6 +401,12 @@ public void rollback() throws org.omg.CORBA.SystemException,
                                     destroyState = false;
                                 }
                                 break;
+                            case XAException.XAER_NOTA:
+                                if (_phaseTwoStarted) {
+                                    // rolled back previously and recovery completed
+                                    destroyState();
+                                    break;
+                                }
```
mmusgrov, line 409: Why do we need to fall through to the next arm of the case here?

mmusgrov: I notice that PR https://github.com/jbosstm/narayana/pull/2326 does the same but with the addition of a `throw new UNKNOWN();` for the XAER_RMFAIL case. Are both PRs candidates for merging?

tomjenkinson: They are both candidates. This one is a bug, the other is for consideration of whether we want to go that way (on top of this). (I'll respond to the first comment in a follow-up)

tomjenkinson: I am reverting the bug introduced in https://github.com/jbosstm/narayana/commit/7f41fcb996ad3492a8718d762b47ce5616499c82 whereby XAER_NOTA no longer throws an exception because it drops into the XAER_RMFAIL handling and breaks out of the switch and therefore bypasses https://github.com/jbosstm/narayana/blob/main/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java#L413-L417

I added a comment to describe the potential issue hopefully then if someone adds another case they will be aware of this

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/resources/jts/orbspecific/XAResourceRecord.java
```diff
@@ -407,6 +401,15 @@ public void rollback() throws org.omg.CORBA.SystemException,
...
+                                // DO NOT add any case handling below here which would break out of the switch and prevent
+                                // XAException.XAER_NOTA from throwing an org.omg.CosTransactions.HeuristicHazard or 
+                                // UNKNOWN exception
```
mmusgrov, line 412: I still think it is confusing to fall through even with the code warning. The only benefit of falling through here is to save three lines of code in the XAException.XAER_NOTA arm.

tomjenkinson: OK, thank you. I'll go ahead and merge this change as what I am doing here is reverting this part you are commenting on back to a point that existed before my original change. I think the code you are referring to was added as this part of a pull request from 2017: https://github.com/jbosstm/narayana/pull/1229/files#diff-c0091e065ebb4dea411f68b3ea81f32d4b0ebe2b1349c2876073d576fdc505b8R422-R427

PR #2328, opened 2024-11-06
JBTM-3926 Move lra component into a new jbosstm project

marcosgopen:
https://issues.redhat.com/browse/JBTM-3926

related to https://github.com/jbosstm/lra/pull/1

https://github.com/jbosstm/jboss-as/pull/92

JDK21 !JDK11

CORE AS_TESTS RTS JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB PERFORMANCE LRA !DB_TESTS


scripts/hudson/narayana.sh
```diff
@@ -385,6 +366,28 @@ function clone_as {
...
+    [ $? -eq 0 ] || fatal "git fetch of pulls failed"
+    git checkout $LRA_BRANCH
+    [ $? -eq 0 ] || fatal "git fetch of pull branch failed"
+    cd ../
+    ./build.sh -f lra/pom.xml -B clean install
+    [ $? -eq 0 ] || fatal "Build of LRA failed"
+    LRA_VERSION=`awk '/lra-parent/ { while(!/<version>/) {getline;} print; }' lra/pom.xml | cut -d \< -f 2|cut -d \> -f 2`
+    OVERRIDE_LRA_VERSION=" -Dversion.org.jboss.narayana.lra=${LRA_VERSION} "
+  fi
+
```
jmfinelli, line 390: I was wondering if it makes sense to build WildFly with the main branch of lra in case `LRA_BRANCH` is not provided. WDYT @marcosgopen?

jmfinelli: I think I sent my previous message to quickly. What's the reason to build WildFly with a different LRA version if the Narayana repo won't test LRA at all? @marcosgopen

marcosgopen: This is an excellent question @jmfinelli. We are at a stage where no lra artifacts are present in maven central. Those artifacts are needed for Wildfly to be built. So at the moment I am building lra for having the artifacts needed by the  WildFly build process.

marcosgopen: When lra artifacts are available on nexus we can rely on WildFly to be able to retrieve the lra artifacts by itself, see  https://github.com/jbosstm/jboss-as/pull/92

jmfinelli: So, if I understand correctly, this script is a temporary patch to build WildFly while we are still in the middle of producing the first (indipendent) LRA artifact and update WildFly's bom. Is my understanding correct?

marcosgopen:
At the moment test are failing because this PR depends on https://github.com/jbosstm/lra/pull/1
So I put it on hold.

tomjenkinson (approved):
Having discussed with Marco the possibility of `LRA_*` types being left in base Narayana I think that being the case should not block the merge or release of the new LRA project. I do think we should raise a JBTM to investigate if it would be appropriate to move these numbers into the LRA project alone in future (and I would ask @marcosgopen to do that) but even if we did  that we would need to consider then how to prevent the numbers being reused in some conflicting manner then.

Thank you!

marcosgopen:
JBTMs created, thanks Tom:
https://issues.redhat.com/browse/JBTM-3949
https://issues.redhat.com/browse/JBTM-3950

PR #2331, opened 2024-11-11
JBTM-3945 handle store write errors

mmusgrov:
https://issues.redhat.com/browse/JBTM-3945

@marcosgopen This PR adds checks for object store write failures. I have only included tests for starting and enlisting with LRAs but I have another dozen or so fault injection test scenarios to write (I'll document them on the JBTM) but the required code changes are in this PR and any bugs I uncover with the extra failure scenarios should be less intrusive so another member of the team can review those while you are away.

!CORE !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE LRA !DB_TESTS !mysql !db2 !postgres oracle

!JDK11 !JDK21

### DIFF
--- a/rts/lra/client/src/main/java/io/narayana/lra/client/internal/NarayanaLRAClient.java
+++ b/rts/lra/client/src/main/java/io/narayana/lra/client/internal/NarayanaLRAClient.java
@@ -16,6 +16,7 @@ import jakarta.ws.rs.NotFoundException;
 import jakarta.ws.rs.POST;
 import jakarta.ws.rs.PUT;
 import jakarta.ws.rs.Path;
+import jakarta.ws.rs.ServiceUnavailableException;
 import jakarta.ws.rs.WebApplicationException;
 import jakarta.ws.rs.client.Client;
 import jakarta.ws.rs.client.ClientBuilder;
@@ -746,20 +747,20 @@ public class NarayanaLRAClient implements Closeable {
                     .get(JOIN_TIMEOUT, TimeUnit.SECONDS);
 
             String responseEntity = response.hasEntity() ? response.readEntity(String.class) : "";
+            // remove it and create tests for PRECONDITION_FAILED and NOT_FOUND
             if (response.getStatus() == Response.Status.PRECONDITION_FAILED.getStatusCode()) {
-                String logMsg = LRALogger.i18nLogger.error_tooLateToJoin(lraId, responseEntity);
+                String logMsg = LRALogger.i18nLogger.error_tooLateToJoin(String.valueOf(lraId), responseEntity);
                 LRALogger.logger.error(logMsg);
                 throw new WebApplicationException(logMsg,
                         Response.status(PRECONDITION_FAILED).entity(logMsg).build());
             } else if (response.getStatus() == NOT_FOUND.getStatusCode()) {
                 String logMsg = LRALogger.i18nLogger.info_failedToEnlistingLRANotFound(
-                        lraId, coordinatorUrl, NOT_FOUND.getStatusCode(), NOT_FOUND.getReasonPhrase(), GONE.getStatusCode(), GONE.getReasonPhrase());
+                        lraId, coordinatorUrl, NOT_FOUND.getStatusCode(), NOT_FOUND.getReasonPhrase(),
+                        GONE.getStatusCode(), GONE.getReasonPhrase());
                 LRALogger.logger.info(logMsg);
-                throw new WebApplicationException(logMsg);
+                throw new WebApplicationException(Response.status(GONE).entity(logMsg).build());
             } else if (response.getStatus() != OK.getStatusCode()) {
-                String logMsg = LRALogger.i18nLogger.error_failedToEnlist(lraId, coordinatorUrl, response.getStatus());
-                LRALogger.logger.error(logMsg);
-                throwGenericLRAException(uri, response.getStatus(), logMsg, null);
+                throw new WebApplicationException(responseEntity, response);
             }
 
             String recoveryUrl = null;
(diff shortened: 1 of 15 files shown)


rts/lra/client/src/main/java/io/narayana/lra/client/internal/NarayanaLRAClient.java
```diff
@@ -737,54 +738,53 @@ public URI enlistCompensator(URI uri, Long timelimit, String linkHeader, StringB
...
                 headers.add("Link", linkHeader);
                 response = client.target(coordinatorUrl)
-                    .path(LRAConstants.getLRAUid(uri))
-                    .queryParam(TIMELIMIT_PARAM_NAME, timelimit)
-                    .request()
-                    .headers(headers)
-                    .async()
-                    .put(Entity.text(compensatorData == null ? linkHeader : data))
-                    .get(JOIN_TIMEOUT, TimeUnit.SECONDS);
+                        .path(LRAConstants.getLRAUid(uri))
```
mmusgrov, line 741: These next 25 lines or so move them to the right by a tab - I did this because the control flow alignment was difficult to parse.

rts/lra/client/src/main/java/io/narayana/lra/client/internal/NarayanaLRAClient.java
```diff
@@ -746,20 +747,20 @@ public URI enlistCompensator(URI uri, Long timelimit, String linkHeader, StringB
                     .get(JOIN_TIMEOUT, TimeUnit.SECONDS);
 
             String responseEntity = response.hasEntity() ? response.readEntity(String.class) : "";
+            // remove it and create tests for PRECONDITION_FAILED and NOT_FOUND
             if (response.getStatus() == Response.Status.PRECONDITION_FAILED.getStatusCode()) {
-                String logMsg = LRALogger.i18nLogger.error_tooLateToJoin(lraId, responseEntity);
+                String logMsg = LRALogger.i18nLogger.error_tooLateToJoin(String.valueOf(lraId), responseEntity);
```
tomjenkinson, line 752: I expect this has a reason to need to be String.valueOf - I am not necessarily asking for that to be explained in the code but I am curious why it needed to changed?

tomjenkinson: Actually, I presume it has something to do with how a URL is interpreted in https://github.com/jbosstm/narayana/pull/2331/files#diff-e4c877e90cd6988c76f4b37dbd005f96444a75919262742fd845d3718f2e2fbaL54 - so I am curious what happens if you pass the object in as a URL (as it was) rather than a String (as it is now)? Do you get an object reference printed out or something?

mmusgrov: I changed the log parameter for the id from URL to String: https://github.com/jbosstm/narayana/pull/2331/files#diff-e4c877e90cd6988c76f4b37dbd005f96444a75919262742fd845d3718f2e2fbaL54

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
```diff
@@ -363,10 +370,10 @@ public int joinLRA(StringBuilder recoveryUrl, URI lra, long timeLimit,
 
                         if (!LRAConstants.AFTER.equals(rel)) {
                             // participants are not allowed to join inactive LRAs
-                            return Response.Status.PRECONDITION_FAILED.getStatusCode();
+                            return Response.Status.GONE.getStatusCode(); // report it as GONE
```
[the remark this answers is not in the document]
marcosgopen: I am also curious to learn more about this. I assume that the lra is already completed and so inactive, in that case the GONE response is more appropriate according to https://github.com/eclipse/microprofile-lra/blob/main/spec/src/main/asciidoc/microprofile-lra-spec.asciidoc?plain=1#L302-L304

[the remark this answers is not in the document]
mmusgrov: I've just successfully reran the tests without the change so I'll revert this bit. But, for completeness, here is the rationale for the change:

I think it's what the code was returning before, it may or may not have been going through this catch block: https://github.com/jbosstm/narayana/pull/2331/files#diff-34b2877655e9cabf2c9bbe0e080b7c5060095f760099cee6dc74b6e4ff57cb65L783, also there is an eclipse microprofile spec test that tends to agree with this view (https://github.com/eclipse/microprofile-lra/blob/2.0/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java#L361). And the analysis I did on JBTM-3331 (https://issues.redhat.com/browse/JBTM-3331?focusedId=25955017&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-25955017) back in 2020 indicated that during the spec process we switched from using 404 Not Found preferring instead to use 410 Gone or 412 Precondition Failed.

marcosgopen (approved):
Thanks for your PR @mmusgrov !
The PR LGTM, I would prefer to remove that bit related to the returning code. But I would be OK to review it in a second PR. If you are going to update the PR with the mentioned change and tests are successful I give my approval for this PR.

rts/lra/client/src/main/java/io/narayana/lra/client/internal/NarayanaLRAClient.java
mmusgrov: > Actually, I presume it has something to do with how a URL is interpreted in https://github.com/jbosstm/narayana/pull/2331/files#diff-e4c877e90cd6988c76f4b37dbd005f96444a75919262742fd845d3718f2e2fbaL54 - so I am curious what happens if you pass the object in as a URL (as it was) rather than a String (as it is now)? Do you get an object reference printed out or something?

It doesn't compile: `incompatible types: java.net.URL cannot be converted to java.lang.String`

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/service/LRAService.java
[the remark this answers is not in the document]
mmusgrov: I have just pushed that change. I also added an extra commit to fix a confusing formatting problem in NarayanaLRAClient#enlistCompensator

rts/lra/coordinator/src/main/java/io/narayana/lra/coordinator/domain/model/LongRunningAction.java
```diff
@@ -770,8 +790,13 @@ public LRAParticipantRecord enlistParticipant(URI coordinatorUrl, String partici
...
                     // need to remember that there is a new participant
-                    deactivate(); // if it fails the superclass will have logged a warning
-                    savedIntentionList = true; // need this clean up if the LRA times out
+                    if (deactivate()) { // if it fails the superclass will have logged a warning
+                        savedIntentionList = true; // need this clean up if the LRA times out
+                    } else {
+                        throw new ServiceUnavailableException(LRALogger.i18nLogger.warn_saveState(DEACTIVATE_REASON));
+                    }
+                } else {
+                    throw new ServiceUnavailableException(LRALogger.i18nLogger.warn_saveState(DEACTIVATE_REASON));
```
tomjenkinson, line 799: Please can you let me know why at this point detecting the participant was null (I presume, not enlisted) should be reported as inability to save state related to deactivation?

tomjenkinson: Thank you!

tomjenkinson (approved):
LGTM - thank you

PR #2335, opened 2024-11-19
[JBTM-3894] AtomicActionRecoveryModule needs to try recovery again when there are problems

jmfinelli:
[JBTM-3894](https://issues.redhat.com/browse/JBTM-3894)

AtomicActionRecoveryModule needs to try recovery again when there are problems

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/AtomicActionRecoveryModule.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/AtomicActionRecoveryModule.java
@@ -47,6 +47,8 @@ public class AtomicActionRecoveryModule implements RecoveryModule {
         boolean AtomicActions = false;
         // Does not block the suspension of the Recovery Manager by default
         this.hasWorkLeftToDo = false;
+        // Does not report problems
+        this.problemDuringRecovery = false;
 
         // uids per transaction type
         InputObjectState aa_uids = new InputObjectState();
@@ -59,6 +61,7 @@ public class AtomicActionRecoveryModule implements RecoveryModule {
             AtomicActions = _recoveryStore.allObjUids(_transactionType, aa_uids);
 
         } catch (ObjectStoreException ex) {
+            problemDuringRecovery = true;
             tsLogger.i18NLogger.warn_recovery_AtomicActionRecoveryModule_1(ex);
         }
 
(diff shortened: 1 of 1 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/AtomicActionRecoveryModule.java
```diff
@@ -253,4 +257,6 @@ public boolean hasWorkLeftToDo() {
 
     private boolean hasWorkLeftToDo;
 
+    private boolean problemDuringRecovery;
```
tomjenkinson, line 260: Please can you explain the advantage of a new property rather than just setting hasWorkLeftToDo in these places? Our API method is called hasWorkLeftToDo() so how does the implementation benefit from having this new variable used? A consequence is that the hasWorkLeftToDo() method seems a bit more complicated now too because it is considering problemDuringRecovery as well as the previous hasWorkLeftToDo

jmfinelli: I think that a dedicated variable that tracks problems is a good separation of concerns: `hasWorkLeftToDo` will track left-over transactions while `problemDuringRecovery` will track exceptions/problems. This is also in line with `XARecoveryModule` (and my other PR #2336)

jmfinelli: > A consequence is that the hasWorkLeftToDo() method seems a bit more complicated now too because it is considering problemDuringRecovery as well as the previous hasWorkLeftToDo

With my modifications, `hasWorkLeftToDo()` will still return when there is "work left to do" but that situation will be associated to having either left-over tranasctions or problems during recovery

tomjenkinson: OK, that is good for me - thank you

tomjenkinson (approved):

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/AtomicActionRecoveryModule.java
jmfinelli: Thanks Tom

PR #2336, opened 2024-11-20
[JBTM-3893] XARecoveryModule now implements RecoveryModule.hasWorkLeftToDo()

jmfinelli:
[JBTM-3893](https://issues.redhat.com/browse/JBTM-3893)


ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
```diff
@@ -412,7 +419,6 @@ protected XARecoveryModule(XARecoveryResourceManager recoveryClass, String logNa
         _logName = logName;
         _recoveryManagerClass = recoveryClass;
         if (_recoveryManagerClass == null) {
-            this.setRecoveryProblems(true);
```
tomjenkinson: Please can you explain why this is no longer classed as a recovery problem?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
```diff
@@ -293,6 +295,11 @@ public String id() {
         return "XARecoveryModule:" + _recoveryManagerClass;
     }
 
+    @Override
+    public boolean hasWorkLeftToDo() {
```
tomjenkinson, line 299: Please can I request some tests similar to https://github.com/jbosstm/narayana/blob/main/ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/recovery/RecoverySuspendTest.java to be added for the XARecoveryModule work left to do checking?

jmfinelli: Yep sure, I was already working on testing but I wanted to start a `Draft` PR ASAP. Thanks for asking :smile:

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
jmfinelli: IMO, this is not a problem during recovery but rather a problem during the initialisation of Narayana.

jmfinelli: Plus, as `setRecoveryProblems(true)` is invoked in the constructor, as soon as `periodicWorkFirstPass()` is invoked from `PeriodicRecovery`, `recoveryProblems` will be set to false (i.e. `setRecoveryProblems(false)`)

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
tomjenkinson: Thank you

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
tomjenkinson: I think that if is null then we will eventually end up in https://github.com/jbosstm/narayana/blob/main/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java#L211 because of a NullPointerException at https://github.com/jbosstm/narayana/blob/main/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java#L198 so that would make `setRecoveryProblems(true)`.

I am not super sure I agree that an invalid configuration should not be considered a problem for recovery but we can think more on this.

jmfinelli: IMO, `recoveryProblem` shouldn't be used to report problems during initialisation (at least, as far as I can tell from other use cases). Moreover, before we end up at line [211](https://github.com/jbosstm/narayana/blob/main/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java#L211), we hit [setRecoveryProblems(false)](https://github.com/jbosstm/narayana/blob/main/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java#L188), which makes setting `recoveryProblems` to true during initialisation pointless

jmfinelli:
AS_TESTS failures are not related to this PR.
cc @tomjenkinson

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
```diff
@@ -0,0 +1,205 @@
...
+        // Make sure that the test environment is ready
+        BytemanControlledXAResource.resetCommitCallCounter();
+        BytemanControlledXAResource.resetRollbackCallCounter();
+        // The following instruction makes the BytemanControlledXAResource commit without failures.
+        BytemanControlledXAResource.setGreenFlag();
+        BytemanControlledXAResource.setCommitReturn(XAResource.XA_OK);
+
+        // BytemanControlledXAResource.getCommitCallCounter() should be 1 as:
+        // - One invocation from the normal commit procedure should pass
+        startTest(1,0, 0);
```
tomjenkinson, line 117: I think for this test to be meaningful from the commit point of view, we would want to see the recovery manager re-issue the commit call. Otherwise we can't be sure the XARecoveryModule is even used in this test (hopefully it is [via the AARM] but maybe we can assert that happened)?

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
```diff
@@ -0,0 +1,205 @@
...
+        // BytemanControlledRecord.getCommitCallCounter() should be 3 as:
+        // - One invocation from the normal commit procedure should fail
+        // - One invocation from the recovery process should fail
+        // - One invocation from the recovery process should pass
+        startTest(1, 2, 0);
+    }
+
+    private void startTest(int numberOfCommits, int numberOfRollbacks, int timeout) {
+
+        createXATransaction(new BytemanControlledXAResource(), timeout);
```
tomjenkinson, line 142: Can we verify after this call that no rollbacks where issued yet? It might be a bit unpredictable though as the PeriodicRecovery might have done it...

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
```diff
@@ -0,0 +1,205 @@
...
+            try {
+                // Let's re-check in 1 second
+                Thread.sleep(1000);
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
+            }
+        }
+    }
+
+    private static class DummyXAResourceOrphanFilter implements XAResourceOrphanFilter {
```
tomjenkinson, line 193: Why can't we use the real orphan filters?

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
```diff
@@ -0,0 +1,205 @@
...
+        // the test set of modules
+        _recoveryConfig.setRecoveryModuleClassNames(Arrays.asList(modules));
+
+        _jtaEnvironmentBean.setXaResourceRecoveryClassNames(
+                List.of(BytemanControlledXAResourceRecovery.class.getName()));
+
+        // obtain a new RecoveryManager with the above config:
+        _manager = RecoveryManager.manager(RecoveryManager.INDIRECT_MANAGEMENT);
+        // recovery can start
+        _manager.startRecoveryManagerThread();
```
tomjenkinson, line 82: Do I recall that recently you were proposing to remove this call as it was not needed?

tomjenkinson: If it is needed, that might help with the asserting rollback was not called thing I was talking about below (i.e. that rollback was not called during the `createXATransaction` call...

jmfinelli: > Do I recall that recently you were proposing to remove this call as it was not needed?

Yep, you do recall correctly. I'll remove that instruction. Thanks for spotting it

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
jmfinelli: I think that this test can be removed. WDYT?

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
jmfinelli: I have reworked this part of the class. WDYT?

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
jmfinelli: Good point. I'll use `NodeNameXAResourceOrphanFilter`. Thanks

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
jmfinelli: I think we should be safe

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
tomjenkinson: What do you think about having the test proving the AARM will commit this (as in commit a recovered XAResource)? I think it would be a useful addition.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
tomjenkinson: Great - thank you

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
tomjenkinson: It would be something like: `testSuspensionWhenThereIsAnXAResourceToCommit`

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
tomjenkinson: I am not spotting it actually.  Apart from the unpredictability I would imagine something like: 
`Assertions.assertEquals(0, BytemanControlledXAResource.getRollbackCallCounter());`

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
```diff
@@ -0,0 +1,199 @@
...
+        TxControl.disable();
+
+        // Makes sure that the transaction reaper completed all transactions
+        TransactionReaper.transactionReaper().waitForAllTxnsToTerminate();
+
+        // Makes sure that all transactions without a timeout have passed the `prepare` phase
+        waitForTransactionsWithoutTimeout();
+
+        // Synchronization is not needed (i.e. async = true) as isWaitForWorkLeftToDo() == true
+        _manager.suspend(true);
```
tomjenkinson, line 161: Please can I ask you to draw my attention how we know that the rollback/commit we might look for occurred during the suspend call?

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
jmfinelli: What I did is:
* `_jtaEnvironmentBean.setOrphanSafetyInterval(periodicRecoveryPeriod);`, which would make the amount of time needed for a transaction to be considered by XARecoveryModule longer
* Start the recovery manager just before the transaction gets created

With these modifications, I don't think we need to `Assertions.assertEquals(0, BytemanControlledXAResource.getRollbackCallCounter());`. WDYT?

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
jmfinelli: What this test does is checking that the new feature of XARecoveryModule isn't blocking the suspension of the Recovery Manager when transactions commit successfully. I proposed to remove this test as I realised that AARM is the actual module that takes care of this logic path...and we already have a test in ArjunaCore. WDYT?

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
jmfinelli: Orphan transactions will be processed only after an interval that tells XARecoveryModule that it can use the collection of OrphanFilters to go through left-over XIDs. Creating/Starting the Recovery Manager just before an XA transaction gets created ensures that the XARecoveryModule will invoke `rollback` after `manager.suspend(true)`

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
tomjenkinson: I think this is different enough from "pure" AARM to warrant it's own test - at least perhaps. IIRC for the AARM test we did not have an XAR in there. So we don't know for sure that if we have an XAR that needs committing the server will hang waiting for it? Maybe it could through RMFAIL a few times in recovery to make this (new) test prove the server will hang when an XAR needs committing

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
tomjenkinson: Ah, OK. Personally I think I would put the assertion in anyway (in case someone has it paused in the debugger and doesn't realise and gets confused) but the gist of what you are saying +1.

It's not mandatory from my point of view to make the change I mentioned, just a nice to have I think

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
tomjenkinson: IIUC then you are relying on the timing of the recovery manager thread running? I wonder if we could use byteman to be sure but I understand the assumption.

That being said - the way these intervals are set up I think it's quite close that the recovery cycle could take place before this call. From what I can see a small delay of perhaps (guessing what to add together) 6 (so the config of 5 and the config of 1 both higher in this file) seconds might be enough to not make this test do what we expect and possibly recover before the suspend starts

I would prefer byteman to be sure of what is happening. But perhaps a more appropriate configurable assumption to use would be https://github.com/jbosstm/narayana/blob/main/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/RecoveryEnvironmentBean.java#L131 - but like I say, that is still an assumption. Really it would be good to see the recovery manager suspend and the rollback call being coordinated by byteman.

jmfinelli: I also thought to use byteman to synchronise when the first recovery scan should take place but I have the feeling that, at the end, we don't really need that level of control. The reason why I think so is that the XID returned by the `BytemanControlledXAResource` during recovery will be considered by the XARecoveryModule only after `periodicRecoveryPeriod * 2`, i.e. 10secs. Thus, even if there are multiple scans in the time window between the creation of the Recovery Manager and the invocation of `manager.suspend()`, the XARecoveryModule will process the leftover XID only during the Recovery Manager's suspension.

jmfinelli: At least, that is my understanding. Please, let me know if I got it wrong :-)

tomjenkinson: So if I understand correctly you are saying that because the recovery will take (say) 10 seconds minimally you expect to reach this instruction before that point and thus it should be considered safe?

Personally I think 10 seconds is not much different to 6 seconds but I suppose it's a bit bigger. I would worry a bit that on CI createXATransaction and waitForAllTxnsToTerminate and waitForTransactionsWithoutTimeout could combine to take longer than 10 seconds.

I don't really like not using Byteman for this and just relying on timeouts but maybe you could argue a 10 second timeout would (generally) be long enough if the instruction `_manager = RecoveryManager.manager(RecoveryManager.INDIRECT_MANAGEMENT);` were to be moved immediately prior to the `_manager.suspend(true);`

jmfinelli: > So if I understand correctly you are saying that because the recovery will take (say) 10 seconds minimally you expect to reach this instruction before that point and thus it should be considered safe?
>
> Personally I think 10 seconds is not much different to 6 seconds but I suppose it's a bit bigger. I would worry a bit that on CI createXATransaction and waitForAllTxnsToTerminate and waitForTransactionsWithoutTimeout could combine to take longer than 10 seconds.

In a production environment, I would agree with you 100pc. In this test class, createXATransaction, waitForAllTxnsToTerminate and waitForTransactionsWithoutTimeout will take much less than a second as there is only one transaction (without timeout) and it is immediately committed (as the test uses local dummy XAResources).

> I don't really like not using Byteman for this and just relying on timeouts but maybe you could argue a 10 second timeout would (generally) be long enough if the instruction `_manager = RecoveryManager.manager(RecoveryManager.INDIRECT_MANAGEMENT);` were to be moved immediately prior to the `_manager.suspend(true);`?

I was thinking about this while I was reading your comment :-) Although I'm a Byteman fanatic, I have the feeling that using a rendezvous in this case is an overkill. Nevertheless, if you have a strong preference for Byteman, I will employ it :-)

jmfinelli: > I kind of think that on CI it is possible for delays in processing to happen so I feel that either moving the start of the recovery manager to just before here or doing Byteman should be done. I am stopping short of saying either are mandatory.

`_manager = RecoveryManager.manager(RecoveryManager.INDIRECT_MANAGEMENT)` moved before `_manager.suspend()`. Thanks Tom

> The main part of the pull request that I think we do need to address is the thread where I talk about testSuspensionWhenThereIsAnXAResourceToCommit - being confident the server waits for transactions (that have an XAR) to commit is definitely an important part to prove in my opinion).

About `testSuspensionWhenThereIsXATxnToRecover`, Byteman makes sure that we pass two checkpoints before the test can succeed:
1. `RULE setwaitForWorkLeftToDoFlag` sets the flag `waitForWorkLeftToDoFlag` only when `PeriodicRecovery.suspendScan` was invoked while `_waitForWorkLeftToDo` was set to true
2. `RULE RULE setGreenFlag` sets the flag `greenFlag` only when point 1. is verified and `XARecoveryModule.hasWorkLeftToDo()` returns true (i.e. there is at least one transaction to recover)
3. `RULE FailRollback` will fail any invocation of `BytemanControlledXAResource.rollback()` until `greenFlag` is false (i.e. until we don't verify that XARecoveryModule sees the transaction to recover)

tomjenkinson: Thank you for accommodating a change on the recovery manager - I will resolve this thread. Just to mention the second point is about the test currently named `testSuspensionWhenThereArentTransactionsToRecover`. I am I still resolving this thread though. Thank you

ArjunaJTA/jta/tests/byteman-scripts/RecoverySuspendTest/recoverySuspendTest_BytemanControlledXAResource.btm
```diff
@@ -0,0 +1,101 @@
...
+# Rule to fail
+# BytemanControlledXAResource.commit()
+# until greenFlag gets set
+########################################################################
+
+RULE FailCommit
+CLASS com.hp.mwtests.ts.jta.common.BytemanControlledXAResource
+METHOD commit
+AT INVOKE deleteFile
+IF com.hp.mwtests.ts.jta.common.BytemanControlledXAResource._xaExceptionCode != 0
```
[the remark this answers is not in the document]
tomjenkinson: Actually I think the first time this method is called we should throw an RMFAIL so we can see the recovery manager commit the resource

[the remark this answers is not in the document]
jmfinelli: I don't think we need to use "greenFlag" here as we always need to fail `BytemanControlledXAResource.commit()` to test `XARecoveryModule`

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
jmfinelli: > I think this is different enough from "pure" AARM to warrant it's own test - at least perhaps. IIRC for the AARM test we did not have an XAR in there. So we don't know for sure that if we have an XAR that needs committing the server will hang waiting for it? Maybe it could through RMFAIL a few times in recovery to make this (new) test prove the server will hang when an XAR needs committing

I think I see your point. Thanks Tom.
AFAIU, AARM is the actual RecoveryModule that takes care of XA transactions. While I can definitely develop a new test to check that AARM makes PeriodicRecovery's suspend wait until all XA transactions are completed but it seems to me that we're not actually testing the new functionality of XARecoveryModule. But maybe I'm missing something here :confused:

jmfinelli: As I said, I see your point. If you prefer that I develop a new test (even though it would be similar to what we already have in AARM), I have no objections.

tomjenkinson: While I believe AARM is the part that will cause the commit of the XAR, I think AARM needs XARM to do it's work and so I don't think it is the only RecoveryModule needing to behave well in order that transactions which have an XAResource can be committed. I am going to look and see if I can be a bit more specific as that might help with the test.

ArjunaJTA/jta/tests/byteman-scripts/RecoverySuspendTest/recoverySuspendTest_BytemanControlledXAResource.btm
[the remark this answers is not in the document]
tomjenkinson: I am not super sure if I agree or not but I will resolve this thread to keep discussion focussed on what I am anticipating might be needed for the commit case.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
tomjenkinson: OK, so what I think this test would do is:

1. Throw XAER_RMFAIL from the commit which should leave the pending transaction in the object store. This is basically the setup to test what I think needs testing.
2. during suspend, make the first call to XAResource::recover(int) method throw an exception so the suspend has to wait because `recoveryProblems` is a true value (set at https://github.com/jmfinelli/narayana/blob/JBTM-3893/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java#L690). This is to verify that the server wont return from suspend when we have incomplete transactions that will commit.
3. Verify that XAResource::commit will be called a second time (this time by recovery) and this must be before the suspend returns

I don't think the existing AARM test on it's own does this, in particular I doubt it does step 2. You can make an argument that the other test case you are proposing in this file could be made to do this check of L690 in a rollback scenario (so step 2), but a rollback scenario wouldn't do step 3 after doing step 2 and so it feels different to me. Maybe it feels different to me because these two steps somehow speak to confidence in the interactions between AARM and XARM combined with the new RecoveryManager::suspend operation.

jmfinelli: Thanks for your comment, @tomjenkinson. Hopefully, the new modifications are close enough to the test you described :crossed_fingers: Please, let me know what you think :-)

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
```diff
@@ -0,0 +1,221 @@
...
+         */
+        startTest(2, 0, 1, 0);
+    }
+
+    @Test
+    @BMScripts(scripts = {
+            @BMScript("RecoverySuspendTest/recoverySuspendTest_BytemanControlledXAResource"),
+            @BMScript("RecoverySuspendTest/recoverySusptenTest_BytemanControlledXAResource_commitFailure"),
+            @BMScript("RecoverySuspendTest/recoverySusptenTest_BytemanControlledXAResource_rollbackFailure"),
+            @BMScript("RecoverySuspendTest/recoverySuspendTest_SkipAtomicActionRecoveryModule"),
```
tomjenkinson, line 125: Please can I ask if this part is strictly necessary? If it is I would be curious to know why and whether not doing AARM functionality might hide a possible problem in the interactions between XARM and AARM? If it's not, I presume it is just some optimization to speed up the test somehow?

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
tomjenkinson: Thank you!

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
jmfinelli: AARM is used only at the end of the test to clean the environment. IOW, instead of cleaning the transaction myself, I let the Recovery Manager (through AARM) commit the transaction (so that the Recovery manager won't get stuck during suspension). DISCLAIMER: this is an artificial scenario as the transaction that the AARM eventually commits has been already rolled back by the XA resource (locally to itself)...so that transaction should really end up with an heuristic outcome :face_with_peeking_eye: Thus, the outcome I "forced" through the use of AARM isn't specs compliant but it is good enough to clean the environment and test the XARM in isolation.

tomjenkinson: Got it - thank you! I think you should add a comment in this bit of the code too (and in the rule if you didn't already). I like where https://github.com/jbosstm/narayana/pull/2336/commits/15ed2341ecaad32336d2ded0a4819578c8550ea8 is going but I did add a comment there with a thought.

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/SimpleResourceXA_RETRYHeuristicRollback.java
```diff
@@ -22,8 +22,11 @@ public void commit(Xid xid, boolean onePhase) throws XAException {
             throw new XAException(XAException.XA_RETRY);
         }
         xid = null;
-        committed = true;
-        throw new XAException(XAException.XA_HEURRB);
+
+        if (!committed) {
+            committed = true;
+            throw new XAException(XAException.XA_HEURRB);
```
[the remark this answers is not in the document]
jmfinelli: Cool, I'll do that. Thanks Tom

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/RecoverySuspendTest.java
jmfinelli: Comment added to the test. Thanks for the suggestion :-)

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/recovery/SimpleResourceXA_RETRYHeuristicRollback.java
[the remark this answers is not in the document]
tomjenkinson: Thank you!

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
```diff
@@ -1136,6 +1149,7 @@ private enum ScanStates {
 
     private final List<XAResourceOrphanFilter> _xaResourceOrphanFilters;
 
+    // Once JBTM-3953 is completed, convert this to normal boolean
```
tomjenkinson, line 1152: If you haven't already, I would recommend you to raise a JBTM for this work. Is that OK?

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/common/BytemanControlledXAResource.java
```diff
@@ -0,0 +1,224 @@
+package com.hp.mwtests.ts.jta.common;
```
tomjenkinson, line 1: Hmm, I think it is missing a copyright header. I think it's possible all the new files here are. Please can I ask you to check  if we put the copyright headers on btm files and if so add it to those, but for the Java ones I would presume we do so please may I ask for that?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
jmfinelli: Yep, it should be [this one](https://issues.redhat.com/browse/JBTM-3953)

ArjunaJTA/jta/tests/classes/com/hp/mwtests/ts/jta/common/BytemanControlledXAResource.java
jmfinelli: Added it. Thanks Tom

tomjenkinson: Thank you

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/recovery/arjunacore/XARecoveryModule.java
jmfinelli: Well, this modification should be part of JBTM-3953

tomjenkinson: Thanks, I was thinking specifically about the conversion to the normal boolean?

tomjenkinson (approved):
Thank you!

PR #2341, opened 2024-11-28
Removed useless instruction

jmfinelli:
JBTM-3894: Removed useless instruction


tomjenkinson (approved):

mmusgrov:
What made it useless?

jmfinelli:
> What made it useless?

Basically, we were trying to start the Recovery Manager's thread twice. However, the second time (i.e. the useless instruction that was removed) we were just checking if the Recovery Manager's thread was alive and then we were skipping the `start()` instruction

mmusgrov:
Ah I see, a useful comment, thanks.

PR #2346, opened 2025-01-15
JBTM-3881 rename the method local variable lastResource because it's …

mmusgrov:
…causing a weird side effect (three byteman based JMS + JTS + byteman tests failed in an external test suite)

the fix is to rename the offending local variable - the reason the bug is weird is because it is the only occurrence of a variable with that name in BasicAction (apart from RecordType.LASTRESOURCE).

https://issues.redhat.com/browse/JBTM-3881

CORE !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !DB_TESTS !mysql !db2 !postgres !oracle

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/BasicAction.java
@@ -2720,8 +2720,8 @@ public class BasicAction extends StateManager
              * the pending list. Otherwise it is moved to another list or
              * dropped if readonly.
              */
-            boolean lastResource = record.typeIs() == RecordType.LASTRESOURCE;
-            if (lastResource && prevLastResource == null) {
+            boolean isLastResource = record.typeIs() == RecordType.LASTRESOURCE;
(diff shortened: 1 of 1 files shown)


mmusgrov:
The symptom is that the record type of the [JTS ExtendedResourceRecord](https://github.com/jbosstm/narayana/blob/7.1.0.Final/ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/resources/ExtendedResourceRecord.java#L225) that gets added to the log for the transaction is read back as  [USER_DEF_FIRST0](https://github.com/jbosstm/narayana/blob/7.1.0.Final/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/RecordType.java#L43) and these by default cannot be instantiated so the record is left in the store during recovery.

tomjenkinson (approved):

mmusgrov:
FYI I tested the fix by building my branch containing the fix and then buildt a version of wildfly that contains the fix:

> cd <wildfly>; ./mvnw install -Dversion.org.jboss.narayana=7.0.2.Final-SNAPSHOT -DskipTests

PR #2347, opened 2025-01-20
JBTM-3973  ObjectStoreBrowser doesn't allow to register external OSBTypeHandler

marcosgopen:
https://github.com/jbosstm/jboss-as/pull/95/

https://issues.redhat.com/browse/JBTM-3973

CORE AS_TESTS RTS JACOCO XTS QA_JTA QA_JTS_OPENJDKORB !PERFORMANCE !DB_TESTS

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/OSBTypeHandler.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/OSBTypeHandler.java
@@ -4,7 +4,9 @@
  */
 package com.arjuna.ats.arjuna.tools.osb.mbean;
 
+import com.arjuna.ats.arjuna.common.Uid;
 import com.arjuna.ats.arjuna.logging.tsLogger;
+import com.arjuna.common.internal.util.ClassloadingUtility;
 
 import java.io.File;
 import java.lang.reflect.Constructor;
@@ -27,6 +29,7 @@ public class OSBTypeHandler {
     private String beanClass; // the JMX mbean representation of the record type
     private String typeName; // the type name {@link com.arjuna.ats.arjuna.coordinator.AbstractRecord#type()}
     HeaderStateReader headerStateReader;
+    private ClassLoader classLoaderToUse; // classLoader to load external bean types
 
     public OSBTypeHandler(boolean enabled, String recordClass, String beanClass, String typeName, String headerStateReaderClassName) {
         this(enabled, true, recordClass, beanClass, typeName, headerStateReaderClassName);
@@ -39,6 +42,17 @@ public class OSBTypeHandler {
         this.beanClass = beanClass;
         this.typeName = typeName;
         this.headerStateReader = headerStateReaderClassName == null ? new HeaderStateReader() : createHeader(headerStateReaderClassName);
+        this.classLoaderToUse = this.getClass().getClassLoader();
+    }
+
+    public OSBTypeHandler(boolean enabled, boolean allowRegistration, String recordClass, String beanClass, String typeName, String headerStateReaderClassName, ClassLoader classLoaderToUse) {
+        this.enabled = enabled;
+        this.allowRegistration = allowRegistration;
+        this.recordClass = recordClass;
+        this.beanClass = beanClass;
+        this.typeName = typeName;
+        this.headerStateReader = headerStateReaderClassName == null ? new HeaderStateReader() : createHeader(headerStateReaderClassName);
+        this.classLoaderToUse = classLoaderToUse;
     }
 
     private static HeaderStateReader createHeader(String headerStateReaderClassName) {
@@ -83,4 +97,27 @@ public class OSBTypeHandler {
         return headerStateReader;
     }
 
+    // This method is intended for the ObjStoreBrowser only
+    // the classLoaderToUse is needed here to load an external bean
+    UidWrapper createMBean(ObjStoreBrowser objStoreBrowser, String beantype, String type, String stateType, Uid uid,
+            boolean registerBean) {
+        UidWrapper w = new UidWrapper(objStoreBrowser, beantype, type, stateType, uid, registerBean);
+
+        OSEntryBean mbean = null;
+
+        try {
+            Class<OSEntryBean> cl = (Class<OSEntryBean>) classLoaderToUse.loadClass(beantype);
+            Constructor<OSEntryBean> constructor = cl.getConstructor(UidWrapper.class);
+            mbean = constructor.newInstance(w);
+        } catch (Throwable e) { // ClassNotFoundException, NoSuchMethodException, InvocationTargetException, IllegalAccessException, InstantiationException
+            tsLogger.i18NLogger.warn_osb_MBeanCtorFail(e);
+            mbean = new OSEntryBean(w);
+        }
+
+        mbean.activate();
+
+        w.setMBean(mbean);
+
+        return w;
+    }
 }
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/ObjStoreBrowser.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/ObjStoreBrowser.java
@@ -26,6 +26,7 @@ import com.arjuna.ats.arjuna.objectstore.ObjectStoreIterator;
 import com.arjuna.ats.arjuna.objectstore.StoreManager;
 import com.arjuna.ats.arjuna.state.InputObjectState;
 import com.arjuna.ats.arjuna.tools.osb.util.JMXServer;
+import com.arjuna.common.internal.util.propertyservice.BeanPopulator;
 
 /**
  * An MBean implementation for walking an ObjectStore and creating/deleting MBeans
@@ -148,6 +149,7 @@ public class ObjStoreBrowser implements ObjStoreBrowserMBean {
     // A system property for defining extra bean types for instrumenting object store types
     // The format is OSType1=BeanType1,OSType2=BeanType2,etc
     public static final String OBJ_STORE_BROWSER_HANDLERS = "com.arjuna.ats.arjuna.tools.osb.mbean.ObjStoreBrowserHandlers";
+    private static volatile ObjStoreBrowser objectStoreBrowser;
     private final String objStoreBrowserMBeanName;
 
     public static HeaderStateReader getHeaderStateUnpacker(String type) {
@@ -264,6 +266,15 @@ public class ObjStoreBrowser implements ObjStoreBrowserMBean {
     }
 
     /**
+     * Expose a method to passing in a type handler that can be used as a factory to create
+     * @param typeName
+     * @param osbTypeHander
+     */
+    public void addOSBTypeHandler(String typeName, OSBTypeHandler osbTypeHander) {
+        osbTypeMap.put(typeName, osbTypeHander);
+    }
+
+    /**
      * @param handler specification for handling object store types
      * @return the previous value associated with type handler, or null if there was no previous handler.
      */
@@ -317,6 +328,8 @@ public class ObjStoreBrowser implements ObjStoreBrowserMBean {
      * In other words, this class should be considered as a singleton class. In fact, if two ObjStoreBrowser instances
      * are created, they may interfere with each other.</p>
      */
+    // Deprecated, use the getInstance method instead since this is a singleton instance
+    @Deprecated
     public ObjStoreBrowser() {
         this(null);
     }
@@ -328,9 +341,26 @@ public class ObjStoreBrowser implements ObjStoreBrowserMBean {
      * In other words, this class should be considered as a singleton class. In fact, if two ObjStoreBrowser instances
      * are created, they may interfere with each other.</p>
      */
-   public ObjStoreBrowser(String logDir) {
+     // Deprecated, use the getInstance method instead since this is a singleton instance
+     // if needed you can set the logDir as 'arjPropertyManager.getObjectStoreEnvironmentBean().setObjectStoreDir(logDir)'
+    @Deprecated(forRemoval = true)
+    public ObjStoreBrowser(String logDir) {
         init(logDir);
         this.objStoreBrowserMBeanName = arjPropertyManager.getObjectStoreEnvironmentBean().getJmxToolingMBeanName();
+    }
+
+
+    public static ObjStoreBrowser getInstance() {
+        ObjStoreBrowser result = objectStoreBrowser;
+        if (result != null) {
+            return result;
+        }
+        synchronized (ObjStoreBrowser.class) {
+            if (objectStoreBrowser == null) {
+                objectStoreBrowser = new ObjStoreBrowser();
+            }
+            return objectStoreBrowser;
+        }
     }
 
     /**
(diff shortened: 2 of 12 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/UidWrapper.java
```diff
@@ -209,4 +209,8 @@ public boolean isAllowRegistration() {
 	public void setAllowRegistration(boolean allowRegistration) {
 		this.allowRegistration = allowRegistration;
 	}
+
+	public void setMBean(OSEntryBean mbean) {
```
tomjenkinson, line 213: @marcosgopen _maybe_ we consider to move this to package local access? What do you think? UidWrapper is part of the public API of Narayana and I am not sure we necessarily want to encourage users to try to set this themselves?

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/ObjStoreBrowser.java
```diff
@@ -341,9 +338,23 @@ public ObjStoreBrowser() {
...
-   public ObjStoreBrowser(String logDir) {
-        init(logDir);
-        this.objStoreBrowserMBeanName = arjPropertyManager.getObjectStoreEnvironmentBean().getJmxToolingMBeanName();
+   private ObjStoreBrowser(String logDir) {
+       init(logDir);
+       this.objStoreBrowserMBeanName = arjPropertyManager.getObjectStoreEnvironmentBean().getJmxToolingMBeanName();
+    }
+
+
+    public static ObjStoreBrowser getInstance(String logDir) {
```
tomjenkinson, line 347: I am not super -1 on this and not sure if there is a better option but it's a bit - I don't know the right words - strange or unintuitive (perhaps?)  -to see the calls with getInstance(null) used. Like I say I don't know what to suggest but basically once initialized this logDir parameter is going to be ignored...

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/OSBTypeHandler.java
```diff
@@ -83,4 +97,25 @@ public HeaderStateReader getHeaderStateReader() {
         return headerStateReader;
     }
 
+    public UidWrapper createMBean(ObjStoreBrowser objStoreBrowser, String beantype, String type, String stateType, Uid uid,
```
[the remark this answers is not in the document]
marcosgopen: Good point, I'll update the method modifier

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/OSBTypeHandler.java
```diff
@@ -83,4 +97,25 @@ public HeaderStateReader getHeaderStateReader() {
...
+            boolean registerBean) {
+        UidWrapper w = new UidWrapper(objStoreBrowser, beantype, type, stateType, uid, registerBean);
+
+        OSEntryBean mbean = null;
+
+        try {
+            Class<OSEntryBean> cl = (Class<OSEntryBean>) classLoaderToUse.loadClass(beantype);
```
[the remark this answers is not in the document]
marcosgopen: Good suggestion, I 'll add a comment

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/UidWrapper.java
marcosgopen: Sure, thanks

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/ObjStoreBrowser.java
tomjenkinson: +1, that makes sense to me thank you!

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/ObjStoreBrowser.java
```diff
@@ -328,9 +338,23 @@ public ObjStoreBrowser() {
      * In other words, this class should be considered as a singleton class. In fact, if two ObjStoreBrowser instances
      * are created, they may interfere with each other.</p>
      */
-   public ObjStoreBrowser(String logDir) {
-        init(logDir);
-        this.objStoreBrowserMBeanName = arjPropertyManager.getObjectStoreEnvironmentBean().getJmxToolingMBeanName();
+   private ObjStoreBrowser(String logDir) {
```
tomjenkinson, line 341: We can't just remove the constructor without deprecating it - people could be using this method already in their apps. Is it possible to leave the old way available for apps that want to use that constructor still?

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/tools/ExposeAllLogsTest.java
```diff
@@ -106,24 +104,8 @@ private void test(boolean exposeAllLogsViaJMX) throws Exception
...
-
-        if (useJMX)
-            osb.setExposeAllRecordsAsMBeans(exposeAllLogs);
-
-        osb.probe();
-    }
-
     // Given a set of MBean names find their corresponding Uids
-    private Map<Uid, ObjectName> getUids(Map<Uid, ObjectName> uids, Set<ObjectName> osEntries) {
+    protected static Map<Uid, ObjectName> getUids(Map<Uid, ObjectName> uids, Set<ObjectName> osEntries) {
```
tomjenkinson, line 108: Please can I ask you to clarify why the scope of the method needed to change? And was the same reason needed for the other getUids and the getProperty methods in this class?

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/ObjStoreBrowser.java
jmfinelli: My +1 goes to make this (and the other) constructor(s) private. This class should only be used through a static get method

tomjenkinson: My point is removing the constructor for this class will break existing applications. It needs to be planned in to allow time for users to migrate

marcosgopen: Yes, that is ok. But since it is preferable to use the getInstance instead I would deprecate it and add a comment.

marcosgopen: Please consider the class is already deprecated and internal. But I am ok to add the deprecation to the method too.

tomjenkinson: I think it's worth separately tracking this deprecation as it's a different use case

jmfinelli: Sorry @tomjenkinson, I didn't see you were asking to deprecate it first and then remove it later. That works for me as well.

+1 to open a different JBTM (unfortunately)

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/tools/ObjStoreBrowserTest.java
```diff
@@ -91,7 +92,8 @@ public void testObjectStoreBrowser() throws Exception {
 	@Test
 	public void basicOSBTest () throws Exception
 	{
-		ObjStoreBrowser osb = new ObjStoreBrowser("os");
+		arjPropertyManager.getObjectStoreEnvironmentBean().setObjectStoreDir("os");
```
marcosgopen, line 95: The envinomentBean property already exists. Since we don't want the StoreDir value to change it would be better to set it in the environment bean only.

marcosgopen:
> AS_TESTS profile tests failed (https://ci-jenkins-csb-narayana.apps.ocp-c1.prod.psi.redhat.com/job/btny-pulls-narayana/PROFILE=AS_TESTS,jdk=openJDK17,label=jnlp-agent/650/): AS tests failed

This error is not related to this PR, but related to testcontainer.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/tools/ObjStoreBrowserTest.java
tomjenkinson: Hmm, something to think about is that the ObjectStoreEnvironmentBean:::setObjectStoreDir could be seen as slightly different use case to what the OSB would use it for in that possibly an OSB user wants to connect to a different object store than the one that the transaction manager in memory is configured to use. That being said, I don't think that use case should be particularly common if at all and so I am not -1 to relying on that field. That being said, we would want a release note for it so we need to make sure the release note for JBTM-3973 users of the added getInstance method to understand the impact and of the environment bean config on the OSB.

marcosgopen: Good point, I'll add a note on the JBTM

marcosgopen: Actually the release note for this should go for the new JBTM that track the deprecation and removal of the constructor accepting the logDir as parameter I think. Do you agree @tomjenkinson ?

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/ObjStoreBrowser.java
marcosgopen: Thanks @tomjenkinson and @jmfinelli . I opened it here: https://issues.redhat.com/browse/JBTM-3974

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/tools/ObjStoreBrowserTest.java
marcosgopen: Thinking again to your message I believe we need a release note for both the JBTMs. For the JBTM this PR addresses I wrote "Deprecating the ObjStoreBrowser constructors in favor of the getInstance method."

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/ObjStoreBrowser.java
```diff
@@ -328,9 +341,26 @@ public ObjStoreBrowser() {
      * In other words, this class should be considered as a singleton class. In fact, if two ObjStoreBrowser instances
      * are created, they may interfere with each other.</p>
      */
-   public ObjStoreBrowser(String logDir) {
-        init(logDir);
-        this.objStoreBrowserMBeanName = arjPropertyManager.getObjectStoreEnvironmentBean().getJmxToolingMBeanName();
+     // Deprecated, use the getInstance method instead since this is a singleton instance
+     // if needed you can set the logDir as 'arjPropertyManager.getObjectStoreEnvironmentBean().setObjectStoreDir(logDir)'
+    @Deprecated(forRemoval = true)
+     private ObjStoreBrowser(String logDir) {
```
tomjenkinson, line 347: I think this should still be public as moving it to private is effectively removing it from the public API. It's not ideal but I think we need to keep it for allowing people to move forwards at least until Narayana 8

marcosgopen: Sorry, I forgot to re setting it to public. I'll update it too. Thanks

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/ObjStoreBrowser.java
```diff
@@ -328,9 +341,26 @@ public ObjStoreBrowser() {
...
+
+
+    public static ObjStoreBrowser getInstance() {
+        ObjStoreBrowser result = objectStoreBrowser;
+        if (result != null) {
+            return result;
+        }
+        synchronized (ObjStoreBrowser.class) {
+            if (objectStoreBrowser == null) {
+                objectStoreBrowser = new ObjStoreBrowser();
```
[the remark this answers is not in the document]
marcosgopen: I don't think that is necessary. The init method overwrites that property if a parameter is passed (https://github.com/jbosstm/narayana/blob/main/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/ObjStoreBrowser.java#L286). 
 If the user does it before the instance is created all should be already set at this point.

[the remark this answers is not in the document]
tomjenkinson: Ah! That is doing something I hadn't considered but you are right basically we are removing the ability to force a call to `ObjectStoreEnvironmentBean::setObjectStoreDir` - thank you

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/tools/ExposeAllLogsTest.java
```diff
@@ -80,17 +88,7 @@ private void test(boolean exposeAllLogsViaJMX) throws Exception
         // get uids via JMX
         getUids(uids2, agent.queryNames(osMBeanName + ",*", null));
 
-        // and validate that there is no MBean corresponding to u
-        assertFalse(uids2.containsKey(u));
```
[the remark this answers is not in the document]
marcosgopen: Thank you  @tomjenkinson for your question. In the original version of the test there were 2 tests:
1. the first ObjStoreBrowser was created with exposeAllLogs disabled, the uids2 was empty, then a new ObjStoreBrowser was created with exposeAllLogs enabled and the uids2 contained the u record.
2. the first ObjStoreBrowser was created with exposeAllLogs disabled, the uids2 was empty, then a new ObjStoreBrowser was created with exposeAllLogs disabled (again) and the uids2 should still have been empty but it was not !! 
This was because even when JMX was not enabled the osb.init() method was 'enabling' it https://github.com/jbosstm/narayana/blob/main/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/ObjStoreBrowser.java#L291-L292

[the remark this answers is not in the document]
marcosgopen: Getting back to your original question: since in the new version of the test the exposeAllLogs is enabled since the beginning (and until the end) the uids2.containsKey(u) is true.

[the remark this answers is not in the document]
tomjenkinson: Thank you, I understand the point about the exposeAllLogs not being false and this list therefore not being empty now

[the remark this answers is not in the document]
tomjenkinson: I also understand that this test was effectively not using the parameter `exposeAllLogsViaJMX` in any meaningful way (as in the behaviour being tested was using the same asserts

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/OSBTypeHandler.java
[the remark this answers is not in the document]
tomjenkinson: Thank you!

tomjenkinson (approved):

marcosgopen:
Merging this even though the AS_TESTS failed because the failure s not related to this PR.

PR #2352, opened 2025-01-29
JBTM-3881 clear ExtendedResourceRecord cache

mmusgrov:
https://issues.redhat.com/browse/JBTM-3881

CORE !AS_TESTS !RTS !JACOCO !XTS !QA_JTA QA_JTS_OPENJDKORB !PERFORMANCE !DB_TESTS !mysql !db2 !postgres !oracle !JDK21

ExtendedResourceRecord [caches the record type](https://github.com/jbosstm/narayana/blob/7.2.0.Final/ArjunaJTS/jts/classes/com/arjuna/ats/internal/jts/resources/ExtendedResourceRecord.java#L235) (type_id) which is causing an issue when it is called early in the prepare phase. This fix adds a method to clear the cached value so that the record is treated as a [RecordType.OTS_ABSTRACTRECORD](https://github.com/jbosstm/narayana/blob/7.2.0.Final/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/RecordType.java#L81) (record type151).

This fix that I have chosen for the QE test failure is not the only possible solution but I suspect it is the least invasive one and therefore less risky (my feeling is that the alternative fixes will not be backwardly compatible since they will probably involve changes to what gets stored in the record).

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
@@ -53,6 +53,14 @@ public abstract class AbstractRecord extends StateManager
 	public abstract int typeIs ();
 
 	/**
+	 * Ask the record type to clear any data it has previously cached so that the data may be
+	 * recomputed, for example if a record has determined that an endpoint it needs has failed
+	 * then it might cache that fact and stop using it.
+	 */
+	public void clearAnyCachedData() {
+	}
+
+	/**
 	 * If this abstract record caused a heuristic then it should return an
 	 * object which implements <code>HeuristicInformation</code>
 	 *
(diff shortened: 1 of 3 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
```diff
@@ -52,6 +52,9 @@ public abstract class AbstractRecord extends StateManager
 
 	public abstract int typeIs ();
 
+	public void clearCache() {
```
tomjenkinson, line 55: I think we should add a Javadoc for this. And how about calling it clearCachedType()? Cache might be a bit too broad?

mmusgrov: Agreed

marekkopecky:
@mmusgrov I made some local tests about this PR (see the details on https://issues.redhat.com/browse/JBEAP-28771 ) and original failures has been fixed by this PR. Thank you! cc @tomjenkinson

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/AbstractRecord.java
mmusgrov: Although it is up to the various record types to decide what cached data they need to clear so clearCachedType is too restrictive.

tomjenkinson (approved):
Thank you for the fix!

PR #2355, opened 2025-01-31
Set the branch name of the Narayana version of WildFly to the current…

tomjenkinson:
NO_TEST


marcosgopen (approved):
I think this PR is correct, but I wonder whether that file is still needed @tomjenkinson

jmfinelli (approved):

PR #2359, opened 2025-02-11
Overload suspendScan

jmfinelli:
[JBTM-3981](https://issues.redhat.com/browse/JBTM-3981)

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/RecoveryManager.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/RecoveryManager.java
@@ -194,22 +194,31 @@ public class RecoveryManager {
      * Suspend the recovery manager. If the recovery manager is in the process of
      * doing recovery scans then it will be suspended afterwards, in order to
      * preserve data integrity.
-     * <p>
-     * Note that this method is also influenced by
-     * {@link com.arjuna.ats.arjuna.common.RecoveryEnvironmentBean#setWaitForWorkLeftToDo}.
-     * For more details, see {@link PeriodicRecovery#suspendScan(boolean)}
-     *
+     * @param async false means wait for the recovery manager to finish any scans before returning.
+     * @param waitForWorkLeftToDo when true, it is important that, before invoking this method,
+     * all transactions will either be terminated by the Transaction Reaper or they have prepared
+     * and a log has been written, otherwise the suspend call may never return.
+     * @throws IllegalStateException if the recovery manager has been shutdown.
+     */
+    public void suspend(boolean async, boolean waitForWorkLeftToDo) {
+        trySuspend(async, waitForWorkLeftToDo);
+    }
+
+    /**
+     * Suspend the recovery manager. If the recovery manager is in the process of
+     * doing recovery scans then it will be suspended afterwards, in order to
+     * preserve data integrity.
      * @param async false means wait for the recovery manager to finish any scans before returning.
      * @throws IllegalStateException if the recovery manager has been shutdown.
      */
     public void suspend(boolean async) {
-        trySuspend(async);
+        trySuspend(async, false);
     }
 
-    public RecoveryManagerStatus trySuspend(boolean async) {
+    public RecoveryManagerStatus trySuspend(boolean async, boolean waitForWorkLeftToDo) {
         checkState();
 
-        PeriodicRecovery.Mode mode = _theImple.trySuspendScan(async);
+        PeriodicRecovery.Mode mode = _theImple.trySuspendScan(async, waitForWorkLeftToDo);
 
         switch (mode) {
             case ENABLED:
@@ -221,6 +230,10 @@ public class RecoveryManager {
             default:
                 throw new IllegalArgumentException("incompatible enum types");
         }
+    }
+
+    public RecoveryManagerStatus trySuspend(boolean async) {
+        return trySuspend(async, false);
     }
 
     /**
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/RecoveryManager.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/RecoveryManager.java
@@ -18,23 +18,19 @@ import com.arjuna.ats.internal.arjuna.recovery.RecoveryManagerImple;
 import com.arjuna.ats.internal.arjuna.recovery.RecoveryManagerStatus;
 import com.arjuna.common.util.ConfigurationInfo;
 
-class ScanThread extends Thread
-{
+class ScanThread extends Thread {
 
-    public ScanThread (RecoveryManagerImple theImple, RecoveryScan callback)
-    {
-	super("RecoveryManagerScanThread");
+    public ScanThread(RecoveryManagerImple theImple, RecoveryScan callback) {
+        super("RecoveryManagerScanThread");
 
-	_theImple = theImple;
-	_callback = callback;
+        _theImple = theImple;
+        _callback = callback;
 
-	setDaemon(true);
+        setDaemon(true);
     }
 
-    public void run ()
-    {
-        if (_theImple != null)
-        {
+    public void run() {
+        if (_theImple != null) {
             _theImple.scan();
 
             if (_callback != null)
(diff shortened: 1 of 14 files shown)


ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/RecoveryManager.java
```diff
@@ -194,22 +194,25 @@ public void waitForTermination() {
...
+        trySuspend(async, waitForWorkLeftToDo);
+    }
+
     public void suspend(boolean async) {
-        trySuspend(async);
+        trySuspend(async, false);
     }
 
-    public RecoveryManagerStatus trySuspend(boolean async) {
+    public RecoveryManagerStatus trySuspend(boolean async, boolean waitForWorkLeftToDo) {
```
tomjenkinson, line 212: If we proceed with this, it's theoretically possible we could emit the warn_feature_not_supported_across_all_recovery_modules at this point.

_However_ I am not convinced that this is good for the user to only learn this information at this point. I wonder if we might keep the need to setWaitForWorkLeftToDo and so it can keep emitting the warning and also if someone calls trySuspend with true for the second parameter when waitForWorkLeftToDo is false we could either log an error or throw an exception.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/logging/arjunaI18NLogger.java
```diff
@@ -1645,13 +1645,6 @@ public void warn_objectstore_JDBCImple_over_max_image_size(int imageSize,
 	@LogMessage(level = WARN)
 	void unexpected_exception(@Cause() Throwable arg0);
 
-	// TODO Delete this as soon as JBTM-3983 has been implemented
-	@Message(id = 12415, value = "You have enabled a feature that has not yet been fully assessed for compliance " +
-			"across all Recovery Modules provided by Narayana. For more details, " +
-			"please refer to https://issues.redhat.com/browse/JBTM-3893.", format = MESSAGE_FORMAT)
-	@LogMessage(level = WARN)
-	void warn_feature_not_supported_across_all_recovery_modules();
```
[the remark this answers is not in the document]
jmfinelli: I removed it as the main modules (i.e. AtomicAction, JTA, and, partially, JTS) have been modified. If we want to keep this message until all modules will be integrated, no problem from my side :-)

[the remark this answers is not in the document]
tomjenkinson: Thank you, I think we should leave it until we know all the modules that we want (from a Narayana point of view, rather than want from a WildFly point of view) to be modified have been done so and we have documented it as complete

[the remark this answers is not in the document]
jmfinelli: I tried to reinstate this message and its uses but realized that:

* We need to keep a global switch that enables/disables the graceful feature of the Recovery Manager, in parallel with the overloaded `suspendScan`, if we want to use the log message when recovery modules are loaded
  * If we go down this path, I think the result might create confusion, as users might wonder which option activates the graceful suspend feature
* (Minor) Using the log message to warn users would require cross-contaminating ArjunaCore with Recovery Modules from other Narayana modules. (At the moment, we only have `_recoveryModules.stream().anyMatch(x -> !x.getClass().getName().equals("com.arjuna.ats.internal.arjuna.recovery.AtomicActionRecoveryModule"))`, but we would need to integrate reporting for other module names here)

IMO, it would be better to describe the graceful feature of the Recovery Manager in the documentation and state clearly which modules can participate. WDYT?

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/RecoveryManager.java
jmfinelli: Please, refer to [this](https://github.com/jbosstm/narayana/pull/2359#discussion_r1979230264) comment. Thanks

tomjenkinson: Having it documented is a great approach to give advance knowledge but I do also think that we should emit the warning at this point (to catch a situation that a user was not already aware) if someone has called this with true with a recovery module configured that we know is not compatible with JBTM-3893.

tomjenkinson: We may have had a similar discussion to that before but I am thinking this is more an implementation detail rather than the public RecoveryModule API if we have a temporary internal marker interface we remove when they are all implemented.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/RecoveryManager.java
```diff
@@ -223,6 +226,10 @@ public RecoveryManagerStatus trySuspend(boolean async) {
         }
     }
 
+    public RecoveryManagerStatus trySuspend(boolean async) {
```
tomjenkinson, line 229: Please can I ask for clarification as to the reason this needs to be added to the public API for this class?

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/PeriodicRecovery.java
```diff
@@ -184,24 +184,28 @@ public void shutdown(boolean async) {
...
      * is true, this parameter is overridden.
+     * @param waitForWorkLeftToDo when true, it is important that, before invoking this method,
+     * all transactions will either be terminated by the Transaction Reaper or they have prepared
+     * and a log has been written, otherwise the suspend call may never return.
      * @return the previous mode before attempting the suspension
      */
-    public Mode suspendScan(boolean async) {
+    public Mode suspendScan(boolean async, boolean waitForWorkLeftToDo) {
+
+        // TODO Delete this as soon as JBTM-3983 has been implemented
```
tomjenkinson, line 198: Thank you!

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/RecoveryManager.java
tomjenkinson: Sorry, I see the reason now!

tomjenkinson, line 229: I think we need a JavaDoc for this public method (I reason it didn't have one before so we can't really say for this method it's a blocker for this one - unless one could argue the introduction of an optional parameter has made it a bit more necessary?)

jmfinelli: Done. Thanks Tom

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/recovery/PeriodicRecovery.java
```diff
@@ -249,6 +253,10 @@ public Mode suspendScan(boolean async) {
         }
     }
 
+    public Mode suspendScan(boolean async) {
```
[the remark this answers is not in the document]
jmfinelli: Done. Thanks Tom

tomjenkinson (approved):
Thanks for the pull request!

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/recovery/RecoveryManager.java
jmfinelli: IIUC, [this](https://github.com/jbosstm/narayana/pull/2359/commits/e5b865b906fbc449e68e43dccb720559ea81c617#diff-7a8476d3b8fa1bb1cb151e68fc1fef8098f622cd1bb51a6f1471f9e37217b0d5R207-R214) should be the javadoc we were after

tomjenkinson: Here I actually referring to `trySuspend` (thinking along the lines of it being a new method, but I realise it is just moved down from above and that it already didn't have a JavaDoc).

I added my approval because basically though you added a new version of this method, the previous method didn't have a JavaDoc. That being said really both methods should now have a JavaDoc as they are public API methods :)

PR #2367, opened 2025-04-30
JBTM-3836 Reaper log message interceptor

mmusgrov:
https://issues.redhat.com/browse/JBTM-3836

CORE !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !DB_TESTS !mysql !db2 !postgres !oracle

Pull requests build and run with JDK17 and JDK21. Add the prefix `!` to disable these defaults.

### DIFF
--- a/narayana-bom/pom.xml
+++ b/narayana-bom/pom.xml
@@ -59,7 +59,7 @@
     <version.org.jboss.jboss-transaction-spi>8.0.0.Final</version.org.jboss.jboss-transaction-spi>
     <version.org.jboss.logging.jboss-logging>3.6.1.Final</version.org.jboss.logging.jboss-logging>
     <version.org.jboss.logging.jboss-logging-tools>3.0.2.Final</version.org.jboss.logging.jboss-logging-tools>
-    <version.org.jboss.logmanager>2.1.19.Final</version.org.jboss.logmanager>
+    <version.org.jboss.logmanager>2.0.1.Final</version.org.jboss.logmanager>
     <version.org.jboss.openjdk-orb>10.1.0.Final</version.org.jboss.openjdk-orb>
     <version.org.jboss.ws>3.0.0.Final</version.org.jboss.ws>
     <version.parsson>1.1.3</version.parsson>
@@ -88,6 +88,11 @@
             <artifactId>netty</artifactId>
           </exclusion>
         </exclusions>
+      </dependency>
+      <dependency>
+        <groupId>org.jboss.logmanager</groupId>
+        <artifactId>log4j2-jboss-logmanager</artifactId>
+        <version>${version.org.jboss.logmanager}</version>
       </dependency>
       <dependency>
         <groupId>org.jboss.logging</groupId>
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/CoordinatorEnvironmentBean.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/CoordinatorEnvironmentBean.java
@@ -8,6 +8,7 @@ import com.arjuna.ats.arjuna.coordinator.CheckedActionFactory;
 import com.arjuna.ats.arjuna.coordinator.TransactionReaper;
 import com.arjuna.ats.arjuna.utils.Utility;
 import com.arjuna.ats.internal.arjuna.coordinator.CheckedActionFactoryImple;
+import com.arjuna.ats.arjuna.logging.ArjunaLoggerInterceptor;
 import com.arjuna.ats.internal.arjuna.objectstore.HashedActionStore;
 import com.arjuna.common.internal.util.ClassloadingUtility;
 import com.arjuna.common.internal.util.propertyservice.FullPropertyName;
@@ -56,6 +57,7 @@ public class CoordinatorEnvironmentBean implements CoordinatorEnvironmentBeanMBe
     private volatile int txReaperZombieMax = TransactionReaper.defaultZombieMax;
     private volatile long txReaperTraceGracePeriod = TransactionReaper.defaultUntracedPeriod;
     private volatile long txReaperTraceInterval = TransactionReaper.defaultTracePeriod;
+    private ArjunaLoggerInterceptor loggerInterceptor;
 
     private volatile int defaultTimeout = 60; // seconds
     private volatile boolean transactionStatusManagerEnable = true;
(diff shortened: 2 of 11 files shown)


narayana-bom/pom.xml
```diff
@@ -59,7 +59,7 @@
     <version.org.jboss.jboss-transaction-spi>8.0.0.Final</version.org.jboss.jboss-transaction-spi>
     <version.org.jboss.logging.jboss-logging>3.6.1.Final</version.org.jboss.logging.jboss-logging>
     <version.org.jboss.logging.jboss-logging-tools>3.0.2.Final</version.org.jboss.logging.jboss-logging-tools>
-    <version.org.jboss.logmanager>2.1.19.Final</version.org.jboss.logmanager>
+    <version.org.jboss.logmanager>2.0.1.Final</version.org.jboss.logmanager>
```
tomjenkinson, line 62: It looks a bit curious to do a downgrade of a dependency? Please can I ask why that was needed?

mmusgrov: We currently don't use logmanager in any of our modules and the one specified here in our current bom, namely  2.1.19.Final, is not available at maven central: https://mvnrepository.com/artifact/org.jboss.logmanager/log4j2-jboss-logmanager

mmusgrov:
@beikov Hi Christian can you use this solution to sort out the Hibernate issue with the excessive logging?

narayana-bom/pom.xml
mmusgrov: I need `org.jboss.logmanager.Logger;`

mmusgrov: These are my options: https://github.com/jboss-logging/log4j2-jboss-logmanager/tags

mmusgrov: It seems like you are urging me to use log4j 1.x but I want log4j 2.x (https://logging.apache.org/log4j/2.x/migrate-from-log4j1.html).

Note that that I am not downgrading the dependency because we don't use logmanager in any of our modules, can you indicate what the issue is with me using 2.0.1.Final?

mmusgrov: If I change the dependency from log4j2-jboss-logmanager version 2.0.1.Final to jboss-logmanager version 2.1.19.Final then I get the error:                                                                          
                                                                                                                                                                                                                    
```                                                                                                                                                                                                                 
java.lang.IllegalStateException: The LogManager was not properly installed (you must set the "java.util.logging.manager" system property to "org.jboss.logmanager.LogManager")                                      
        at org.jboss.logmanager.Logger.getLogger(Logger.java:57)                                                                                                                                                    
```                                                                                                                                                                                                                 
                                                                                                                                                                                                                    
And if I then do what the error says and set the system property, ie ` -Djava.util.logging.manager=org.jboss.logmanager.LogManager` and, although the tests pass, I do get the following horrible error on the command line:
                                                                                                                                                                                                                    
```                                                                                                                                                                                                                 
Could not load Logmanager "org.jboss.logmanager.LogManager"                                                                                                                                                         
java.lang.ClassNotFoundException: org.jboss.logmanager.LogManager                                                                                                                                                   
```                                                                                                                                                                                                                 

Any thoughts on what other changes I need to fix that?

tomjenkinson: As I understand it then, it was a bug that `version.org.jboss.logmanager` was defined before. As such it works for me to select a version that exists.

mmusgrov:
@beikov and @scottmarlow I've put this on hold pending a review from one of you.

beikov:
From the Hibernate ORM side, I don't know if I can give you a lot of insight. A log filter like you are proposing is IMO something that must be configured/setup in Wildfly/Quarkus, so the question is whether that interface fits their needs or is even necessary. 
AFAIU, Narayana uses java.util.logging and jboss-logging has an integration for that, so it should be possible to write a filter for this logger already in Wildfly/Quarkus, hence there should be no need for a new API like that. The Wildfly/Quarkus integration should IMO just attach a filter for the `com.arjuna.ats.arjuna` logger that filters out reaper messages for timeouts.

In the Zulip topic I was more talking about the usefulness of the log message in the first place. If a JDBC statement is cancelled (which it is since [AG-231](https://issues.redhat.com/browse/AG-231) is fixed), which will then lead to an exception in the main app thread with all the context that one needs, what is the point of logging in the reaper? It's IMO much more appropriate to let the exception be handled by the user or bubble up to the runtime environment to be logged there.

mmusgrov:
So it is specific log messages (ie all the log messages preceding the throw which might require some clever look ahead coding) that you want suppressed. If it is something that each integrator should handle then the issue needs to be raised against each integration. I will let @scottmarlow comment before withdrawing the PR (and the JBTM issue).

scottmarlow:
My vote is to leave this open a little longer and consider if there could be a WildFly wrapper that implements a Hibernate extension that is passed into the Hibernate Persistence Unit as an integration property that could leverage the https://github.com/jbosstm/narayana/pull/2367 change.  

Separate from this pull request, we improve a lot by using the WildFly/Narayana  (now WildFly TM wrapper) transaction synchronization registry level monitoring that could help ORM to avoid loading data in a transaction that was rolled back in a background (e.g. reaper) thread.  This is currently not handled by Hibernate ORM but I think could easily be used to inform Hibernate ORM that a transaction has been rolled back.  The WildFly Persistence/JPA container does use [transaction synchronization registry level monitoring for detecting when certain background tx reaper (or foreground events) have occurred](https://github.com/wildfly/wildfly/blob/main/jpa/subsystem/src/main/java/org/jboss/as/jpa/transaction/TransactionUtil.java#L99). See WildFly code of what I mean at https://github.com/wildfly/wildfly/blob/main/jpa/subsystem/src/main/java/org/jboss/as/jpa/transaction/TransactionUtil.java#L99

scottmarlow:
> > _Independent_ from this pull request, we _could_ improve a lot by using the WildFly/Narayana (now WildFly TM wrapper) transaction synchronization registry level monitoring that could help ORM to avoid loading data in a transaction that was rolled back in a background (e.g. reaper) thread. This is currently not handled by Hibernate ORM but I think could easily be used to inform Hibernate ORM that a transaction has been rolled back.
> 
> @scottmarlow Who would do this work? We are happy to accept PRs if it requires changes on the Narayana side /cc @beikov

I think that the work involved would be to leverage the https://github.com/wildfly/wildfly/blob/main/jpa/subsystem/src/main/java/org/jboss/as/jpa/transaction/TransactionUtil.java#L99 logic from within Hibernate ORM in the form of a pre-check in a data load (perhaps once just prior to data load and once at tail end of the data load.)  

We added a similar check to Hibernate ORM around 10 years ago or so but I think the check was incorrect (we only checked for a thread name change or something like that) so we likely removed it.

mmusgrov:
Ok thanks @scottmarlow - in that case I think we can close this PR as not required. If there aren't any objections then I will close it later on in the week.

PR #2373, opened 2025-06-09
Revert JBTM-3859 on 6.0

tomjenkinson:
https://issues.redhat.com/browse/JBTM-3859

CORE AS_TESTS RTS !JACOCO XTS QA_JTA QA_JTS_OPENJDKORB !PERFORMANCE !DB_TESTS mysql db2 postgres oracle

Pull requests build and run with JDK17 and JDK21. Add the prefix `!` to disable these defaults.

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/ObjStoreBrowser.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/ObjStoreBrowser.java
@@ -477,6 +477,34 @@ public class ObjStoreBrowser implements ObjStoreBrowserMBean {
     }
 
     /**
+     * Update registered MBeans based on the current set of Uids.
+     * @param allCurrUids any registered MBeans not in this collection will be deregistered
+     */
+    private void unregisterRemovedUids(Map<String, Collection<Uid>> allCurrUids) {
+
+        for (Map.Entry<String, List<UidWrapper>> e : registeredMBeans.entrySet()) {
+            String type = e.getKey();
+            List<UidWrapper> registeredBeansOfType = e.getValue();
+            Collection<Uid> currUidsOfType = allCurrUids.get(type);
+
+            if (currUidsOfType != null) {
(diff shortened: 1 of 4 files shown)


mmusgrov:
@tomjenkinson Could we not just ask them to take 6.0.4.Final since the 6.0.5.Final only adds changes for JBTM-3859 (the fix that is causing the test failure)  and JBTM-3973 (which only adds support LRA MBeans which isn't needed).

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/ObjStoreBrowser.java
```diff
@@ -485,12 +490,12 @@ public void setExposeAllRecordsAsMBeans(boolean exposeAllLogs) {
...
         Map<String, Collection<Uid>> currUidsForType = new HashMap<>();
 
-        // recreate every MBean so clear the cache first
-        unregisterMBeans();
-
         for (String type : getTypes())
             currUidsForType.put(type, getUids(type));
 
+        // if there are any beans in registeredMBeans that don't appear in new list and unregister them
+        unregisterRemovedUids(currUidsForType); //unregisterMBeans();
```
mmusgrov, line 497: I don't understand where this change comes from (which commit is it from)?

mmusgrov: Ah sorry, it's part of the fix for JBTM-3859 MBean for JTA transaction heuristics is imprecise

mmusgrov (approved):
I'm approve the revert of JBTM-3859.

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/tools/osb/mbean/ObjStoreBrowser.java
tomjenkinson: Sorry, for some reason the commit message of one of the commits was not good and that wouldn't have helped. Thanks for finding it

PR #2380, opened 2025-06-24
Allow smoother importing to an IDE type

tomjenkinson:
CORE !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !DB_TESTS mysql db2 postgres oracle

Depends: https://github.com/jbosstm/jboss-as/pull/98

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -380,19 +380,6 @@
                 <pluginExecution>
                   <pluginExecutionFilter>
                     <groupId>org.codehaus.mojo</groupId>
-                    <artifactId>build-helper-maven-plugin</artifactId>
-                    <versionRange>[1.7,)</versionRange>
-                    <goals>
-                      <goal>add-source</goal>
-                    </goals>
-                  </pluginExecutionFilter>
-                  <action>
-                    <ignore></ignore>
-                  </action>
-                </pluginExecution>
-                <pluginExecution>
-                  <pluginExecutionFilter>
-                    <groupId>org.codehaus.mojo</groupId>
                     <artifactId>idlj-maven-plugin</artifactId>
                     <versionRange>[1.2.1,)</versionRange>
                     <goals>


ArjunaCore/arjunacore/pom.xml
```diff
@@ -41,6 +41,7 @@
         <artifactId>build-helper-maven-plugin</artifactId>
         <executions>
           <execution>
+            <?m2e execute onConfiguration?>
```
mmusgrov: @tomjenkinson I've not seen this xml syntax before - what does it do?

tomjenkinson: It allows the IDE I am wanting to work with to actually add the listed folders as source folders and thus allows the IDE to resolve certain files without showing syntax errors. It stands (I think) roughly for "Maven 2 Eclipse". There is some reference here: https://eclipse.dev/m2e/documentation/release-notes-17.html#new-syntax-for-specifying-lifecycle-mapping-metadata.

mmusgrov:
The AS build failures are because 1.0.2.Final-SNAPSHOT of lra is missing from the local repo (we updated our jboss-as clone). Ah @tomjenkinson I see that you have raised https://github.com/jbosstm/jboss-as/pull/98 to address that but that would mean we can no longer change the WildFly microprofile-lra subsystem and test with it so we should look for an alternate fix.

ArjunaCore/arjunacore/pom.xml
tomjenkinson: I have put the pull request on hold, I think it might be that I need to delete https://github.com/jbosstm/narayana/blob/main/pom.xml#L380-L392 instead

tomjenkinson: I am removing hold, it seems it was that older IDE handling which was the cause so I pushed an update

mmusgrov: Interesting link, thanks. I'll resolve this conversation then.

mmusgrov:
> > > I am not sure if I agree with "no longer change the WildFly microprofile-lra subsystem and test with it so we should look for an alternate fix" - I think the Narayana LRA project will update the pom.xml of the jbosstm/jboss-as fork to set the version number when it is tested? I think it does that here: https://github.com/jbosstm/lra/blob/main/scripts/hudson/narayana.sh#L308
> > 
> > 
> > An example alternate fix would be to build the lra repo (minus the tests) prior to building our jboss-as clone.
> 
> I don't think we want to couple the building of jbosstm/narayana to jbosstm/lra. That would then by cyclical because jbosstm/lra depends on jbosstm/narayana (https://github.com/jbosstm/lra/blob/main/pom.xml#L85).

OK thanks, that all makes sense.

mmusgrov (approved):

PR #2381, opened 2025-07-04
Use Apache Maven GPG Plugin in Narayana

jmfinelli:
Addresses [JBTM-3995](https://issues.redhat.com/browse/JBTM-3995)

!CORE !AS_TESTS !RTS !JACOCO !XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !DB_TESTS


build.sh
```diff
@@ -7,60 +7,21 @@
...
-  MAVEN_OPTS="$MAVEN_OPTS --add-modules java.corba"
-  MAVEN_OPTS="$MAVEN_OPTS --add-modules java.xml.bind"
-  MAVEN_OPTS="$MAVEN_OPTS --add-modules java.xml.ws"
-  export MAVEN_OPTS
-fi
-
-if [ -z "$MAVEN_OPTS" ]
-then
-	if [ $JAVA_VERSION -ge "9" ]; then
-		MAVEN_OPTS="$MAVEN_OPTS -Xms1303m -Xmx1303m"
```
tomjenkinson, line 55: I agree with the bulk of the changes in this file as BPA is not needed and nor is JDK 9 support itself.

From what I can see here, unless it is set somewhere else, I think with this change we are no longer setting our "preferred" Xms Xmx settings? Do I follow correctly? I have a remark on PRESERVE_WORKING_DIR I will mention later.

build.sh
```diff
@@ -168,16 +129,13 @@ main() {
     #  Export some stuff for maven.
     export MVN MVN_OPTS MVN_GOAL
 
-    echo "$MVN $MVN_OPTIONS $MVN_GOAL $ADDIT_PARAMS"
-
-    # workaround in case 'mvn -f' is not supported
-    if [ "$PRESERVE_WORKING_DIR" = "true" ]; then cd "$ORIG_WORKING_DIR"; fi
+    echo "${BASH_INTERPRETER} $MVN $MVN_OPTIONS $MVN_GOAL $ADDIT_PARAMS"
```
tomjenkinson, line 131: I am not sure who set PRESERVE_WORKING_DIR before, but now it won't allow the cd into ORIG_WORKING_DIR as part of the script execution?

ArjunaCore/arjuna/pom.xml
```diff
@@ -171,11 +171,6 @@
           </excludes>
         </configuration>
       </plugin>
-      <plugin>
```
tomjenkinson: I am not sure if it's right or wrong :) But I will say that maven-install-plugin and I think the deployment part of our build choices was complicated and an aim was to only deploy the main artifacts and their truly needed dependencies (in other words the uber jars and their parent poms and presumably the bom).

I don't know if this pull request changes this (or indeed if that is the current status) but I remember somehow spending some time try to make sure only certain dependencies were deployed to Maven at some point and so wanted to mention as I do remember it being fiddly and so might account for the lack or normalisation already?

pom.xml
```diff
@@ -540,6 +545,19 @@
               </execution>
             </executions>
           </plugin>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-gpg-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>sign-artifacts</id>
```
tomjenkinson, line 553: I guess this might need some wiki/documentation somewhere (maybe the wiki on GitHub or maybe even README.md) so that people performing a the release know how to configure their environment? Would you be able to share some notes on that, please? I can imagine it might not be straightforward if it's in the wiki as it would need to be a placeholder. But in that case maybe some notes in [the JBTM](https://issues.redhat.com/browse/JBTM-3995)?

jmfinelli: Thanks for asking this question, Tom. The plan is to add some notes to the release process we have in our GH's wiki as soon as we approve/merge this PR.

build.sh
jmfinelli: I would like to ask what the rationale was for using Xms and Xmx in the first place. The build.sh script only builds Narayana, so it seems to me that tuning the JVM for this task is a bit of an overkill. Also, JVM tuning should generally be done only when strictly necessary and with a specific purpose in mind (e.g., reducing GC frequency, controlling memory footprint, improving performance, etc.). In this case, I don't understand the purpose of setting -Xms1303m -Xmx1303m.

build.sh
jmfinelli: I'm not sure why `PRESERVE_WORKING_DIR` was introduced. Also, I don't think we use this env var anywhere. Moreover, `mvnw` can only be invoked from one place, i.e. the root of Narayana's code base. Sorry if I missed something obvious.

.mvn/wrapper/maven-wrapper.properties
```diff
@@ -14,6 +14,6 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.8.8/apache-maven-3.8.8-bin.zip
-distributionSha256Sum=2e181515ce8ae14b7a904c40bb4794831f5fd1d9641107a13b916af15af4001a
-wrapperUrl=https://repo.maven.apache.org/maven2/org/apache/maven/wrapper/maven-wrapper/3.2.0/maven-wrapper-3.2.0.jar
+wrapperVersion=3.3.2
```
marcosgopen, line 17: +1

marcosgopen:
@jmfinelli What about squashing the commits 'sortpom' related ?

pom.xml
tomjenkinson: I think this topic is generally resolved  - thank you @jmfinelli!

PR #2385, opened 2025-07-22
Updated snapshot deployment and place to deploy to

tomjenkinson:
NO_TEST

Updated where to deploy snapshots to and updated to allow the narayana-release-process.sh script to start to work for Snapshots where tagging is not expected

### DIFF
--- a/narayana-release-process.sh
+++ b/narayana-release-process.sh
@@ -1,12 +1,19 @@
 #!/bin/bash
 
-# You need to run this script with 2 arguments.
+# You need to run this script with 1 or 2 arguments.
 # Arguments are then tranformed to env varibales CURRENT and NEXT
 # Release process releases version ${CURRENT} and prepares github with commits to ${NEXT}-SNAPSHOT
 #
+# 1 argument: `./narayana-release-process.sh snapshot`
 # 2 arguments: `./narayana-release-process.sh CURRENT NEXT`
 
-if [ $# -ne 2 ]; then
+if [ $# -eq 1 ]; then
+  if [ "$1" != "snapshot" ]; then
+    echo "With one argument only snapshot deployment is allowed"
+    exit -1
+  fi
+  SNAPSHOT=true
+elif [ $# -ne 2 ]; then
   echo 1>&2 "$0: usage: CURRENT NEXT"
   exit 2
 else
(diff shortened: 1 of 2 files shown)


narayana-release-process.sh
```diff
@@ -1,101 +1,115 @@
...
+    git log -n 5
+    read -p "Did the log before look OK?" ok
+    if [[ $ok == n* ]]
     then
-      git status
       exit
     fi
-    git status | grep "ahead"
-    if [[ $? != 1 ]]
+    set -e
```
mmusgrov, line 69: What's the rationale for not putting this on line 2, which is more standard, of the script.

narayana-release-process.sh
```diff
@@ -1,101 +1,115 @@
...
   then
+    echo 1>&2 documentation: Tag '$CURRENT' did not exist
     exit
   fi
-  set -e
-  
-  read -p "Until JBTM-3891 is resolved, please review with project team to ensure that it is safe to release. Please check for failed CI jobs before continuing. Continue? y/n " NOBLOCKERS
-  if [[ $NOBLOCKERS == n* ]]
+
+  rm -rf $PWD/localm2repo
```
mmusgrov: Remark: I'm not asking for a change, I'm just remarking (mostly for my own future benefit), that DONOTDELETELOCALM2REPO only applies to the source repo.

narayana-release-process.sh
tomjenkinson: It's been moved from https://github.com/jbosstm/narayana/pull/2385/files#diff-530cdb1e8542754f0d5c3b24ac1598f52957ed33177c8ca69b719666077c3bbeL57

So it's not exactly a change brought in here you can see the proposed place https://github.com/jbosstm/narayana/blob/cfa668220a3c61f224e6dab9630cfd56e76c708e/narayana-release-process.sh#L69 is not really too different to where it was https://github.com/jbosstm/narayana/blob/main/narayana-release-process.sh#L57 (basically just protected by the SNAPSHOT check now - but it causes a big diff

tomjenkinson: At a guess it is basically to try to disallow errors in (cd ./scripts/ ; ./pre-release.sh $CURRENT $NEXT) but that doesn't seem quite right so maybe it is a relic from when more of the script was interactive vs automated and thus having some parts fail on error and some not would be more evident?

That being said, I would basically try to argue that I am not materially changing this behaviour from how it was before.

narayana-release-process.sh
tomjenkinson: Actually you would be right to have asked for a change :) I have wrapped this with the same protection. Thanks for mentioning this topic as it helped me to spot it.

narayana-release-process.sh
mmusgrov: I assume it's not a material change though the script complexity makes me pause:                                                                                                                                    

originally the usage was around the reading of input from the user:                                                                                                                                                 
                                                                                                                                                                                                                    
```                                                                                                                                                                                                                 
set -e                                                                                                                                                                                                              
read -p ...                                                                                                                                                                                                         
...                                                                                                                                                                                                                 
set +e                                                                                                                                                                                                              
...                                                                                                                                                                                                                 
```                                                                                                                                                                                                                 
                                                                                                                                                                                                                    
which makes little sense. A possibility is that the original author was intending the `-e` to apply to the read command (see the read section of the set manual page `man set` where `-e` tells bash to use readline) and confused it with the set command (see the set section of the manual page `man set` which tells bash to "exit immediately on a non-zero status from the pipeline").                                                                                    

Anyway, since I've partially lost the ability to reliably parse the script as its complexity has risen over the years, I still approve of the PR since it has improved it, ie on the surface it does not appear to have made it worse.

tomjenkinson: Thank you, my guess is this set - and set + stuff was to deal with some more complicated scripting when it was first added and we maybe simplified that out but left the error handling around?

PR #2386, opened 2025-07-29
New wording for the DCO paragraph

jmfinelli:
New DCO paragraph in CONTRIBUTING.md

NO_TEST

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,8 @@ We use a four step process for contributions:
 
 All files in this repository are subject to the Apache 2.0 license.
 
-By contributing to this project you agree to the [Developer Certificate of Origin (DCO)](https://developercertificate.org/). The DCO was created by the Linux Kernel community and is a simple statement that you, as a contributor, have the legal right to make the contribution (we adopt the "CoreOS method" approach to using the DCO in which there is no per-commit signoff ceremony).
+By contributing to this project you agree to the Developer Certificate of Origin (DCO).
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -22,7 +22,8 @@ We use a four step process for contributions:
 
 All files in this repository are subject to the Apache 2.0 license.
 
-By contributing to this project you agree to the [Developer Certificate of Origin (DCO)](https://developercertificate.org/). The DCO was created by the Linux Kernel community and is a simple statement that you, as a contributor, have the legal right to make the contribution (we adopt the "CoreOS method" approach to using the DCO in which there is no per-commit signoff ceremony).
+By contributing to this project you agree to the Developer Certificate of Origin (DCO).
```
tomjenkinson, line 25: I think it should be `[DCO](DCO)` (please see an example DCO at https://github.com/metal3-io/metal3-docs/blob/main/DCO)

tomjenkinson: Sorry, we have the DCO file already (https://github.com/jbosstm/narayana/blob/main/dco.txt) so maybe it should just be `[DCO](DCO)` that would be enough (or maybe it can't find a dco.txt and the filename needs to be specified perhaps as `[dco.txt](DCO)`?

jmfinelli: I didn't use [DCO](DCO) as at line 28 we already specify [DCO](dco.txt)

tomjenkinson: That works for me, thank you

tomjenkinson (approved):

PR #2388, opened 2025-08-22
JBTM-2493

vedpathakyb:
https://issues.redhat.com/browse/JBTM-2493  Fixed this issue. 

1. Added one comment at line 77 in the class "https://github.com/jbosstm/narayana/blob/main/rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java#L77"

2. Added try, catch block in the class org.jboss.narayana.rest.integration.test.functional.RecoveryManagerTestCase.java so that it shall throw an exception during commit if there is any.

### DIFF
--- a/rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
+++ b/rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
@@ -73,7 +73,7 @@ public final class RecoveryManager {
             recoverParticipants();
         }
     }
-
+/* Participant implementations must implement either PersistableParticipant Or java.io.Serializable*/
     public void persistParticipantInformation(final ParticipantInformation participantInformation) {
         if (LOG.isTraceEnabled()) {
             LOG.trace("RecoveryManager.persistParticipantInformation: participantInformation=" + participantInformation);
--- a/rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/functional/RecoveryManagerTestCase.java
+++ b/rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/functional/RecoveryManagerTestCase.java
@@ -34,9 +34,12 @@ public class RecoveryManagerTestCase {
 
         ParticipantsContainer.getInstance().clear();
         ParticipantsManagerFactory.getInstance().setBaseUrl("");
-        RecoveryManager.getInstance().persistParticipantInformation(participantInformationBefore);
-        RecoveryManager.getInstance().registerDeserializer(APPLICATION_ID + "2", new TestParticipantDeserializer());
-
+        try {
+            RecoveryManager.getInstance().persistParticipantInformation(participantInformationBefore);
+            RecoveryManager.getInstance().registerDeserializer(APPLICATION_ID + "2", new TestParticipantDeserializer());
+        } catch (RuntimeException e) {
+            throw new RuntimeException(e);
+        }
         final ParticipantInformation participantInformationAfter = ParticipantsContainer.getInstance()
                 .getParticipantInformation(participantId);
         Assert.assertNull(participantInformationAfter);


rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/functional/RecoveryManagerTestCase.java
```diff
@@ -34,9 +34,12 @@ public void testRecoveryWithoutDeserializer() {
...
         ParticipantsContainer.getInstance().clear();
         ParticipantsManagerFactory.getInstance().setBaseUrl("");
-        RecoveryManager.getInstance().persistParticipantInformation(participantInformationBefore);
-        RecoveryManager.getInstance().registerDeserializer(APPLICATION_ID + "2", new TestParticipantDeserializer());
-
+        try {
+            RecoveryManager.getInstance().persistParticipantInformation(participantInformationBefore);
+            RecoveryManager.getInstance().registerDeserializer(APPLICATION_ID + "2", new TestParticipantDeserializer());
+        } catch (RuntimeException e) {
+            throw new RuntimeException(e);
```
tomjenkinson: This looks a bit unusual in that you would propose to catch the RTE but throw a new one? Please can I ask you to clarify the intention?

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
```diff
@@ -73,7 +73,7 @@ public void registerDeserializer(final String applicationId, final ParticipantDe
             recoverParticipants();
         }
     }
-
+/* Participant implementations must implement either PersistableParticipant Or java.io.Serializable*/
```
tomjenkinson: This looks to need some further indentation, please.

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/functional/RecoveryManagerTestCase.java
vedpathakyb: My mistake. I will revert those changes. As per [JBTM-2493],  Can I remove the try catch in   persistParticipantInformation method of org.jboss.narayana.rest.integration.RecoveryManager.java while calling the recoveryStore.write_committed in line number 95 .

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
vedpathakyb: Acknowledged

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
```diff
@@ -87,17 +89,17 @@ public void persistParticipantInformation(final ParticipantInformation participa
...
             final Uid uid = new Uid(participantInformation.getId());
 
             recoveryStore.write_committed(uid, PARTICIPANT_INFORMATION_RECORD_TYPE, state);
             // to identify the uid from the participant persisted into the object store in order to delete it later
             persistedParticipants.put(participantInformation.getId(), uid);
-        } catch (Exception e) {
+        /*} catch (Exception e) {
             RESTATLogger.atI18NLogger.warn_persistParticipantInformationRecoveryManager(e.getMessage(), e);
-        }
+        }*/
```
jmfinelli: As I wasn't the original reporter for JBTM-2493, I may be missing some context.

I am open to removing the try-catch block and letting the caller handle the exception. If we adopt this approach, I suggest we apply it consistently across the `RecoveryManager` class.

That said, @mmusgrov likely has a deeper understanding of the original issue. He is unavailable until next week though, so I recommend we wait for his input before proceeding.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
```diff
@@ -73,8 +73,10 @@ public void registerDeserializer(final String applicationId, final ParticipantDe
             recoverParticipants();
         }
     }
-
-    public void persistParticipantInformation(final ParticipantInformation participantInformation) {
+       /*
+       * Participant implementations must implement either PersistableParticipant Or java.io.Serializable
+       */
+    public void persistParticipantInformation(final ParticipantInformation participantInformation) throws IOException, ObjectStoreException {
```
jmfinelli:
```suggestion
    /**
     * Participant implementations must implement either {@link PersistableParticipant} or {@link java.io.Serializable}
     * @param participantInformation
     * @throws IOException
     * @throws ObjectStoreException
     */
    public void persistParticipantInformation(final ParticipantInformation participantInformation) throws IOException, ObjectStoreException {
```

It would be nice to add some wording to explain what `participantInformation` does and why `IOException` and `ObjectStoreException` are thrown

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/functional/RecoveryManagerTestCase.java
```diff
@@ -24,7 +27,7 @@ public class RecoveryManagerTestCase {
     private static final String APPLICATION_ID = "org.jboss.narayana.rest.integration.test.functional.RecoveryManagerTestCase";
 
     @Test
-    public void testRecoveryWithoutDeserializer() {
+    public void testRecoveryWithoutDeserializer() throws ObjectStoreException, IOException {
```
jmfinelli: I'm not entirely sure, but if it's important that we throw `ObjectStoreException` and `IOException`, should we add a few tests specifically for those scenarios?

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/functional/RecoveryManagerTestCase.java
```diff
@@ -36,7 +39,6 @@ public void testRecoveryWithoutDeserializer() {
         ParticipantsManagerFactory.getInstance().setBaseUrl("");
         RecoveryManager.getInstance().persistParticipantInformation(participantInformationBefore);
         RecoveryManager.getInstance().registerDeserializer(APPLICATION_ID + "2", new TestParticipantDeserializer());
-
```
jmfinelli, line 39: Sorry to be pedantic, but could I ask you to move all the style changes into a separate commit?
If you'd rather not, would you mind reverting those specific changes instead? Thanks

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
mmusgrov: Agreed. The information is saved in the object store and made available during the completion phase of the transaction. IO exceptions are normally wrapped by ObjectStoreException so it would be interesting to know why it is needed here since it may be indicative of a bug elsewhere in the rest-at code.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
```diff
@@ -87,17 +89,17 @@ public void persistParticipantInformation(final ParticipantInformation participa
...
+        //try {
             final RecoveryStore recoveryStore = StoreManager.getRecoveryStore();
             final OutputObjectState state = getParticipantInformationOutputState(participantInformation);
             final Uid uid = new Uid(participantInformation.getId());
 
             recoveryStore.write_committed(uid, PARTICIPANT_INFORMATION_RECORD_TYPE, state);
             // to identify the uid from the participant persisted into the object store in order to delete it later
             persistedParticipants.put(participantInformation.getId(), uid);
-        } catch (Exception e) {
+        /*} catch (Exception e) {
```
mmusgrov: Ah this may be where the IO exception is coming from. Maybe the commented out exception catch block needs to be wrap the io exception and rethrow as an ObjectStoreException.

vedpathakyb: Is the below way correct to wrap the IOException inside ObjectStoreException

try {
            final RecoveryStore recoveryStore = StoreManager.getRecoveryStore();
            final OutputObjectState state = getParticipantInformationOutputState(participantInformation);
            final Uid uid = new Uid(participantInformation.getId());

            recoveryStore.write_committed(uid, PARTICIPANT_INFORMATION_RECORD_TYPE, state);
            // to identify the uid from the participant persisted into the object store in order to delete it later
            persistedParticipants.put(participantInformation.getId(), uid);
        } catch (IOException e) {
            RESTATLogger.atI18NLogger.warn_persistParticipantInformationRecoveryManager(e.getMessage(), e);
            throw new ObjectStoreException(e);
        }
    }

mmusgrov: I've taken a broader look at the PR. In general it looks good apart from the exception reporting, and if possible a test. There is a [dedicated API for participants](https://github.com/jbosstm/narayana/tree/7.2.2.Final/rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/api) so we should reported IO and object store exceptions using that, so in this example I would suggest `throw new ParticipantException(e);`.

Similarly in the various other places in this PR replacing `throws ObjectStoreException` and `throws IOException` with `throws ParticipantException `and wrap the specific exceptions inside an  instance of ParticipantException.

mmusgrov (changes requested):
The ideas in the PR look good and ought to address the reported issue. But as per my comment above will you change the exception handling to use the API, ie wrap specific exceptions inside instances of ParticipantException. Testing exception reporting might be tricky but do let us know if you need help with that.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
mmusgrov: In the past [we've had reports](https://issues.redhat.com/browse/JBTM-3990) that logging an exception and then re-throwing it is an anti-pattern. So to avoid that the code would need to be:

```
   ...
    } catch (IOException e) {
            String msg = RESTATLogger.atI18NLogger.warn_persistParticipantInformationRecoveryManager(e.getMessage(), e);

            throw new ParticipantException(msg, e);
```
Doing it this way would require you to change the atI18NLogger.java [warn_persistParticipantInformationRecoveryManager method](https://github.com/jbosstm/narayana/blob/7.2.2.Final/rts/at/util/src/main/java/org/jboss/jbossts/star/logging/atI18NLogger.java#L123) to return a String. This log message is only ever used in this catch block so I would be okay with changing the log message signature to return a String but others, such as @jmfinelli for example, may have a different opinion so I'd check with them first.

jmfinelli: I agree 100% with you @mmusgrov. This PR looks good to me apart from what you highlighted: once we fix the issue around exception handling and we add a test, I think that we're ready to go :-)

vedpathakyb: @mmusgrov , Seems it is not possible to change return value of method warn_participantCreateParticipantsRecoveryManager  of class org.jboss.jbosssts.star.logging.atI18NLogger.java . It can only be void, not something else. I am getting below compile time error: 
------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.14.0:compile (default-compile) on project restat-util: Compilation failure
[ERROR] /home/yvedpath/NotBackedUp/JavaProjects/JBTM/narayana/rts/at/util/src/main/java/org/jboss/jbossts/star/logging/atI18NLogger.java:[123,12] Message logger methods can only have a void return type.
[ERROR] 
------------------------------------------------

mmusgrov: Hi Yogesh, thanks for getting back onto this fix. You need to change the signature of the logger method to return a String, ie change method atI18NLogger.java [warn_persistParticipantInformationRecoveryManager method](https://github.com/jbosstm/narayana/blob/7.2.2.Final/rts/at/util/src/main/java/org/jboss/jbossts/star/logging/atI18NLogger.java#L123) to:

```
    @Message(id = 27023, value = "Participant Exception while recreateParticipantInformation in Recovery Manager. '%s'")
    String warn_participantCreateParticipantsRecoveryManager(String cause, @Cause Throwable t);
```

Notice that the "@LogMessage" annotation has been removed and the return value has been changed to a String.

You will need to rebuild the util logger after making the change (this can be done by running mvn package` in the top level restat folder or cd into the util folder)

Note that this log message is only ever used in this catch block so I think, and Manuel agrees, that it is okay to change the signature

mmusgrov: This behaviour is because JBoss Logging does more than only [generate logging messages](https://docs.redhat.com/en/documentation/red_hat_jboss_enterprise_application_platform/6.4/html/development_guide/sect-jboss_logging_tools):

```
JBoss Logging Tools is a Java API that provides support for the internationalization and localization of log messages, exception messages, and generic strings. In addition to providing a mechanism for translation, JBoss Logging tools also provides support for unique identifiers for each log message.
```

and in this example we just want to localise the String together with a unique identifier.

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/integration/BasicIntegrationTestCase.java
```diff
@@ -5,24 +5,23 @@
...
+import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
+import com.arjuna.ats.arjuna.exceptions.ObjectStoreException;
 import jakarta.xml.bind.JAXBException;
 
+import org.jboss.jbossts.star.logging.RESTATLogger;
+import org.jboss.narayana.rest.integration.api.*;
```
mmusgrov: Our style rules (we use the maven-checkstyle-plugin) disallow importing everything from a package. You can test your PR for style compliance by enabling the plugin:

`mvn clean install -DskipTests -Dcheckstyle.skip=false`

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantResource.java
```diff
@@ -4,6 +4,7 @@
  */
 package org.jboss.narayana.rest.integration;
 
+import com.arjuna.ats.arjuna.exceptions.ObjectStoreException;
```
mmusgrov: There is a [dedicated API for participants](https://github.com/jbosstm/narayana/tree/7.2.2.Final/rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/api) so we should report IO and object store exceptions using that, so in this example I would suggest `throw new ParticipantException(e);`.

Similarly in the various other places in this PR where ObjectStoreException and IOException are thrown.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantResource.java
```diff
@@ -33,6 +34,8 @@
 
 import com.arjuna.ats.arjuna.common.arjPropertyManager;
 
+import java.io.IOException;
+
```
mmusgrov, line 36: See previous comment.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantResource.java
```diff
@@ -90,7 +93,7 @@ public Response getStatus(@PathParam("participantId") final String participantId
     @Consumes(TxMediaType.TX_STATUS_MEDIA_TYPE)
     @Produces(TxMediaType.TX_STATUS_MEDIA_TYPE)
     public Response terminate(@PathParam("participantId") final String participantId, final String content)
-            throws HeuristicException {
+            throws HeuristicException, ObjectStoreException, IOException {
```
mmusgrov: Just throw ParticipantException

Please can you also do the same change every other place in this PR where you are throwing ObjectStoreException or IOException.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
```diff
@@ -73,38 +73,37 @@ public void registerDeserializer(final String applicationId, final ParticipantDe
...
         }
     }
-
-    public void persistParticipantInformation(final ParticipantInformation participantInformation) {
+       /*
+       *  Participant implementations must implement either {@link PersistableParticipant} or {@link java.io.Serializable}
+        * @param participantInformation
+        * @throw ParticipantException
+        */
+    public void persistParticipantInformation(final ParticipantInformation participantInformation) throws ParticipantException {
```
mmusgrov: None of the AT integration API currently has any javadoc so just documenting this method looks very odd. I would suggest that you raise a separate issue to report that the API is missing javadocs.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
```diff
@@ -73,38 +73,37 @@ public void registerDeserializer(final String applicationId, final ParticipantDe
...
         if (LOG.isTraceEnabled()) {
             LOG.trace("RecoveryManager.persistParticipantInformation: participantInformation=" + participantInformation);
         }
-
```
mmusgrov, line 81: Our PR rules disallow formatting changes unless they are in separate commits.

Similarly in the various other places where this PR has introduced formatting changes to parts of the code unrelated to the fix. Please can you go through the PR and revert those similar formatting/style changes,.

Also I don't understand the purpose of removal of the newline.

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/functional/ParticipantsManagerTestCase.java
```diff
@@ -54,7 +53,12 @@ public void testReportHeuristic() throws MalformedURLException {
         final ParticipantInformation participantInformation = ParticipantsContainer.getInstance().getParticipantInformation(participantId);
         participantInformation.setStatus(TxStatus.TransactionPrepared.name());
 
-        participantsManager.reportHeuristic(participantId, HeuristicType.HEURISTIC_ROLLBACK);
+        try {
+            participantsManager.reportHeuristic(participantId, HeuristicType.HEURISTIC_ROLLBACK);
+        } catch (Exception e) {
```
mmusgrov: This is a junit test so we don't need to catch and rethrow. I would just let the original exception leak out the method so that it appears in the test reports as is.

Similarly in the other tests where the catch and rethrow idiom has been applied.

In general we'd prefer the PR to be targeted to only the issue and to avoid adding in extra changes and improvements. If you notice things in the code that you think could be improved could you add an RFE JIRA for them.

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/functional/RecoveryManagerTestCase.java
mmusgrov: +1

mmusgrov (changes requested):
Thanks for the updated PR. I've ask for a few changes around exception reporting and  the fix should be targeted specifically to just the reported issue and to avoid any extra changes, even if they are beneficial (since those would normally go in a separate commit or separate issue).

mmusgrov:
> 
> Let me know if the way of handling ParticipantsInformation is proper. Once it is done, I will start working on writing test case.

I have added the Hold label to allow more time to add the test case.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
```diff
@@ -73,38 +73,37 @@ public void registerDeserializer(final String applicationId, final ParticipantDe
...
             final OutputObjectState state = getParticipantInformationOutputState(participantInformation);
             final Uid uid = new Uid(participantInformation.getId());
-
             recoveryStore.write_committed(uid, PARTICIPANT_INFORMATION_RECORD_TYPE, state);
             // to identify the uid from the participant persisted into the object store in order to delete it later
             persistedParticipants.put(participantInformation.getId(), uid);
         } catch (Exception e) {
-            RESTATLogger.atI18NLogger.warn_persistParticipantInformationRecoveryManager(e.getMessage(), e);
+            String msg = RESTATLogger.atI18NLogger.warn_persistParticipantInformationRecoveryManager(e.getMessage(), e);
+            throw new ParticipantException(msg, e);
```
mmusgrov, line 99: @vedpathakyb asked
> Let me know if the way of handling ParticipantsInformation is proper. 

Hi Yogesh, this is exactly the way I think it should be coded, nice, thanks.

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/integration/BasicIntegrationTestCase.java
vedpathakyb: I tested with mvn clean install -DskipTests -Dcheckstyle.skip=false.. It asked changes for more than one places. After making all those changes, I am getting 

Caused by: org.apache.maven.plugins.checkstyle.exec.CheckstyleExecutorException: There are 20 errors reported by Checkstyle 10.21.4 with narayana-checkstyle/checkstyle.xml ruleset.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
vedpathakyb: Removed those  added lines of javadoc

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantResource.java
vedpathakyb: done . Please check

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantResource.java
vedpathakyb: Done. please check

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
vedpathakyb: Reverted most of the changes. Please check

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/functional/ParticipantsManagerTestCase.java
vedpathakyb: Changed. Please check

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/integration/BasicIntegrationTestCase.java
mmusgrov: You should only run the style checker on the module you are changing, the rest of the codebase contains older code that we haven't style checked. So in this PR just the following should check your changes:

`mvn compile -DskipTests -Dcheckstyle.skip=false -f integration/pom.xml`

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantResource.java
mmusgrov: I see that you have removed them from this file, which is good. Note that I said the same change needs to be applied "in the various other places in this PR where ObjectStoreException and IOException are thrown". ie the same change needs to be applied to ParticipantsManager, ParticipantsManagerImpl, etc.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantResource.java
```diff
@@ -187,7 +190,7 @@ private Vote prepare(final ParticipantInformation participantInformation) throws
                 RESTATLogger.atI18NLogger.warn_prepareParticipantResource(e.getMessage(), e); // JBTM-3990
             }
             participantInformation.setStatus(TxStatus.TransactionActive.name());
-            throw e;
+            throw (e);
```
mmusgrov: We disallow formatting changes that are unrelated to the PR.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantResource.java
```diff
@@ -90,7 +93,7 @@ public Response getStatus(@PathParam("participantId") final String participantId
     @Consumes(TxMediaType.TX_STATUS_MEDIA_TYPE)
     @Produces(TxMediaType.TX_STATUS_MEDIA_TYPE)
     public Response terminate(@PathParam("participantId") final String participantId, final String content)
-            throws HeuristicException {
+            throws ParticipantException,HeuristicException {
```
mmusgrov: I notice that in some places you separate the classes in the throws clause with and without spaces. In the rest of the module we separate them with a space so I'd suggest using that for consistency. *Similarly* elsewhere in the PR.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantsManagerImpl.java
```diff
@@ -110,7 +113,7 @@ public void registerDeserializer(final String applicationId, final ParticipantDe
     }
 
     @Override
-    public void reportHeuristic(String participantId, HeuristicType heuristicType) {
+    public void reportHeuristic(String participantId, HeuristicType heuristicType) throws ObjectStoreException, IOException {
```
mmusgrov: See my previous review comment in ParticipantResource.java about using these exceptions - I'd suggest changing the signature to `throw new ParticipantException(e);` like you did above.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
mmusgrov: Thanks. We'll add the javadoc for the whole API in a separate issue. I'll leave this comment unresolved for now until the issue is created and the PR accepted.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
mmusgrov: Line 81 has still been removed in the latest change.

There are other places where a line has been removed, please can you *revert all of them*.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
```diff
@@ -74,37 +74,33 @@ public void registerDeserializer(final String applicationId, final ParticipantDe
...
         }
     }
-
```
mmusgrov, line 102: Another deleted line. There are *other places similar to this* in the PR.

tools/src/main/java/io/narayana/perf/Averager.java
```diff
@@ -5,7 +5,11 @@
 
 package io.narayana.perf;
 
-import java.util.*;
+import java.util.Arrays;
```
mmusgrov: Please only change files related to the issue, ie you don't need to fix problems in code that is unrelated the current issue, JBTM-2493. See my earlier comment about how to run the style checker on just  the module that you have changed (ie `mvn compile -DskipTests -Dcheckstyle.skip=false -f integration/pom.xml`).

This comment also applies to the other files in the tools module.

tools/src/main/java/io/narayana/perf/Measurement.java
```diff
@@ -31,7 +44,7 @@ public class Measurement<T> implements Serializable {
     private int numberOfWarmupCalls = 0; //number of iterations before starting the measurement
     private String info;
 
-    RegressionChecker config; // holds file based measurment data
+    RegressionChecker config; // holds file based measurement data
```
mmusgrov: Please only change code related to the issue.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantResource.java
mmusgrov: When the PR is ready to be tested we need to see a run the arquillian profile (-Parq) passing (this profile tests the changes against the WildFly RTS subsystem). I'm concerned that the signature of a public API is being changed and any code using the API may not expect the new exception type.

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/functional/ParticipantsManagerTestCase.java
```diff
@@ -48,14 +46,15 @@ public void testSetBaseUrl() {
     }
 
     @Test
-    public void testReportHeuristic() throws MalformedURLException {
+    public void testReportHeuristic() throws MalformedURLException , IOException, ObjectStoreException {
```
mmusgrov: I don't think the test should be throwing these new exceptions. It would be better to catch them at the place they can be thrown in the test and use `Assert.fail("Unexpected exception thrown: " + exception message);` or something similar.

*Similarly in the other tests where you have done the same.*

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/functional/RecoveryManagerTestCase.java
mmusgrov: +1 When the PR is closer to being ready, so I can see what need to be tested, I will have a think about how best to test "those specific scenarios".

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/functional/RecoveryManagerTestCase.java
mmusgrov: @vedpathakyb don't forget to also address other reviewers comments as well as mine.

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/integration/BasicIntegrationTestCase.java
```diff
@@ -177,7 +176,12 @@ public void testHeuristicRollbackBeforePrepare() throws JAXBException {
                     txSupport.getDurableParticipantEnlistmentURI(), p);
         }
 
-        ParticipantsManagerFactory.getInstance().reportHeuristic(lastParticipantid, HeuristicType.HEURISTIC_ROLLBACK);
+        try {
+            ParticipantsManagerFactory.getInstance().reportHeuristic(lastParticipantid, HeuristicType.HEURISTIC_ROLLBACK);
+        } catch (Exception e) {
```
mmusgrov: See [my comment](https://github.com/jbosstm/narayana/pull/2388#discussion_r2348873006) from my previous review. We don't need to catch and rethrow in unit tests. Also we don't need to internationalise/localise messages generated in unit tests. This comment also applies wherever else the same technique has been applied.

mmusgrov:
@vedpathakyb When you are ready will you remove the "Hold" label so that I know to review the next set of changes.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantResource.java
vedpathakyb: I have made changes.

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/functional/RecoveryManagerTestCase.java
vedpathakyb: Resolved.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
vedpathakyb: Resolved.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
vedpathakyb: As per https://github.com/jbosstm/narayana/pull/2388#discussion_r2315701486 The approach of removing try-catch is not followed. Instead of that, throwing ParticpantException is followed.

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/functional/RecoveryManagerTestCase.java
vedpathakyb: Style changes reverted.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
vedpathakyb: Made changes.

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/integration/BasicIntegrationTestCase.java
vedpathakyb: Resolved.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantResource.java
vedpathakyb: Checked and made changes in all files in org.jboss.narayana.rest.integration

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
vedpathakyb: Reverted almost all new changes I had made.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantResource.java
vedpathakyb: Almost all files are updated. Along with the ParticipantResource.java about formatting change.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantResource.java
vedpathakyb: separated with space. Did not find any other occurrence of giving space.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantsManagerImpl.java
vedpathakyb: Done.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
vedpathakyb: Added both lines. Checked and added else where in the PR. Hope all lines are aligned now.

tools/src/main/java/io/narayana/perf/Averager.java
vedpathakyb: Almost reverted all changes made outside the module.

tools/src/main/java/io/narayana/perf/Measurement.java
vedpathakyb: This is done

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantResource.java
vedpathakyb: ack. I will need your inputs on how to proceed with this.

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/functional/ParticipantsManagerTestCase.java
vedpathakyb: I have removed both those exceptions.

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/integration/BasicIntegrationTestCase.java
vedpathakyb: This is done

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
vedpathakyb: As per https://github.com/jbosstm/narayana/pull/2388#discussion_r2372788201 , I have removed doc

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantsManagerImpl.java
```diff
@@ -110,7 +110,7 @@ public void registerDeserializer(final String applicationId, final ParticipantDe
     }
 
     @Override
-    public void reportHeuristic(String participantId, HeuristicType heuristicType) {
+    public void reportHeuristic(String participantId, HeuristicType heuristicType)  {
```
mmusgrov: Extra white space. If you back out this change then ParticipantsManagerImpl.java‎ does not need to be part of your patch.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
```diff
@@ -30,7 +29,7 @@
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
-import java.io.IOException;
+//import java.io.IOException;
```
mmusgrov: Just remove the import statement.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantResource.java
```diff
@@ -90,7 +91,7 @@ public Response getStatus(@PathParam("participantId") final String participantId
     @Consumes(TxMediaType.TX_STATUS_MEDIA_TYPE)
     @Produces(TxMediaType.TX_STATUS_MEDIA_TYPE)
     public Response terminate(@PathParam("participantId") final String participantId, final String content)
-            throws HeuristicException {
+            throws ParticipantException, HeuristicException {
```
mmusgrov: I've just realised that ParticipantException is a [RuntimeException](https://devdocs.io/openjdk~17/java.base/java/lang/runtimeexception) and therefore does not need to be declare in the method signature. If it should be a checked exception then we would need to change the signature to `ParticipantException extends Exception` or minimally add the javadoc to methods that throw it indicating under what conditions the caller should be explicitly handling it. I have updated the [issue tracker for the javadocs](https://issues.redhat.com/browse/JBTM-2894) with this information. But for the purposes of this PR I would just back out your changes to the public method signatures throughout the PR, @jmfinelli do you think this is a reasonable way forward.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
```diff
@@ -96,60 +95,75 @@ public void persistParticipantInformation(final ParticipantInformation participa
...
             persistedParticipants.put(participantInformation.getId(), uid);
         } catch (Exception e) {
-            RESTATLogger.atI18NLogger.warn_persistParticipantInformationRecoveryManager(e.getMessage(), e);
+            String msg = RESTATLogger.atI18NLogger.warn_persistParticipantInformationRecoveryManager(e.getMessage(), e);
+            throw new ParticipantException(msg, e);
         }
     }
 
-    public void removeParticipantInformation(final ParticipantInformation participantInformation) {
+      public void removeParticipantInformation(final ParticipantInformation participantInformation) {
```
mmusgrov: Extra formatting.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
```diff
@@ -96,60 +95,75 @@ public void persistParticipantInformation(final ParticipantInformation participa
...
-            return ((PersistableParticipant) participant).getRecoveryState();
+        private byte[] getParticipantBytes(final Participant participant) throws ParticipantException {
+        try {
+            if (participant instanceof Serializable) {
+                return serializeParticipant((Serializable) participant);
+            } else if (participant instanceof PersistableParticipant) {
+                return ((PersistableParticipant) participant).getRecoveryState();
+            }
+        } catch (Exception e) {
+            String msg = RESTATLogger.atI18NLogger.warn_persistParticipantInformationRecoveryManager(e.getMessage(), e);
```
mmusgrov: The code in this catch block is identical to the catch block in getParticipantInformationOutputState on line 135 which is the only caller. I would recommend that you don't catch any exception and let getParticipantInformationOutputState handle it.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
```diff
@@ -96,60 +95,75 @@ public void persistParticipantInformation(final ParticipantInformation participa
...
+    private byte[] serializeParticipant(final Serializable participant) throws ParticipantException {
+        try {
+            final ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+            final ObjectOutputStream objectOutputStream = new ObjectOutputStream(byteArrayOutputStream);
+            objectOutputStream.writeObject(participant);
 
-        return byteArrayOutputStream.toByteArray();
+            return byteArrayOutputStream.toByteArray();
+        } catch (Exception e) {
+            String msg = RESTATLogger.atI18NLogger.warn_persistParticipantInformationRecoveryManager(e.getMessage(), e);
```
mmusgrov: The code in this catch block is identical to the catch block in getParticipantBytes on line 148 which is the only caller. I would recommend that you don't catch any exception here and let the caller of getParticipantBytes, ie getParticipantInformationOutputState, handle it.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
```diff
@@ -177,26 +191,24 @@ private void recoverParticipants() {
                         }
                     }
                 }
-            } catch (ObjectStoreException e) {
-                RESTATLogger.atI18NLogger.warn_recoverParticipantsRecoveryManager(e.getMessage(), e);
-            } catch (IOException e) {
-                RESTATLogger.atI18NLogger.warn_ioRecoverParticipantsRecoveryManager(e.getMessage(), e);
+            } catch (Exception e) {
+                String msg = RESTATLogger.atI18NLogger.warn_persistParticipantInformationRecoveryManager(e.getMessage(), e);
+                throw new ParticipantException(msg, e);
```
mmusgrov: Remark: Note that this exception will leak out into TransactionalService.registerDeserializer, ie the interface methods in ParticipantsManager can now throw ParticipantException.

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/functional/ParticipantsManagerTestCase.java
```diff
@@ -48,14 +49,15 @@ public void testSetBaseUrl() {
...
-    public void testReportHeuristic() throws MalformedURLException {
+    public void testReportHeuristic() throws MalformedURLException  {
         final String participantId = new Uid().toString();
         registerParticipant(participantId, new LoggingParticipant(new Prepared()));
         final ParticipantInformation participantInformation = ParticipantsContainer.getInstance().getParticipantInformation(participantId);
         participantInformation.setStatus(TxStatus.TransactionPrepared.name());
 
         participantsManager.reportHeuristic(participantId, HeuristicType.HEURISTIC_ROLLBACK);
 
+
```
mmusgrov: Extra line. Generally we don't allow reformatting in PRs. Similarly in the other places that this PR is reformatting the code.,

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/integration/BasicIntegrationTestCase.java
mmusgrov: The code is still calling `throw new ParticipantException(msg, e);`

tools/src/main/java/io/narayana/perf/Averager.java
```diff
@@ -4,7 +4,6 @@
  */
 
 package io.narayana.perf;
-
```
mmusgrov, line 7: The changes to the files in the tools module are unrelated and need to be backed out.

mmusgrov (changes requested):
The requirement in the issue tracker (https://issues.redhat.com/browse/JBTM-2493) is: 
 
> "during commit processing RecoveryManager#persistParticipantInformation needs to throw an exception if the participant cannot be written to the object store.".

Wouldn't a PR based on *just* the following be sufficient:

```
diff --git a/rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java b/rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
index c4446d2d5c..4819214b16 100644
--- a/rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
+++ b/rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
@@ -97,6 +97,7 @@ public final class RecoveryManager {
             persistedParticipants.put(participantInformation.getId(), uid);
         } catch (Exception e) {
             RESTATLogger.atI18NLogger.warn_persistParticipantInformationRecoveryManager(e.getMessage(), e);
+            throw new ParticipantException(e.getMessage());
         }
     }
```

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantResource.java
jmfinelli: +1 from my side. Thanks @mmusgrov

tools/src/main/java/io/narayana/perf/Averager.java
vedpathakyb: Reverted the changes.

rts/at/integration/src/test/java/org/jboss/narayana/rest/integration/test/functional/ParticipantsManagerTestCase.java
vedpathakyb: ack. Removed the extra line.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
vedpathakyb: Two methods getParticipantBytes & serializeParticipant are throwing IOException. I tested removing but it was compile error.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantsManagerImpl.java
vedpathakyb: Oh ok. Extra space removed.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/ParticipantResource.java
vedpathakyb: Removed ParticipantException from method signature.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
vedpathakyb: removed the extra space.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
vedpathakyb: Updated the getParticipantBytes method signature.  I even tried removing IOException but that was not getting compiled.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
vedpathakyb: Removed the catch block.

rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
vedpathakyb: Yes. Do we need to add "throws" registerDeserializer method signature of ParticipantManager.java

vedpathakyb:
> > > > The requirement in the issue tracker (https://issues.redhat.com/browse/JBTM-2493) is:
> > > > > "during commit processing RecoveryManager#persistParticipantInformation needs to throw an exception if the participant cannot be written to the object store.".
> > > > 
> > > > 
> > > > Wouldn't a PR based on _just_ the following be sufficient:
> > > > ```
> > > > diff --git a/rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java b/rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
> > > > index c4446d2d5c..4819214b16 100644
> > > > --- a/rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
> > > > +++ b/rts/at/integration/src/main/java/org/jboss/narayana/rest/integration/RecoveryManager.java
> > > > @@ -97,6 +97,7 @@ public final class RecoveryManager {
> > > >              persistedParticipants.put(participantInformation.getId(), uid);
> > > >          } catch (Exception e) {
> > > >              RESTATLogger.atI18NLogger.warn_persistParticipantInformationRecoveryManager(e.getMessage(), e);
> > > > +            throw new ParticipantException(e.getMessage());
> > > >          }
> > > >      }
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > And perhaps using the logger message string as the exception message.
> > > 
> > > 
> > > @vedpathakyb Thanks for your latest PR update. Did you notice my overall comment in the review ^
> > 
> > 
> > @mmusgrov Thanks for the inputs. As per my understanding, handling of ParticipantException should be removed from all methods except for the RecoveryManager#persistParticipantInformation. In that case, methods like removeParticipantInformation,recoverParticipants will have to catch ObjectStoreException. Can I change this two methods to catch ObjectStoreException?
> 
> @vedpathakyb - I expect what is trying to be avoided is to remove the public API signature, something like [vedpathakyb/narayana@vedpathakyb:JBTM-2493...tomjenkinson:narayana:JBTM-2493](https://github.com/vedpathakyb/narayana/compare/vedpathakyb:JBTM-2493...tomjenkinson:narayana:JBTM-2493)

I have made the required changes.

PR #2390, opened 2025-09-10
[SCRIPT] GitHub Actions updated

jmfinelli:
When the author of a PR pushes a new commit, some GH Actions are not triggered. The root cause is that the list of types for `pull_request` should also contain `synchronize`.

NO_TEST

### DIFF
--- a/.github/workflows/core-test.yml
+++ b/.github/workflows/core-test.yml
@@ -3,7 +3,7 @@ name: CORE
 on:
   push:
   pull_request:
-    types: [opened, labeled, unlabeled]
+    types: [ opened, edited, synchronize, ready_for_review, labeled, unlabeled, reopened ]
   workflow_dispatch:
     
 # Only run the latest job


.github/workflows/core-test.yml
```diff
@@ -3,7 +3,7 @@ name: CORE
 on:
   push:
   pull_request:
-    types: [opened, labeled, unlabeled]
+    types: [ opened, edited, synchronize, ready_for_review, labeled, unlabeled, reopened ]
```
mmusgrov, line 6: The issue description says that the root cause of the issue (of GH actions not triggering on new commits) is that "synchronize" is missing. What issue are the extra triggers addressing?

jmfinelli: `edited` will have us cover when we change the description of the pull request
`ready_for_review` will trigger tests when we change from draft to ready for review
`reopened` is basically a default event that `pull_requests` uses when `types` are not specified...I thought it was good to include it among the other events I specified to cover all possibilities

PR #2407, opened 2025-09-22
[JBTM-3997] another rework...

jmfinelli:
...and again another rework

CORE !AS_TESTS RTS !JACOCO XTS !QA_JTA !QA_JTS_OPENJDKORB !PERFORMANCE !DB_TESTS

### DIFF
--- a/XTS/localjunit/crash-recovery-tests/pom.xml
+++ b/XTS/localjunit/crash-recovery-tests/pom.xml
@@ -18,13 +18,6 @@
   <build>
     <plugins>
       <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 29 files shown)


XTS/localjunit/crash-recovery-tests/pom.xml
```diff
@@ -68,6 +61,7 @@
 
   <profiles>
     <profile>
+      <id>IBM JDK</id>
```
marcosgopen: @jmfinelli Could you use an id without ' ' (space char)?

jmfinelli: Sure, no problem

PR #2412, opened 2025-09-25
[SCRIPT] Improved release scripts

jmfinelli:
NO_TESTS

### DIFF
--- a/scripts/post-release.sh
+++ b/scripts/post-release.sh
@@ -23,7 +23,7 @@ fi
 echo "Proceeding..."
 
 set -e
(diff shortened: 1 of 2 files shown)


scripts/post-release.sh
```diff
@@ -34,6 +34,6 @@ do
   echo ""
 
   cd $REPO
-  git push origin $BRANCH --tags || fatal
+  git push origin main --tags || fatal
```
marcosgopen: What about maintenance branches tags?

jmfinelli: Fair, I didn't think about that. I'll add BRANCH back

PR #2537, opened 2026-02-18
JBTM-845 Datagrid ObjectStore

mmusgrov:
https://issues.redhat.com/browse/JBTM-845

CORE JDK25

A SlotStore implementation backed by an infinispan cache.

The test uses it with an in-VM cluster to verify that the store is correctly replicated across the cluster.

I'll follow up with a quickstart and some performance tests to tease out any other requirements - infinispan comes with an impressive set of configs that I'd like to experiment with.

This store is an in-memory datastore and can be backed be a cluster of infinispan nodes to maintain data availability provided the caches are suitably configured to manage replication of data across the cluster.

If the store is to be used with a recovery manager it is important that the environment is configured such that the usual caveats are maintained:
- jgroups-raft, or equivalent, is used to ensure strict consistency as required by a CP system
- only one recovery manager acts as the leader
- if the leader fails automatic failover chooses a new leader while avoiding split brain scenarios
- etc.

Maintaining these requirements is non-trivial and requires extra support from the environment and I'd expect that https://github.com/jbosstm/narayana-proposals/pull/5 will address that requirement.


ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/infinspan/InfinispanSlots.java
```diff
@@ -0,0 +1,104 @@
+/*
+   Copyright The Narayana Authors
+   SPDX-License-Identifier: Apache-2.0
+ */
+package com.arjuna.ats.internal.arjuna.objectstore.slot.infinspan;
```
marcosgopen: package r/infinspan/infinispan

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/infinispan/ClusterMemberId.java
```diff
@@ -0,0 +1,52 @@
...
+import java.nio.charset.StandardCharsets;
+
+/*
+ * a member of a recovery group
+ * the pair <failoverGroupId>:<nodeId> are managed by one Recovery Manager
+ */
+public class ClusterMemberId {
+    String nodeId;
+    // keys in the same failover group support migrate semantics within the group
+    String failoverGroupId;
```
marcosgopen: I think that the nodeId is unique key to identify the node. So I am wondering what is the scope for the failoverGroupId?
If I have to guess I would say that it is for Infinispan distributed caches. That is interesting. The Infinispan reference for that would be https://infinispan.org/docs/stable/titles/configuring/configuring.html#distributed-caches_clustered-caches where "Infinispan attempts to keep a fixed number of copies of any entry in the cache, configured as numOwners"
I haven't looked deeply into it as in my proposal I propose to use the  "Replicated caches" instead, where 'Infinispan replicates all entries in the cache to all nodes in the cluster.' (see https://infinispan.org/docs/stable/titles/configuring/configuring.html#replicated-caches_clustered-caches).

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/infinispan/InfinispanSlots.java
```diff
@@ -0,0 +1,104 @@
...
+            load(cache.keySet()); // set up the slot keys
+        } catch (Exception e) {
+            throw new IOException(e);
+        }
+    }
+
+    @Override
+    public void write(int slot, byte[] data, boolean sync) throws IOException {
+        try {
+            cache.put(slots[slot], data);
```
marcosgopen, line 128: I am analyzing the BackingSlots interface and I can see it uses 'byte[] data' (array of byte) and I have an observation:
 I think that the array of byte is useful when writing on the disk in an efficient way, but when using the infinispan cache we are using an abstraction of that as it is infinispan that takes care of persisting the data (see https://infinispan.org/docs/stable/titles/configuring/configuring.html#persistence). In this specific context I wonder if using 'byte []' is still convenient.

ArjunaCore/arjuna/pom.xml
```diff
@@ -18,6 +18,7 @@
 
   <properties>
     <skipTests>false</skipTests>
+    <version.infinispan>16.0.5</version.infinispan>
```
tomjenkinson: I think this should go in the root pom.xml

tomjenkinson: I wonder if we can import the WildFly pom to get the same dependency version: https://github.com/wildfly/wildfly/blob/main/pom.xml - I realise it's the same now but maybe we would diverge unless someone commits to tracking it. This is not just for this dependency so please think of it is an open remark

marcosgopen: Maybe using the 'wildfly-common-ee-dependency-management' bom we would inherit the correct infinispan version (see https://github.com/wildfly/wildfly/blob/main/boms/common-ee/pom.xml#L3083-L3089)

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/infinispan/ClusterMemberId.java
mmusgrov: The cache config is not part of the feature since that is managed by the user of store, similar to how we configure the object store when we integrate with things like WildFly and quarkus.

This PR is just providing the store type, how it get's used is not part of the PR. That said the test I have provided uses CacheMode.REPL_SYNC.

Similarly the failoverGroupId is metadata that is available to the user of the store. I based the implementation of this store on the RedisStore that [blogged about](https://jbossts.blogspot.com/2024/06/some-experiments-in-migrating.html). It was useful for that store but I acknowledge that it isn't necessarily needed for the Infinispan store - I'll have a think and see if it makes sense to abstract it into an interface.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/infinspan/InfinispanSlots.java
```diff
@@ -0,0 +1,104 @@
...
+            return cache.get(slots[slot]);
+        } catch (Exception e) {
+            throw new IOException(e);
+        }
+    }
+
+    @Override
+    public void clear(int slot, boolean sync) throws IOException {
+        try {
+            // TODO evicting cache entries has some caveats
```
tomjenkinson: Would seem to need it's own JBTM if it gets merged with the TODO

ArjunaCore/arjuna/pom.xml
jmfinelli: +1

jmfinelli: I’d prefer to keep our dependency management independent from WildFly. Of course, we can try to align this dependency’s version (and others) with WildFly’s, but our CI should be robust enough to catch failures caused by version differences.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/infinispan/InfinispanSlots.java
mmusgrov: I'm not sure, but Infinispan is a key value store where the key and value are "owned" by the business logic (in this case we are the business logic).

ArjunaCore uses byte streams to manage [object state](https://javadoc.io/static/org.jboss.narayana.jta/narayana-jta/7.3.3.Final/com/arjuna/ats/arjuna/state/OutputBuffer.html) so that's why we need to use byte arrays.

mmusgrov: The place in the code where we write a new copy of the object's committed state is https://github.com/jbosstm/narayana/blob/7.3.3.Final/ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/SlotStoreAdaptor.java#L271

ArjunaCore/arjuna/pom.xml
tomjenkinson: OK!

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/infinispan/ClusterMemberId.java
tomjenkinson: So will we advise our users with specific cache config to use, or not? I think I infer from what is said that people are able to configure how ever they like, but is it possible to at least suggest a known good configuration (like if REPL_SYNC is known to be ACID we might suggest that?)

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/infinispan/InfinispanClusterTest.java
```diff
@@ -0,0 +1,388 @@
...
+import java.lang.invoke.VarHandle;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class InfinispanClusterTest {
```
mmusgrov, line 48: @jmfinelli has asked for some tests showing how to configure write-through caches, ie caches backed a file system and/or a database etc. I'm adding this comment here so that I can include the request in the next round of reviews.

mmusgrov (commented):
Adding a private comment from Manuel so that we don't loose sight of it.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/infinispan/InfinispanClusterTest.java
mmusgrov: In addition Tom and Manuel both asked for some performance comparisons between this new infinispan backed slot store, in its various configurations, and the existing DiskSlots slot store implementation. I'll also throw in a comparison with the existing file system store which seems to be the most used even though the Artemis Journal based store is faster so that may be worth comparing with too.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/infinispan/InfinispanSlots.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        slots = new byte[slotStoreConfig.getNumberOfSlots()][];
+        infinispanSlotKeyGenerator = config.getSlotKeyGenerator();
+        infinispanSlotKeyGenerator.init(config);
+
+        try {
+            cache = config.getCache();
+
+            load(cache.keySet()); // set up the slot keys
```
marcosgopen: keySet() has no ordering guarantee. Data can be in different order after restart: e.g cache.get(slots[1]) value before restart can be mapped to cache.get(slots[x]) after

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/infinispan/InfinispanReplicatedAndPersistentTest.java
```diff
@@ -0,0 +1,88 @@
...
+        Assertions.assertTrue(Files.exists(store1.path()), "infinispan persistence store 1 was not created");
+        Assertions.assertTrue(Files.exists(store2.path()), "infinispan persistence store 2 was not created");
+
+        // verify that the slot stores at each node have the same values
+        byte[] value1 = store1.slots().read(0);
+        byte[] value2 = store2.slots().read(0);
+        byte[] value3 = store1.slots().read(1);
+        byte[] value4 = store2.slots().read(1);
+
+        Assertions.assertArrayEquals(value1, value2);
```
marcosgopen, line 68: Please consider comparing the read value to kv1.value/kv2.value to avoid null == null assertions:
e.g. Assertions.assertArrayEquals(value1,  kv1.value); and so on

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/infinispan/InfinispanSlots.java
mmusgrov: It's just a set of keys so doesn't need to be ordered, see my answer below to you question about generateUniqueKey to see how the slot table works.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/infinispan/InfinispanReplicatedAndPersistentTest.java
mmusgrov: It a good spot. The reason the values are null is because I start the slot store and then I bypass the slot store by writing directly to the cache so the store would not see the new entries until it was restarted. The test is meant to be minimal so I don't go through the store write route (other tests do that) so the test fix is to move the store initialisation until after the direct cache writes. I'll push that change in my next update.

marcosgopen: I see, thanks. Alternatively you can just load() the slots so that the store can see the new entries

mmusgrov: The load() should be under the control of the SlotStoreAdaptor, the API for users (generally that should be the TM and the RM) is the object store api ie ObjectStoreAPI

mmusgrov: And the API to the infinispan store is ` interface BackingSlots` which does not have a load method and, besides, the interface is for use by the SlotStoreAdaptor and using it independently of the transaction/recovery system would be a problem - I'll add a note but also notice that the infinispan package (com.arjuna.ats.internal.arjuna.objectstore.slot.infinispan) is internal so would never be called by users.

mmusgrov: I notice that I made the same error in InfinispanGrouperTest#testWithMetadata so thanks much appreciated for spotting the issue.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/infinispan/InfinispanSlots.java
mmusgrov: My response below that discusses how the SlotStoreAdaptor works with the backing slot implementations may help shed light on this question.

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/infinspan/InfinispanSlots.java
mmusgrov: fixed by commit d24ef9945 JBTM-845 fix typo in package name

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/infinispan/ClusterMemberId.java
mmusgrov: fixed by commit 6d613bf45 JBTM-845 Add a configurable interface for creating infinispan slot store keys (the replacement is an interface called InfinispanSlotKeyGenerator which users supply via the config)
but see important tweaks added in some of the other commits.

I didn't include structured advice on which specific cache config to use depending upon which use cases are required but I've bookmarked the request (I think it's best if I include that in a follow up post to https://jbossts.blogspot.com/ if you agree). Later on we can then migrate that to our docs.

ArjunaCore/arjuna/pom.xml
mmusgrov: 26b920490 JBTM-845 define infinispan version in root pom

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/infinspan/InfinispanSlots.java
mmusgrov: Fixed with commit 3168a3f32 JBTM-845 Add a test for persistent cache stores
The fix was to replace `cache.evict` with `cache.remove` which removes it from all (replicated) caches.
The [javadoc for evict](https://infinispan.org/docs/dev/apidocs/org/infinispan/Cache.html#evict(K)) explains the semantics of evict vs put (quote "Evicts an entry from the memory of the cache. Note that the entry is not removed from any configured cache stores or any other caches in the cluster (if used in a clustered mode). Use BasicCache.remove(Object) to remove an entry from the entire cache system").

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/infinispan/InfinispanClusterTest.java
mmusgrov: The first comment was fixed with commit 3168a3f32 JBTM-845 Add a test for persistent cache stores
The feature is added by test InfinispanReplicatedAndPersistentTest#testWriteThroughCache by disabling   passivation on the PersistenceConfigurationBuilder (this happens in InfinispanTestBase.createCacheManager based on the persistence flag passed in (the javadoc does explain but the [docs do](https://infinispan.org/docs/stable/titles/configuring/configuring.html#how-passivation-works_persistence).

I am working on the performance comparisons and will follow up in a separate issue/pr.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/infinispan/InfinispanConfigTest.java
```diff
@@ -0,0 +1,59 @@
...
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class InfinispanConfigTest extends InfinispanTestBase {
```
mmusgrov, line 28: I forgot to document what this test is for, hopefully it is self explanatory but I'll add some if required (basically it tests configuring an infinispan store using the standard infinispan xml config whereas the other tests all use programmatic config).

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/logging/arjunaI18NLogger.java
```diff
@@ -1660,6 +1660,11 @@ public void warn_objectstore_JDBCImple_over_max_image_size(int imageSize,
 	@LogMessage(level = WARN)
 	void warn_coordinator_BasicAction_73(String cause);
 
+	@Message(id = 12418, value = "The number of actual transaction logs, {0}, exceeds the number of configured " +
+			"slots, {1}. The number of configured slots should equal the maximum number of unresolved transactions " +
+	        "expected at any given time, including those in-flight and awaiting recovery.", format = MESSAGE_FORMAT)
+	String get_infinipan_too_few_slots(int actual, int configured);
```
marcosgopen: typo get_infinipan_too_few_slots/get_infinispan_too_few_slots

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/infinispan/InfinispanStoreEnvironmentBean.java
```diff
@@ -0,0 +1,147 @@
...
+
+    public Cache<byte[], byte[]> getCache() {
+        return cache;
+    }
+
+    public void setCache(Cache<byte[], byte[]> cache) {
+        this.cache = cache;
+
+        if (isIgnoreReturnValues()) {
+            cache.getAdvancedCache().withFlags(Flag.IGNORE_RETURN_VALUES);
```
marcosgopen: I think 'cache.getAdvancedCache().withFlags(Flag.IGNORE_RETURN_VALUES);' does not update the cache, but returns a new flagged cache. So maybe 
this.cache = cache.getAdvancedCache().withFlags(Flag.IGNORE_RETURN_VALUES); ?

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/infinispan/InfinispanClusterTest.java
```diff
@@ -0,0 +1,476 @@
...
+     * creates a cache manager, the intent is to have one such manager per cluster node
+     *
+     * @param nodeName Name of the current node. This is a friendly name to make logs, etc. make more sense.
+     *                 Defaults to a combination of host name and a random number
+     *                 (to differentiate multiple nodes on the same host) although any name can be used
+     */
+    private DefaultCacheManager createCacheManager(String nodeName) {
+        GlobalConfigurationBuilder globalConfig = GlobalConfigurationBuilder.defaultClusteredBuilder();
+
+        globalConfig.transport().nodeName(nodeName).clusterName(CLUSTER_NAME);
```
marcosgopen: It is strange, but for me locally test passes only if I replace globalConfig.transport().nodeName(nodeName).clusterName(CLUSTER_NAME);
with
globalConfig.transport().nodeName(nodeName).clusterName(CLUSTER_NAME)
                .addProperty("configurationFile", "jgroups.xml");

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/infinispan/InfinispanTestBase.java
```diff
@@ -0,0 +1,306 @@
...
+
+    static DefaultCacheManager createCacheManager(String nodeName,
+                                                   CacheMode cacheMode,
+                                                   int numOwners, // used with CacheMode.CacheMode
+                                                   Grouper<WrappedByteArray> grouper,
+                                                   boolean persistence,
+                                                   boolean partitionResilience) {
+        GlobalConfigurationBuilder globalConfig = GlobalConfigurationBuilder.defaultClusteredBuilder();
+
+        globalConfig.transport().nodeName(nodeName).machineId(nodeName).clusterName(CLUSTER_NAME);
```
marcosgopen: Same here:
for me locally test passes only if I replace globalConfig.transport().nodeName(nodeName).clusterName(CLUSTER_NAME);
with
globalConfig.transport().nodeName(nodeName).clusterName(CLUSTER_NAME)
                .addProperty("configurationFile", "jgroups.xml");

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/infinispan/InfinispanClusterTest.java
mmusgrov: I pushed a test (InfinispanStoreComparisonTest) to compare the new store type against the other disk based slot stores, the filesystem store and the volatile store. All the stores offer similar performance (that said my run did show a slight improvement on the two DiskSlot implementations), including the volatile store which is strange so I'm investigating that. I configured the infinispan store with a write-through cache which means that it also writes to the filesystem which enables a better comparison with the other file based stores (DiskSlots, MappedDiskSlots and ShadowNoFileLockStore). I was surprised that the new store wasn't slower than the other disk based stores because the new store is doing extra work - and it will be slower still on a clustered cache since the network and/or IPC (if multiple caches run on a single node) overhead will kick in:

== com.arjuna.ats.internal.arjuna.objectstore.slot.infinispan.InfinispanSlots: throughput: 53561.863281 (18670 ms)                                                                                                  
== com.arjuna.ats.internal.arjuna.objectstore.TwoPhaseVolatileStore: throughput: 54692.628906 (18284 ms)                                                                                                            
== com.arjuna.ats.internal.arjuna.objectstore.ShadowNoFileLockStore: throughput: 54558.351563 (18329 ms)                                                                                                            
== com.arjuna.ats.internal.arjuna.objectstore.slot.MappedDiskSlots: throughput: 53524.593750 (18683 ms)                                                                                                             
== com.arjuna.ats.internal.arjuna.objectstore.slot.DiskSlots: throughput: 53199.980469 (18797 ms)

mmusgrov: I tuned the number of threads used to run the benchmark setting it to the number of available processors and the results now more accurately reflect, in my opinion, the overhead of using write through infinispan caches, although there is still an oddity in that the volatile store is only the fourth fastest so I will continue my investigations:

== com.arjuna.ats.internal.arjuna.objectstore.slot.infinispan.InfinispanSlots: throughput: 51480.050781 (1554 ms 20 threads 80000 actions)
== com.arjuna.ats.internal.arjuna.objectstore.TwoPhaseVolatileStore: throughput: 72137.062500 (1109 ms 20 threads 80000 actions)
== com.arjuna.ats.internal.arjuna.objectstore.ShadowNoFileLockStore: throughput: 77444.335938 (1033 ms 20 threads 80000 actions)
== com.arjuna.ats.internal.arjuna.objectstore.slot.MappedDiskSlots: throughput: 77669.906250 (1030 ms 20 threads 80000 actions)
== com.arjuna.ats.internal.arjuna.objectstore.slot.DiskSlots: throughput: 74906.367188 (1068 ms 20 threads 80000 actions)

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/infinspan/InfinispanSlots.java
mmusgrov: Can this conversation be resolved now?

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/infinispan/ClusterMemberId.java
mmusgrov: Can this conversation be resolved now?

ArjunaCore/arjuna/pom.xml
mmusgrov: Can this conversation be resolved now?

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/infinspan/InfinispanSlots.java
mmusgrov: Can this conversation be resolved now?

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/infinispan/InfinispanClusterTest.java
mmusgrov: @jmfinelli Can this conversation be resolved now?

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/infinispan/InfinispanSlots.java
mmusgrov: Can this conversation be resolved now?

ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/logging/arjunaI18NLogger.java
mmusgrov: Can this conversation be resolved now?

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/infinispan/InfinispanClusterTest.java
mmusgrov: Can this conversation be resolved now?

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/infinispan/InfinispanTestBase.java
mmusgrov: Can this conversation be resolved now?

ArjunaCore/arjuna/classes/com/arjuna/ats/internal/arjuna/objectstore/slot/infinispan/InfinispanStoreEnvironmentBean.java
mmusgrov: But if that was the case then it would imply that infinispan makes a fresh copy of the configuration, discards the old copy and associates the new copy with the cache which seems inefficient given how large the config can get. I couldn't find the property in the debugger (the cache data structure is quite complex so I may have been looking in the wrong place - I dumped the hundreds of attributes of the cache config (including the getAdvancedCache config) and it wasn't there.

Anyway I applied your suggested fix since it is only applied when the store is constructed so will not impact performance (and trust that it works).

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/infinispan/InfinispanGrouperTest.java
```diff
@@ -0,0 +1,437 @@
...
+        /*
+         * verify that a new primary is selected if the current one is down
+         */
+
+        // stop the primary
+        primaryStore.stop();
+
+        String newPrimary = dm.getCacheTopology().getDistribution(aKey).primary().getMachineId();
+        // a new primary store should have been chosen since the old one was removed from the cluster when we stopped it:
+        Assertions.assertNotEquals(primary, newPrimary);
```
marcosgopen, line 325: I can see this assertion fails intermittently. I think it is a timing issue.

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/infinispan/InfinispanStoreComparisonTest.java
```diff
@@ -0,0 +1,161 @@
...
+import io.narayana.perf.WorkerWorkload;
+import org.infinispan.commons.marshall.WrappedByteArray;
+import org.infinispan.configuration.cache.CacheMode;
+import org.infinispan.distribution.group.Grouper;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.io.IOException;
+
+public class InfinispanStoreComparisonTest extends InfinispanTestBase {
```
tomjenkinson: I think this test should be in https://github.com/jbosstm/performance/ ?

tomjenkinson: A version of a file similar to https://github.com/jbosstm/performance/blob/main/narayana/ArjunaJTA/jta/tests/classes/com/arjuna/ats/jta/xa/performance/HQStoreBenchmark.java

tomjenkinson: We can then use the existing performance tests to see the difference in performance of the new store, vs the existing stores

tomjenkinson: Possibly we should also extend that harness to be able to run a second ISPN cache so we can then assess more like a user would experience in real scenarios.

tomjenkinson: We will need something like https://github.com/jbosstm/performance/pull/193 to be in to add a the performance test in a separate pull to that repo, but I think it's better to have it there

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/infinispan/InfinispanGrouperTest.java
mmusgrov: ok, I'll see what the documentation says about how long it can take for the DistributionManager to detect the topology change (and I know there's a topology change listener that could help too).

ArjunaCore/arjuna/tests/classes/com/hp/mwtests/ts/arjuna/objectstore/infinispan/InfinispanClusterTest.java
jmfinelli: +1 to what Tom said [here](https://github.com/jbosstm/narayana/pull/2537#discussion_r2939769142).

mmusgrov:
I've added the Hold label so that I can trigger a build of the PR via the perf and quickstart repos, hope it works.

mmusgrov:
The latest push (commits 63677ded0b and 05edc5a773) includes:                                                                                                                                                       
- documentation                                                                                                                                                                                                       
- improved configuration                                                                                                                                                                                            
- clearly marking the feature as experimental in line with our procedure for marking such features                                                                                                                  
  (see the update to the documentation and the ARJUNA012419 message)                                                                                                                                                                 
- moved the performance testing over to the perf repo                                                                                                                                                               
                                                                                                                                                                                                                    
I've also pushed the [quickstart](https://github.com/jbosstm/quickstart/pull/629) and the [performance tests](https://github.com/jbosstm/performance/pull/204) to their respective repos.

marcosgopen (approved):
LGTM, thanks Mike!

mmusgrov:
> LGTM, thanks Mike!

@marcosgopen Thanks, in that case I will rebase and merge.
The associated [performance](https://github.com/jbosstm/performance/pull/204) pull request had a successful GH Action test run (again I appreciate the help you provided on those two) - is there anything else we need to add to it. FYI I started a similar JGroups based solution this week so I hope to follow up with a PR for that work next week.

marcosgopen:
> > LGTM, thanks Mike!
> 
> @marcosgopen Thanks, in that case I will rebase and merge. The associated [performance](https://github.com/jbosstm/performance/pull/204) pull request had a successful GH Action test run (again I appreciate the help you provided on those two) - is there anything else we need to add to it. FYI I started a similar JGroups based solution this week so I hope to follow up with a PR for that work next week.

Thanks @mmusgrov , looking forward to the Jgroup based solution!

PR #3008, opened 2026-06-02
Update RFE template

jmfinelli:


.github/ISSUE_TEMPLATE/request-for-enhancement.md
```diff
@@ -0,0 +1,23 @@
...
+labels: type/enhancement
+assignees: ''
+type: Feature
+
+---
+
+**Proposing a major feature? If your request requires design planning and/or discussion with the core team, please submit it via the [narayana-proposals](https://github.com/jbosstm/narayana-proposals) workflow.**
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
```
marcosgopen, line 14: I wonder if it is a problem maybe we should open a bug instead? Maybe 'related to a new feature' or 'related to an enhancement/improvement' would be better?

jmfinelli: My interpretation of the default GitHub template is that a _problem_ is not necessarily the same as a _bug_. In other words, a problem could refer to something you would like the application to be able to do or address, rather than an issue caused by incorrect behavior.

marcosgopen: Fair point

marcosgopen (approved):

PR #3009, opened 2026-06-02
Update CONTRIBUTING.md following GitHub integration

jmfinelli:
NO_TESTS

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,7 +5,9 @@ Want to contribute? Great!
 We try to make it easy, and all contributions, even the smaller ones, are more than welcome.
 This includes bug reports, fixes, documentation, examples...
 
-If you are looking for an issue to work on and haven't had much previous experience with Narayana then you could choose one that has the label [good-first-issue or hacktoberfest](https://issues.redhat.com/browse/JBTM-2493?filter=12421681) or one whose "Estimated Difficulty" field is `Low`. For Hacktoberfest we have created a zulip stream called [hacktoberfest](https://narayana.zulipchat.com/#narrow/stream/406889-hacktoberfest/topic/stream.20events/near/393204842) for discussions. If you want to take an issue then ping a team member (or add a message to the zulip stream) and we will update the assignee field.
+If you are looking for an issue to work on and haven't had much previous experience with Narayana then you could choose one that has the label [good-first-issue](https://github.com/jbosstm/narayana/issues?q=state%3Aopen%20label%3Agood-first-issue) or [hacktoberfest](https://github.com/jbosstm/narayana/issues?q=state%3Aopen%20label%3Ahacktoberfest).
+For `hacktoberfest` we have created a zulip stream called [hacktoberfest](https://narayana.zulipchat.com/#narrow/stream/406889-hacktoberfest/topic/stream.20events/near/393204842) for discussions.
+If you want to take on an issue, simply assign it to yourself. If you don't have the necessary permissions, leave a comment saying, "I'd like to work on this."
 
 On the other hand, if you are set to tackle a big, complicated issue or an issue that will have a high impact on the code base, it is highly recommended to follow the [Request for enhancement (RFE) [workflow](https://github.com/jbosstm/narayana-proposals/blob/main/README.md). Choosing to employ this workflow before starting the development of the selected RFE will help the Narayana core team and yourself to clarify the requirements, the constraints, and the overall design to tackle the issue in advance. In other words, employing the RFE workflow will make sure that the Narayana core team and yourself are on the same page before starting coding.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -43,22 +45,11 @@ Although copyright notices on contributions are not strictly necessary we do ask
...
+To report the issue use the [GitHub issue tracker](https://github.com/jbosstm/narayana/issues). Don't forget to indicate which version of Narayana, Java and Maven you are using.
 
 ## Making open source more inclusive
 
-Red Hat is committed to eradicating problematic language in all of our interactions, including code, documentation, web properties, etc (https://www.redhat.com/en/blog/making-open-source-more-inclusive-eradicating-problematic-language). Although the Narayana project has no plans to retrospectively update its materials we do ask that all contributors be cognizant of the language they use in all communications related to Narayana.
+In accordance with the [Commonhaus Foundation Code of Conduct](https://www.commonhaus.org/policies/code-of-conduct/), we are committed to using welcoming and inclusive language in all our interactions, including, but not limited to, code, documentation, and web properties. Although the Narayana project does not plan to retroactively update historical materials, we expect all contributors to uphold these standards in all Narayana-related communications and across all our communication channels.
```
marcosgopen, line 52: +1

CONTRIBUTING.md
```diff
@@ -160,18 +150,40 @@ Also don't forget the documentation (reference documentation for features, javad
 
 ## Update the issue with the correct release information:
 
-If the change would result in behaviour in Narayana that is incompatible with the current release stream for Narayana, then the relevant JIRA issue `Fix Version/s` must be set to the `<version+1>.next`. For example if the root pom.xml is currently `<version>7.0.3.Final-SNAPSHOT</version>` then the JIRA issue should have a `Fix Version/s` of `8.next` added. When resolving the issue in JIRA we encourage the assignee to consider adding a release note describing the impact of the change.
+If the change would result in behaviour in Narayana that is incompatible with the current release stream for Narayana, then the relevant GitHub issue must be included in the milestone `<major_version+1>.next`. For example if the root `pom.xml` is currently `<version>7.0.3.Final-SNAPSHOT</version>` then the GitHub issue should be included in the milestone `8.next`. When resolving the issue we encourage the assignee to consider adding a release note describing the impact of the change.
```
marcosgopen: +1 for mentioning the milestone, is it the GitHub milestone? If so could you add the link to it here?

CONTRIBUTING.md
```diff
@@ -160,18 +150,40 @@ Also don't forget the documentation (reference documentation for features, javad
...
+
+Don't forget to indicate which version of Narayana, Java and Maven you are using.
+
+Additionally, for more involved items of work, if you are able to estimate the work required to fix the issue or RFE please provide, as best you can, details using the following fields:
+
+- *Original and Remaining Estimate* for estimating resources
+- *Estimated Difficulty* for estimating complexity
 
-When the github Pull Request has passed all relevant CI checks and has been Approved by a reviewer the code can be merged. If you don't have permission to do this then ping one of the team who will then merge it. Once merged the issue must be updated in the issue tracker (if you don't have permission then a team member will do this):
+When the GitHub Pull Request has passed all relevant CI checks and has been approved by a reviewer the code can be merged.
```
marcosgopen: This line looks very similar to line 155, is that intended?

CONTRIBUTING.md
```diff
@@ -160,18 +150,40 @@ Also don't forget the documentation (reference documentation for features, javad
...
+
+When the GitHub Pull Request has passed all relevant CI checks and has been Approved by a reviewer, the code can be merged. If you don't have permission to do this then ping one of the team who will then merge it.
+
+## Product-related issues
+
+> [!NOTE]
+> This section is for the core team only!
+
+When product-specific issues are reported upstream, you must create a corresponding tracking ticket in the [JBTM issue tracker](https://issues.redhat.com/projects/JBTM).
+Please fill out as many relevant fields as possible, specifically:
```
marcosgopen: Not sure if a JBTM needs to be created in this case or maybe a JBEAP (or a quarkus one ) would be a better fit, which must be linked to the upstream issue? 

Could you explain me why this would be for the core team only?

I think a JBTM should be created for a security related (internal) issue only. In that case I agree that is for the core team only, even if a user can report it to narayana-security

CONTRIBUTING.md
jmfinelli: Yep, it is. Basically, this section should guide the core team when they need to handle JBTMs.

CONTRIBUTING.md
jmfinelli: Thanks for your reply.

> Not sure if a JBTM needs to be created in this case or maybe a JBEAP (or a quarkus one ) would be a better fit, which must be linked to the upstream issue?

Product-specific Narayana community issues should be cloned into the JBTM project. The resulting JBTM issue can then be linked to the corresponding product Jira issue. This section should provide guidance to the core team on how to handle community issues that also affect a product, ensuring the appropriate issue tracking relationships are established.

> I think a JBTM should be created for a security related (internal) issue only. In that case I agree that is for the core team only, even if a user can report it to narayana-security

Yep, I agree. This section is intended to cover the scenario where an issue is first reported in the community project and subsequently needs to be tracked internally as well. The opposite case (i.e. where an issue originates internally and later needs to be cloned upstream) is outside the scope of this section. 😉

marcosgopen: > Product-specific Narayana community issues should be cloned into the JBTM project. The resulting JBTM issue can then be linked to the corresponding product Jira issue. This section should provide guidance to the core team on how to handle community issues that also affect a product, ensuring the appropriate issue tracking relationships are established.

AFAIR we talked about that in the past for the LRA project and we agreed that a JBTM is not needed as it would be a duplicate of the GitHub issue in this case. Do you think this is a requirement from a specific product? From my point of view the product related issue (e.g. a JBEAP) can link directly the narayana GH issue without having the need to link a JBTM which links the GH issue.

jmfinelli: I backed out this part. As we discussed, I will add it to our Wiki. Thanks Marco

CONTRIBUTING.md
jmfinelli: Done. Thanks

marcosgopen (approved):
LGTM, thanks @jmfinelli

CONTRIBUTING.md
```diff
@@ -102,8 +93,8 @@ To contribute, use GitHub Pull Requests (PRs), from your **own** fork.
...
 
-1. The Pull Request title is properly formatted: `JBTM-XYZ Subject`
-2. The Pull Request *should* contain a link to the JIRA issue(s) at the start of the PR description (only minor changes to script/text files are exempt from this rule). If the engineer wishes to address multiple issues and they are closely related then they can be addressed in a single PR. The JIRA must contain sufficient information to enable the reader to understand what the issue is, so at a minimum the description field of the JIRA must be present and legible/clear.
+1. The Pull Request title is properly formatted: `[#XXXX] Subject`
+2. The Pull Request *should* contain a link to the GitHub issue(s) at the start of the PR description (only minor changes to script/text files are exempt from this rule). If the engineer wishes to address multiple issues and they are closely related then they can be addressed in a single PR. The GitHub issue(s) must contain sufficient information to enable the reader to understand what the issue is, so at a minimum the description field of the issue must be present and legible/clear.
```
mmusgrov, line 97: Ah yes of course, these needed updating.

PR #3061, opened 2026-08-13
[JBTM-576] : update excluded unit tests

Ohifeme:
issue: #2807 

Stress tests (LogStressTest2, TransactionTest2) are now runnable on demand via mvn test -Pstress-tests rather than being silently excluded.

### DIFF
--- a/ArjunaCore/arjuna/pom.xml
+++ b/ArjunaCore/arjuna/pom.xml
@@ -566,8 +566,6 @@
             <artifactId>maven-surefire-plugin</artifactId>
             <configuration>
               <excludes>
-                <!-- stress tests are not unit tests -->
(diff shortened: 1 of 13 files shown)


jmfinelli:
Hi @Ohifeme, thank you very much for your PR!

I have a couple of questions:
* Do you think we should address CodeRabbit's review, or are its comments safe to ignore?
* I was also wondering whether we should improve the CI for enabling the `stress-tests` profile.
* It also looks like the PR doesn't cover all the tests mentioned in [Jonathan's comment](https://github.com/jbosstm/narayana/issues/2807#issuecomment-4594274814). Am I understanding that correctly? If so, should we consider this PR a (very welcome) incremental step toward completing #2807?

PR #3068, opened 2026-08-21
[#2966] Avoid illegal initCause on InvalidTransactionException in JTS…

tomjenkinson:
https://github.com/jbosstm/narayana/issues/2966

CORE AS_TESTS !RTS !JACOCO !XTS !QA_JTA QA_JTS_OPENJDKORB !PERFORMANCE !DB_TESTS mysql db2 postgres oracle

### DIFF
--- a/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/TransactionManagerImple.java
+++ b/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/TransactionManagerImple.java
@@ -149,9 +149,7 @@ public class TransactionManagerImple extends BaseTransaction implements
 	        }
 	        catch (org.omg.CosTransactions.InvalidControl e1)
 	        {
-                InvalidTransactionException invalidTransactionException = new InvalidTransactionException();
-                invalidTransactionException.initCause(e1);
-                throw invalidTransactionException;
+                throw new InvalidTransactionException(e1.getMessage());
 	        }
 	        catch (org.omg.CORBA.SystemException e2)
 	        {
--- a/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/TransactionManagerImple.java
+++ b/ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/TransactionManagerImple.java
@@ -149,7 +149,9 @@ public class TransactionManagerImple extends BaseTransaction implements
 	        }
 	        catch (org.omg.CosTransactions.InvalidControl e1)
 	        {
-                throw new InvalidTransactionException(e1.getMessage());
+                InvalidTransactionException itc = new InvalidTransactionException(e1.getMessage());
+                itc.detail = e1;
+                throw itc;
 	        }
 	        catch (org.omg.CORBA.SystemException e2)
 	        {
--- a/ArjunaJTS/jtax/tests/classes/com/hp/mwtests/ts/jta/jts/twophase/TransactionManagerImpleUnitTest.java
+++ b/ArjunaJTS/jtax/tests/classes/com/hp/mwtests/ts/jta/jts/twophase/TransactionManagerImpleUnitTest.java
@@ -11,6 +11,8 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.CoreMatchers.startsWith;
 
 import jakarta.transaction.InvalidTransactionException;
 import jakarta.transaction.Transaction;
@@ -87,7 +89,13 @@ public class TransactionManagerImpleUnitTest extends TestBase
             // JBTM-4013: before the fix, initCause on InvalidTransactionException
             // (which extends RemoteException) threw IllegalStateException here
             // instead of propagating as InvalidTransactionException
-            assertEquals("IDL:omg.org/CosTransactions/InvalidControl:1.0", e.getMessage());
+            // RemoteException.getMessage() appends nested exception when detail is set,
+            // so we check that the message starts with the repo ID
+            assertThat(e.getMessage(), startsWith("IDL:omg.org/CosTransactions/InvalidControl:1.0"));
+            // Full message includes the nested exception detail:
+            // "IDL:omg.org/CosTransactions/InvalidControl:1.0; nested exception is: \n\torg.omg.CosTransactions.InvalidControl: IDL:omg.org/CosTransactions/InvalidControl:1.0"
+            assertTrue("message should contain nested exception",
+                e.getMessage().contains("; nested exception is:"));
             assertNotNull("cause should be preserved via detail", e.getCause());
             assertTrue(e.getCause() instanceof org.omg.CosTransactions.InvalidControl);
         }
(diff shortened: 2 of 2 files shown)


marcosgopen:
@tomjenkinson could you please link the issue in the description to make it simpler following along (even if the issue number is already present in the PR title)? Thanks

I believe it is https://github.com/jbosstm/narayana/issues/2966

ArjunaJTS/jtax/tests/classes/com/hp/mwtests/ts/jta/jts/twophase/TransactionManagerImpleUnitTest.java
```diff
@@ -48,4 +56,83 @@ public void testNested () throws Exception
...
+            // InvalidControl is thrown naturally when resuming a remote
+            // transaction whose coordinator is no longer reachable (e.g. the
+            // remote ORB crashed). Reproducing that requires a multi-node
+            // setup, so Byteman injects the fault directly into resumeWrapper.
+            tmi.resume(tx);
+            fail("Expected InvalidTransactionException");
+        }
+        catch (InvalidTransactionException e)
+        {
+            // JBTM-4013: before the fix, initCause on InvalidTransactionException
```
marcosgopen, line 89: +1

ArjunaJTS/jtax/tests/classes/com/hp/mwtests/ts/jta/jts/twophase/TransactionManagerImpleUnitTest.java
```diff
@@ -48,4 +56,83 @@ public void testNested () throws Exception
...
+        };
+
+        try
+        {
+            tmi.resume(foreignTransaction);
+            fail("Expected InvalidTransactionException");
+        }
+        catch (InvalidTransactionException e)
+        {
+            // resume must reject transaction types it does not recognise
```
marcosgopen, line 126: +1

ArjunaJTS/jtax/tests/classes/com/hp/mwtests/ts/jta/jts/twophase/TransactionManagerImpleUnitTest.java
```diff
@@ -48,4 +56,83 @@ public void testNested () throws Exception
...
+        }
+    }
+
+    @Test
+    public void testInvalidTransactionExceptionInitCauseIsDisabled ()
```
marcosgopen: I am not sure this test is needed. I understand the initCause throws IllegalStateException, I think it is out of scope to add a test for it. What it would be good would be to test that the message included in the InvalidControl exception is preserved in the InvalidTransactionException. WDYT? @tomjenkinson

tomjenkinson: That is a great suggestion! I'll get it updated

ArjunaJTS/jtax/tests/classes/com/hp/mwtests/ts/jta/jts/twophase/TransactionManagerImpleUnitTest.java
```diff
@@ -48,4 +56,63 @@ public void testNested () throws Exception
...
+            // setup, so Byteman injects the fault directly into resumeWrapper.
+            tmi.resume(tx);
+            fail("Expected InvalidTransactionException");
+        }
+        catch (InvalidTransactionException e)
+        {
+            // JBTM-4013: before the fix, initCause on InvalidTransactionException
+            // (which extends RemoteException) threw IllegalStateException here
+            // instead of propagating as InvalidTransactionException
+            assertEquals("IDL:omg.org/CosTransactions/InvalidControl:1.0", e.getMessage());
```
tomjenkinson: @marcosgopen do you think this looks better? You asked for the message being preserved in your earlier feedback

marcosgopen: Yes, that does look better and it is fine if all we need here is to know the message of the exception (e.g. IDL:omg.org/CosTransactions/InvalidControl:1.0), and this might be the case.
However be aware that using e1.getMessage we lose track of the exception chain and stack trace.

If we care about the stack trace maybe this can be better:
`throw new RemoteException(e1.toString(), e1);`

marcosgopen: Oh, I can see that changing the exception type will break the contract so your fix is probably the best we can achieve.

marcosgopen: What coderabbit suggests above is:

```
InvalidTransactionException itc = new InvalidTransactionException(e1.getMessage());
itc.detail = e1;  //maybe even itc.addSuppressed(e1);
throw itc;
```
This would preserve the original exception stack. Coderabbit also suggests "Inject a known message in the Byteman rule and assert the mapped message and  getCause() value." However, maybe the message is just enough for this case, so this might not be needed."

tomjenkinson: Do you think it looks better?

tomjenkinson: Hmm, it's complaining

ArjunaJTS/jtax/classes/com/arjuna/ats/internal/jta/transaction/jts/TransactionManagerImple.java
```diff
@@ -149,9 +149,9 @@ public void resume (Transaction which) throws InvalidTransactionException,
 	        }
 	        catch (org.omg.CosTransactions.InvalidControl e1)
 	        {
-                InvalidTransactionException invalidTransactionException = new InvalidTransactionException();
-                invalidTransactionException.initCause(e1);
-                throw invalidTransactionException;
+                InvalidTransactionException itc = new InvalidTransactionException(e1.getMessage());
+                itc.detail = e1;
+                throw itc;
```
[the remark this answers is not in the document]
tomjenkinson: @marcosgopen WDYT? I am not super sure about this. I am also wondering about using the addSuppressed instead?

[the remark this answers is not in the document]
marcosgopen: I would prefer setting the detail field as more semantically correct for this purpose. And maybe in the test 'testResumeInvalidControlThrowsInvalidTransactionException' you can assert that the e.getMessage() contains the string 'IDL:omg.org/CosTransactions/InvalidControl:1.0' instead? (I assume there are no other test checking the format of the error message).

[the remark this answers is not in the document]
tomjenkinson: Is that what I am doing as it stands in this pull?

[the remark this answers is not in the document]
marcosgopen: I can still see the 'assertEquals' (line 90?) which would cause the failure of the test I believe. If the last assert (`            assertTrue(e.getCause() instanceof org.omg.CosTransactions.InvalidControl);` passes) I think we are good .

marcosgopen (approved):
LGTM, Thanks Tom!

PR #3074, opened 2026-08-25
[#2945] Move Javadoc from MBean implementation classes to MBean inter…

tomjenkinson:
#2945 
CORE

### DIFF
--- a/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/CoordinatorEnvironmentBean.java
+++ b/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/CoordinatorEnvironmentBean.java
@@ -79,14 +79,6 @@ public class CoordinatorEnvironmentBean implements CoordinatorEnvironmentBeanMBe
 
     private volatile boolean finalizeBasicActions = false;
 
-    /**
-     * Returns true if anynchronous commit behaviour is enabled.
-     *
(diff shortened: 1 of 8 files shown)


marcosgopen (approved):
Thank you Tom! LGTM

@jmfinelli do you think we should also create a GH issue for moving javadoc from MBEans implementation from `com/arjuna/ats/arjuna/coordinator/ ` `com/arjuna/ats/internal/arjuna/objectstore/` `and com/arjuna/ats/arjuna/tools/osb` ?
Specifically for:
- HornetqJournalEnvironmentBean → HornetqJournalEnvironmentBeanMBean
- SlotStoreEnvironmentBean → SlotStoreEnvironmentBeanMBean

jmfinelli:
@tomjenkinson, I think this is a good opportunity to address coderabbit's review.

@marcosgopen, WDYT if we use this PR to move the remaining javadocs?

jmfinelli (changes requested):
@tomjenkinson, I think this is a good opportunity to address coderabbit's review.

marcosgopen:
> @tomjenkinson, I think this is a good opportunity to address coderabbit's review.
> 
> @marcosgopen, WDYT if we use this PR to move the remaining javadocs?

I think it is a good idea, but I let Tom decide as the current PR is fully consistent with the issue description at the moment.

PR #3095, opened 2026-09-17
[#3094] Close CMR prepare Connection when immediate cleanup is enabled

### LINKED ISSUE #3094: CMR immediate cleanup leaks prepare Connection (afterCompletion closes a different one)
## Description

When `JTAEnvironmentBean.performImmediateCleanupOfCommitMarkableResourceBranches` is `true`, `CommitMarkableResourceRecord` leaks one pooled JDBC `Connection` per successful 2PC commit.

WildFly's transactions subsystem defaults CMR `immediate-cleanup` to **true**, so this is the path most WildFly CMR users hit. Narayana's own bean default is **false**, which is probably why this has gone unreported.

## Environment

- Narayana `main` (same logic in 7.3.4.Final / WildFly 39.0.1)
- Connectable last-resource JDBC datasource (CMR) plus a second XA resource (e.g. JMS)
- IronJacamar pool with CCM: connections stay `InUse` until `IJ000453`

## Expected

The `Connection` obtained in `topLevelPrepare()` is returned to the pool after the XA commit, including when immediate cleanup is enabled.

## Actual

1. `topLevelPrepare()` keeps `preparedConnection = connectableResource.getConnection()` after the `INSERT` into `xids`.
2. `commit()` only calls `removeConnection()` when `!isPerformImmediateCleanupOfBranches`. With immediate cleanup the prepare connection is never closed.
3. The interposed `afterCompletion(STATUS_COMMITTED)` does `connectableResource.getConnection()` again, `DELETE`s the row, and `close()`s **that** connection — a different pool handle.
4. `topLevelAbort()` still closes via `removeConnection()` in `finally`, so the leak is the successful 2PC + immediate-cleanup path only.

The `notifyCommitMarkableResourceRecoveryModuleOfCompleteBranches` path is the `else if` of immediate cleanup. It does not close JDBC; with immediate cleanup **false**, `commit()` already closed `preparedConnection`.

## Code (`main`)

- Immediate-cleanup synch (new connection + close of local var only):  
  https://github.com/jbosstm/narayana/blob/main/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/CommitMarkableResourceRecord.java#L177-L254
- Prepare stores `preparedConnection`; commit skips `removeConnection()` when immediate cleanup is on:  
  https://github.com/jbosstm/narayana/blob/main/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/CommitMarkableResourceRecord.java#L381-L496
- `removeConnection()`:  
  https://github.com/jbosstm/narayana/blob/main/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/CommitMarkableResourceRecord.java#L508-L516

## Workaround

Set `immediate-cleanup="false"` so `commit()` closes the prepare connection and let `CommitMarkableResourceRecordRecoveryModule` garbage-collect `xids` (optionally with `batch-size` for the DELETE IN lists).

## Suggested fix

Either always call `removeConnection()` in `commit()`'s `finally` (afterCompletion can keep using a fresh connection for DELETE), or reuse `preparedConnection` for that DELETE and close it in the synch `finally`.

## Notes

Happy to test a patch. Observed under WildFly with MariaDB `connectable="true"` + `JmsXA`.

ychoudhry-getwell:
- commit() skipped removeConnection() when immediate cleanup was on, so the Connection from topLevelPrepare() never returned to the pool. The afterCompletion DELETE already uses a different getConnection() handle and closing that one does not release the prepare Connection.
- Always close preparedConnection in commit()'s finally, matching abort.
- AfterCompletion still opens its own Connection for the xids DELETE.

Fixes https://github.com/jbosstm/narayana/issues/3094


marcosgopen (approved):
LGTM, thanks Yogesh!
@jmfinelli do you know why the checks haven't run?

ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/resources/arjunacore/CommitMarkableResourceRecord.java
jmfinelli: While I was going through #3094, I noticed that this message is wrong. Could you fix this as well? Thanks

ychoudhry-getwell: Sure, updated the log message.

jmfinelli:
Just a quick note: we typically squash all commits into a single commit before merging.

ychoudhry-getwell:
> Just a quick note: we typically squash all commits into a single commit before merging.

commits are now squashed.

jmfinelli (approved):

